PR #216, opened 2016-03-11
Removed custom react-hot-loading devServer for webpack-dev-server

dmsnell:
The hot reloader was adding messy `ProxyComponent` wrappers around
everything and added unecessary complexity to the build process. This
commit removes that in favor of a config-less `webpack-dev-server` tool
instead.

This may lose some fluidity in propagating changes, but
webpack-dev-server now support auto-reloading anyway, so this new way is
probably good enough.

cc: @roundhill


dmsnell:
@roundhill I have been enough problems from the tacky react-hot-loader setup we had that I would like to consider this PR again. Please let me know how you feel about it.

roundhill:
Seeing some strange behavior with this branch. If I make a change, I see the app reload but the change isn't applied. If I save the file again or refresh the browser, then I see it.

dmsnell:
Abandoning.

PR #218, opened 2016-03-12
Replace emoji in title with non-bolded variants

### LINKED ISSUE #107: Emojis don't appear in the note list
I found some inconsistent emoji behavior. In the electron app, emojis don't appear unless the font weight is normal (bold weight is causing the issue):

screen shot 2015-12-10 at 12 05 35 pm

Testing in Chrome, the emojis don't render:
screen shot 2015-12-10 at 12 10 09 pm

Not sure what the difference is between the two.

Electron Version 0.35.4 (0.35.4)
Chrome Version 47.0.2526.80 (64-bit)

dmsnell:
Resolves #107 

Chromium doesn't display emoji and some other characters which have a font-weight other than the default set. We might be able to solve this by loading a webfoot, but this solution also is working somewhat. I've noticed some issues with certain Emoji that behave differently than others, such as the  :heart: or :airplane: which are appearing as a box in most cases but sometimes correctly.

This patch also doesn't fix the preview pane when Markdown is enabled, but I'm done for the day. Will need to run that preview through the `unbold()` or whatever also to fix there.

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -3,6 +3,8 @@ import marked from 'marked';
 import Textarea from 'react-textarea-autosize';
 import { noop, get } from 'lodash';
 
+import { unbold } from './note-display-mixin';
+
 const uninitializedNoteEditor = { focus: noop };
 
 export default React.createClass( {
@@ -79,7 +81,7 @@ export default React.createClass( {
 	},
 
 	renderMarkdown( divStyle ) {
-		var markdownHTML = marked( this.state.content );
+		var markdownHTML = marked( unbold( this.state.content ) );
 
 		return (
 			<div className="note-detail-markdown theme-color-bg theme-color-fg"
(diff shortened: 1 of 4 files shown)


dmsnell:
This PR fixes one major bug: [bolded Emoji disappear in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=441946&q=label%3AMovedFrom-41&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified). However, it still leaves open another major bug, [some Emoji simply don't display at all](https://bugs.chromium.org/p/chromium/issues/detail?id=551843).

The latest Canary build of Chrome works and displays the Emoji, so I'm not sure where that leaves us in fixing this with a workaround. I'm not sure exactly yet how we would work around this issue either, besides using something like Tweemoji, which replaces Emoji with images. I'm not a fan of that solution.

roundhill:
This fix unfortunately was causing performance issues for my account that has hundreds of notes in it. Like we discussed, let's just un-bold the list title for now and cross our fingers that Electron updates chromium to the version with the emoji fix soon.

PR #221, opened 2016-03-14
Change "Sign Out" to "Log Out"

### LINKED ISSUE #219: Change sign out to log out
Under Settings, it should be "Log Out" instead of "Sign Out". Also, need to make sure this change applies to the Electron app menus.

dmsnell:
Resolves #219 

@drw158 the Electron menu shows "Preferences" but doesn't have a sign-out option, just quit. Do you think we need to add this or were you just wanting to make sure we don't mix terminology?

### DIFF
--- a/lib/dialogs/settings.jsx
+++ b/lib/dialogs/settings.jsx
@@ -83,7 +83,7 @@ export default React.createClass( {
 						</div>
 
 						<ul className="dialog-actions">
-							<li><button type="button" className="button button-primary" onClick={this.props.onSignOut}>Sign Out</button></li>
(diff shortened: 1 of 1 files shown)


drw158:
I assume that Preferences opens the Settings modal (I haven't tested the latest package). I think that is fine.

I just wanted to make sure we didn't mix terminology, and be consistent with Log in/ Log out.

drw158:
:+1: to the change

dmsnell:
> I assume that Preferences opens the Settings modal (I haven't tested the latest package). I think that is fine.

This is correct, @drw158 - it all matches.

PR #223, opened 2016-03-14
Center Preview/Edit mode selector in Safari

### LINKED ISSUE #147: Markdown toggle not centered in Safari 9
screen shot 2016-02-05 at 12 03 30 pm

dmsnell:
Resolves #147

Previously the mode bar in the note editor wasn't centered in Safari 9.
This was probably due to `margin: 0 auto` being set without having a
`width` set as well.

Now, the mode bar has been nested inside another `<div />` to setup
relative positioning and translations to make sure the bar gets centered
in all browsers.

cc: @drw158 please review the way I accomplished this with CSS

### DIFF
--- a/scss/editor.scss
+++ b/scss/editor.scss
@@ -55,7 +55,7 @@
 	display: flex;
 	align-items: center;
 	padding: 0 15px;
-	
+
 	@media only screen and ( max-width: $single-column ) {
 		width: 100%;
 		justify-content: space-between;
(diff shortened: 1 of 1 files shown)


drw158:
That works, but I think it will be simplest to use flexbox here.

1. Remove the extra div you created.
2. Only CSS change that is needed:

```
.note-editor-mode-bar {
     display: flex;
     justify-content: center;
}
```

The margin, display, and flex properties on that class can be removed.

Seems to work well.

dmsnell:
Genius @drw158!

roundhill:
:shipit:

PR #225, opened 2016-03-14
Hide publish link if note not published

### LINKED ISSUE #130: Idea: hide link field if not published
link

dmsnell:
Resolves #130

Previously in the share dialog, the published link field would appear
even if the note wasn't being shared.

Now, the pushlished link field only appears if the published toggle is
on. Additionally, the copy link only enables if a shortlink is
available, but not during the waiting time to get that link.

cc: @drw158 @roundhill

### DIFF
--- a/lib/dialogs/share.jsx
+++ b/lib/dialogs/share.jsx
@@ -168,18 +168,25 @@ export default React.createClass( {
 								Anyone with the public link will be able to view the latest version.
 							</p>
 						</div>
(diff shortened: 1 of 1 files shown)


drw158:
This is great, I didn't think we'd ever get around to this. The UI/UX works as expected. Thanks for doing this!

roundhill:
Looks good :shipit:

PR #226, opened 2016-03-14
Manually sort notes when updating store

### LINKED ISSUE #144: Sorting doesn't work in Firefox
Just a note that if we ever want to release this on the web, the sorting doesn't work in Firefox.

dmsnell:
Resolves #144

Previously, the sort or the notes was determined by **indexdb**, which
apparently didn't work as expected in Firefox.

Now, the notes are manually sorted in JavaScript to guarantee sort
order. This works in Firefox and should work universally.

I changed the note filtering function a bit too.

@roundhill you might want to verify the performance with a big bucket of notes

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -17,13 +17,7 @@ import TagsIcon from './icons/tags'
 import NoteDisplayMixin from './note-display-mixin'
 import analytics from './analytics'
 import classNames	from 'classnames'
-import {
-	get,
-	has,
-	includes,
-	matchesProperty,
-	noop
-} from 'lodash';
+import { noop, get, has } from 'lodash';
 
 let ipc = getIpc();
 
@@ -59,19 +53,6 @@ const isElectron = ( () => {
 
 	return () => foundElectron;
 } )();
-
-/* Note filters */
-const untrashedAndVisibleTrash = showTrash =>
-	matchesProperty( 'data.deleted', showTrash );
-
-const matchingSelectedTag = tag => note =>
-	! tag || includes( note.data.tags, tag.data.name );
-
-const matchesSearch = regexp => note =>
-	! regexp || regexp.test( get( note, 'data.content', '' ) );
-
-const matchesAll = ( ...filters ) => o =>
-	filters.reduce( (t, c) => t && c(o), true );
 
 export const App = connect( mapStateToProps, mapDispatchToProps )( React.createClass( {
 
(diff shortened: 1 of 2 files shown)


roundhill:
Just tried this out with my account (a few hundred notes), and it didn't seem to work in Firefox. Clicks on the notes list stopped working as well, so I'm wondering if it is taking too much to process the array...

Some log stuffs:

```
"Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `ProxyComponent`. See https://fb.me/react-warning-keys for more information." app.js line 799 > eval:45:9

Warning: Failed propType: Required prop `key` was not specified in `ProxyComponent`. Check the render method of `ProxyComponent`. app.js line 799 > eval:45:9

TypeError: event.srcElement is undefined
 app.js line 3289 > eval:79:1
```

dmsnell:
Thanks for looking into this @roundhill. The `key` warnings were already there and that darn `ProxyComponent` has to do with the way the local dev environment was setup with the React hot loader. If you run via `webpack-dev-server` or what-not then the proxy components disappear to give more useful information.

I wouldn't be surprised if the broken clicks were from processing in the main thread. It would be nice to know how often these get sorted as I think a JS solution is best, but maybe the architecture isn't well suited for sort-once-then-reuse.

dmsnell:
@roundhill this is a strange thing. I _was_ able to see the messages you saw, and they weren't there in **master**, but they didn't make any sense to me. Also, after I ran this without the funky proxy stuff going on, I didn't get the errors. Could you run the following sequence and see if you still get the messages and flakey notes? (by the way, I was never able to reproduce the clicks not working)

```bash
git checkout master
git pull -p
git checkout issue/144-no-firefox-sort
git checkout HEAD~1
NODE_ENV=development node `npm bin`/webpack-dev-server --hot --progress --inline --port 4000 --history-api-fallback
```

I have taken the liberty of adding another commit which removed the messages for me, but I would like you to try without that "revert" commit and see if this wasn't more of a funny relationship with the react hot loader. On the other hand, you could try with a production build as well since the proxy wrapper should only be here in development builds.

Thanks!

roundhill:
Alas, I am still seeing the same behavior in Firefox :cry: It does seem to be working well in Electron, though.

dmsnell:
Thanks for the review @roundhill - and happy Easter!

Did this also fail out in Firefox on the branch tip, without backing up a commit?

dmsnell:
Abandoning since sort is fine in Chrome

PR #229, opened 2016-03-16
WIP: Provide stable tag sort

### LINKED ISSUE #166: Deleting tags causing reordering for some reason
Deleting certain tags seem to randomly sort them around.

tags delete

dmsnell:
Resolves #166

The goal is to alphabetically sort the tags but place tags with a
specific index at those index locations and push the tags without a
specific index down the line

On hold for the moment while attending other priorities.


dmsnell:
Abandoning since sorting is fine in Firefox

PR #242, opened 2016-03-21
Adds copy button to note info panel.

### LINKED ISSUE #241: Info: add a way to copy the published link for the info menu
Right now clicking is disabled.

We can enable selection, or add copy button.

roundhill:
I pretty much copied (ha!) how the copy button worked in the publish panel.

@drw158 The copy button didn't really fit next to the Url so I put it beneath it:

screen shot 2016-03-21 at 4 56 37 pm

Fixes #241

### DIFF
--- a/lib/dialogs/share.jsx
+++ b/lib/dialogs/share.jsx
@@ -3,6 +3,7 @@ import createHash from 'create-hash/browser';
 import { includes } from 'lodash';
 import TabbedDialog from '../tabbed-dialog';
 import ToggleControl from '../controls/toggle';
+import { isEmpty } from 'lodash';
 
 const shareTabs = [ 'collaborate', 'publish' ];
 
@@ -38,6 +39,10 @@ export default React.createClass( {
 		}
 
 		this.copyUrlElement.focus();
+	},
+
+	getPublishURL( url ) {
+		return isEmpty( url ) ? null : `http://simp.ly/p/${url}`;
 	},
 
 	onAddCollaborator( event ) {
@@ -111,7 +116,7 @@ export default React.createClass( {
 		const { note } = this.props.params;
 		const data = note && note.data || {};
 		const isPublished = includes( data.systemTags, 'published' );
-		const publishURL = isPublished && data.publishURL !== '' && data.publishURL && `http://simp.ly/publish/${data.publishURL}` || null;
+		const publishURL = this.getPublishURL( data.publishURL );
 
 		switch ( tabName ) {
 			case 'collaborate':
--- a/lib/note-info.jsx
+++ b/lib/note-info.jsx
@@ -3,6 +3,7 @@ import { includes } from 'lodash';
 import ToggleControl from './controls/toggle';
 import moment from 'moment';
 import CrossIcon from './icons/cross';
+import { isEmpty } from 'lodash';
 
 export default React.createClass( {
 
@@ -22,6 +23,22 @@ export default React.createClass( {
 		this.props.onOutsideClick( false );
 	},
 
+	copyPublishURL: function() {
+		this.publishUrlElement.select();
+
+		try {
+			document.execCommand( 'copy' );
+		} catch ( err ) {
+			return;
+		}
+
+		this.copyUrlElement.focus();
+	},
+
+	getPublishURL: function( url ) {
+		return isEmpty( url ) ? null : `http://simp.ly/p/${url}`;
+	},
+
 	render: function() {
 		const { note, markdownEnabled } = this.props;
 		const data = note && note.data || {};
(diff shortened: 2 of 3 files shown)


drw158:
I just committed a change so that it would be on one line. The text is truncated, but I think that is ok, and hopefully we can get the urls a bit shorter one day :)

roundhill:
Added the new shorter `p` url, has space now:

screen shot 2016-03-22 at 1 06 32 pm

lib/dialogs/share.jsx
```diff
@@ -111,7 +111,7 @@ export default React.createClass( {
 		const { note } = this.props.params;
 		const data = note && note.data || {};
 		const isPublished = includes( data.systemTags, 'published' );
-		const publishURL = isPublished && data.publishURL !== '' && data.publishURL && `http://simp.ly/publish/${data.publishURL}` || null;
+		const publishURL = isPublished && data.publishURL !== '' && data.publishURL && `http://simp.ly/p/${data.publishURL}` || null;
```
dmsnell: Now would be a good time to clean up this line. What's that `null` doing there anyway?

How about something like this:
```js
const publishURL = ! isEmpty( data.publishURL ) && data.publishURL;
```

With this, `publishURL` will either be a non-empty string, or it will be `undefined`, both of which probably should work fine for the rest of the file because `if ( ! publishURL )` will still evaluate to `false` and we test for the section stuff with `isPublished`

I did not test this.

lib/note-info.jsx
```diff
@@ -30,7 +42,7 @@ export default React.createClass( {
 		const isPinned = includes( data.systemTags, 'pinned' );
 		const isMarkdown = includes( data.systemTags, 'markdown' );
 		const isPublished = includes( data.systemTags, 'published' );
-		const publishURL = isPublished && data.publishURL !== '' && data.publishURL && `http://simp.ly/publish/${data.publishURL}` || null;
+		const publishURL = isPublished && data.publishURL !== '' && data.publishURL && `http://simp.ly/p/${data.publishURL}` || null;
```
dmsnell: Similar place as above to simplify. This just seems like a crazy line to parse and understand what the code's intent was. I keep thinking, "am I missing something? why is this such a long construct?"

roundhill:
@dmsnell I simplified it a bit more in https://github.com/Automattic/simplenote-electron/pull/242/commits/aaa38d876915d2d5ac968691d190969019b1505b. It's still a big line. The `null` is needed I think because it is used as the `value` of an `<input>`, and will display the text `false` if it is a boolean.

lib/dialogs/share.jsx
```diff
@@ -111,7 +112,7 @@ export default React.createClass( {
 		const { note } = this.props.params;
 		const data = note && note.data || {};
 		const isPublished = includes( data.systemTags, 'published' );
-		const publishURL = isPublished && data.publishURL !== '' && data.publishURL && `http://simp.ly/publish/${data.publishURL}` || null;
+		const publishURL = ! isEmpty( data.publishURL ) && `http://simp.ly/p/${data.publishURL}` || null;
```
dmsnell: Thanks for simplifying. I guess I meant that the `&&` `||` with `null` was funny. Though I remember seeing the use in `value`, I though there was already protection against this.

Might still consider using a ternary here or in the view below since it does exactly what this line is doing…

```js
const publishURL = ! isEmpty( data.publishURL )
	? `http://simp.ly/p/${data.publishURL}`
	: '';
```

or handle it in the view
```js
const publishURL = ! isEmpty( data.publishURL ) && `http://simp.ly/p/${data.publishURL}`;

…

// looks kinda ugly, but somewhere something is going to be ugly anyway ¯\_(ツ)_/¯
value={ publishURL && publishURL }
```

roundhill: > // looks kinda ugly, but somewhere something is going to be ugly anyway ¯\_(ツ)_/¯
> value={ publishURL && publishURL }

lol! I guess that looks better...

dmsnell: There another approach that may be more elegant too at the cost of a bit more mental work. What we're really wanting is a function that gives us a default value like `get( object, path, default )` except that we want it to use the default whenever the value is string-empty, a `getWith( object, path, default[ predicate, defaultValue ] )`

Iteration 1 - make a simple function, maybe the best option here

```js
const getPublishUrl = url => isEmpty( url ) ? null : `http://simp.ly/p/${url}`;

…

// good and good - now it's null if the string is empty
const publishURL = getPublishUrl( data.publishURL );
```

Iteration 2 - make our reusable getWith()

```js
const getWith = ( object, path, predicate, default ) => {
	const value = get( object, path );
	return predicate( value ) ? value : default;
};

…
const notEmpty = negate( isEmpty );
const publishURL = getWith( data, 'publishURL', notEmpty, null );
```

dmsnell:
after a `squash` I say :shipit:

lib/dialogs/share.jsx
roundhill: > Iteration 1 - make a simple function, maybe the best option here

I went with this, created a function.

PR #246, opened 2016-03-23
Autoupdate feature for OS X, manual update notification for the rest

rodrigoi:
To test, make and run :smile: 

This also bumps version to `v1.0.0-rc.1` and forces update for testing.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -1,11 +1,18 @@
+'use strict';
+
 var app = require( 'app' );	 // Module to control application life.
 var Menu = require( 'menu' );
 var BrowserWindow = require( 'browser-window' );	// Module to create native browser window.
 var path = require('path')
 
-module.exports = function main( url ) {
+require( 'module' ).globalPaths.push( path.resolve( path.join( __dirname ) ) );
+
+module.exports = function main() {
+	var url = 'file://' + path.join( __dirname, '..', 'dist', 'index.html' );
+
 	// Report crashes to our server.
 	require( 'crash-reporter' ).start();
+	require( './updater' )();
 
 	// Keep a global reference of the window object, if you don't, the window will
 	// be closed automatically when the JavaScript object is GCed.
--- a/desktop/config-updater.json
+++ b/desktop/config-updater.json
@@ -1,0 +1,8 @@
+{
+	"forceUpdate": true,
+	"updater": {
+		"url": "https://app.simplenote.com/desktop/{platform}/version?compare={version}",
+		"delay": 2000,
+		"interval": 600000
+	}
+}
--- a/desktop/config/index.js
+++ b/desktop/config/index.js
@@ -1,0 +1,12 @@
+'use strict';
+
+const pkg = require( '../../package.json' );
+let config = require( '../config-updater.json' );
+
+// Merge in some details from package.json
+config.name = pkg.productName;
+config.description = 'Simplenote Desktop';
+config.version = pkg.version;
+config.author = pkg.author;
+
+module.exports = config;
--- a/desktop/index.js
+++ b/desktop/index.js
@@ -1,0 +1,1 @@
+var app = require('./app')();
--- a/desktop/platform/index.js
+++ b/desktop/platform/index.js
@@ -1,0 +1,28 @@
+'use strict';
+
+/**
+ * Module variables
+ */
+let platform = false;
+
+function Platform() {
+	this.platform = false;
+}
+
+Platform.prototype.isOSX = function() {
+	return process.platform === 'darwin';
+};
+
+Platform.prototype.isWindows = function() {
+	return process.platform === 'win32';
+};
+
+Platform.prototype.isLinux = function() {
+	return process.platform === 'linux';
+};
+
+if ( ! platform ) {
+	platform = new Platform();
+}
+
+module.exports = platform;
--- a/desktop/updater/index.js
+++ b/desktop/updater/index.js
@@ -1,0 +1,51 @@
+'use strict';
+
+/**
+ * External Dependencies
+ */
+const electron = require( 'electron' );
+const app = electron.app;
+const dialog = electron.dialog;
+
+/**
+ * Internal dependencies
+ */
+const platform = require( '../platform' );
+const config = require( '../config' );
+const AutoUpdater = require( './auto-updater' );
+const ManualUpdater = require( './manual-updater' );
+
+let updater = false;
+
+function urlBuilder( version ) {
+	let platformString = 'osx';
+
+	if ( platform.isWindows() ) {
+		platformString = 'windows';
+	} else if ( platform.isLinux() ) {
+		platformString = 'linux';
+	}
+
+	if ( config.forceUpdate ) {
+		version = '0.0.1';
+	}
+
+	return config.updater.url.replace( '{platform}', platformString ).replace( '{version}', version );
+}
+
+module.exports = function() {
+	app.on( 'will-finish-launching', function() {
+		let url = urlBuilder( app.getVersion() );
+
+		if ( platform.isOSX() ) {
+			updater = new AutoUpdater( url );
+		} else {
+			updater = new ManualUpdater( url );
+		}
+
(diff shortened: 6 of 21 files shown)


desktop/platform/index.js
```diff
@@ -0,0 +1,37 @@
...
+Platform.prototype.isOSX = function() {
+	return process.platform === 'darwin';
+};
+
+Platform.prototype.isWindows = function() {
+	return process.platform === 'win32';
+};
+
+Platform.prototype.isWindows10 = function() {
+	return parseInt( os.release(), 10 ) >= 10;
```
dmsnell: should we not call this something like `isAtLeastWindows10` or redo the comparison so we aren't doing `>=`?

desktop/updater/index.js
```diff
@@ -0,0 +1,50 @@
...
+		platformString = 'windows';
+	} else if ( platform.isLinux() ) {
+		platformString = 'linux';
+	}
+
+	if ( config.forceUpdate ) {
+		version = '0.0.1';
+	}
+
+	return config.updater.url + platformString + '/version?compare=' + version;
```
dmsnell: it would be nice to see us using a URL path builder here just to prevent issues with the slashes and query parts. for example, if we mistype coming in `config.updater.url`

```js
import joinUrl from 'url-join';

return joinUrl( config.updater.url, platformString, 'version?compare=' + version );
```

or a step further…

```js
import url from 'url';
import joinUrl from 'url-join';

return url.format( {
	...url.parse( joinUrl( config.updater.url, platformString, 'version' ) ),
	query: { compare: version }
} );
```

Unfortunately it doesn't look like `url` handles joining path pieces in a way that prevents double slashes.

desktop/updater/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+		if ( platform.isOSX() ) {
+			updater = new AutoUpdater( url );
+		} else {
+			updater = new ManualUpdater( url );
+		}
+
+		// Start one straight away
+		setTimeout( updater.ping.bind( updater ), config.updater.delay );
+		setInterval( updater.ping.bind( updater ), config.updater.interval );
```
dmsnell: There's probably a valid reason to have the updater call itself or wrap this in a function that recursively calls `setTimeout` because `setInterval()` will attempt to call the given function every interval on the interval. Granted, the config is set at 10 minutes, but supposing there were network trouble or some freak delay or a misconfigured `config` maybe we wouldn't want to pile up those calls.

desktop/updater/manual-updater/index.js
```diff
@@ -0,0 +1,55 @@
...
+	https.get( this.url, function( response ) {
+		let body = '';
+
+		response.on( 'data', function( chunk ) {
+			body += chunk;
+		} );
+
+		response.on( 'end', function() {
+			if ( body ) {
+				let update = JSON.parse( body );
```
dmsnell: this could throw an exception, which is probably something we want to guard against with a `try` block so that a network or server issue doesn't kill the app

desktop/platform/index.js
rodrigoi: I'll do something better. I'll remove the function. Not even calypso is using it :disappointed:

roundhill:
Looks good, I only tested on OS X though. @rodrigoi were you able to test on linux/windows?

rodrigoi:
@roundhill yes, works fine on Windows and Linux.

The automatic updates only work on OS X. For all others, we have to open a browser window to initiate a download, and we can control the page we show the user on the backend.

screen shot 2016-03-23 at 18 59 50
screen shot 2016-03-23 at 19 02 33

roundhill:
Cool, I think we're good to go once the test stuff is removed and it is pointed to production.

rodrigoi:
@dmsnell I've refactored the `setInterval` out. Now the `AutoUpdater` and `ManualUpdater` emit a `end` event in case of error, update unavailable or a cancel click on the dialog window, and we use that event to run a new interval. Let me know if you are happy with those changes. To test the auto updater, you'll have to pack the app (`make osx`, `make win32` or `make linux`).

desktop/app.js
```diff
@@ -3,9 +3,14 @@ var Menu = require( 'menu' );
 var BrowserWindow = require( 'browser-window' );	// Module to create native browser window.
 var path = require('path')
 
-module.exports = function main( url ) {
+require( 'module' ).globalPaths.push( path.resolve( path.join( __dirname ) ) );
+
+module.exports = function main() {
+	var url = 'file://' + path.join( __dirname, '..', 'dist', 'index.html' );
```
dmsnell: I hate to sound naggy, but we should be careful here using `path.join()` to form a URL. Maybe it's not a problem, but of course the Windows path separator is a `\` and not a `/` and I'm not sure if that will cause issues. Maybe it's fine since it's `file://` but generally I would recommend specifically using a URL path joiner.

desktop/config-updater.json
```diff
@@ -0,0 +1,8 @@
+{
+	"forceUpdate": true,
+	"updater": {
+		"url": "https://app.simplenote.com/desktop/{platform}/version?compare={version}",
```
dmsnell: :smile:

desktop/config/index.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
```
dmsnell: Are we using this elsewhere in the app? Do we even need to?

I _am_ aware that certain parts of the code can't be run through Babel and what not, so please pardon me if this _has_ to be vanilla JS.

desktop/index.js
```diff
@@ -0,0 +1 @@
+var app = require('./app')();
```
dmsnell: Why does this file exist? Do we plan on having other non-desktop index files that do something else? it seems like we should probably just include `./app` wherever it is we include `desktop`

desktop/platform/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+Platform.prototype.isLinux = function() {
+	return process.platform === 'linux';
+};
+
+if ( ! platform ) {
+	platform = new Platform();
+}
+
+module.exports = platform;
```
dmsnell: A little complicated?

```js
const os = require( 'os' );

const platformInfo = ( { platform } ) => ( {
	isOSX: platform === 'darwin',
	isWindows: platform === 'win32',
	isLinux: platform === 'linux'
} );

module.exports = platformInfo( process );
```

seems simpler _and_ it's testable if we were to export the `platformInfo` function.

desktop/platform/index.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+/**
+ * Internal dependencies
+ */
+const os = require( 'os' );
```
dmsnell: Does this have some hidden behavior? It doesn't get called anywhere else in this code…

dmsnell:
Lookin' good @rodrigoi

rodrigoi:
@dmsnell I've cleaned up that `os` reference that was unused, but sadly destructuring parameters are not supported by electron. Everything in `desktop` run as a node.js app and to support that syntax we'll have to transpile the code. I'd rather leave it as it is for now.

The `index.js` file is more or less a convention. We could set `app.js` on `package.json` and run `main()` directly, but I'd rather not have that module call itself.

desktop/app.js
rodrigoi: `path.join` will indeed use `\` on Windows, but Chromium handles that just fine. It changes the separator to `/`, and also adds the third slash to `file://`. There's no need for us to detect the OS and replace `\` with `/`.

dmsnell: yeah - I wouldn't ever recommend that we detect and replace the slashes :wink:

this is fine. I think ideally we'd use a URL-builder, but if IE/Chrome/Firefox in Windows handle the wrong slash it's okay to ship.

desktop/config/index.js
rodrigoi: Everything on `desktop` runs as a standard node.js app, and indeed, it has to be vanilla JS.
I'd like to add a build step in the future to transpile the desktop code to enable some more advanced syntax features. Maybe for `v1.0.1` :smile:

PR #247, opened 2016-03-23
Fix Typing Lag

roundhill:
Updates `note-detail` to handle the saving of the note itself. The note will be saved if 3 seconds have passed since the user stopped typing, or if a different note is selected.

@dmsnell what do you think of this solution? Seemed best performance-wise to not have the textarea be directly tied to the note prop.

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -1,9 +1,11 @@
 import React, { PropTypes } from 'react';
 import marked from 'marked';
 import Textarea from 'react-textarea-autosize';
-import { noop, get } from 'lodash';
+import { noop, get, debounce } from 'lodash';
+import analytics from './analytics'
 
 const uninitializedNoteEditor = { focus: noop };
+const saveDelay = 2000;
 
 export default React.createClass( {
 
@@ -14,16 +16,14 @@ export default React.createClass( {
 		onChangeContent: PropTypes.func.isRequired
 	},
 
-	getInitialState: function() {
-		var note = this.props.note;
-
-		return {
-			content: get( note, 'data.content', '' )
-		};
+	componentWillMount: function() {
+		this.queueNoteSave = debounce( this.saveNote, saveDelay );
+		this.noteEditor = uninitializedNoteEditor;
 	},
 
-	componentWillMount: function() {
-		this.noteEditor = uninitializedNoteEditor;
+	componentDidMount: function() {
+		// Ensures note gets saved if user abruptly quits the app
+		window.addEventListener( 'beforeunload', this.queueNoteSave.flush );
 	},
 
 	initializeNoteEditor: function( noteEditor ) {
@@ -34,22 +34,28 @@ export default React.createClass( {
 		return note && note.id;
 	},
 
-	componentWillReceiveProps: function( { note } ) {
-		this.setState( {
-			content: get( note, 'data.content', '' )
-		} );
+	componentWillReceiveProps: function() {
+		this.queueNoteSave.flush();
 	},
 
 	componentDidUpdate: function() {
-		// Let's focus the editor for new and empty notes
 		const { note } = this.props;
-		const { content } = this.state;
-		if ( this.isValidNote( note ) && content === '' ) {
-			this.noteEditor.focus();
+		const content = get( note, 'data.content', '' );
+		if ( this.isValidNote( note ) && this.noteEditor ) {
+			this.noteEditor.value = content;
+
+			// Let's focus the editor for new and empty notes
+			if ( content === '' ) {
+				this.noteEditor.focus();
+			}
 		}
 	},
 
-	onPreviewClick( event ) {
+	componentWillUnmount: function() {
+		window.removeEventListener( 'beforeunload', this.queueNoteSave.flush );
+	},
+
+	onPreviewClick: function( event ) {
 		// open markdown preview links in a new window
 		for ( let node = event.target; node != null; node = node.parentNode ) {
 			if ( node.tagName === 'A' ) {
@@ -58,6 +64,15 @@ export default React.createClass( {
 				break;
 			}
 		}
+	},
+
+	saveNote: function() {
+		const { note } = this.props;
+
+		if ( ! this.isValidNote( note ) ) return;
+
+		this.props.onChangeContent( note, this.noteEditor.value );
+		analytics.tracks.recordEvent( 'editor_note_edited' );
 	},
 
 	render: function() {
@@ -79,7 +94,8 @@ export default React.createClass( {
 	},
 
 	renderMarkdown( divStyle ) {
-		var markdownHTML = marked( this.state.content );
+		const { content = '' } = this.props.note.data;
+		const markdownHTML = marked( content );
 
 		return (
 			<div className="note-detail-markdown theme-color-bg theme-color-fg"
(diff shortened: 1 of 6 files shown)


lib/note-detail.jsx
```diff
@@ -34,22 +30,34 @@ export default React.createClass( {
...
-	componentWillReceiveProps: function( { note } ) {
-		this.setState( {
-			content: get( note, 'data.content', '' )
-		} );
+	componentWillReceiveProps: function() {
+		const { note } = this.props;
+
+		if ( ! this.isValidNote( note ) || ! this.noteEditor ) return;
+
+		if ( saveTimerId ) clearTimeout( saveTimerId );
```
dmsnell: `clearTimeout()` is always safe to call, even if the timer handle is `null`. I like to avoid the conditionals just to make the code smoother

lib/note-detail.jsx
```diff
@@ -34,22 +30,34 @@ export default React.createClass( {
...
+
+			// Let's focus the editor for new and empty notes
+			if ( content === '' ) {
+				this.noteEditor.focus();
+			}
 		}
 	},
 
-	onPreviewClick( event ) {
+	onPreviewClick: function( event ) {
```
dmsnell: intentional change?

lib/note-detail.jsx
```diff
@@ -60,6 +68,21 @@ export default React.createClass( {
 		}
 	},
 
+	onEditorChanged: function() {
+		clearTimeout( saveTimerId );
+		// Save the note after user stops typing
+		saveTimerId = delay( this.saveNote, saveDelay );
+	},
```
dmsnell: I believe you are looking for `debounce()` which would let you remove the timer handles and delay variables.

```js
onEditorChanged: debounced( this.saveNote, saveDelay )
```

What `debounce()` does is wait until the delay period has finished before actually running the function. If more calls to this function come in during that waiting period, it resets the clock. Additionally, it has a few options which can still help us out, such as the `maxWait` property. `maxWait` will guarantee that the function gets called at least as often as the number of milliseconds given in. That is, we could tell the app to wait to save the note until the user stops typing for three seconds, but if they continue to type, we could save it at least every twenty seconds.

```js
import { debounce } from 'lodash';
onEditorChanged: debounced( this.saveNote, saveDelay, { maxWait: 30000 } );
```

lib/note-detail.jsx
```diff
@@ -90,16 +114,12 @@ export default React.createClass( {
...
-		var valueLink = {
-			value: this.state.content,
-			requestChange: this.props.onChangeContent.bind( null, note )
-		};
+		const note = this.props.note;
 
 		return (
 			<Textarea ref={ this.initializeNoteEditor } className="note-detail-textarea theme-color-bg theme-color-fg"
 				disabled={ !!( note && note.data.deleted ) }
-				valueLink={ valueLink }
```
dmsnell: I'm a big fan of removing the `valueLink` stuff!

lib/note-detail.jsx
```diff
@@ -90,16 +114,12 @@ export default React.createClass( {
...
+				onChange={ this.onEditorChanged }
```
dmsnell: given that we have a succinct helper to work with, we could actually eliminate `onEditorChanged` and make the call here:

```js
onChange={ debounce( this.saveNote, saveDelay, { maxWait: minimumSaveInterval } ) }
```

lib/note-detail.jsx
```diff
@@ -34,22 +30,34 @@ export default React.createClass( {
...
+
+		if ( ! this.isValidNote( note ) || ! this.noteEditor ) return;
+
+		if ( saveTimerId ) clearTimeout( saveTimerId );
+
+		// Save note if content changes haven't been saved yet
+		if ( note.data.content !== this.noteEditor.value ) {
+			this.props.onChangeContent( note, this.noteEditor.value );
+			analytics.tracks.recordEvent( 'editor_note_edited' );
+		}
```
dmsnell: Is the purpose of this function to make sure that an incoming update from the server doesn't overwrite the content that we have already changed but not yet dispatched? In other words, if we have stopped typing for two seconds and we get an update from the server, is this code supposed to prevent that update from wiping out what I did two second ago?

lib/note-detail.jsx
```diff
@@ -34,22 +30,34 @@ export default React.createClass( {
...
+		const content = get( note, 'data.content', '' );
+		if ( this.isValidNote( note ) && this.noteEditor ) {
+			this.noteEditor.value = content;
+
+			// Let's focus the editor for new and empty notes
+			if ( content === '' ) {
+				this.noteEditor.focus();
+			}
 		}
 	},
```
dmsnell: I'm not sure if you are aware, but `componentDidUpdate( prevProps, prevState )` actually has access to the old props. It looks like we could accomplish in one spot what we are doing in the two lifecycle methods here.

If we use something like the debounce function from **lodash**, this would be a good spot to call its `flush` method.

```js
componentDidUpdate() {
	this.onEditorChanged.flush();

	…
}
```

that `flush()` method won't do anything if no changes are pending. In other words, if I haven't changed the note, then I haven't requested the update, so when `flush()` gets called [it has nothing to run](https://github.com/lodash/lodash/blob/4.6.1/lodash.js#L8824) and just returns. The only **catch** I see is that `this.saveNote()` relies on `this.props`, which when called from `componentDidUpdate()` are the new ones. I'm not sure if that matters or not.

dmsnell:
@roundhill here are a few hefty comments. This is great work you are doing - thanks for fixing this potentially embarrassing performance issue!

lib/note-detail.jsx
roundhill: I gave debounce a try but it didn't do the trick for me. It would delay the first execution, but then all of the rest of the queued calls to the functions would fire after the 3 second delay. Is that what `maxWait` can help with?

lib/note-detail.jsx
roundhill: :+1:

roundhill:
- [ ] Bug: if you start typing, then close the browser or quit the app you will lose those changes.

lib/note-detail.jsx
roundhill: :+1: Nice, I missed that it had `prevProps`.

lib/note-detail.jsx
roundhill: This was my attempt to save the note content before a change happens, or if the user switches notes. So if you start typing, then immediately switch notes before the timer fires.

I will test that server change scenario, I think there might be content loss in that case!

lib/note-detail.jsx
roundhill: Yeah, even though it's cool these days to not do that, I was matching it up with the rest of the class ;)

lib/note-detail.jsx
dmsnell: Not sure exactly what behavior you are seeing, @roundhill, but I made a quick [example on jsbin](http://jsbin.com/jeyoxibula/3/edit?js,console)

roundhill:
@dmsnell Ready for another look. Got `debounce` working and its `flush` method helped the code be a lot simpler!

dmsnell:
@roundhill I'm happy with this if you are happy with my contributions!

dmsnell:
we should squash it first though!

roundhill:
I am, and huge thanks for the assistance once again!

PR #250, opened 2016-03-24
New menu items and keyboard shortcuts for New Note and Search

### LINKED ISSUE #249: App is Missing a Few Keyboard Shortcuts
Let's add:
* File->New Note (cmd or ctrl + N)
* Focus search field (cmd or ctrl + L)

roundhill:
I had to add a special case for `newNote` in the `onAppCommand` method because it needs to pass the note bucket to the actions method.

Search field focus works by adding a new boolean state that instantly gets set back to false once the field has been focused.

Fixes #249

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -114,7 +114,12 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 
 	onAppCommand: function( command ) {
 		if ( command != null && typeof command === 'object' && command.action != null && this.props.actions.hasOwnProperty( command.action ) ) {
-			this.props.actions[ command.action ]( command );
+			// newNote expects a bucket to be passed in, but the action method itself wouldn't do that
+			if ( command.action === 'newNote' ) {
+				this.onNewNote();
+			} else {
+				this.props.actions[ command.action ]( command );
+			}
 		}
 	},
 
@@ -147,6 +152,12 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 	onNotePrinted: function() {
 		this.props.actions.setShouldPrintNote( {
 			shouldPrint: false
+		} );
+	},
+
+	onSearchFocused: function() {
+		this.props.actions.setSearchFocus( {
+			searchFocus: false
 		} );
 	},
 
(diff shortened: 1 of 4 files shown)


drw158:
Heh, I tried to test this locally but realized that the shortcuts are used by the browser.

lib/app.jsx
```diff
@@ -114,7 +114,12 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 
 	onAppCommand: function( command ) {
 		if ( command != null && typeof command === 'object' && command.action != null && this.props.actions.hasOwnProperty( command.action ) ) {
-			this.props.actions[ command.action ]( command );
+			// Special case for newNote since it needs the note bucket prop from this class
```
dmsnell: I'm confused about the comment here. Is the reason we have the special case because we need to preserve the `this` context by calling `onNewNote()` as a method on `this`?

roundhill: No, sorry if I wasn't clear. If you look at the (newNote)[https://github.com/Automattic/simplenote-electron/blob/issue/249-keyboard-shortcuts/lib/flux/app-state.js#L261] action, it requires that the bucket object gets passed to it. These `appCommands` usually call the action methods directly which is why I added this case.

dmsnell: oh okay, that makes sense. I wish some of this could have been a little less-custom in the way we have our app state, but at least we're in a much better spot than the notes client :smile:

maybe we could clarify in the comment, something like
> newNote expects a bucket to be passed in, but the action method itself wouldn't do that

lib/auth.jsx
```diff
@@ -20,7 +20,7 @@ export default React.createClass( {
 	render() {
 		const { isAuthenticated, authPending } = this.props;
 		const submitClasses = classNames( 'button', 'button-primary', {
-			'pending': authPending
+			pending: authPending
 		} );
```
rodrigoi: some syntactic sugar. Destructuring assignment lets's you assign an extracted value to a new variable name.

```
const { isAuthenticated, authPending: pending } = this.props;
const submitClasses = classNames( 'button', 'button-primary', { pending } );
```

roundhill:
@dmsnell / @rodrigoi updated and squashed. Cheers

dmsnell:
just do it @roundhill

roundhill:
> just do it @roundhill

http://img2.etsystatic.com/004/0/6956076/il_fullxfull.369240182_fuxt.jpg

PR #252, opened 2016-03-25
Updating to use npm hosted simperium package

beaucollins:
updates the bucket store API to use the isIndexing parameter.

See Simperium/node-simperium#30

### DIFF
--- a/lib/simperium/bucket-store.js
+++ b/lib/simperium/bucket-store.js
@@ -67,7 +67,7 @@ BucketStore.prototype.get = function(id, callback) {
 	});
 };
 
-BucketStore.prototype.update = function(id, data, callback) {
+BucketStore.prototype.update = function(id, data, isIndexing, callback) {
 	var beforeIndex = this.beforeIndex || function() { return arguments[0]; },
 			self = this;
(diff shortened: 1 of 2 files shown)


lib/simperium/bucket-store.js
```diff
@@ -67,7 +67,7 @@ BucketStore.prototype.get = function(id, callback) {
 	});
 };
 
-BucketStore.prototype.update = function(id, data, callback) {
+BucketStore.prototype.update = function(id, data, isIndexing, callback) {
```
dmsnell: I didn't see any changes to where we actually call this function. I guess it's just good practice to declare it to match the library call?

dmsnell:
@beaucollins does this address any errors or problems we have been experiencing?

lib/simperium/bucket-store.js
beaucollins: This method is called by the Simperium bucket object. I was sincing errors in the console that `callback` was not a function.

dmsnell:
@beaucollins probably good after conflicts are resolved

beaucollins:
I rebased and deployed to http://automattic.github.io/simplenote-electron/

PR #257, opened 2016-03-27
Refactor settings app state, enable two-way communication to menus, add view menu items

### LINKED ISSUE #236: View menu: inconsistent menu names
- `Sort Order` should be `Sort Type`
- Sort Order is missing
- Note Display is missing
- Sort Type options should be 'Last modified`, `Last created`, and `Alphabetical`

Mac OS X 
1.0.1

screen shot 2016-03-18 at 10 33 28 am

### LINKED ISSUE #239: Display Settings: clicking on the row is not responsive
Mac OS X 1.0.1
and Windows 10

Clicking on the radio buttons under Settings > Display is fast and responsive, but clicking on the row is not. Seems to affect mainly the first row for some reason, "Last modified".

sn error2

dmsnell:
The existing app store and reducers have been pretty complicated and
confusing for me to figure out. They are definitely inspired by a Redux
model, but have taken the things that are great about Redux and spoiled
them: action creators that are just simple objects, stateless reducers,
composability.

This bigger PR is attempting to replace all of the custom code with
simpler reducers and a more proper use of the `connect()` function to
prevent having to pass so many props down the component chain.

Resolves #239 wherein a typo was breaking the label-input connection. Previously, clicking on the row/label for "sort by last modified" didn't selected the radio button. This typo evidences why having large blocks of repetitive text can lead to bugs, since it was very hard to see the typo due to that mess.

<del>Eases the path for #236 (see comments in the issue for why it is troublesome).</del>
Resolves #236


dmsnell:
Please note that I have removed the storage of the app settings in `localStorage` temporarily while restructuring this code. A more ideal way to handle that caching is through a Redux store enhancer, or middleware. Once this patch is stable and ready to go, it will have that component in `store.js`

See the [redux-localstorage](https://github.com/elgerlambert/redux-localstorage) module for inspiration.

dmsnell:
This was one of those…

lib/tabbed-dialog.jsx
```diff
@@ -30,8 +30,8 @@ export default React.createClass( {
 			<Dialog className="settings" {...dialog}>
```
rodrigoi: `Dialog` has a required `key` prop.

lib/tabbed-dialog.jsx
```diff
@@ -30,8 +30,8 @@ export default React.createClass( {
...
 				<nav className="dialog-tabs theme-color-border">
 					<ul>
-						{tabs.map( tab =>
-							<li key={tab}>
+						{tabs.map( ( tab, key ) =>
+							<li key={key}>
```
rodrigoi: there are a bunch of spaces before and after things missing on this file.

desktop/app.js
```diff
@@ -49,7 +52,11 @@ module.exports = function main() {
 		}
 
 		// Uncomment me to debug in the electron window
-		// mainWindow.openDevTools();
+		mainWindow.openDevTools();
```
rodrigoi: jus a reminder that this should be commented out before merging :wink:

lib/tabbed-dialog.jsx
dmsnell: Yeah I wasn't sure how far to track this down. I believe that the problem existed before this branch, and it should be getting the key from that `dialog` prop. I can continue to dig in before we merge. Thanks!

lib/tabbed-dialog.jsx
dmsnell: This is related to the comment above. I came in to fix the key issue here since it was relatively easy. It seems like most of the files follow the ideology that spaces are expensive and should be avoided.

lib/tabbed-dialog.jsx
dmsnell: > Dialog has a required key prop.

Required `key` prop has been removed from `Dialog`

:wink:

desktop/app.js
dmsnell: gone

roundhill:
A wild mystery `0` has appeared!

screen shot 2016-03-28 at 11 13 30 am

roundhill:
The note info toggles don't appear to be updating when clicked:

1. Open the note info panel
2. Click on 'Pin' or 'Markdown' toggles
3. State seems to update but the toggle does not change.

dmsnell:
> A wild mystery 0 has appeared!

Wild mystery 0 has gone back into hiding! (in abba943)

dmsnell:
> The note info toggles don't appear to be updating when clicked:

The panel was getting its props from within `props.appState` and even though the items inside that object changed, as a JavaScript object, I think the app was seeing no changes since the reference to that object was the same.

Not sure why this PR is bringing this up, but I fixed it by pushing `selectedNote` and `notes` into `this.props` and not just `this.props.appState`, meaning that the app should now automatically re-render when it sees those change.

roundhill:
Working well. Let's :shipit:

PR #260, opened 2016-03-27
Sort/filter notes in `connect()` on state changes

### LINKED ISSUE #144: Sorting doesn't work in Firefox

dmsnell:
Resolves #144 - sorting broken in firefox
Replaces #226 - manually sort notes
Depends on #257 - app state refactoring

This PR pulls in the sorting and filtering logic of the app into the `connect()` function that ties the app into the Redux app state. The principle is that if the notes in the app state don't change, we won't need sort and filter agin. I haven't checked to make sure there's not a performance gap here, but I did test it to make sure this didn't break things and it worked fine in a large test account.

Before merging:
 - [ ] Merge #257 
 - [ ] Take out or fix the sort order on notes exiting `indexdb` for performance


dmsnell:
Killing this PR because the rebase is too awful and it's probably not an important bug to fix since we're not releasing a web version.

PR #276, opened 2016-04-01
Making the source list resizable

beaucollins:
I'm not confident I'm using the correct reducer here. It might be `appState` where I want to stash this info.

Other things to consider:
- minimum and maximum widths (percentage of window?)
- after certain minimum threshold it just hides the sidebar?

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -70,7 +70,10 @@ function mapDispatchToProps( dispatch, { noteBucket } ) {
 			'increaseFontSize',
 			'resetFontSize',
 			'setNoteDisplay',
-			'toggleMarkdown'
+			'toggleMarkdown',
+			'setSourceListWidth',
+			'setSourceListResizing',
+			'setSourceListCanResize'
 		] ), dispatch ),
 		setSortType: thenReloadNotes( settingsActions.setSortType ),
 		toggleSortOrder: thenReloadNotes( settingsActions.toggleSortOrder )
@@ -83,6 +86,24 @@ const isElectron = ( () => {
 
 	return () => foundElectron;
 } )();
+
+const when = ( condition ) => ( action, fallback = noop ) => ( ... args ) => (
+	condition( ... args ) ? action( ... args ) : fallback( ... args )
+)
+
+const preventDefaultThen = ( fn ) => ( e ) => {
+	e.preventDefault()
+	return fn( e )
+}
+
+const whenCanResize = when( ( { canResize } ) => canResize );
+const whenIsResizing = when( ( { isResizing } ) => isResizing );
+
+const detectCanResize = ( { onResizable, width } ) => ( e ) => onResizable( Math.abs( e.clientX - width ) < 3 )
+
+const makeResizable = whenIsResizing( ( { onUpdateWidth } ) => ( e ) => onUpdateWidth( e.clientX ), detectCanResize )
+const startResizing = whenCanResize( ( { onStartResizing } ) => preventDefaultThen( () => onStartResizing() ) )
+const stopResizing = whenIsResizing( ( { onStopResizing } ) => preventDefaultThen( onStopResizing ) )
 
 export const App = connect( mapStateToProps, mapDispatchToProps )( React.createClass( {
 
@@ -452,11 +473,16 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			'touch-enabled': ( 'ontouchstart' in document.body ),
 		} );
 
+		const { sourceList } = this.props.settings;
+		const { canResize, isResizing } = sourceList;
+
 		const mainClasses = classNames( 'simplenote-app', {
 			'note-open': selectedNote,
 			'note-info-open': state.showNoteInfo,
 			'navigation-open': state.showNavigation,
-			'is-electron': isElectron()
+			'is-electron': isElectron(),
+			'can-resize': canResize,
+			'is-resizing': isResizing
 		} );
 
 		return (
@@ -473,7 +499,10 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 					</ContextMenu>
 				}
 				{ state.authorized ?
-						<div className={mainClasses}>
+						<div className={mainClasses}
+							onMouseMove={ makeResizable( { isResizing, width: sourceList.width, onResizable: this.props.setSourceListCanResize, onUpdateWidth: this.props.setSourceListWidth } ) }
+							onMouseDown={ startResizing( { canResize, onStartResizing: this.props.setSourceListResizing } ) }
+							onMouseUp={ stopResizing( { isResizing, onStopResizing: () => this.props.setSourceListResizing( false ) } ) } >
 							<NavigationBar
 								onSelectAllNotes={() => this.props.actions.selectAllNotes() }
 								onSelectTrash={() => this.props.actions.selectTrash() }
@@ -489,7 +518,10 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 								selectedTag={state.tag}
 								tags={state.tags}
 								onOutsideClick={this.onToolbarOutsideClick} />
-							<div className="source-list theme-color-bg theme-color-fg">
+							<div className="source-list theme-color-bg theme-color-fg" ref={ ( ref ) => {
+								if ( !ref ) return;
+								ref.style.flexBasis = sourceList.width + 'px';
+							} }>
 								<div className="search-bar theme-color-border">
 									<button className="button button-borderless" onClick={() => this.props.actions.toggleNavigation() }>
 										<TagsIcon />
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -99,7 +99,11 @@ const preventDefaultThen = ( fn ) => ( e ) => {
 const whenCanResize = when( ( { canResize } ) => canResize );
 const whenIsResizing = when( ( { isResizing } ) => isResizing );
 
-const detectCanResize = ( { onResizable, width } ) => ( e ) => onResizable( Math.abs( e.clientX - width ) < 3 )
+const detectCanResize = ( { onResizable, sourceListSelector } ) => ( e ) => {
+	const node = document.querySelector( sourceListSelector )
+	if ( !node ) return;
+	onResizable( e.clientX - Math.abs( node.getBoundingClientRect().right ) < 3 )
+}
 
 const makeResizable = whenIsResizing( ( { onUpdateWidth } ) => ( e ) => onUpdateWidth( e.clientX ), detectCanResize )
 const startResizing = whenCanResize( ( { onStartResizing } ) => preventDefaultThen( () => onStartResizing() ) )
@@ -500,7 +504,7 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 				}
 				{ state.authorized ?
 						<div className={mainClasses}
-							onMouseMove={ makeResizable( { isResizing, width: sourceList.width, onResizable: this.props.setSourceListCanResize, onUpdateWidth: this.props.setSourceListWidth } ) }
+							onMouseMove={ makeResizable( { isResizing, sourceListSelector: '.source-list', onResizable: this.props.setSourceListCanResize, onUpdateWidth: this.props.setSourceListWidth } ) }
 							onMouseDown={ startResizing( { canResize, onStartResizing: this.props.setSourceListResizing } ) }
 							onMouseUp={ stopResizing( { isResizing, onStopResizing: () => this.props.setSourceListResizing( false ) } ) } >
 							<NavigationBar
--- a/lib/flux/settings-reducer.js
+++ b/lib/flux/settings-reducer.js
@@ -49,7 +49,7 @@ const markdownEnabled = ( state = false, action ) => {
 	return action.markdownEnabled;
 };
 
-const sourceList = ( state = { width: 330, isResizing: false, canResize: false, node: null }, { type, width, canResize, isResizing } ) => {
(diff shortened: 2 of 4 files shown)


lib/app.jsx
```diff
@@ -84,6 +87,28 @@ const isElectron = ( () => {
...
+const preventDefaultThen = ( fn ) => ( e ) => {
+	e.preventDefault()
+	return fn( e )
+}
+
+const whenCanResize = when( ( { canResize } ) => canResize );
+const whenIsResizing = when( ( { isResizing } ) => isResizing );
+
+const detectCanResize = ( { onResizable, sourceListSelector } ) => ( e ) => {
+	const node = document.querySelector( sourceListSelector )
```
beaucollins: @dmsnell do you know of a better way to get a reference to a node in "state-less" way? I'm trying to use a reference to the `.source-list` node to figure out where the right edge of the node is in order to activate resizing.

lib/app.jsx
```diff
@@ -84,6 +87,28 @@ const isElectron = ( () => {
 	return () => foundElectron;
 } )();
 
+const when = ( condition ) => ( action, fallback = noop ) => ( ... args ) => (
+	condition( ... args ) ? action( ... args ) : fallback( ... args )
+)
```
dmsnell: I couldn't find anywhere here that we use `fallback`, so I wonder if we couldn't leave it out and make this simpler.

```js
const when = condition => action => ( ...args ) => condition( ...args ) ? action : null;
```

I'm going to add some comments about this below, too, because I'm not sure if we need it.

lib/app.jsx
dmsnell: With **React** we can pass a function in on a components `ref` prop and use that to eliminate the need for these things.

```js
updateDescription() {
	this.displayParagraph.innerHTML = 'Bazinga!;
}

render() {
	return (
		<div>
			<h1>Test</h1>
			<p ref={ node => ( this.displayParagraph = node ) }>Gobble gobble!</p>
		</div>
	);
}
```

I think we can eliminate the checks if `this.displayParagraph` exists (unless we might programmatically remove it) because it should be initialized when the component mounts. Maybe it won't be there at `componentWillMount`

Hope this helps!

lib/app.jsx
```diff
@@ -84,6 +87,28 @@ const isElectron = ( () => {
...
+
+const detectCanResize = ( { onResizable, sourceListSelector } ) => ( e ) => {
+	const node = document.querySelector( sourceListSelector )
+	if ( !node ) return;
+	onResizable( e.clientX - Math.abs( node.getBoundingClientRect().right ) < 3 )
+}
+
+const makeResizable = whenIsResizing( ( { onUpdateWidth } ) => ( e ) => onUpdateWidth( e.clientX ), detectCanResize )
+const startResizing = whenCanResize( ( { onStartResizing } ) => preventDefaultThen( () => onStartResizing() ) )
+const stopResizing = whenIsResizing( ( { onStopResizing } ) => preventDefaultThen( onStopResizing ) )
```
dmsnell: I feel like we should just include [Ramda](http://ramdajs.com/0.20.1/docs/) in this project since its helpers would be awesome in places like this. I find these sequences difficult to read because although they feel like a chain or sequence of functions, they are actually passing a bunch of functions through the parameters and I've had to go over it a few times before realizing what exactly they are doing.

```js
import {
	compose,
	ifElse,
	invoker,
	prop,
	tap,
	when
} from 'ramda';

const preventDefault = tap( invoker( 'preventDefault' ) );

const makeResizable = ifElse(
	prop( 'isResizing' ), 
		( { onStartResizing } ) => compose( onStartResizing, prop( 'clientX' ) ),
		detectCanResize
);

const startResizing = when( prop( 'canResize' ), compose( prop( 'onStartResizing' ), preventDefault ) );

const stopResizing = when( prop( 'isResizing' ), compose( prop( 'onStopResizing' ), preventDefault ) );
```

That's an idea using Ramda. I like what you are going for with the functional style, but I think it would be simpler/clearer without the middle step.

lib/app.jsx
```diff
@@ -489,7 +522,10 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 								selectedTag={state.tag}
 								tags={state.tags}
 								onOutsideClick={this.onToolbarOutsideClick} />
-							<div className="source-list theme-color-bg theme-color-fg">
+							<div className="source-list theme-color-bg theme-color-fg" ref={ ( ref ) => {
+								if ( !ref ) return;
+								ref.style.flexBasis = sourceList.width + 'px';
```
dmsnell: Just realized you are already using this form. If you want to hold the reference, just save it on `this`

dmsnell:
@beaucollins `settings` seems to be the right place since it's not related to the note account in any way - in other words, it's a matter that only relates to this particular client, right?

dmsnell:
@beaucollins a question on terminology - source list is the note list on the left, isn't it?

lib/flux/settings-reducer.js
```diff
@@ -49,11 +49,25 @@ const markdownEnabled = ( state = false, action ) => {
 	return action.markdownEnabled;
 };
 
+const sourceList = ( state = { width: 330, isResizing: false, canResize: false }, { type, width, canResize, isResizing } ) => {
```
dmsnell: I'm a bit leery of the length of this line already. I _like_ how the action is restructured, but I think, "what if we add one more property?" when do we split it to a new line?

In reducers like these, I often prefer to keep `action` in the props then restructure inside. As far as the initial state goes I have less of an opinion, but I'm going to go ahead and give a counter-example.

```js
const initialSourceList = {
	width: 330,
	isResizing: false,
	canResize: false
}
const sourceList( state = initialSourceList, action ) {
	// now if we get more, we just start putting one item
	// per line - we could do this already too I suppose
	const { type, width, canResize, isResizing } = action;

	// or for more obvious diffs
	const {
		canResize,
		isResizing,
		type,
		width
	} = action;

	…
}
```

lib/flux/settings-reducer.js
```diff
@@ -49,11 +49,25 @@ const markdownEnabled = ( state = false, action ) => {
...
+		case 'setSourceListCanResize':
+			return Object.assign( {}, state, { canResize } )
+		case 'setSourceListResizing':
+			return Object.assign( {}, state, { isResizing } )
+		case 'setSourceListWidth':
+			return Object.assign( {}, state, { width } );
+		default:
+			return state;
+	}
+};
```
dmsnell: This could probably be its own composed reducer, since we're starting to bundle together different logic. I realize that duplicate code is becoming a problem here, so maybe we need a helper function.

```js
import {
	equals,
	ifElse,
	prop
} from 'ramda';

const simpleState = ( type, propName, default ) => 
	( state = default, action ) =>
		ifElse( equals( type, prop( 'type', action ) ), prop( propName, action ), state );

const sourceList = combineReducers( {
	canResize: simpleState( 'setSourceListCanResize', 'canResize', false ),
	isResizing: simpleState( 'setSourceListResizing', 'isResizing', false ),
	width: simpleState( 'setSourceListWidth', 'width', 330 )
} );
```

drw158:
This is really cool. 

The main design challenge is determining when we trigger the 1 column styles we have in place (hide the sidebar). Nylas has some complex panel manipulation going on; maybe we can learn from them. https://github.com/nylas/N1

N1 has resizable panels, but if you resize the _window_ small enough, it overrides the user's setting and gradually decreases the panel width until the minimum _widow_ width is reached.

The mac app is a good starting point for min and max width — 240px / 380px.

Percentage of window would be cool, but I think users would get frustrated that the width would change when they adjust the window size.

Might be nice to use the `col-resize` cursor

lib/app.jsx
```diff
@@ -473,7 +503,10 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 					</ContextMenu>
 				}
 				{ state.authorized ?
-						<div className={mainClasses}>
+						<div className={mainClasses}
+							onMouseMove={ makeResizable( { isResizing, sourceListSelector: '.source-list', onResizable: this.props.setSourceListCanResize, onUpdateWidth: this.props.setSourceListWidth } ) }
+							onMouseDown={ startResizing( { canResize, onStartResizing: this.props.setSourceListResizing } ) }
+							onMouseUp={ stopResizing( { isResizing, onStopResizing: () => this.props.setSourceListResizing( false ) } ) } >
```
dmsnell: Looks like a great opportunity for a reactive observable, which could potentially allow us to remove `isResizing` from our store. I'm a bit confused as to why we need `isResizing` in the first place, as it doesn't appear to impact styling at all. Is this only to detect the mouse movements? If so, maybe it should remain contained here in this component.

```js
const Draggable = React.createClass( {
	componentWillMount() {
		this.dragging = false;
	},

	startDragging( e ) {
		e.preventDefault();

		this.dragging = true;
	},

	stopDragging( e ) {
		e.preventDefault();

		this.dragging = false;
	},

	render() {
		const { children, onDrag } = this.props;

		return React.createElement( children, {
			onMouseDown: this.startDragging,
			onMouseUp: this.stopDragging,
			onMouseMove: onDrag
		} ;
	}
} );

<Draggable onDrag={ e => onUpdateWidth( e.clientX ) }>
	<div className={ mainClasses }>
		<NavigationBar …>
		…
	</div>
</Draggable>
```

dmsnell:
Thanks for tackling this @beaucollins! I left a few comments, feel free to ping me to discuss them with me.

drw158:
Also, the resize icon should only show up when the cursor is hovering on the border.

lib/app.jsx
beaucollins: `canResize` means their mouse is over the area where they can initiate dragging with a click.

`isResizing` means they are currently holding the mouse button down.

The `is-resizing` CSS class isn't being used so can be removed.

beaucollins:
@dmsnell yes, a source list is a common name for the list of content that accompanies a detail view. It was also the class name already used (`.source-list`).

beaucollins:
@drw158 much of the collapsing and size limits can be controlled via CSS. Currently I'm using `flex-basis` to determine the width of the column, so `max-width` and `min-width` combined with media queries might be able to take care of all that logic.

lib/app.jsx
beaucollins: This is mostly a product of avoiding the introduction of component state.

lib/app.jsx
beaucollins: Ok, the problem I'm having then is that the `ref` doesn't exist when setting the `onMouseMove` listener. But attempting to trigger a re-render in the `ref` handler in order to provide the `ref` to the `onMouseMove` listener seems to trigger a `render` loop.

dmsnell:
@beaucollins are you still interesting in pushing this through? it looks like we might need some updates before that happens.

lib/app.jsx
dmsnell: Let's create a method on the class to set the ref like this…

```js
storeSourceList( ref ) => this.sourceList = ref;

<div className="source-list …" ref={ this.storeSourceList }>…</div>
```

and then the mouse move handler can directly reference `this.sourceList`. if it doesn't exist yet then we can safely ignore the move event because the underlying node won't exist yet either. the handler can safely reference `this.sourceList`

also, wtf is going on with setting the stye `ref.style.flexBasis` _inside of the `ref` function_?

lib/flux/settings-reducer.js
dmsnell: even better yet, why store complex state in a single reducer when we have `combineReducers()`?

```js
const width = ( state = 330, { type, width } ) =>
	type === SOURCE_LIST_WIDTH_SET && state !== width
		? width
		: state;

const isResizing = ( state = false, { type, isResizing } ) =>
	type === SOURCE_LIST_IS_RESIZING_SET && state !== isResizing
		? isResizing
		: state;

const canResize = ( state = false, { type, canResize } ) =>
	type === SOURCE_LIST_CAN_RESIZE_SET && state !== canResize
		? canResize
		: state;

export default combineReducers( {
	canResize,
	isResizing,
	width,
} );
```

This seems much more testable and Redux-y

lib/app.jsx
dmsnell: @beaucollins I'm all for removing component state, but I'm not sure `isResizing` belongs in the global app state either. How do you feel about the kind of wrapper component I provided above?

If the app is rendering the data contained in the state tree, then I'm not sure anything beyond the required data should be there. If we persisted the entire state and reloaded it when `isResizing = true` then wouldn't we end up in a wrong state?

dmsnell:
@beaucollins should we move with this or close it out?

dmsnell:
Also @beaucollins I have taken the liberty of rebasing against **master** but in the process I ran into some oddities. The rebased and squashed commit has landed in #428

beaucollins:
> a note about style, I'd like to see us use semicolon's as if they were required. personal preferences aside, I agree with Crockford that not including them is a bug: it's a behavior which can lead to hard-to-track down errors in the code. let's leave those in if you don't mind.

I'm happy to abide by any rules configured in `.eslintrc`.

dmsnell:
> I'm happy to abide by any rules configured in .eslintrc. I'd rather not see them but won't put up a fight. If eslint --fix adds them for me, even better.

I can look into adding that. You may notice that my personal projects tend to leave them out, but they are an unreasonable risk in team projects, particularly with `return`

```js
return
	<div>
		<p>I never appear because this is dead code</p>
	</div>;
```

beaucollins:
I tend to always wrap my `jsx` in parens. Something makes me uneasy to have those stabby  angle braces floating free with nothing to stop them.

```jsx
return (
    <div>
        <p>I never appear because this is dead code</p>
    </div>
);
```
I feel safer.

dmsnell:
> I tend to always wrap my jsx in parens

yep. so do I. but that that doesn't prevent it from happening otherwise 😢

dmsnell:
Closing for rebased version in #428

PR #281, opened 2016-04-03
Make linux builder more robust

envygeeks:
This makes building on Linux a bit nicer by making so that we can use bundler if we are clean Rubyist's, so that we have a generic name, and more categories for more generic searching, and so that we can launch simplenote from our vterm.

### DIFF
--- a/resources/build-scripts/package-linux.js
+++ b/resources/build-scripts/package-linux.js
@@ -44,6 +44,7 @@ cp.execSync( "cp -r release/Simplenote-linux-x64 release/tmp/usr/share/simplenot
 cp.execSync( "mv release/tmp/usr/share/simplenote/Simplenote release/tmp/usr/share/simplenote/simplenote", onErrorBail );	// rename binary to wpcom
 cp.execSync( "cp resources/linux/simplenote.desktop release/tmp/usr/share/applications/", onErrorBail );
 cp.execSync( "cp resources/images/icon_256x256.png release/tmp/usr/share/pixmaps/simplenote.png", onErrorBail );
+cp.execSync( "mkdir -p release/tmp/usr/bin && cd release/tmp/usr/bin && ln -sf /usr/share/simplenote/simplenote" );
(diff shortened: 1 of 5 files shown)


resources/build-scripts/package-linux.js
```diff
@@ -44,6 +44,7 @@ cp.execSync( "cp -r release/Simplenote-linux-x64 release/tmp/usr/share/simplenot
 cp.execSync( "mv release/tmp/usr/share/simplenote/Simplenote release/tmp/usr/share/simplenote/simplenote", onErrorBail );	// rename binary to wpcom
 cp.execSync( "cp resources/linux/simplenote.desktop release/tmp/usr/share/applications/", onErrorBail );
 cp.execSync( "cp resources/images/icon_256x256.png release/tmp/usr/share/pixmaps/simplenote.png", onErrorBail );
+cp.execSync( "mkdir -p release/tmp/usr/bin && cd release/tmp/usr/bin && ln -s ../share/simplenote/simplenote" );
```
lemuelroberto: @envygeeks, thank you for the this change that also solves a $PATH's problem on XUbuntu 14.04, while it doesn't have '/usr/share' in it's $PATH.

But I'd like to suggest you to point the link to it's real path target ('/usr/share/simplenote/simplenote') instead of pointing it to '../share/simplenote/simplenote'.

You could do something like this:

```
cp.execSync( "mkdir -p release/tmp/usr/bin", onErrorBail);

cp.execSync( "ln -s /usr/share/simplenote/simplenote release/tmp/usr/bin", onErrorBail);
```

envygeeks: Sure, I'm happy to make that change, we'll just have to force it so that build servers don't break.

envygeeks: @lemuelroberto Done.

envygeeks:
Closing.  Project has bad maintainership.

Issue #264, opened 2016-03-30
Open links in default browser


hindsholm:
When clicking on a link in a note, the page opens in a new window with a simple webview (or similar).
It would be very nice to have the option of opening links in the default browser instead.

drw158:
Interesting, that is unexpected. I'll file as a bug. Thanks!

hindsholm:
I didn't know that it was a bug, I thought it was intentional :)
But glad to hear that it will be fixed...

drw158:
Documentation for the solution to this: https://github.com/atom/electron/blob/master/docs/api/shell.md#shellopenexternalurl

roundhill:
Closed via #283

PR #283, opened 2016-04-03
Open link previews on external browser instead of webview.

efegurkan:
Links should be opened on user's default browser instead of a new
electron window. This fixes issue #264.

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -2,7 +2,8 @@ import React, { PropTypes } from 'react';
 import marked from 'marked';
 import Textarea from 'react-textarea-autosize';
 import { noop, get, debounce } from 'lodash';
-import analytics from './analytics'
+import analytics from './analytics';
+import { viewExternalUrl } from './utils/url-utils';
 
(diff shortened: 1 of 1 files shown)


lib/note-detail.jsx
```diff
@@ -3,6 +3,7 @@ import marked from 'marked';
 import Textarea from 'react-textarea-autosize';
 import { noop, get, debounce } from 'lodash';
 import analytics from './analytics'
```
roundhill: First off, thanks for the PR! This makes sense to me to add. Not your fault, but would you mind adding a semicolon to the previous line? I know it's not needed but it looks off with the new import you added.

efegurkan:
Hi @roundhill

Added semicolon as requested. Making linter to fail in these cases for these kind of issues might be a good idea I believe, don't know what you guys are thinking about that. Anyways that's out of the scope of this conversation.

Cheers.

roundhill:
> Making linter to fail in these cases

Yes, that would be good. Thanks for fixing that.

roundhill:
:shipit:

Issue #248, opened 2016-03-23
Pressing enter doesn't do anything when trying to log in


drw158:

dmsnell:
I saw this behavior too and it bothered me, but I'm terrible at creating issues. Thanks @drw158

Sushubh:
Noticed this too!

PR #284, opened 2016-04-03
Pressing enter logs in user now.

efegurkan:
This fixed issue #248.

Enter key bindings were missing on auth.jsx. I added onKeyDown props to
both of the inputs to handle the case of if user enters password, or case of
typing invalid credentials.

In onLogin method added event type check to determine if the event is
keypress or mouseclick. If it is a keypress and it is not Enter key,
just keeps continue regular event handling.
I placed check before preventDefault to make sure the event handling works as
expected on other keypresses than Enter key.

### DIFF
--- a/lib/auth.jsx
+++ b/lib/auth.jsx
@@ -30,13 +30,23 @@ export default React.createClass( {
 						<label className="login-field theme-color-border" htmlFor="login-field-username">
 							<span className="login-field-label">Email</span>
 							<span className="login-field-control">
-								<input ref={ ref => this.usernameInput = ref } id="login-field-username" type="email" />
+								<input
+									ref={ ref => this.usernameInput = ref }
+									id="login-field-username"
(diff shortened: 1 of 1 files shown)


efegurkan:
Sorry for doing that. But I believe this project is low prioritized and no one has time, no ?

dmsnell:
@efegurkan please accept our apologies for letting this slide. thanks for your contribution - it looks great!

we _are_ hurrying along and have a todo list much bigger than we are currently able to complete. it _was_, however important to us to get this out there for people who have been waiting on it and to make it open source so that we could all benefit from the various contributions.

thanks again!

efegurkan:
@dmsnell Yeah its fine, sorry for whining actually 😞.  Thanks.

dmsnell:
No worries @efegurkan! We're so excited that people love Simplenote enough to contribute to it

PR #302, opened 2016-04-20
Silently fail if the updater check fails instead of showing a dialog.

### LINKED ISSUE #295: Error when offline - Windows
When the app is offline in Windows an error message displays intermittently even though it's still possible to use the app:

capture

If it's possible to use the app in offline mode then it would be good to remove this error notification.

roundhill:
Fixes #295 

cc @rodrigoi, this look ok?


rodrigoi:
:shipit:

PR #348, opened 2016-07-26
Removed unused popover dependency

bostrt:
Use of react-popover was discontinued in b7bb4a2665398503e39ca36fb7d7b1408178892b.


dmsnell:
@bostrt please accept my apology for allowing this to collect dust for so long! thanks for the PR and nice catch - this shouldn't have slipped through review way back when. we're pleased to receive your contribution!

dmsnell:
cc: @roundhill

PR #356, opened 2016-08-08
Update Electron build tools

rodrigoi:
Updates version numbers of the electron build tools to the latest versions.

Also clears $APP_DATA on windows in between installs to prevent white screen after update.

cc: @roundhill


roundhill:
Seems ok, I packaged it up fine along with #358. Shall we merge both?

I do notice a few deprecation errors that we should fix sometime.

PR #358, opened 2016-08-12
Update React and Electron packages

roundhill:
Updates `react` and `electron-prebuilt` packages to the latest and greatest.

Fixes imports based on latest electron.

cc @rodrigoi


dmsnell:
I didn't see anything out of the ordinary here, @roundhill, but I also didn't test it.

rodrigoi:
Tried it and I didn't find any regressions. The build works with versions of node up to `5.12.0` and npm up to `3.8.6`. I recommend we add `engines` to *package.json*:

```json
engines: {
  "node": "5.12.0",
  "npm": "3.8.6"
}
```

other than that, this looks awesome. :shipit:

Issue #368, opened 2016-08-26
Link detection like in iOS or Mac app


nfcampos:
Hi,

This looks great!
I'd like to see link detection (like in the mac and ios apps) added. Is that something you'd want?

From a cursory look at the code for the `NodeDetail` component it seems that doing this would require replacing the `textarea` element with a `<div contenteditable />` (maybe using https://github.com/facebook/draft-js) because text inside a `textarea` is not style-able in any way (as  far as i know). Is this something you'd be ok with? If so I can take a stab at this

Nuno

Also Reported in:
 - #985556-zen
 - #1118368-zen
 - #1194082-zen

roundhill:
Yes, please! I think I tried `contenteditable` for a while but it gave me problems which of course I don't recall. Go for it!

nfcampos:
Cool, I'll open a PR when I have something!
In the meantime, would you accept a PR updating this to use babel 6? the version you're using currently doesn't work with node 6

roundhill:
Maybe, I think we'd need to test our desktop app packager to see if it works with node 6. cc @rodrigoi

nfcampos:
you could keep using node 5, updating babel would simply allow me (and everyone else) to develop using node 6, but it'd still be totally compatible with node 5, unless I misunderstood what you wrote?

nfcampos:
but honestly I can just use node 5 for this, it doesnt realy matter to me

dmsnell:
@nfcampos it's probably best to work within node 5 for this PR. an update would be an entire PR by itself is my guess, and probably also something that wouldn't happen quickly enough to be in there before this work gets merged

nfcampos:
ok, no problem

nfcampos:
opened PR replacing textarea with draft.js editor :) the link detection, styling will follow

memeplex:
The web app neither shows urls as links.

tmchartrand:
this would be great on Linux too!

katinthehatsite:
A user in 985556-zen also asked for the feature to be added on Windows and reported links not being detected.

ChrissiePollock:
Requested again in #1118368-zen.

serabi:
Requested . again in #1194082-zen

laoyuan:
need this!

crankygeologist:
I would likely switch from the native Mac version to the web app or Electron flavor if they had if link detection.

roman-yagodin:
Vote for it!

[...]

PR #370, opened 2016-08-26
replaced textarea-based note editor with Draft.js

nfcampos:
- replaced textarea note editor with Draft.js
- ~~so far this PR should have no user-facing change at all (if it has then I did something wrong), this is the stepping stone to be able to style links different from regular text (for #368)~~ I've added link highlighting!
- ~~should I do the link detection/styling in this PR or would it be easier to review this first separately?~~

## other changes
- updated versions of react-dom and react-addons-* to match react version
  - generally it's not a great idea to keep these out of sync
- bumped webpack version to get rid of sourcemap warnings in the console on chrome
  - it was impossible to use the console for debugging otherwise, if you'd like me to undo this before merging let me know


nfcampos:
I've added the url decorator now! :)

nfcampos:
(this is apparently not working in safari for now (but works in chrome with no problems as far as I can tell). I'll investigate)

dmsnell:
> not working in safari for now

there will likely be complications in Internet Explorer as well, though thankfully the Electron shell is running the Chrome codebase. `draft-js` depends on the polyfill for things like `String.prototype.startsWith` and I can't remember if we already have that in the app.

dmsnell:
@nfcampos thanks for all this great work. I wonder if it wouldn't be good to split the editor into its own and new component, where all of the editor logic could be split from the note detail component.

lib/note-detail.jsx
```diff
@@ -19,36 +19,61 @@ export default React.createClass( {
...
+		return get( note, 'data.content', '' )
+			!== editorState.getCurrentContent().getPlainText( '\n' )
+	},
+
+	handleEditorStateChange( editorState ) {
+		this.setState( { editorState } );
+		// only queue note save if it was the content that changed
+		if ( this.hasContentChanged( editorState, this.props.note ) ) {
+			this.queueNoteSave();
+		}
```
dmsnell: There are a couple of notes here:

 - since the editor content is immutable, we should be able to detect if there were edits made by comparing `this.state.editorState.getCurrentContent()` to `editorState.getCurrentContent()`
 - let's make sure we are explicit about what sequencing we want here. the call to `this.queueNoteSave()` could actually occur before `setState()` has updated the component state. if we want this to happen in sequence, we can use the second argument to `setState()` and pass in a function to call once the state has updated

lib/note-detail.jsx
```diff
@@ -19,36 +19,61 @@ export default React.createClass( {
 
 	componentWillMount: function() {
 		this.queueNoteSave = debounce( this.saveNote, saveDelay );
-		this.noteEditor = uninitializedNoteEditor;
+		this.focus = () => this.refs.editor.focus();
```
dmsnell: It would be great to include some more context in the name here. Maybe something like `focusEditor`

lib/url-decorator.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { CompositeDecorator } from 'draft-js';
+import urlRegex from 'url-regex';
+import normalizeUrl from 'normalize-url';
+import { omit } from 'lodash';
+import { viewExternalUrl } from './utils/url-utils';
+
+const urlRe = urlRegex();
+const markdownLinkRe = /\(|\]:\s$/m;
```
dmsnell: I'm a bit confused here at what we are wanting to match with this pattern. Could we clarify either through reworking the RegExp or by leaving an explanatory comment?

lib/url-decorator.jsx
```diff
@@ -0,0 +1,51 @@
...
+		component: React.createClass( {
+			onClick: function( event ) {
+				event.preventDefault();
+				viewExternalUrl( normalizeUrl( this.props.decoratedText ) );
+			},
+
+			render: function() {
+				return (
+					<a
+						{...omit( this.props, 'entityKey', 'offsetKey', 'decoratedText' )}
```
dmsnell: this might be a bit permissive. we can probably know up-front which attributes we want to pass, no?

nfcampos:
@dmsnell good point about splitting out the editor into its own component, I'd been thinking about that actually, I'll change it

dmsnell:
> I've added the url decorator now! :)

what should we be seeing with links? could you provide some before/after screenshots? I tried this out but wasn't sure _what_ I should be expecting.

dmsnell:
thanks @nfcampos - this is awesome work and I look forward to taking a second pass after you have had a chance to see my feedback

nfcampos:
screenshot after this PR
screen shot 2016-08-27 at 22 01 00
screenshot before 
screen shot 2016-08-27 at 22 01 51

dmsnell:
thanks for the screenies @nfcampos - I think I missed that because I was trying with a URL like  something.com and I didn't see any decoration.

nfcampos:
@dmsnell yeah I didn't write the url regexp but from my tests it looks like just a domain is not enough, you need the www before it, we can probably change that if you think it's worth it

dmsnell:
@nfcampos probably worth doing 100% before releasing it. we're probably also not in a rush for this, so there is time to figure it out :wink:

nfcampos:
@dmsnell I'll be away from a computer during the next week so I'll probably address your comments only after that, in the meantime if you find any problem or have any other comment leave it here and I'll take a look at everything then

nfcampos:
@dmsnell about splitting the editor into its own component, what name should I give it? the obvious name `NoteEditor` is already taken...

lib/url-decorator.jsx
nfcampos: I'll add a comment and some tests for the regexp which should make it clearer

lib/note-detail.jsx
nfcampos: good point, will change

lib/note-detail.jsx
nfcampos: 1) form looking at the draft.js docs it appears that a `ContentState` contains info about both the contents and the current `SelectionState` of the editor. To avoid triggering a save when all the user changed was the selection, I think we need to compare the actual content which for our purposes (since that's how it is saved) is what's returned from `getPlainText`. Now the question is, do we want to compare against the note in the props, or against the previous `ContentState`?

2) great point about doing this in a setState callback, will change

dmsnell:
> what name should I give it?

There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.

:wink:

You can work out whichever name you think is appropriate. One that immediately comes to my mind would be something like `NoteContentEditor`

lib/url-decorator.jsx
nfcampos: there was actually a mistake in this pattern, I've fixed it now. let me know if the comment is clear enough

lib/url-decorator.jsx
nfcampos: actually there's no reason to pass any props at all I think, so I've just removed that line

nfcampos:
@dmsnell I believe this is ready for a second pass :)

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,68 @@
...
+		const { content } = this.props;
+		const contentState = ContentState.createFromText( content, '\n' );
+		return {
+			editorState: EditorState.createWithContent( contentState, urlDecorator )
+		}
+	},
+
+	handleEditorStateChange( editorState ) {
+		const contentHasChanged = plainTextContent( editorState )
+			!== plainTextContent( this.state.editorState )
```
dmsnell: I'm curious if you have tried comparing the content from the model without converting it. `draft-js` is an immutable-based system and so I would think that we could accomplish this by a simple reference check…

```js
const hasChangedContent = editorState.getCurrentContent() !== this.state.editorState.getCurrentContent();
```

also I have rearranged the variable name to keep the boolean-ish verb out front

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,68 @@
...
+			// if the new content isn't different from what's rendered
+			// don't change it because we'd lose the SelectionState
+			&& plainTextContent( this.state.editorState ) !== nextProps.content
+		) {
+			const contentState = ContentState.createFromText( nextProps.content, '\n' );
+			this.setState( {
+				editorState: EditorState.createWithContent( contentState, urlDecorator )
+			} );
+		}
+	},
```
dmsnell: using some destructuring and renaming here might help keep things clear here

```js
componentWillReceiveProps( { content: oldContent } ) {
	const { content: newContent } = this.props;
	const { editorState } = this.state;

	if ( newContent === oldContent ) {
		return;
	}

	if ( newContent === plainTextContent( editorState ) ) {
		return;
	}

	this.setState( {
		editorState: EditorState.createWithContent(
			ContentState.createFromText( newContent, '\n' ),
			urlDecorator
		)
	} );
}
```

lib/note-content-editor.jsx
nfcampos: I haven't done that but for a different reason — `contentState` (despite its name, this is actually not great naming on the part of `draft-js` I guess, see https://facebook.github.io/draft-js/docs/api-reference-content-state.html#content) contains info about both content and selection state.

lib/note-content-editor.jsx
dmsnell: This method is well-worth considering. What does it mean to have the props change and what does it mean to create a new `EditorState` object. I think that for one, we will lose our undo history. Do we want to be able to undo changes that pour in from outside of this particular client (for example, I make a change on my phone). I'm sure we could find a way to convey that information and use `EditorState.push()` instead with a specific action type, such as `'update-from-simperium'`.

Further, could we map the Simperium deltas into individual editor change sets such as found in the `Modifier` module. I'm sorry that I don't know these things off of the top of my head, such as if we even have access here to the Simperium diff.

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+	focus() {
+		if ( !this.state.editorState.getSelection().getHasFocus() ) {
+			this.refs.editor.focus();
+		}
```
dmsnell: Is there a reason go check if it's not already focussed instead of simply always focusing it?

```js
focus() {
	this.editor.focus();
}
```

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,68 @@
...
+	},
+
+	render() {
+		return (
+			<Editor
+				ref='editor'
```
dmsnell: Let's not introduce new code with this style of `ref` attribute. Instead, let's use the function syntax. The React project strongly suggests this as well (for reasons which can be enumerated)

```js
import { invoke } from 'lodash';

saveEditorRef( ref ) {
	this.editor = ref;
}

focus() {
	invoke( this, 'editor' );
}

<Editor ref={ this.saveEditorRef } … />
```

Notice that we broke a convention of working through `this.setState()` which in this case is appropriate due to the stateful nature of the DOM elements with respect to the React component's _backing instance_.

lib/note-content-editor.jsx
nfcampos: there was a weird issue where the cursor would sometimes jump to the right when you were already focused and the `focus` method was called, this fixed it.

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,68 @@
...
+				spellCheck
+				stripPastedStyles
+				textAlignment='left'
```
dmsnell: isn't this the default value for `textAlignment`? since we're not overwriting the value, maybe leave it out?

nfcampos: you're quite right, I'll remove it

lib/note-detail.jsx
```diff
@@ -107,13 +99,17 @@ export default React.createClass( {
 	},
 
 	renderEditable( divStyle ) {
-		const note = this.props.note;
+		const { content = '' } = this.props.note.data;
```
dmsnell: if we're just destructuring one property _and_ still allowing for a `TypeError` from the nested access, let's just use `_.get` instead…

```js
import { get } from 'lodash';

const content = get( this.props, 'note.data.content', '' );
```

lib/note-detail.jsx
```diff
@@ -107,13 +99,17 @@ export default React.createClass( {
...
 		return (
-			<Textarea ref={ this.initializeNoteEditor } className="note-detail-textarea theme-color-bg theme-color-fg"
-				disabled={ !!( note && note.data.deleted ) }
-				onChange={ this.queueNoteSave }
-				style={ divStyle } />
+			<div
+				className="note-detail-textarea theme-color-bg theme-color-fg"
+				style={ divStyle }>
+				<NoteContentEditor
+					ref='editor'
```
dmsnell: Please see earlier note about `ref` syntax

lib/note-detail.jsx
nfcampos: I did this actually to keep with the `renderMarkdown` method above, which does exactly this
(no chance of a `TypeError` here, see `if ( ! this.isValidNote( note ) ) return;` before it in `render`)

lib/note-detail.jsx
nfcampos: will change

lib/note-content-editor.jsx
nfcampos: you're right, this is well worth considering, another situation that would produce a props change would be the use of the revisions ui to select an older revision, do we want to for any reason to differentiate the action type of this v. getting a new note from the api?

lib/note-content-editor.jsx
nfcampos: will change `ref`

lib/url-decorator.jsx
```diff
@@ -0,0 +1,53 @@
...
+import urlRegex from 'url-regex';
+import normalizeUrl from 'normalize-url';
+import { viewExternalUrl } from './utils/url-utils';
+
+const urlRe = urlRegex();
+/*
+- markdownLinkRe is a regexp to be applied to the substring preceding a matched url, ie. `wholeString.slice(0, wholeString.indexOf(matchedUrl))`
+- it tests whether the matched url is preceded by the charcters that define a link in markdown, ie. `(` or `]: `
+*/
+const markdownLinkRe = /(?:\(|\]:\s+)$/;
```
dmsnell: can we word-wrap these comments so they are shorter lines? also, let's use `//` instead of `/* */`

lib/note-content-editor.jsx
nfcampos: done

lib/note-content-editor.jsx
nfcampos: done

lib/note-detail.jsx
nfcampos: done

lib/url-decorator.jsx
nfcampos: done

lib/url-decorator.jsx
```diff
@@ -0,0 +1,53 @@
...
+export function normalizeUrlsInMarkdownString( str ) {
+	return str.replace( urlRe, ( match, offset ) => {
+		var normalized = normalizeUrl( match );
+		if ( markdownLinkRe.test( str.slice( 0, offset ) ) ) {
+			return normalized;
+		}
+		// replace url with markdown link only if it's not already part of one
+		return `[${match}](${normalized})`;
+	} );
+}
```
dmsnell: This function could get really expensive as we add URLs and as the document length increases. At a minimum I believe that we could rearrange it to scan through the text instead of iterating over the entire document-up-until-the-start-of-the-URL.

Maybe through the use of splitting on the RegExp matches we can build a recursive parser that would eliminate the overhead of starting fresh each time. It's something I'll think about more and follow-up here, but feel free to add your own ideas.

I do not believe that the performance of this function as written will cut it, unfortunately 😢 but we will prevail!

lib/url-decorator.jsx
```diff
@@ -0,0 +1,53 @@
...
+			render: function() {
+				return (
+					<a
+						onClick={this.onClick}
+						href={this.props.decoratedText}>
+						{this.props.children}
+					</a>
+				);
+			}
+		} ),
```
dmsnell: It would be really helpful to have this component created elsewhere. Maybe just in the module scope of this file, but possible in another file altogether. Further, if we simply export the _strategy_ from here it's trivial enough to create the decorator in the note content editor itself, which would have certain advantages for code comparability and organization.

package.json
```diff
@@ -41,33 +41,35 @@
...
     "react-onclickoutside": "4.5.0",
-    "react-textarea-autosize": "3.3.0",
     "react-transform-catch-errors": "1.0.2",
     "react-transform-hmr": "1.0.4",
     "redbox-react": "1.2.2",
     "sass-loader": "3.1.2",
     "simperium": "0.2.4",
     "style-loader": "0.12.4",
-    "webpack": "1.12.14",
+    "webpack": "1.13.2",
```
dmsnell: was there a reason to update webpack? I'm not entirely opposed to doing this, though I find certain value in doing it separately from feature updates if we can help it.

package.json
```diff
@@ -41,33 +41,35 @@
...
     "lodash": "4.6.1",
     "marked": "0.3.5",
     "moment": "2.12.0",
+    "normalize-url": "1.6.0",
     "promise": "7.1.1",
     "react": "15.3.0",
-    "react-addons-pure-render-mixin": "0.14.3",
-    "react-addons-update": "0.14.3",
-    "react-dom": "0.14.3",
+    "react-addons-pure-render-mixin": "15.3.0",
```
dmsnell: hmmm…this reminds me that we should get rid of this soon

dmsnell:
@nfcampos sorry it took me a week to get back with you on this. It's great work. There were two big concerns that popped into my head while reviewing: we appear to be making plenty of full conversions between the internal `draft-js` state and plaintext every time the editor gets updated; we appear to have a very heavy (and untested) URL decorator.

As important as the URL parsing is, we would definitely want a test suite to accompany the function to be explicit about cases we think could break it. Further, it would be doubly-amazing if we could rely less on Regexs for the parsing. I'm still a bit fuzzy on what's happening with the MarkdownRe.

Please let me know how I can be a help with this. Your PR is a super-appreciated contribution to the project and we're happy to have you on board, even if we're a bit slow! :smile: 👍

lib/note-content-editor.jsx
dmsnell: okay. right. I should have realized that

lib/note-content-editor.jsx
dmsnell: it's definitely semantically different to scroll the history vs get updates from the API. big :question: really

lib/note-content-editor.jsx
dmsnell: that sounds suspicious, and it sounds like this "fix" could actually be hiding the problem. did you find others with this problem too?

package.json
nfcampos: see my comment in the first post of this PR, can definitely roll back before this is merged

lib/note-detail.jsx
dmsnell: > no chance of a TypeError here

hehe, I only wish…

lib/url-decorator.jsx
nfcampos: I agree, this version of this function definitely wasn't one of my best creations.

how about this?

```js
const mdRe = new RegExp(
	'(\\(|\\]:\\s+)?(' + urlRe.source + ')',
	urlRe.flags
);

export function normalizeUrlsInMarkdownString( str ) {
	return str.replace( mdRe, function( _, markdownLinkChars, url ) {
		var normalizedUrl = normalizeUrl( url );

		if ( markdownLinkChars !== undefined ) {
			// if the first capturing group of the regex matched
			// then this url is part of a markdown link
			return markdownLinkChars + normalizedUrl;
		}

		// replace the matched url with a markdown link, to make it clickable
		return `[${url}](${normalizedUrl})`;
	} )
}
```

basically this is modifying `urlRe` to also capture either `(` or `]: ` if it appears right before the url. If either one is there it means the url is part of a markdown link so we shouldn't replace it with a markdown link to make it clickable because it is already clickable. makes sense? (this is definitely much faster)

nfcampos:
@dmsnell no problem about the time :)

I actually meant to add a test suite for `normalizeUrlsInMarkdownString` but I didn't find any tests anywhere in the repo so I wasn't sure what test runner you guys wanted,etc. if you point me in the right direction I can definitely write some tests for it

(btw, this isn't actually explicit anywhere so might as well say it, the reason `normalizeUrlsInMarkdownString` exists is so that all urls that are highlighted and clickable in edit mode are also clickable in markdown preview mode)

let me know if the version of `normalizeUrlsInMarkdownString` in my comment above is clearer, I'll try to make it clearer if it isn't there yet, but I don't there's any alternative to using regexs in some way for doing the url parsing

lib/url-decorator.jsx
nfcampos: could you expand on why you'd like this component to be defined elsewhere?
do you mean to use it for something else as well?
if no, this does look to me like the clearest way of creating a decorator for a draft-js `Editor`, define the strategy and its component right next to each other (if not for any other reason because this way it's really clear what `decoratedText`is inside the component, it's exactly what you called the callback with in the strategy).

in terms of organisation I think what could be done is put `note-content-editor` and `url-decorator` in their own folder, `url-decorator` is only really meant to be used by `note-content-editor`. what do you think?

lib/note-detail.jsx
nfcampos: ah you're right, I was looking at the wrong place. should I change this both in `renderMarkdown` and `renderEditable` then?

lib/url-decorator.jsx
dmsnell: I'm still concerned about the strength of this matcher to the URLs. I'm going to think about it some. this iteration is definitely better though.

why isn't our Markdown parser detecting the URLs… shouldn't Markdown automatically linkify all the things?

also, what about `<a href="wordpresscom">Yahoo</a>` - what happens here?

package.json
dmsnell: roger. sorry for not reading the comment

dmsnell:
> I didn't find any tests anywhere in the repo

that's a sad thing. okay, we have had good success with Mocha, chai, enzyme, etc… but that certainly doesn't belong in this PR.

what if you simply created a basic test file we could run on the command line for now…

```js
import normalizr from 'thing';

const tests = [
	// input, expected output
	[ 'this is simplenote.com, fun!', 'this is [simplenote.com](http://simplenote.com), fun!' ],
	[ … ],
	…
];

tests.forEach( ( [ input, expected ] ) => {
	if ( expected !== normalizr( input ) ) {
		throw `Test failure\n\texpected: ${ expected }\n\tactual: ${ normalizr( input ) }`;
	}
} );
```

something like that can go a long way and lead into a more proper testing framework

lib/url-decorator.jsx
nfcampos: > why isn't our Markdown parser detecting the URLs… shouldn't Markdown automatically linkify all the things?

Regular markdown only linkifies things that are defined with the link syntax, either `[abc](www.abc.com)` or `[abc][1]`. Github flavoured markdown is the one that linkifies everything.

> also, what about <a href="wordpresscom">Yahoo</a> - what happens here?

well in that particular case you get shown a link with an invalid href because that doens't match the url regexp. however, if that href was valid this would actually break :( and replace the href with a markdown url...
I'm starting to think it's actually not a great idea to be writing this function, how about using a different `markdown` parser (instead of `marked`) that actually linkifies everything (eg. https://github.com/markdown-it/markdown-it)?

dmsnell: > I'm starting to think it's actually not a great idea to be writing this function, how about using a different markdown parser (instead of marked) that actually linkifies everything

another good :question:

do we _need_ to linkify everything if indeed Markdown wouldn't imply it? why should the Simplenote behavior change the way markdown works?

nfcampos: the only reason I did this to linkify everything is because it was kind of weird to see some urls highlighted in edit mode, switch to preview mode and now they're not urls anymore.
but you do have a point that this is no longer markdown (it's probably one of the myriad flavours of markdown out there).
what do you think is more important *consistency between edit and preview mode* or *sticking to pure markdown output*?
-

package.json
nfcampos: no problem

lib/note-content-editor.jsx
nfcampos: you were right, this was working around the problem described here https://github.com/facebook/draft-js/issues/73#issuecomment-191394559
i've removed it and added the proper fix

lib/note-content-editor.jsx
nfcampos: (a 3rd situation, and this the only one where we really have to create a new `EditorState` is when the Editor starts rendering a different note because the user selected a different one in the sidebar, here we definitely don't want undo support)

lib/url-decorator.jsx
dmsnell: @roundhill @beaucollins @jleandroperez ^^^ :question:

lib/url-decorator.jsx
dmsnell: > could you expand on why you'd like this component to be defined elsewhere?

mainly I would like to see their definitions outside of the object passed in to create the decorator. if they are separate, we can test them (even though we don't have tests yet) and it can be a little easier to find them.

the call to create the decorator isn't that big of a deal either. if we think that something like this is confusion, it's probably more of a naming thing than anything else…

```js
new CompositeDecorator( {
	strategy: findLinks,
	component: Link
} );
```

nfcampos: > mainly I would like to see their definitions outside of the object passed in to create the decorator. 

makes sense, I've changed this

nfcampos:
I'm thinking that maybe this PR could be split into 2, first replacing the editor with Draft.js, second adding the url-decorator. This way other improvements that could be made once the editor is Draft.js-based (eg. auto-continue lists, list items indent/outdent with tab, search term highlighting) don't have to wait for the url-decorator. Thoughts @dmsnell ?

dmsnell:
> this PR could be split into 2

that sounds great @nfcampos !

nfcampos:
ok, so I'll remove the url-decorator from this one. do I open the url-decorator PR only once this one is merged, since the diff won't make sense against the current master?

dmsnell:
> ok, so I'll remove the url-decorator from this one. do I open the url-decorator PR only once this one is merged, since the diff won't make sense against the current master?

you can do it any way that suits you, though I tend to avoid branching PRs off of other PRs because of the merge mess

nfcampos:
alright, I'v removed the url decorator from this PR. This PR is now just replacing the textarea editor with Draft.js. All behaviour should be the same as before. I believe the only open question now is what to do in `componentWillReceiveProps`.

lib/note-content-editor.jsx
dmsnell: @nfcampos could you surround these in parentheses and line up the args a bit? this is just a silly styling thing but I would like to avoid the hanging equality comparison. Alternatively we could use a couple of helper variables and make it short enough on a single line…

```js
const nextContent = plainTextContent( editorState );
const prevContent = plainTextContent( this.state.editorState );

const hasChangedContent = nextContent !== prevContent;
```

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,80 @@
...
+		this.editor = ref
+	},
+
+	handleEditorStateChange( editorState ) {
+		const contentHasChanged = plainTextContent( editorState )
+			!== plainTextContent( this.state.editorState )
+
+		this.setState( { editorState }, () => {
+			if ( contentHasChanged ) {
+				this.props.onChangeContent( plainTextContent( editorState ) );
```
dmsnell: if we use the helper variable above, we can just pass it here too -> `onChangeContent( nextContent )`

also, we could rearrange this here with another variable. it's styling purely; my preference would be something like this…

```js
import { noop } from 'lodash';

const announceChanges = contentHasChanged
	? () => this.props.onChangeContent( nextContent )
	: noop;
```

in fact, if that's the only thing that we use `hasContentChanged` for, we could join the expressions…

```js
const announceChanges = nextContent !== prevContent
	? () => this.props.onChangeContent( nextContent )
	: noop;

this.setState( { editorState }, announceChanges );
```

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,80 @@
...
+	},
+
+	render() {
+		return (
+			<Editor
+				ref={this.saveEditorRef}
+				spellCheck
+				stripPastedStyles
+				onChange={this.handleEditorStateChange}
+				editorState={this.state.editorState} />
```
dmsnell: small nitpick, sorry - can we move the `/>` to the next line aligned with `<Editor`?

nfcampos: sure thing, done

lib/note-content-editor.jsx
nfcampos: done

dmsnell:
@nfcampos I'm going to be testing this as soon as I can and try to run the gauntlet. I have been familiarizing myself with the code and really want to prioritize this, even over my tab/indentation PRs. In the past week or two I have been distracted trying to resolve the sync issues, unfortunately.

You are amazing. :smile:

dmsnell:
@nfcampos how would you feel about converting this into a React "class" with `export class NoteContentEditor extends Component {`…

don't worry about it if you don't want to. we can do that later - I know there's a ton of less-than-exemplary code in here, but I'm simply trying to move properly going forward

nfcampos:
Thanks :)
I can convert this to a class no problem, in fact that's how I usually write react components.

dmsnell:
@nfcampos I also noticed that the url normalization is still here for Markdown. can we pull that from this PR as well so we can make sure we test the editor change in isolation? thanks!

nfcampos:
@dmsnell that was my mistake, I had already removed it a week or two ago but did something stupid with a rebase and lost that commit. Solved now. Sorry about the ridiculous commit history now, it makes next to no sense, should I just squash it all into one commit?

nfcampos:
@dmsnell nevermind, after some careful cherry-picking, we now have a sane commit history again :)

dmsnell:
Looks great @nfcampos - we're all testing it on our ends. Hopefully we will have this out this week! (no guarantees)

So far it seems to work well on my machine.

nfcampos:
(updated the draft.js version to the latest)

package.json
nfcampos: @dmsnell I don't think this is actually being used anywhere...

lib/note-content-editor.jsx
```diff
@@ -0,0 +1,77 @@
...
+		let newEditorState = EditorState.createWithContent(
+			ContentState.createFromText( newContent, '\n' )
+		)
+
+		// avoids weird caret position if content is changed
+		// while the editor had focus, see
+		// https://github.com/facebook/draft-js/issues/410#issuecomment-223408160
+		if ( oldEditorState.getSelection().getHasFocus() ) {
+			newEditorState = EditorState.moveFocusToEnd( newEditorState )
+		}
```
dmsnell: I'm not sure if this is really something to respond to or not, but that `let` dangling there makes me a little uneasy. I've seen others simply create a chain of new editor states like this…

```js
const newEditorState = EditorState.createWithContent( … );

const editorStateWithProperFocus = oldEditorState.getSelection().getHasFocus()
	? EditorState.moveFocusToEnd( newEditorState )
	: newEditorState;

this.setState( { editorState: editorStateWithProperFocus } );
```

I'll let you make the call @nfcampos if you like this or not. So far things are going well in testing.

dmsnell:
@nfcampos when I'm on this branch I'm finding that I can't scroll within the editor window. I don't remember seeing this before but I'm not sure if I was even trying with notes long enough to require scrolling. Can you test this out?

 - Open a long note that extends beyond the view
 - Try and scroll. For me it doesn't work. It does in **master**

nfcampos:
@dmsnell oops, this was correct before but i accidentally reverted it, fixed now

dmsnell:
> fixed now

fixed for me too - thanks!

nfcampos:
while you're testing I'd suggest to try the following things:
- type in a note, with pauses of varying length (to allow the note to be synced), and make sure focus is never lost and the cursor never jumps to another place in the note
- paste text in a note and see if it looks ok
- copy text from a note, paste it somewhere else, and see if it looks ok
- try to create a couple of really long notes (eg. I've tried a chapter from the great gatsby) and see if it still works ok with no pauses, etc

hope this helps, I've tried all of these and they work correctly

lib/note-content-editor.jsx
nfcampos: I'd usually agree with not liking to reassign other things to the same variable, but I think in this case this is clear as is

beaucollins:
Thanks for the work on this!

Taking it for a bit of a test drive and one thing missing in the draft-js version is a context menu when right clicking the editor. Looks like this is an Electron integration thing specifically.

Current master using Electron:
screen shot 2016-10-06 at 1 10 09 pm

This PR using Electron (no context menu):
screen shot 2016-10-06 at 1 14 52 pm

This PR using Chrome:
screen shot 2016-10-06 at 1 13 32 pm

roundhill:
The text doesn't appear to be centering any longer on wide screens:

screen shot 2016-10-06 at 1 57 57 pm

Once the editor gets to a certain width, it should keep the content centered and add equal padding on each side.

nfcampos:
@beaucollins latest commit should have context menu

nfcampos:
@roundhill it appears to work for me, can you give more detail?

screen shot 2016-10-06 at 22 39 10

roundhill:
> @roundhill it appears to work for me, can you give more detail?

Nevermind :) I was looking at the web app because I had no `config` file...

beaucollins:
@nfcampos all fixed up!

nfcampos:
@roundhill that redirecting to the web app thing was actually pretty confusing for me when I first cloned this repo ;)

dmsnell:
what is this redirect we are speaking of?

nfcampos:
if you don't have a config file and open the development version of this repo served by webpack you end up being redirected to app.simplenote.com

roundhill:
> if you don't have a config file and open the development version of this repo served by webpack you end up being redirected to app.simplenote.com (open it in chrome)

We should probably just kill that for now, have it default to the react app...

roundhill:
The cursor on the editor should be a text select cursor, on this branch it is showing the mouse arrow cursor.

nfcampos:
@roundhill fixed

dmsnell:
yay @nfcampos - it's happening!

thanks thanks thanks for your awesome contributions. I'll need to abandon my tab and list PRs or update them in the very least, but that's okay because I'm sure they'll be easier to handle now

nfcampos:
@dmsnell great :)

I have some stuff done for tabs and lists, just needs a rebase, and I'll open a PR with it

beeblebrox3:
Do you have a date when this will be available in the app?
Thanks!

PR #374, opened 2016-09-01
Markdown: Remove global setting

### LINKED ISSUE #259: Removing the markdown toggle
Once I enabled markdown in settings I expected it to just work in my notes, so I was surprised that an extra step was needed to further enable it on individual notes. That feels like more effort than should be necessary.

roundhill:
Matches up behavior to the iOS/Android app where we always show the markdown toggle in the note info panel. 

The original global setting is still used to set the markdown flag for a new note. The 'Writing' section of `Settings` has been completely removed since it only contained the Markdown setting.

Fixes #259

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -70,7 +70,7 @@ function mapDispatchToProps( dispatch, { noteBucket } ) {
 			'increaseFontSize',
 			'resetFontSize',
 			'setNoteDisplay',
-			'toggleMarkdown',
+			'setMarkdown',
 			'setAccountName'
 		] ), dispatch ),
 		setSortType: thenReloadNotes( settingsActions.setSortType ),
@@ -214,6 +214,9 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			noteBucket: this.props.noteBucket,
 			note, markdown
 		} );
+
+		// Update global setting to set markdown flag for new notes
+		this.props.setMarkdown( markdown );
 	},
 
 	onNotesIndex: function() {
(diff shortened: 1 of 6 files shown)


dmsnell:
@roundhill is this ready for review?

roundhill:
> @roundhill is this ready for review?

@dmsnell yes, added the label. Thanks!

dmsnell:
@roundhill I'm confused at the global setting. Will we be leaving any place to adjust it after this? In other words, if I disable the setting then checkout this branch I don't think I can ever make it so that new notes start with Markdown support. On the other hand, if I enable the global Markdown setting, it appears like every new note on this branch will always have it enabled. If this is true, would it make more sense to eliminate the global setting from the app and use some simpler method for determining the new note policy, such as "use whatever the most recent note used"

roundhill:
> Will we be leaving any place to adjust it after this?

Sorry for ay confusion, but what I did was repurpose that boolean. It now is set _whenever any note_ sets the individual markdown toggle. This is how the app knows whether or not to have markdown enabled or not when a _new_ note is created.

You can see it being set here: https://github.com/Automattic/simplenote-electron/blob/7fd5f61170d0ad4ca58410a0fd260285217e015d/lib/app.jsx#L212-L220

roundhill:
And also, whether that boolean is set or not, all notes will now show the checkbox in the note info panel.

dmsnell:
> It now is set whenever any note sets the individual markdown toggle.

@roundhill that did confuse me and I missed it when testing. it seems reasonable enough though.

:shipit:

PR #382, opened 2016-09-12
escape fiter str before creating filter regexp

nfcampos:
currently if the filter string is eg. `(` search doesn't work because an un-escaped `(` is an invalid regexp

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -1,6 +1,7 @@
 import React, { PropTypes } from 'react';
 import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
+import escapeStringRegexp from 'escape-string-regexp'
 import appState from './flux/app-state'
 import browserShell from './browser-shell'
 import { ContextMenu, MenuItem, Separator } from './context-menu';
@@ -319,7 +320,7 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 		var regexp;
 
 		if ( filter ) {
-			regexp = new RegExp( filter, 'gi' );
+			regexp = new RegExp( escapeStringRegexp( filter ), 'gi' );
 		}
 
 		function test( note ) {
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -1,7 +1,6 @@
 import React, { PropTypes } from 'react';
 import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
-import escapeStringRegexp from 'escape-string-regexp'
 import appState from './flux/app-state'
 import browserShell from './browser-shell'
 import { ContextMenu, MenuItem, Separator } from './context-menu';
@@ -317,10 +316,9 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 
 	filterNotes: function() {
 		var { filter, showTrash, notes, tag } = this.props.appState;
-		var regexp;
 
 		if ( filter ) {
-			regexp = new RegExp( escapeStringRegexp( filter ), 'gi' );
+			filter = filter.toLowerCase();
 		}
 
 		function test( note ) {
@@ -332,7 +330,7 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			if ( tag && note.data.tags.indexOf( tag.data.name ) === -1 ) {
 				return false;
 			}
-			if ( regexp && !regexp.test( note.data.content || '' ) ) {
+			if ( filter && ( note.data.content || '' ).toLowerCase().indexOf( filter ) === -1 ) {
 				return false;
 			}
 			return true;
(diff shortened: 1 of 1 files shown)


dmsnell:
thanks for another submission @nfcampos - this PR raises an interesting question. it appears that escaping those characters effectively eliminates our RegEx-based search, except it actually does so by performing two `RegExp` operations where none could do `¯\_(ツ)_/¯`

the bigger question is whether or not we should have a `RegExp` search at all and whether people are currently using it this way.

unfortunately, I do not think we will be able to allow both a `RegExp` input in the search field _and_ automatically escape the special characters because the escaping function has no way of knowing which use was intended in writing something like `(`

nfcampos:
yeah I was assuming this wasn't the intended behaviour (especially because the iOS and Mac apps don't allow regex search).
if we wanted to support both regex and regular search one possibility would be to only do regex search when the search string starts and ends with `/`, eg. typing `2$` searches for notes that contain the string `2$`, typing `/2$/` searches for notes whose last char is `2`.

dmsnell:
> only do regex search when the search string starts and ends with /

that's not a bad idea, although it could still be confusing to have the search magically start working differently based on its input. when I compare to something like most text editors, they have a separate checkbox or toggle to indicate that the input is a `RegExp`

if, however, the mobile apps don't do this search, I find it hard to believe that we would demand it here.

nfcampos:
but even you end up deciding you do want regexp search the line where the `RegExp` line is created needs to be inside a `try` block, currently the app just crashes there if the typed string is an invalid regexp

dmsnell:
> currently the app just crashes there if the typed string is an invalid regex

how embarrassing! probably at this point we should eliminate the `RegExp` search altogether. I'm not ready to push it quite yet as something in my memory seems to indicate that there was a particular reason it was put in there, but I haven't found that reason yet

dmsnell:
@drw158 I forgot to ask you about this one. it seems very appropriate to kill the RegExp and make this a plain-text search. can you remember any reason we added a RegExp or think of any reason to keep it?

cc: @roundhill

roundhill:
I don't think we intentionally added anything for regexp specifically, so I'm a go to merge this.

dmsnell:
@nfcampos let's kill it! no more RegExp search :smile:

nfcampos:
@dmsnell for some reason i forgot to add escape-string-regexp to the dependencies but other than that this should be good to go then

dmsnell:
@nfcampos instead of adding that library, would you be interested in making the search a non-regexp? just a plain string search

nfcampos:
for search that makes sense (and probably be marginally faster actually) but to support highlighting of occurrences of the search term in notes in the future a regexp with the g flag like we have now is the way to go, no?

dmsnell:
I'm not sure about that @nfcampos - it looks like the search is currently contained within `filterNotes()`. If we end up moving forward to highlight the occurrences of the search terms, that will probably need to happen within the note content display.

nfcampos:
yes I realise that, my point is that those 2 places would then have two different ways of doing the same search behaviour the app supports (exact match, case insensitive), one of the places would do 
```js
new RegExp( escape( filterStr ), 'gi' ).test( content )
```
and the other would do
```js
content.toLowerCase().includes( filterStr.toLowerCase() ) 
```
is that desirable?

dmsnell:
> those 2 places would then have two different ways of doing the same search behaviour

@nfcampos I don't think we have those two places, right? only a single place right now that searches notes?

nfcampos:
> @nfcampos I don't think we have those two places, right? only a single place right now that searches notes?

@dmsnell yes only one of those exists currently

dmsnell:
@nfcampos let's just replace the `RegExp` for now and replace it with a simple string search. I think we can address it in the future if we add a second piece of search code.

nfcampos:
@dmsnell done

lib/app.jsx
```diff
@@ -331,7 +330,7 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			if ( tag && note.data.tags.indexOf( tag.data.name ) === -1 ) {
 				return false;
 			}
-			if ( regexp && !regexp.test( note.data.content || '' ) ) {
+			if ( filter && ( note.data.content || '' ).toLowerCase().indexOf( filter ) === -1 ) {
```
dmsnell: yay Chrome! let's use `string.includes( search )`

```js
if ( filter && ( note.data.content || '' ).toLowerCase().includes( filter ) ) {
```

dmsnell:
@nfcampos this all looks good to me. let's see if @roundhill or @drw158 or @rodrigoi or @beaucollins  could give this a look-over and a 👍 as well

rodrigoi:
looks good to me too! :shipit:

dmsnell:
thanks @rodrigoi - @nfcampos I'm merging!

PR #386, opened 2016-09-21
Insert a tab character while tabbing in the editor

### LINKED ISSUE #299: Allow tab key to indent
In the Windows app the tab key changes focus to the tags fields. It would be better if the tab key let the user add an indentation to the note.

_Tracking Tickets_
- #2642910-t
- #2641835-t
- #2653851-t
- #2654072-t
- #2670773-t
- #2708420-t
- #2711658-t
- #2724788-t
- #2770231-t
  -  #2849381
- https://twitter.com/user/status/749495005866258432

dmsnell:
Resolves #299 

Previously, when editing a note and hitting the tab key the app would advance the focus to the next element. In this case, that element happened to be the tag list. Since previous Simplenote apps have all allowed inserting a tab character as seamlessly as we can do in normal text editors, this was a bug in the Electron version that broke our assumptions.

This PR updates the `NoteDetail` component and intercepts tab presses so that they do what's expected of them. It raises some accessibility questions since the tab behavior is odd in order to better support keyboard navigation of the site (screen-readers and blind users depend on this). On the other hand, none of the other Simplenote apps behave this way and we aren't presenting what would be considered a website even though it runs in a (hidden) browser.

**Questions**
 - [x] Is there a better way to intercept the keyboard events and make sure that they originate from the editor than `'TEXTURE' === event.target.nodeName` ?

**Testing**
Open a note and try to hit the tab key. A tab character should be inserted whereas in **master** the tab key will advance focus to the tag list. Make sure that you can undo a tab insertion. Try to see if the tab key interacts unexpectedly with other parts of the app (the keypress event is currently global on `window.addEventListener`.

cc: @roundhill @rodrigoi @jleandroperez

### DIFF
--- a/devServer.js
+++ b/devServer.js
@@ -18,7 +18,7 @@ app.get( '*', function( req, res ) {
 	res.send( compiler.outputFileSystem.readFileSync( path.join( __dirname, 'dist', 'index.html' ) ) );
 } );
 
-app.listen( 4000, 'localhost', function( err ) {
+app.listen( 4000, '0.0.0.0', function( err ) {
 	if ( err ) {
 		console.log( err );
 		return;
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -1,64 +1,58 @@
-import React, { PropTypes } from 'react';
-import marked from 'marked';
+import React, { Component, PropTypes } from 'react';
 import Textarea from 'react-textarea-autosize';
-import { noop, get, debounce } from 'lodash';
 import analytics from './analytics';
+import marked from 'marked';
+import { get, debounce, invoke } from 'lodash';
 import { viewExternalUrl } from './utils/url-utils';
 
-const uninitializedNoteEditor = { focus: noop };
 const saveDelay = 2000;
 
-export default React.createClass( {
+const isValidNote = note => note && note.id;
 
-	propTypes: {
+export default class NoteDetail extends Component {
+	static propTypes = {
 		note: PropTypes.object,
 		previewingMarkdown: PropTypes.bool,
 		fontSize: PropTypes.number,
 		onChangeContent: PropTypes.func.isRequired
-	},
+	};
 
-	componentWillMount: function() {
+	constructor( props ) {
+		super( props );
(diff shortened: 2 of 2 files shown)


nfcampos:
@dmsnell this appears to add a tab character on <kbd>tab</kbd> and do nothing on <kbd>shift+tab</kbd>, right?

dmsnell:
correct @nfcampos - I suppose we should catch both actions if we can. currently we are simply inserting a tab and so I'm not entirely confident that we can remove indentation, but I will check and see what I can get in quickly.

the plan is still to build more behavior on your PR though once it's in. this is mainly to improve things in the meantime

nfcampos:
if it's not easy to remove indentation then I'd suggest for now just not doing anything if this is true https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/shiftKey

dmsnell:
> if it's not easy to remove indentation then I'd suggest for now just not doing anything if this is true

thanks for the catch @nfcampos - updated in 5df67d9

drw158:
Mac OS X 
Electron, Chrome

Sometimes hitting the tab key inserts a space, sometimes a tab character.

sn-tab

I can get consistent results by using this phrase: `consectetur adipiscing`. It seems to insert tabs based on the characters that are around the cursor? Sometimes hitting tab actually inserts a _small_ space.

Select a paragraph of text then hitting tab inserts a tab at the beginning of the paragraph. I think the default OS behavior is to replace the entire selection with a tab.

screen shot 2016-10-07 at 10 27 42 am

I kind of like how it inserts the tab at the beginning, but it was unexpected. Worth seeing what other editors do.

Bug: 
- go to Trash and select a note
- make sure you are in Edit mode
- click inside the note content
- hit tab
- info panel is opened and can't be closed
- reason: it's tabbing through the links in the panel

Suggested fix: when you are "focused" on the editor, tab should not tab through links. When you are in All Notes in preview mode, tabbing goes through links too.

dmsnell:
closing as irrelevant due to #370

PR #387, opened 2016-09-22
Automatically insert - and * in lists

### LINKED ISSUE #371: Markdown: New list item (* or -) not appearing automatically on new line of a list
When editing a Markdown list in a Android app the editor adds new list item character (\* or -) automatically when hitting enter at the end of a list item.

This doesn't happen on the Windows desktop app.

dmsnell:
Depends on #386 
Resolves #371

Previously, there was a missing behavior in this app that was present in
the other Simplenote apps: when creating a new line after a line
starting with a list bullet, the new line should start with the same
bullet at the same level of indentation.

This PR adds that behavior into the app.

**Testing**
Create a list in a note, such as the following…

```
List:
	- One
	- Two
```

On either of those lines, if you hit _enter_ to create a new line, the tab and dash should come with it as expected. Attempt to indent or outdent those lines. Try to add a second line with _enter, enter_ and the list prefix should disappear, returning the line to normal.

cc: @roundhill @rodrigoi @drw158

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -1,64 +1,58 @@
-import React, { PropTypes } from 'react';
-import marked from 'marked';
+import React, { Component, PropTypes } from 'react';
 import Textarea from 'react-textarea-autosize';
-import { noop, get, debounce } from 'lodash';
 import analytics from './analytics';
+import marked from 'marked';
+import { get, debounce, invoke } from 'lodash';
 import { viewExternalUrl } from './utils/url-utils';
 
-const uninitializedNoteEditor = { focus: noop };
 const saveDelay = 2000;
 
-export default React.createClass( {
+const isValidNote = note => note && note.id;
 
-	propTypes: {
+export default class NoteDetail extends Component {
+	static propTypes = {
 		note: PropTypes.object,
 		previewingMarkdown: PropTypes.bool,
 		fontSize: PropTypes.number,
 		onChangeContent: PropTypes.func.isRequired
-	},
+	};
 
-	componentWillMount: function() {
+	constructor( props ) {
+		super( props );
+
+		this.storeNoteEditor = r => this.noteEditor = r;
 		this.queueNoteSave = debounce( this.saveNote, saveDelay );
-		this.noteEditor = uninitializedNoteEditor;
-	},
+	}
 
-	componentDidMount: function() {
+	componentDidMount = () => {
 		// Ensures note gets saved if user abruptly quits the app
 		window.addEventListener( 'beforeunload', this.queueNoteSave.flush );
 		window.addEventListener( 'keydown', this.insertTabCharacter );
-	},
+	};
 
-	initializeNoteEditor: function( noteEditor ) {
-		this.noteEditor = noteEditor;
-	},
-
-	isValidNote: function( note ) {
-		return note && note.id;
-	},
-
-	componentWillReceiveProps: function() {
+	componentWillReceiveProps = () => {
 		this.queueNoteSave.flush();
-	},
+	};
 
-	componentDidUpdate: function() {
+	componentDidUpdate = () => {
 		const { note } = this.props;
 		const content = get( note, 'data.content', '' );
-		if ( this.isValidNote( note ) && this.noteEditor ) {
+		if ( isValidNote( note ) && this.noteEditor ) {
 			this.noteEditor.value = content;
 
 			// Let's focus the editor for new and empty notes
 			if ( content === '' ) {
-				this.noteEditor.focus();
+				invoke( this.noteEditor, 'focus' );
 			}
 		}
-	},
+	};
 
-	componentWillUnmount: function() {
+	componentWillUnmount = () => {
 		window.removeEventListener( 'beforeunload', this.queueNoteSave.flush );
 		window.removeEventListener( 'keydown', this.insertTabCharacter );
-	},
+	};
 
-	onPreviewClick: function( event ) {
+	onPreviewClick = event => {
 		// open markdown preview links in a new window
 		for ( let node = event.target; node != null; node = node.parentNode ) {
 			if ( node.tagName === 'A' ) {
@@ -67,9 +61,9 @@ export default React.createClass( {
 				break;
 			}
 		}
-	},
+	};
 
-	insertTabCharacter( event ) {
+	insertTabCharacter = event => {
 		if ( 'Tab' !== event.code ) {
 			return;
 		}
@@ -82,7 +76,7 @@ export default React.createClass( {
 			selectionStart,
 			selectionEnd,
 			value,
-		} = this.noteEditor
+		} = this.noteEditor;
 
 		this.noteEditor.value = [
 			value.substring( 0, selectionStart ),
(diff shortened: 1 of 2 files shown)


nfcampos:
(this is another one that is better with draft.js, see https://github.com/nfcampos/simplenote-electron/blob/lists/lib/note-content-editor.jsx#L73)

lib/note-detail.jsx
```diff
@@ -1,62 +1,166 @@
...
+		}
+
+		const {
+			selectionStart,
+			value,
+		} = this.noteEditor;
+
+		const prevLineStart = value.lastIndexOf( '\n', selectionStart - 1 ) + 1;
+		const prevLine = value.substring( prevLineStart, selectionStart - 1 );
+		const prevIndent = prevLine.match( /^(\s+)([-*])/ );
```
nfcampos: do you mean `[-\*]` instead of `[-*]`?

lib/note-detail.jsx
```diff
@@ -1,62 +1,166 @@
...
+		// if inserting at a cursor position
+		if ( ! event.shiftKey && selectionStart === selectionEnd ) {
+
+			// handle list indents too
+			const currentLineStart = value.lastIndexOf( '\n', selectionStart - 1 ) + 1;
+			const currentLine = value.substring( currentLineStart, selectionStart );
+			console.log( `currentLine: '${ currentLine }'` );
+
+			// line is a list point
+			if ( /^\s+[-*]\s?$/.test( currentLine ) ) {
```
nfcampos: it's not an amazing idea to maintain two separate (and similar) regexps for the same thing in each of the two functions, they're bound to get out of sync

lib/note-detail.jsx
dmsnell: @nfcampos did you find a problem with it? inside of the character list `[]` some of the RegExp symbols take on new meaning, some are even position-dependent. In this case `*` matches a literal `*` and _doesn't mean_ "any number of `-`"

nfcampos: ah you're right, nevermind then

lib/note-detail.jsx
dmsnell: which two are you speaking of? one of the RegExps is for determining if the previous line _has_ a list item while the other is for determining if the previously line _is only_ the list item bullet without following text.

nfcampos: `/^(\s+)([-*])([\s\S]+)?/` I mean it's better to use one regexp for both things, if the third group doesn't match then you have "only the list item bullet without following text"

lib/note-detail.jsx
```diff
@@ -1,62 +1,166 @@
...
+
+		// The previous line did not start with
+		// indentation, so continue as normal
+		if ( null === prevIndent ) {
+			return true;
+		}
+
+		const indent = prevIndent[ 1 ];
+		const bullet = prevIndent[ 2 ];
+		const nextIndent = `\n${ indent }${ bullet } `;
```
nfcampos: what's the point of `indent` and `bullet` being in separate groups if you then concatenate them?

lib/note-detail.jsx
nfcampos: (actually it should be `/^(\s+)([-*])\s?([\s\S]+)?/` to allow for a space after the bullet)

lib/note-detail.jsx
dmsnell: good call. originally I was planning on doing something with these, but now I'm not. I'll take them out

lib/note-detail.jsx
nfcampos: yeah maybe it's not worth it, i definitely see the point of simpler regexps, i just dont like the possibility of getting bugs from someone changing one of these without changing both

lib/note-detail.jsx
```diff
@@ -1,62 +1,166 @@
...
 		fontSize: PropTypes.number,
 		onChangeContent: PropTypes.func.isRequired
-	},
+	};
+
+	constructor( props ) {
+		super( props );
 
-	componentWillMount: function() {
+		this.storeNoteEditor = r => this.noteEditor = r;
```
nfcampos: since you started using class properties these could all be turned into those and the constructor could go away

lib/note-detail.jsx
```diff
@@ -1,62 +1,166 @@
...
-	},
-
-	initializeNoteEditor: function( noteEditor ) {
-		this.noteEditor = noteEditor;
-	},
-
-	isValidNote: function( note ) {
-		return note && note.id;
-	},
+		window.addEventListener( 'keydown', this.interceptTabPresses );
```
nfcampos: these can't be regular react event handlers on `Textarea`?

lib/note-detail.jsx
dmsnell: removed in 3fbfd77

lib/note-detail.jsx
```diff
@@ -1,62 +1,164 @@
...
 
-	componentWillMount: function() {
+		this.storeNoteEditor = r => this.noteEditor = r;
 		this.queueNoteSave = debounce( this.saveNote, saveDelay );
-		this.noteEditor = uninitializedNoteEditor;
-	},
+	}
 
-	componentDidMount: function() {
+	componentDidMount = () => {
```
nfcampos: is there a reason lifecycle methods like this one aren't regular methods ie. `componentDidMount() {`? they definitely don't need to be bound to the instance

lib/note-detail.jsx
dmsnell: @nfcampos the tab is a strange character on the web. the tab key doesn't trigger any of the React keypress events from the textarea because it's designed as a way to advance the input focus across the page. these behaviors we're adding actually break web accessibility in one sense because we're changing what the tab key does. on the other hand, this is more of an app than a webpage and the other apps already hold this behavior.

in short, no we couldn't do it with a React event because it won't fire when we need it

lib/note-detail.jsx
dmsnell: They would have to be bound because they access `this`, and I prefer this syntax over creating the bind statements in the constructor.

lib/note-detail.jsx
dmsnell: `queueNoteSave` would still be problematic the way its declared. we could get around this by moving `debounce` to where `saveNote` is defined, but for the time being I want to leave it where it is.

lib/note-detail.jsx
dmsnell: after playing around with composing the RegExps, I think I want to leave them alone for now. I don't see much gain in combining them and I don't see much gain in composing them when there are just two uses of them and those uses aren't identical and they aren't complicated.

thoughts @rodrigoi ?

lib/note-detail.jsx
nfcampos: hmm, it does trigger `keydown`, see here a quick example, press tab while focused on that textarea https://esnextb.in/?gist=10f5e3b63353827d220fbe6c4bfe7c05

lib/note-detail.jsx
nfcampos: nope, you only need to bind class methods to an instance when you use them "detached" from the a reference to the instance, ie.
```js
const obj = {
  abc: 2,

  boo() {
    console.log(this.abc)
  }
}

obj.boo() // logs `2`
// this is akin to how react internally calls lifecycle methods

const boo = this.boo

boo() // fails because `this` is not defined
// because you're not calling this method on any object (and the function isn't previously bound to any context), so there's no context to pass to the function
// this is akin to how event handler end up being called by react, because by doing `<div onClick={this.boo} />` you're passing a reference to the function boo, like in `const boo = this.boo` above
```

long story short, do bind event handlers, don't bind lifecycle methods

lib/note-detail.jsx
dmsnell: I guess I assumed it was a web thing in general. It's definitely not coming through in the Electron app. Your suggestion was actually my first attempt but then the event never fired. I couldn't seem to easily track down whatever might be intercepting it either.

I'd be more than happy to get rid of the window listener if we can. If you check out this branch and test locally you should be able to see what I'm seeing though.

nfcampos: hmm, it works for me https://github.com/nfcampos/simplenote-electron/commit/a0873f388ca3325d556e994cd20aa09cfde93412 running it in chrome, weird

nfcampos:
some stuff isn't right with shift+tab

see this gif http://www.giphy.com/gifs/3o6Zt9ncYZ8oUFjnfa

i first press tab, works as expected, then press shift-tab, and some text I had deleted (I think) gets inserted instead of the tab being removed

lib/note-detail.jsx
dmsnell: aha, so strange but the `onKeyDown` event sends the tab key while `onKeyPress` skips it. I'm updating the PR. thanks again @nfcampos

dmsnell: refactored in f984d70

dmsnell:
> some stuff isn't right with shift+tab

@nfcampos I updated the <kbd>shift+tab</kbd> behavior in 2031dc6. Thanks for finding that!

lib/note-detail.jsx
dmsnell: TIL. thanks @nfcampos!

refactored in 0d00ddf

lib/note-detail.jsx
```diff
@@ -65,56 +173,78 @@ export default React.createClass( {
...
-		);
-	},
+		document.execCommand( 'insertText', false, indented );
+
+		this.noteEditor.selectionStart = firstLineStart;
+	};
 
-	renderEditable( divStyle ) {
-		const note = this.props.note;
+	render = () => {
```
nfcampos: this can also be `render() {` like the other lifecycle methods

lib/note-detail.jsx
```diff
@@ -1,62 +1,170 @@
...
+			value,
+		} = this.noteEditor;
+
+		if ( hasModKeys( [ 'alt', 'ctrl', 'meta', 'shift' ], event ) ) {
+			return true;
+		}
+
+		const prevLineStart = value.lastIndexOf( '\n', selectionStart - 1 ) + 1;
+		const prevLine = value.substring( prevLineStart, selectionStart - 1 );
+		const prevIndent = prevLine.match( /^(\s+[-*])/ );
```
nfcampos: since this is a single capturing group with the whole regexp inside it there's no need for it to be a group at all, you can remove the `(` and `)` and use `prevIndent[ 0 ]` below

nfcampos:
another issue, pressing enter with the cursor at the end of line with an indented bullet list item does the wrong thing, see http://www.giphy.com/gifs/3o7TKqHIejvq2zAs36

nfcampos:
also, the behaviour described in the title seems to not be working for lines with bullet items without any indentation, see http://www.giphy.com/gifs/26uf9DOyE3ankSeWc

dmsnell:
> lines with bullet items without any indentation

thanks @nfcampos - we don't consider these lines to be actual lists unless I am wrong, so we wouldn't want a newline to add the `-` or `*`

dmsnell:
> pressing enter with the cursor at the end of line with an indented bullet list item does the wrong thing

good catch! I will make sure that gets corrected before a merge

nfcampos:
> thanks @nfcampos - we don't consider these lines to be actual lists unless I am wrong, so we wouldn't want a newline to add the - or *

@dmsnell ah, i didn't know that. the mac and ios apps do consider those actual lists, i thought the behaviour here would be the same

dmsnell:
> the mac and ios apps do consider those actual lists

haha, another win for reviews. I tested in the Mac app and it cleared it out, but I failed to add anything after the bullet (so of course it wasn't a list when I hit enter).

will update this behavior as well! thanks again

drw158:
I might be doing something wrong, but I couldn't get the new feature to work at all.

Chrome and Electron
Mac OS X

dmsnell:
Closing as irrelevant due to #370

PR #389, opened 2016-09-23
Autocomplete tag names when adding new tags

### LINKED ISSUE #19: Tag autocomplete as you type
While adding tags to a note it should show a list of existing tags to choose from as you are typing.

### LINKED ISSUE #113: Tag field should remain active after adding a tag
1. Add a new tag
2. Observe that the tag field is no longer focused, it should remain focused so that you can continue adding more tags easily.

### LINKED ISSUE #340: Tags duplicates
SimpleNote allows duplicated tags (no difference in case neither in right spaces). This must be avoided.
As soon as the user types the first letters, SimpleNote should suggest the possible existing tags.
Thank you.

dmsnell:
Resolves #19 
Resolves #113 
Resolves #340 

Entering new tags in this app can be a glitchy process.
- Although the app only shows unique tags, it allows storing new tags with identical names. You can see this by adding the same tag twice and opening the note in a native app. That tag will be duplicated (as initially reported in #340)
- Without autocomplete, it's easy to add duplicate tags and it's easy to create new tags that should be identical to an existing tag (adding `ofinterest` when `of-interest` already exists)

This PR attempts to resolve these issues and clean up the tagging experience. It provides an autocomplete list and prevents creating tags with the same name as existing tags. It also adds a new behavior: clicking on a selected tag will remove it (a double-click on an unselected tag removes it).

tagautocomplete

**Caveats**
<del>Although the autosuggest component I imported seems to have a well-designed API, it appears to have no way to externally trigger a refocus of the input field. For now, that means that the tag input field _does not_ refocus after selecting and deleting an existing tag. The focus should return to the input after adding new tags, however.</del> Resolved in 17e453f with the help of the component author.

**Testing**
Run the gamut on tag entry and removal.
- Do the tags added correspond to the tags displayed in the other apps?
- If you have a tag selected and change the tags in another app, does the selection behavior meet your expectations? (try adding new tags, deleting other tags, deleting the tag which is selected in the Electron app)
- Can you create duplicate tags (tags with identical names)?
- Does the autocompleter feel right? look right?

**Known bugs**
 - ~The tag input field sits below the rendered tag items~
 - ~The tag input field is hidden and hard to find~
 - ~Deleting tags is finicky; works sometimes, others not~
 - ~Tag suggestion can get slightly out of alignment with actual input~

cc: @roundhill @drw158

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -1,9 +1,11 @@
-import { partition } from 'lodash';
 import update from 'react-addons-update'
 import ActionMap from './action-map'
 import throttle from '../utils/throttle'
 import analytics from '../analytics'
 import { util as simperiumUtil } from 'simperium'
+import {
+	partition,
+} from 'lodash';
 
 const typingThrottle = 3000;
 
@@ -435,7 +437,7 @@ var actionMap = new ActionMap( {
 						return;
 					}
 
-					let working = state.note.data;
+					let working = note.data;
 
 					// diff of working and original will produce the modifications the client has currently made
 					let working_diff = simperiumUtil.change.diff( original, working );
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -1,11 +1,9 @@
+import { partition } from 'lodash';
 import update from 'react-addons-update'
 import ActionMap from './action-map'
 import throttle from '../utils/throttle'
 import analytics from '../analytics'
 import { util as simperiumUtil } from 'simperium'
-import {
-	partition,
-} from 'lodash';
 
 const typingThrottle = 3000;
 
--- a/lib/tag-chip.jsx
+++ b/lib/tag-chip.jsx
@@ -1,29 +1,15 @@
-import React from 'react'
-import classNames from 'classnames'
+import React from 'react';
+import classNames from 'classnames';
+import { noop } from 'lodash';
 
-export default React.createClass( {
+export const TagChip = ( { onSelect = noop, selected, tag: tagName } ) => (
+	<div
+		className={ classNames( 'tag-chip', { selected } ) }
+		data-tag-name={ tagName }
+		onClick={ onSelect }
+	>
+		{ tagName }
+	</div>
+);
 
-	getDefaultProps: function() {
-		return {
-			tag: '',
-			selected: false,
-			onSelect: function() {}
-		};
-	},
-
-	onClick: function( e ) {
-		e.preventDefault();
-		this.props.onSelect( e );
-	},
-
-	render: function() {
-		return (
-			<a
-				className={classNames( 'tag-chip', { selected: this.props.selected } )}
-				href="#"
-				onClick={this.onClick}>
-				{this.props.tag}
-			</a>
-		)
-	}
-} );
+export default TagChip;
--- a/lib/tag-field.jsx
+++ b/lib/tag-field.jsx
@@ -1,12 +1,19 @@
-import React, { PropTypes } from 'react'
-import TagChip from './tag-chip'
+import React, { PropTypes } from 'react';
+import TagChip from './tag-chip';
+import TagInput from './tag-input';
 import classNames from 'classnames';
 import analytics from './analytics';
+import {
+	difference,
+	invoke,
+	union,
+} from 'lodash';
 
 export default React.createClass( {
 
 	propTypes: {
-		tags: PropTypes.array,
+		unusedTags: PropTypes.arrayOf( PropTypes.string ),
+		usedTags: PropTypes.arrayOf( PropTypes.string ),
 		onUpdateNoteTags: PropTypes.func.isRequired
 	},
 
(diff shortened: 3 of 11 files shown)


lib/flux/app-state.js
```diff
@@ -448,7 +451,7 @@ var actionMap = new ActionMap( {
 					}
 
 					// working is the state of the note in the editor
-					let working = state.note.data;
+					let working = get( state, 'note.data', {} );
```
dmsnell: was causing an error

lib/tag-chip.jsx
```diff
@@ -1,29 +1,14 @@
...
+export const TagChip = ( { onSelect = noop, selected, tag: tagName } ) => (
+	<div
+		className={ classNames( 'tag-chip', { selected } ) }
+		onClick={ onSelect }
+	>
+		{ tagName }
+	</div>
+);
+
+export default TagChip;
```
dmsnell: Refactored for updated styling and pattern

lib/tag-chip.jsx
```diff
@@ -1,29 +1,14 @@
...
-} );
+import React from 'react';
+import classNames from 'classnames';
+import { noop } from 'lodash';
+
+export const TagChip = ( { onSelect = noop, selected, tag: tagName } ) => (
+	<div
+		className={ classNames( 'tag-chip', { selected } ) }
+		onClick={ onSelect }
+	>
```
dmsnell: Not sure why this was previously and <a />, but I couldn't think of any reason to keep it like that

lib/tag-field.jsx
```diff
@@ -18,12 +23,30 @@ export default React.createClass( {
...
+		if (
+			-1 === selectedTag || // no tag is selected
+			prevTags === nextTags // the tags are identical
+		) {
+			return;
+		}
+
+		this.setState( {
+			selectedTag: nextTags.indexOf( prevTags[ selectedTag ] ),
+		} );
```
dmsnell: this previously reset the tag selection whenever the props changed. I think this was to prevent switching the selection unexpectedly (index vs name)

for example, tag `Simplenote` was selected at index `2`, a new tag was added externally and that new tag comes before `Simplenote` in the list. Selected tag is still index `2` but is now a different tag, maybe `Electron`.

we don't need to be this cautious. If the tags don't change at all or if we don't even have a selected tag we can skip this. if we do have a selected tag, we just need to find where it ends up in the new list and update the selection index.

lib/tag-field.jsx
```diff
@@ -32,18 +55,24 @@ export default React.createClass( {
...
-	clearTextField: function() {
-		this.refs.tag.value = '';
-	},
-
 	addTag: function( tags ) {
-		tags = tags.trim().replace( /\s+/g, ',' ).split( ',' );
-		tags = this.props.tags.concat( tags );
-		this.props.onUpdateNoteTags( tags );
+		const newTags = tags.trim().replace( /\s+/g, ',' ).split( ',' );
+		this.props.onUpdateNoteTags( union( this.props.tags, newTags ) );
```
dmsnell: it was concatenation that was providing duplicate tags. using a set union instead prevents that.

drw158:
Hm not seeing the autocomplete. Maybe because of this error?

`ERROR in ./lib/tag-input.jsx`

dmsnell:
@drw158 can you try again? I have updated this. I didn't see that error you got either but I wonder if it had to do with other changes we have recently merged

drw158:
Got it to work! That is really nice. I think in the spirit of simplicity, we should try something inline.

http://www.jqueryscript.net/demo/Simple-Inline-Autocomplete-Autosuggest-Plugin-suggest-js/

Can't find live demo of a Material example, but they have an image:

dmsnell:
@drw158 to confirm, inline would break with the mac app

drw158:
That’s fine. The mac app has a combination of inline and the popover. I
think inline will be good. We can see how it feels.

drw158:
The suggestion text should be just a tiny bit lighter, maybe lighten($gray, 10%)

Other than that and the known issues you outlined, I think it's simple and elegant. A few quirks:

- Tab usually autocompletes the word for this type of thing
- Enter use to confirm the tag and make it into the pill shape
- Not sure if clicking twice to delete would be expected behavior.

dmsnell:
> Not sure if clicking twice to delete would be expected behavior.

@drw158 while technical easy to change, if we remove the click-select-click-delete behavior we have no confirmation before removing a tag. granted, there is note history (just found a bug related to notes not sync'ing after adding/deleting a tag) and someone could type the tag back in as well. still they may not have seen which tag was removed.

in the meantime I will remove the confirmation step so that a single click removes the tags. we can play with it and see how it feels

dmsnell:
> Tab usually autocompletes the word for this type of thing
> Enter use to confirm the tag and make it into the pill shape

this is a divergence from the Mac app, but we can make it happen

dmsnell:
> The suggestion text should be just a tiny bit lighter, maybe lighten($gray, 10%)

on it. you're constantly increasing my technical load in this PR :smile:

dmsnell:
@drw158 one issue with not selecting the tags before deleting them is that it currently mirrors the keyboard method for deleting tags

 - have the cursor focused in the tag entry with existing tags
 - backspace to clear out the entry
 - backspace again and it should select the last tag
 - backspace again and it confirms and deletes the tag

should we maintain selection for the keyboard but remove it for the mouse?

dmsnell:
@drw158 I think this is ready for some more of your review. I'm really struggling with the CSS to get the input and the tag list to line up right. you can see this really badly when there are no tags. adding a new tag makes the whole line shift up and then the tag list and input text are out of alignment with each other.

dmsnell:
@drw158, @roundhill, @Copons, and @rodrigoi 

I have updated this PR and made big changes to the way the autocomplete was rendered which resolved the funky issues. If I am testing this and seeing it correctly I think that it is ready for final review and merging!

@drw158 can you review `Enter a tag name…` as placeholder text and let me know what you think? Without the placeholder it seemed to me too hard to know that this was a tag entry area. Also, I wasn't entirely sure what words would be best to use.

Thanks all!

lib/tag-input.jsx
```diff
@@ -0,0 +1,138 @@
...
+		const suggestion = value.length && tagNames.find( startsWith( value ) );
+
+		return (
+			<div className="tag-input"
+				onClick={ this.focusInput }
+			>
+				<div
+					ref={ this.storeInput }
+					className="tag-input__entry"
+					contentEditable="true"
```
Copons: This produces a warning:

> Warning: A component is `contentEditable` and contains `children` managed by React. It is now your responsibility to guarantee that none of those nodes are unexpectedly modified or duplicated. This is probably not intentional.

React has a `suppressContentEditableWarning` prop to be added here that does exactly what it says.

dmsnell: thanks! I did a quick online search and didn't find this.

dmsnell: suppressed in b11197b

dmsnell:
@copons in addition to suppressing the warning (thanks for the tip!) I have closed off what I consider a bug in b11197b

previously we were able to create duplicate tags if they differed in the case of the letter. for example, `Recipe` was considered different than `recipe` even though they would auto-complete and were essentially the same. I have changed this so that tags of different cases are considered identical.

maybe @drw158 can review this decision. I think it's the right one.

dmsnell:
@copons I added the right-arrow auto-complete in 16a2e05

if the cursor is at the far-right of the text then hitting the right-arrow should do the same as <kbd>tab</kbd>; otherwise it should behave as usual with left/right cursor navigation.

roundhill:
> Tab usually autocompletes the word for this type of thing

Can we have the tab keypress also add the tag? Currently it autocompletes, but you still need to press enter to actually add it.

dmsnell:
> Can we have the tab keypress also add the tag? Currently it autocompletes, but you still need to press enter to actually add it.

This was changed [per request](https://github.com/Automattic/simplenote-electron/pull/389#issuecomment-264316082) by @drw158.

Either way is okay

drw158:
Sounds good to me.

dmsnell:
> Sounds good to me.

@drw158 what sounds good to you?

drw158:
Sorry, this sounds good to me:

>Can we have the tab keypress also add the tag?

dmsnell:
As of a2ae4e4 the tab key now auto-adds the tag to the note

roundhill:
Working great now, thanks for this cool feature @dmsnell!

:shipit:

drw158:
Sorry Dennis, I never got back to you about the placeholder text. Before this, the placeholder was `Add tags…` 

We should just use that. I'll create an issue.

PR #390, opened 2016-09-23
Create docker-compose.yml to allow building from inside Docker

dmsnell:
Previously I have had issues managing `node` versions between this and
other projects on my machine. By building inside of a Docker container
we can eliminate these issues.

This patch allows starting the app with `docker-compose up dev` instead of
`npm start` which runs inside the contained environment.

cc: @roundhill - I was tired of running the long `docker` command each time. this shouldn't hurt anything in the app, just adds a new way of starting it.

cc: @rodrigoi @vindl

### DIFF
--- a/devServer.js
+++ b/devServer.js
@@ -18,7 +18,7 @@ app.get( '*', function( req, res ) {
 	res.send( compiler.outputFileSystem.readFileSync( path.join( __dirname, 'dist', 'index.html' ) ) );
 } );
 
-app.listen( 4000, 'localhost', function( err ) {
+app.listen( 4000, '0.0.0.0', function( err ) {
 	if ( err ) {
 		console.log( err );
 		return;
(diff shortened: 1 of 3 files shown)


nfcampos:
this is great!

devServer.js
```diff
@@ -18,7 +18,7 @@ app.get( '*', function( req, res ) {
 	res.send( compiler.outputFileSystem.readFileSync( path.join( __dirname, 'dist', 'index.html' ) ) );
 } );
 
-app.listen( 4000, 'localhost', function( err ) {
+app.listen( 4000, '0.0.0.0', function( err ) {
```
nfcampos: you have this diff in all your PRs I think?

dmsnell: yeah it should be rebased out when #369 gets merged in. it's required in order for the dev server to serve connections from within the Docker container

roundhill:
LGTM :shipit:

PR #391, opened 2016-10-01
Finds note from `state.notes` instead of using `state.note`

dmsnell:
When `onNoteUpdate` was being called it was looking for `state.note` but
that didn't exist in at least some calls. I have not fully understood
the `state.note` variable or when it should exist.

In this patch, I have manually scanned the stores list of notes to try
and find it, working around the need for `state.note` to be set.

This is fixing some exceptions that were raised. I am not sure how many
other issues it will resolve but it was preventing the call to
`updateNoteContent`

cc: @roundhill @rodrigoi @beaucollins

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -429,7 +429,6 @@ var actionMap = new ActionMap( {
 			creator( { noteBucket, noteId, original, data, patch, isIndexing } ) {
 				return ( dispatch, getState ) => {
 					var state = getState().appState;
-					var note = state.note;
 
 					if ( isIndexing ) {
 						// Increase index counter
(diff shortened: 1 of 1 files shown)


roundhill:
Looks ok as a band-aid for now :shipit:

dmsnell:
I agree @roundhill - thanks for the review.

My recent experiences with Elm have been making me scream with things like this. We wouldn't be able to get it to compile if we tried accessing something that may not be there at runtime.

Were I to classify this, I would probably mark it as a result of an overly-confusing Redux/data implementation. I'm not sure why we ended up with this crazy `creator()` system or why _everything_ ended up in one giant file, but I can't seem to make good sense of the structure of the app state or when certain pieces should end up being updated.

lib/flux/app-state.js
```diff
@@ -449,6 +448,13 @@ var actionMap = new ActionMap( {
 					}
 
 					// working is the state of the note in the editor
+					const note = state.notes.find( n => n.id === noteId );
+
+					if ( ! note ) {
+						console.error( `Cannot find note (id=${id})!` );
```
nfcampos: @dmsnell variable `id` is not defined, I think you meant `noteId`? maybe the `no-undef` rule should be added to `.eslintrc` to avoid this?

PR #392, opened 2016-10-02
Several small updates to remove warnings and errors

dmsnell:
There were previously quite a few annoying messages in the console when running the build. This cleans those up.

cc: @roundhill


roundhill:
:shipit:

PR #393, opened 2016-10-02
Replace value link in tag list for controlled state changes

dmsnell:
Related to #392 

Removes the use of `valueLink` in the `TagList` so that React will stop complaining in the console. `valueLink` is deprecated. I did not merge this into that other PR because I felt like the changes here have a higher risk value and aren't simpler version-bumps etc…

> Despite the larger PR, only a few lines really change here. I have updated the style of the component into a `class` component which has muddied the diff.

**Teting**
 - Open the tag list
 - Click _Edit_
 - Rename a tag.

Look for issues with the tag list. For example, at one point during the development, renaming a tag resulted in the app showing `[Object object]` instead of the tag name (this has been fixed).

cc: @roundhill

### DIFF
--- a/lib/tag-list.jsx
+++ b/lib/tag-list.jsx
@@ -1,18 +1,48 @@
-import React, { PropTypes } from 'react'
+import React, { Component, PropTypes } from 'react'
 import classNames from 'classnames'
 import EditableList from './editable-list'
 import { get } from 'lodash'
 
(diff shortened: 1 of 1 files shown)


rodrigoi:
Verified! :shipit:

dmsnell:
Thanks @rodrigoi!

PR #394, opened 2016-10-02
Fix/revision slider

dmsnell:
Fixes some problems with the revision selector. Could be the reason behind the syncing issues.

Previously, the revision slider would open up in a strange state and hold onto that state until a different revision were explicitly selected. That is to say, the selection was `-1` and 'Latest' was showing up in position `1` on the slider.

**Before**
broken revision slider

**After**
fixed revision slider

**Testing**

Start with a fresh login and cleared browser state (open the DevTools and clear the application data, cache, IndexedDB, etc…).

Open a note and click on the history button to open the revision slider. In **master** it should load with `Latest` in the second position, position `1`, and that revision should probably be the second-most-recent revision. After interacting with the slider _or_ after navigating back to the list view and reselecting the note, the slider should have `Latest` at the far-right. It should remain to be the second-most-recent revision, however, with the latest actual revision found on the far left. Position `1` may still be selected and now that position should be a blank note, the original version before any text was entered.

In _this branch_ none of this unexpected behavior should be present. The revisions should load instantly in the proper order with `Latest` on the far right and with `Latest` actually being the most-recent revision.

**Further Work**

We need to get a test build of this built and tested. I think that there's a good chance this will take care of many of our sync issues. Actually, I believe that this could be the smokescreen hiding the fact that sync'ing may not be broken, but when loading notes the Electron app ends up reverting to the original blank revision.

cc: @roundhill @rodrigoi

### DIFF
--- a/lib/revision-selector.jsx
+++ b/lib/revision-selector.jsx
@@ -1,16 +1,38 @@
-import React from 'react'
+import React, { PropTypes } from 'react'
 import moment from 'moment'
+import { property } from 'lodash';
 
-export default React.createClass( {
+const lastUpdateOf = property( 'data.modificationDate' );
+const revisionSorter = ( a, b ) => lastUpdateOf( a ) - lastUpdateOf( b );
 
+const sortedRevisions = revisions => (
+	revisions
+		.slice()
+		.sort( revisionSorter )
+);
+
+export const RevisionSelector = React.createClass( {
 	mixins: [
 		require( 'react-onclickoutside' )
 	],
 
 	getInitialState() {
 		return {
+			revisions: sortedRevisions( this.props.revisions ),
 			selection: Infinity,
 		}
+	},
+
+	componentWillReceiveProps( { revisions: nextRevisions } ) {
+		const { revisions: prevRevisions } = this.props;
+
+		if ( nextRevisions === prevRevisions ) {
+			return;
+		}
+
+		this.setState( {
+			revisions: sortedRevisions( nextRevisions ),
+		} );
 	},
 
 	componentDidUpdate( { revisions: prevRevisions } ) {
@@ -38,11 +60,12 @@ export default React.createClass( {
 	},
 
 	onAcceptRevision: function() {
-		const idx = this.state.selection;
-		const revisions = this.sortedRevisions();
-		const revision = revisions[ idx ];
+		const {
+			revisions,
+			selection,
+		} = this.state;
 
-		this.props.onSelectRevision( revision );
+		this.props.onSelectRevision( revisions[ selection ] );
 		this.resetSelection();
 	},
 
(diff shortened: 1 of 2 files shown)


lib/revision-selector.jsx
```diff
@@ -1,56 +1,89 @@
...
 import moment from 'moment'
+import { property } from 'lodash';
 
-export default React.createClass( {
+const lastUpdateOf = property( 'data.modificationDate' );
+const revisionSorter = ( a, b ) => lastUpdateOf( a ) - lastUpdateOf( b );
 
-	getDefaultProps: function() {
+const sortedRevisions = revisions => (
+	revisions
```
nfcampos: this could simply be `orderBy( revisions, 'data.modificationDate', 'desc' )` (orderBy comes from lodash)

lib/revision-selector.jsx
```diff
@@ -1,56 +1,89 @@
...
-		require( 'react-onclickoutside' )
-	],
+	componentDidUpdate( { revisions: prevRevisions } ) {
+		const { revisions: nextRevisions } = this.props;
+
+		if ( prevRevisions !== nextRevisions ) {
+			// I'm not sure why exactly, but
+			// this control wasn't refreshing
+			// after loading in the revisions
+			// the first time.
```
nfcampos: how could code in `componentDidUpdate` solve something that happens "after loading in the revisions the first time.", seeing as `componentDidUpdate` doesn't get called after the first render, only for subsequent renders?

dmsnell: boy @nfcampos I wish I knew. the rendered HTML ends up having the intended `value` value, but it doesn't render the slider in the right spot. If you clone this and run it, you should be able to comment out this line and see that it ends up incorrect.

also, it's not _after the component mounts_, but _after loading the revisions_. There are typically a few renders that run _before_ the revisions load, since the revisions end up being pulled out of IndexedDB which is asynchronous.

lib/revision-selector.jsx
dmsnell: good call. I should have done that to start with.

dmsnell: updated in fa71bd1, though I used `asc` order :smile:

lib/revision-selector.jsx
dmsnell: @nfcampos I think what's happening is simply that it's not rendering. you can log the state changes and the render; you can actually open up the HTML view in the dev tools and watch the `<input value={ value } />` change _in the rendered HTML_ and for some reason it doesn't update visually.

As much as I am hesitant to think of these explanations, I suspect some bug in the `input` component or with React rendering it. I noticed some related issues in IE11 with `input` and plain-old HTML

Thanks as always for your feedback!

roundhill:
LGTM :shipit:

PR #398, opened 2016-10-07
Autocomplete lists

nfcampos:
Here it is, this PR includes list autocompletion behaviour (modelled it on the Mac app).

oct-08-2016 00-15-24

### DIFF
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -11,14 +11,16 @@ function plainTextContent( editorState ) {
 	return editorState.getCurrentContent().getPlainText( '\n' )
 }
 
+function getCurrentBlock( editorState ) {
+	const key = editorState.getSelection().getFocusKey();
+	return editorState.getCurrentContent().getBlockForKey( key );
+}
+
 function indentCurrentBlock( editorState ) {
 	const selection = editorState.getSelection();
 	const selectionStart = selection.getStartOffset();
 
-	const content = editorState.getCurrentContent();
-	const block = content.getBlockForKey( selection.getFocusKey() );
-	const line = block.getText();
-
+	const line = getCurrentBlock( editorState ).getText();
 	const atStart = line.trim() === '-' || line.trim() === '*';
 	const offset = atStart ? 0 : selectionStart;
 
@@ -26,7 +28,7 @@ function indentCurrentBlock( editorState ) {
 	const afterInsert = EditorState.push(
 		editorState,
 		Modifier.replaceText(
-			content,
+			editorState.getCurrentContent(),
 			selection.isCollapsed()
 				? selection.merge( {
 					anchorOffset: offset,
@@ -52,15 +54,12 @@ function outdentCurrentBlock( editorState ) {
 	const selection = editorState.getSelection();
 	const selectionStart = selection.getStartOffset();
 
-	const content = editorState.getCurrentContent();
-	const block = content.getBlockForKey( selection.getFocusKey() );
-	const line = block.getText();
-
+	const line = getCurrentBlock( editorState ).getText();
 	const atStart = line.trim() === '-' || line.trim() === '*';
 	const rangeStart = atStart ? 0 : selectionStart - 1;
 	const rangeEnd = atStart ? 1 : selectionStart;
-	const prevChar = block.getText().slice( rangeStart, rangeEnd );
 
+	const prevChar = line.slice( rangeStart, rangeEnd );
 	// there's no indentation to remove
 	if ( prevChar !== '\t' ) {
 		return editorState
@@ -70,7 +69,7 @@ function outdentCurrentBlock( editorState ) {
 	const afterRemove = EditorState.push(
 		editorState,
 		Modifier.removeRange(
-			content,
+			editorState.getCurrentContent(),
 			selection.merge( {
 				anchorOffset: rangeStart,
 				focusOffset: rangeEnd,
@@ -86,6 +85,56 @@ function outdentCurrentBlock( editorState ) {
 			anchorOffset: selectionStart - 1, // -1 because 1 char was removed
 			focusOffset: selectionStart - 1,
 		} )
+	);
+}
+
+function finishList( editorState ) {
+	// remove `- ` from the current line
+	const withoutBullet = EditorState.push(
+		editorState,
+		Modifier.removeRange(
+			editorState.getCurrentContent(),
+			editorState.getSelection().merge( {
+				anchorOffset: 0,
+				focusOffset: getCurrentBlock( editorState ).getLength(),
+			} )
+		),
+		'remove-range'
+	);
+
+	// move selection to the start of the line
+	return EditorState.forceSelection(
+		withoutBullet,
+		withoutBullet.getCurrentContent().getSelectionAfter()
+	);
+}
+
+function continueList( editorState, listItemMatch ) {
+	// create a new line
+	const withNewLine = EditorState.push(
+		editorState,
+		Modifier.splitBlock(
+			editorState.getCurrentContent(),
+			editorState.getSelection()
+		),
+		'split-block'
+	);
+
+	// insert `- ` in the new line
+	const withBullet = EditorState.push(
+		withNewLine,
+		Modifier.insertText(
+			withNewLine.getCurrentContent(),
+			withNewLine.getCurrentContent().getSelectionAfter(),
+			listItemMatch[0]
+		),
+		'insert-characters'
+	);
+
+	// move selection to the end of the new line
+	return EditorState.forceSelection(
+		withBullet,
+		withBullet.getCurrentContent().getSelectionAfter()
 	);
 }
 
@@ -171,6 +220,29 @@ export default class NoteContentEditor extends React.Component {
 		)
 	}
 
+	handleReturn = () => {
+		// matches lines that start with `- ` or `* `
+		// preceded by 0 or more tab characters
+		const listItemRe = /^\t*[-*]\s/;
+
+		const { editorState } = this.state;
+		const line = getCurrentBlock( editorState ).getText();
+
+		const trimmedLine = line.trim()
+		if ( trimmedLine === '-' || trimmedLine === '*' ) {
+			this.handleEditorStateChange( finishList( editorState ) );
+			return 'handled'
+		}
+
+		const listItemMatch = line.match( listItemRe )
+		if ( listItemMatch ) {
+			this.handleEditorStateChange( continueList( editorState, listItemMatch ) );
+			return 'handled'
+		}
+
+		return 'not-handled';
+	}
+
 	render() {
 		return (
 			<Editor
@@ -180,6 +252,7 @@ export default class NoteContentEditor extends React.Component {
 				onChange={this.handleEditorStateChange}
 				editorState={this.state.editorState}
 				onTab={this.onTab}
+				handleReturn={this.handleReturn}
 			/>
 		);
 	}


nfcampos:
@dmsnell here it is

dmsnell:
thanks @nfcampos! how hard would it be to split this into separate PRs for the two features? It would be much easier to tackle if we only had to consider one thing at a time :smile:

nfcampos:
@dmsnell hmm, the thing is you never really want to have one feature without the other, list autocompletion is not that useful if you can't indent list items, indentation is useful mostly for list items, no?

nfcampos:
or if you really want to split it I'd suggest making it one PR at a time, starting with the lists thing

dmsnell:
> or if you really want to split it I'd suggest making it one PR at a time, starting with the lists thing

my thoughts were to hook up the tab key first so that it indents properly and then we can worry about auto-completion

nfcampos:
@dmsnell opened the tab key PR at #400

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,41 @@
 import React, { PropTypes } from 'react';
-import { Editor, EditorState, ContentState } from 'draft-js';
+import {
+	Editor, EditorState, ContentState, Modifier, SelectionState
```
nfcampos: in this PR on in the other?

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,41 @@
...
+	},
+
+	splitBlock( editorState, selection = editorState.getSelection() ) {
+		return EditorState.push(
+			editorState,
+			Modifier.splitBlock( editorState.getCurrentContent(), selection ),
+			'split-block'
+		);
+	}
+}
```
dmsnell: Is there a reason to hold these inside an object instead of simply defining them in the module scope?

lib/note-content-editor.jsx
```diff
@@ -63,6 +93,144 @@ export default class NoteContentEditor extends React.Component {
...
 
+	handleReturn = () => {
+		let editorState = this.state.editorState;
+		const key = editorState.getSelection().getFocusKey();
+		const block = editorState.getCurrentContent().getBlockForKey( key );
+		const line = block.getText();
+
+		// matches lines that start with `- ` or `* `
+		// preceded by 0 or more tab characters
+		const listItemRe = /^\t*[-\*]\s/;
```
dmsnell: Not a problem, but we don't need to escape the `*` inside the character list.

lib/note-content-editor.jsx
```diff
@@ -63,6 +93,144 @@ export default class NoteContentEditor extends React.Component {
...
+			// move selection to the start of the line
+			editorState = EditorState.forceSelection(
+				editorState,
+				SelectionState.createEmpty( key )
+			);
+
+			this.handleEditorStateChange( editorState );
+
+			return 'handled';
+		} else if ( ( listItemMatch = line.match( listItemRe ) ) ) {
```
dmsnell: Instead of carrying the conditional through with indentation, why don't we create a new function to handle the behavior which is here in the block…

```js
const trimmedLine = line.trim();
if ( trimmedLine === '-' || trimmedLine === '*' ) {
	return removeBulletAndEndList( editorState, block );
}

const listItemMatch = line.match( listItemRe );
if ( listItemMatch ) {
	const [ bullet ] = listItemMatch;

	return addNewListItem( editorState, bullet );
}
```

lib/note-content-editor.jsx
```diff
@@ -63,6 +93,144 @@ export default class NoteContentEditor extends React.Component {
...
+			// if current line is just a bullet point and whitesapce,
+			// then clear the line (finishing a list)
+
+			// remove `- ` from the current line
+			editorState = EditorStateModifier.removeRange(
+				editorState,
+				SelectionState.createEmpty( key ).merge( {
+					focusOffset: block.getLength()
+				} )
+			)
```
dmsnell: I'm still a fan of creating separate variables here for each step. what if instead of reassigning the `editorState` we created a sequence like this…

```js
const withoutPreviousBullet = removeRange( editorState, … );
const withSelectionOnNewline = forceSelection( withoutPreviousBullet, … );
```

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,41 @@
...
+
+	removeRange( editorState, selection = editorState.getSelection() ) {
+		return EditorState.push(
+			editorState,
+			Modifier.removeRange( editorState.getCurrentContent(), selection ),
+			'remove-range'
+		);
+	},
+
+	splitBlock( editorState, selection = editorState.getSelection() ) {
```
dmsnell: This could probably have a more descriptive name, like `splitBlockAtSelection()`

lib/note-content-editor.jsx
```diff
@@ -63,6 +93,144 @@ export default class NoteContentEditor extends React.Component {
...
+			return
+		}
+
+		const content = editorState.getCurrentContent();
+		const block = content.getBlockForKey( selection.getFocusKey() );
+		const selectionStart = selection.getStartOffset();
+
+		if ( !e.shiftKey ) {
+			const addTabAtStart = block.getText().trim() === '-'
+				|| block.getText().trim() === '*';
```
dmsnell: Let's shorten this so that it fits comfortably on a single line or use some parenthesis or helper functions so that we're not stuck with an awkward split.

```js
const trimmedLine = block.getText().trim();
const addTabAtStart = trimmedText === '-' || trimmedText === '*';

const addTabAtStart = (
	trimmedText === '-' ||
	trimmedText === '*'
);

const addTabAtStart = [ '-', '*' ].contains( trimmedText );
```

lib/note-content-editor.jsx
```diff
@@ -63,6 +93,144 @@ export default class NoteContentEditor extends React.Component {
...
+					focusOffset: offset,
+				} ),
+				'\t'
+			);
+
+			// move selection to where it was
+			editorState = EditorState.forceSelection(
+				editorState,
+				SelectionState.createEmpty( block.getKey() ).merge( {
+					anchorOffset: selectionStart + 1, // +1 because 1 char was added
```
dmsnell: thanks for the nice comment!

dmsnell:
Thanks again for this great PR @nfcampos. I would love to see some of the code broken into separate pieces - split into functions that are shorter and do fewer things. It would also be nice to see less reassignment of variables.

Let me know if I can help out in any way.

lib/note-content-editor.jsx
nfcampos: I kept the same names as the `Modifier` methods exported by `draft-js` (which operate on `ContentState` objects, the difference with these is that they do the same operations on `EditorState` objects by using `push` with the corresponding `changeType`

lib/note-content-editor.jsx
nfcampos: see this comment for explanation why I wrote it like this https://github.com/Automattic/simplenote-electron/pull/398/files#r82646392

lib/note-content-editor.jsx
nfcampos: what do you think about this https://github.com/Automattic/simplenote-electron/pull/398/files#r82647506 ?

lib/note-content-editor.jsx
nfcampos: this was fixed in the other PR

lib/note-content-editor.jsx
dmsnell: hmm… good for consistency but also a little confusing since we're using the same name for a different function.

why not simply use `splitBlock` as it comes? shouldn't this get a different name to indicate that it does something different?

nfcampos:
@dmsnell i've rewritten this on top of the latest master, this is ready for your review again

lib/note-content-editor.jsx
```diff
@@ -171,6 +220,29 @@ export default class NoteContentEditor extends React.Component {
 		)
 	}
 
+	handleReturn = () => {
+		// matches lines that start with `- ` or `* `
+		// preceded by 0 or more tab characters
+		const listItemRe = /^\t*[-*]\s/;
```
dmsnell: Is a line a valid list if it starts with a dash or star but no space? What about spaces instead of tabs? I thought spaces were valid too

nfcampos: 1. in the mac app at least yes it it is a valid list even if if it has no indentation
2. I made it work only with tab characters for simplicity but it can be changed

dmsnell: @drw158 looks lie @nfcampos found/pointed out another inconsistency among the apps. what are your thoughts here on what _should_ define a list of items?

drw158: I haven't tried out the PR, but no indent should be required imo. Meaning:

1. type dash character + some text (With space or no space, doesn't matter. Tab creates indented list)
2. press enter
3. new line is created with dash

This is how the Mac app works, and I think it's fine. iOS app works the same way.

drw158: On second thought, I think a space after the character (`-` or `*`) should be required

nfcampos: That is then the exact behaviour implemented in this PR

lib/note-content-editor.jsx
```diff
@@ -171,6 +220,29 @@ export default class NoteContentEditor extends React.Component {
...
+			return 'handled'
+		}
+
+		const listItemMatch = line.match( listItemRe )
+		if ( listItemMatch ) {
+			this.handleEditorStateChange( continueList( editorState, listItemMatch ) );
+			return 'handled'
+		}
+
+		return 'not-handled';
```
dmsnell: are these strings really needed or are they mainly for debugging? I don't think we have to return any values here since this gets asynchronously called on input. we can still return early in 233 and 239 with `return this.handleEditorStateChange( … )` but just leave this function without a proper `return`

lib/note-content-editor.jsx
```diff
@@ -89,6 +88,56 @@ function outdentCurrentBlock( editorState ) {
...
+	);
+
+	// move selection to the start of the line
+	return EditorState.forceSelection(
+		withoutBullet,
+		withoutBullet.getCurrentContent().getSelectionAfter()
+	);
+}
+
+function continueList( editorState, listItemMatch ) {
```
dmsnell: I'm suspicious of the implicit state being passed in `listItemMatch` and curious why we need it. it requires a RegExp match object but only uses the first item. what if we required this to be the bullet+prefix and called it something like that

```js
function continueList( editorState, listPrefix ) { … }
```

then the calling code would read like this:
```js
const listItem = matcher.match( line );
if ( listItem ) {
	return … this.continueList( editorState, listItem[ 0 ] );
}
```

or better yet use some destructuring and comments…
```js
// list item has prefix up to the bullet and following space then arbitrary text
//  - thing
const listItem = pattern.match( line );
if ( listItem ) {
	const [ listPrefix ] = listItem;
	return this.handle… this.continueList( editorState, listPrefix );
}
```

dmsnell:
@nfcampos sorry you have been waiting so long on me to respond. I've been pretty busy and mobile lately. I have a couple more comments here but I think the behavior is 💯. after this PR merges I would like to examine the modularity here and see if we can't restructure some files for better organization.

thanks again for your wonderful contributions!

lib/note-content-editor.jsx
nfcampos: I was following this paragraph in the documentation https://facebook.github.io/draft-js/docs/api-reference-editor.html#cancelable-handlers-optional

lib/note-content-editor.jsx
nfcampos: good point, will change.

lib/note-content-editor.jsx
dmsnell: nice @nfcampos - I was unaware of those. I'm a bit nervous about the string values here for reasons like how I just mistook them. I wonder if constant keywords would be better here, something such as `DRAFTJS_HANDLER_CONTINUE` or `DRAFTJS_HANDLER_FINISH` - what do you think?

roundhill:
@dmsnell @nfcampos It's hard to tell where this one left off. Is it waiting on changes?

roundhill:
Merging for now, but we did find one bug to be fixed in a follow-up PR:

If you type a space and then a `-` or `*`, the autocomplete fails to add a new bullet when you press enter.

PR #399, opened 2016-10-10
Replace dependency on electron-prebuilt with electron

nfcampos:
per http://electron.atom.io/blog/2016/08/16/npm-install-electron

the electron-prebuilt package is deprected and will stop receiving new updates at the end of 2016


dmsnell:
This looks great. I think it's ready to 🚢 - @roundhill want to see it also?

roundhill:
👍

PR #400, opened 2016-10-10
added indent/outdent behaviour to editor

nfcampos:
The tab part of #398

### DIFF
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -1,9 +1,31 @@
 import React, { PropTypes } from 'react';
-import { Editor, EditorState, ContentState } from 'draft-js';
+import {
+	Editor, EditorState, ContentState, SelectionState, Modifier
+} from 'draft-js';
 import { invoke, noop } from 'lodash';
 
 function plainTextContent( editorState ) {
 	return editorState.getCurrentContent().getPlainText( '\n' )
+}
+
+const EditorStateModifier = {
+	insertText( editorState, selection, characters ) {
+		return EditorState.push(
+			editorState,
+			Modifier.insertText(
+				editorState.getCurrentContent(), selection, characters
+			),
+			'insert-characters'
+		);
+	},
+
+	removeRange( editorState, selection = editorState.getSelection() ) {
+		return EditorState.push(
+			editorState,
+			Modifier.removeRange( editorState.getCurrentContent(), selection ),
+			'remove-range'
+		);
+	},
 }
 
 export default class NoteContentEditor extends React.Component {
@@ -63,6 +85,78 @@ export default class NoteContentEditor extends React.Component {
 		invoke( this, 'editor.focus' );
 	}
 
+	onTab = ( e ) => {
+		// prevent moving focus to next input
+		e.preventDefault()
+
+		let editorState = this.state.editorState;
+		const selection = editorState.getSelection();
+
+		if ( !selection.isCollapsed() ) {
+			return
+		}
+
+		const content = editorState.getCurrentContent();
+		const selectionStart = selection.getStartOffset();
+		const block = content.getBlockForKey( selection.getFocusKey() );
+		const line = block.getText();
+
+		const atStart = line.trim() === '-' || line.trim() === '*';
+
+		if ( !e.shiftKey ) {
+			// inserting a tab character
+
+			const offset = atStart ? 0 : selectionStart
+
+			// add tab
+			editorState = EditorStateModifier.insertText(
+				editorState,
+				SelectionState.createEmpty( block.getKey() ).merge( {
+					anchorOffset: offset,
+					focusOffset: offset,
+				} ),
+				'\t'
+			);
+
+			// move selection to where it was
+			editorState = EditorState.forceSelection(
+				editorState,
+				SelectionState.createEmpty( block.getKey() ).merge( {
+					anchorOffset: selectionStart + 1, // +1 because 1 char was added
+					focusOffset: selectionStart + 1,
+				} )
+			);
+		} else {
+			// outdenting
+
+			const rangeStart = atStart ? 0 : selectionStart - 1;
+			const rangeEnd = atStart ? 1 : selectionStart;
+			const prevChar = block.getText().slice( rangeStart, rangeEnd );
+
+			if ( prevChar === '\t' ) {
+				// remove tab
+				editorState = EditorStateModifier.removeRange(
+					editorState,
+					SelectionState.createEmpty( block.getKey() ).merge( {
+						anchorOffset: rangeStart,
+						focusOffset: rangeEnd,
+					} )
+				);
+
+				// move selection to where it was
+				editorState = EditorState.forceSelection(
+					editorState,
+					SelectionState.createEmpty( block.getKey() ).merge( {
+						anchorOffset: selectionStart - 1, // -1 because 1 char was removed
+						focusOffset: selectionStart - 1,
+					} )
+				);
+			}
+		}
+
+		this.handleEditorStateChange( editorState );
+	}
+
 	render() {
 		return (
 			<Editor
@@ -71,6 +165,7 @@ export default class NoteContentEditor extends React.Component {
 				stripPastedStyles
 				onChange={this.handleEditorStateChange}
 				editorState={this.state.editorState}
+				onTab={this.onTab}
 			/>
 		);
 	}
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -1,6 +1,10 @@
 import React, { PropTypes } from 'react';
 import {
-	Editor, EditorState, ContentState, SelectionState, Modifier
+	ContentState,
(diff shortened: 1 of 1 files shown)


dmsnell:
@nfcampos I'm going to refrain from adding too many comments here since I think I covered most of it in #398. Please let me know if this is unclear.

nfcampos:
@dmsnell i'll go through all your comments soon. can you just tell me if you'd like me to update this PR or the other, keeping both in sync is not really possible

nfcampos:
(this one has a few changes relative to the other one, which I can easily port to the other one if you'd like)

dmsnell:
> can you just tell me if you'd like me to update this PR or the other? keeping both in sync is not really possible

let's do this one since it appears to be logically prior to the other.

nfcampos:
@dmsnell please see the last commit and let me know if this change makes sense to you (the idea is for EditorStateModifier to return an EditorState with the SelectionState that is "natural" after the operation. Please let me know if this makes the onTab function short enough.

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,68 @@
...
+} from 'draft-js';
 import { invoke, noop } from 'lodash';
 
 function plainTextContent( editorState ) {
 	return editorState.getCurrentContent().getPlainText( '\n' )
 }
 
+// contains methods that mirror the methods on Modifier with 2 differences:
+// 1. operate on EditorState instances
+// 2. move selection to the natural place following the modification
```
dmsnell: I like this comment, but I find it a big lacking in content. For example, I would assume that the normal `insertText` function in `draft-js` would make the claim that it moves the selection to the natural place following a modification, but we have a different definition of `natural`

(sorry for duplicating the same comment here as in the other PR), but I would be strongly in favor of not hiding these methods inside the `EditorStateModifier` object. Even simply extracting them without changing their names would be better, though it's possible that there could be more fitting names.

We can then turn the comments into proper JSDoc DocBlock comments explaining the behaviors, such as "Moves the selection to the end of the insertion"

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,68 @@
...
+const EditorStateModifier = {
+	insertText( editorState, selection, characters ) {
+		// this function assumes selection is collapsed
```
dmsnell: why? is this what we want it to assume? what happens if it isn't?

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,68 @@
...
+			anchorOffset: prevOffset + offsetDiff,
+			focusOffset: prevOffset + offsetDiff,
+		} )
+
+		return EditorState.forceSelection( afterInsert, nextSelection );
+	},
+
+	removeRange( editorState, selection = editorState.getSelection() ) {
+		// this function assumes startKey and endKey are the same
+		// ie. selection doesn't span more than one block
```
dmsnell: see comment above on related note

lib/note-content-editor.jsx
```diff
@@ -63,6 +120,54 @@ export default class NoteContentEditor extends React.Component {
...
+			const prevChar = block.getText().slice( rangeStart, rangeEnd );
+
+			if ( prevChar === '\t' ) {
+				const newEditorState = EditorStateModifier.removeRange(
+					editorState,
+					selection.merge( { anchorOffset: rangeStart, focusOffset: rangeEnd } )
+				);
+
+				this.handleEditorStateChange( newEditorState );
+			}
```
dmsnell: I probably commented on this in the other PR, so again, sorry for the dupes.

can we split these into separate `indentRange` and `outdentRange` functions? they don't have to hold those names, but it would be nice to have them in separate functions. further, after making sure that other modifier keys aren't pressed (<kbd>ctrl</kbd>, <kbd>meta</kbd>, <kbd>alt</kbd>), we can simplify this function here to

```js
return this.handleEditorStateChange(
	e.shiftKey
		? outdentRange( editorState )
		: indentRange( editorState )
);
```

the general principle here is twofold:
 - keep functions as short and single-minded as possible
 - keep event handlers even shorter so that they are primarily dispatchers to other functions instead of functions themselves which do things. (we should be able to easily indent or outdent a range without having to call the `onTab` handler)

dmsnell:
@nfcampos you continue to be awesome here. thanks for being so responsive to my comments. I reviewed #398 before I hit this one and I should have waited. I'll try and remember to leave it until this is merged in so I don't send a bunch of noise your way.

lib/note-content-editor.jsx
nfcampos: well, `Modifier.insertText` itself works only on collapsed selections ( see https://facebook.github.io/draft-js/docs/api-reference-modifier.html#inserttext )

lib/note-content-editor.jsx
nfcampos: this one assumes this because my uses of it so far never need to operate on ranges that span across lines

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,68 @@
...
+// 1. operate on EditorState instances
+// 2. move selection to the natural place following the modification
+const EditorStateModifier = {
+	insertText( editorState, selection, characters ) {
+		// this function assumes selection is collapsed
+
+		const afterInsert = EditorState.push(
+			editorState,
+			Modifier.insertText(
+				editorState.getCurrentContent(), selection, characters
```
nfcampos: sure thing

lib/note-content-editor.jsx
dmsnell: importantly though, `insertText` _will_ overwrite a non-collapsed section in production - just throws an error when it does

it would be nice here to either mimic that behavior or simply remove the note about assuming. if we send in a selection I think the general assumption is that things will write over it.

this brings me back to wondering about the utility of these functions attempting to mimic `draft-js`. we're clearly _not_ trying to match behavior by means of messing with the selection. we could do ourselves well by breaking that notion and asserting "this is _our_ insert text function"

lib/note-content-editor.jsx
dmsnell: maybe a good reason to investigate further :wink:

lib/note-content-editor.jsx
```diff
@@ -1,11 +1,68 @@
...
+		// move selection forward by nr of chars that were added
+		const nextSelection = prevSelection.merge( {
+			anchorOffset: prevOffset + offsetDiff,
+			focusOffset: prevOffset + offsetDiff,
+		} )
+
+		return EditorState.forceSelection( afterInsert, nextSelection );
+	},
+
+	removeRange( editorState, selection = editorState.getSelection() ) {
```
dmsnell: when I did my first couple passes I missed this, but I'm suspicious of the default value for `selection` here

I _do_ like using proper defaults on parameters, but just as with destructuring objects, it's easy to jump a little too far too quickly. in this case, we've got an input arg dependent on the arg passed in (or not passed in) _and_ dependent on another input arg passed in. would make for some odd debugging at best when something goes wrong.

constant values used as defaults: great
function calls as defaults: okay sometimes
dynamic, dependent, side-effecting actions as defaults: :shiver:

dmsnell:
> Maybe this is specific enough to indent/outdent that this should just live inside an indent/outdent function and there's no point in creating these custom modifier methods?

I would wager that this is more the case

nfcampos:
@dmsnell I've reorganised like you suggested

dmsnell:
@roundhill or @rodrigoi or @drw158 I'm ready to merge this. would either/all of you mind testing it to smoke-test?

this should provide normal expected tabbing behavior. we can compare to a word processor's behavior or other apps.

roundhill:
It works fine for indenting with the cursor on its own, but I noticed that nothing happens when text is selected. There's inconsistent behavior in the other apps as well:

Native macOS app: Deletes the highlighted test when tab is pressed.
app.simplenote.com: Adds a tab at the beginning of the line when tab is pressed on highlighted text.

Which behavior do we want here? I think I expected the text to be deleted.

dmsnell:
> Which behavior do we want here? I think I expected the text to be deleted.

My expectation when highlighting multiple lines and hitting tab is for all the lines to be indented…

rodrigoi:
> My expectation when highlighting multiple lines and hitting tab is for all the lines to be indented…

We are used to code editors. Text editors usually delete the highlighted text when hitting tab (they replace the highlighted text with any key really).

nfcampos:
Alright, so the consensus is, when text is selected, replace the selected text with a tab character?

nfcampos:
and when text is selected and the user hits `shift-tab` do nothing?

roundhill:
> Alright, so the consensus is, when text is selected, and the user hits tab, replace the selected text with a tab character?
> and when text is selected and the user hits shift-tab do nothing?

Yes, let's go with that!

nfcampos:
I've made the changes mentioned above

roundhill:
Looks great, thanks for your work on this @nfcampos! :shipit:

PR #402, opened 2016-10-11
Refactor/app state from weird to redux

dmsnell:
The existing app state is stored in an odd Redux-inspired model. It replaces the nice simple purity of Redux with several layers of confusion.

This PR starts migrating the existing app state into a proper Redux system and removes callback via prop-passing with `connect()` and `mapDispatchToProps`.

To keep the PR size down I have only changed a couple of state items: auth and the selected tag/collection.

This should bring in no functional changes, so **to test** mess around with login/logout and selecting tags/all notes/trashed notes in the tag list.

**Do not merge until after #398** is in **master**

cc: @roundhill @rodrigoi


lib/tag-field.jsx
```diff
@@ -2,6 +2,9 @@ import React, { PropTypes } from 'react'
 import TagChip from './tag-chip'
 import classNames from 'classnames';
 import analytics from './analytics';
+import {
```
nfcampos: the changes in this file are not related to the rest of the PR, correct?

lib/tag-field.jsx
```diff
@@ -131,7 +135,7 @@ export default React.createClass( {
 					onKeyDown={this.onKeyDown}
 					onBlur={this.onBlur}>
 					<input className="hidden-tag" tabIndex="-1" ref="hiddenTag" onKeyDown={this.preventTyping} />
-					{this.props.tags.map( ( tag, index ) =>
+					{ uniq( tags ).map( ( tag, index ) =>
```
nfcampos: shouldn't this happen when the tags are loaded into state, otherwise you'll end up needing to use `uniq` everywhere you render tags, which kind of defeats the purpose of having a central place to store state

dmsnell: it was a drive-by fix. there's a bigger issues I've discovered with tags wherein we're getting multiple tags with the same label. you are absolutely right in that we need to fix this at its root, but I threw this in in the meantime to prevent the errors from crashing things

dmsnell:
Note: This PR is big. We will either split it up or do something with it. I was without internet and did a lot of work locally in the meantime.

dmsnell:
After feeling like the history of this PR was too mangled and long, I have squashed it into a single commit and it should now be easier to work with. It has been rebased against the current **master** as well.

lib/app.jsx
```diff
@@ -476,7 +482,11 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			appState: state,
 			authIsPending,
 			isAuthorized,
+			selectAllNotes,
+			selectedCollection,
```
roundhill: My editor informs me that this is never used.

lib/navigation-bar.jsx
```diff
@@ -77,3 +96,16 @@ export default React.createClass( {
 		);
 	}
 } );
+
+const mapDispatchToProps = ( dispatch, { onSelectAllNotes, onSelectTrash } ) => ( {
```
roundhill: 🎆

lib/app.jsx
dmsnell: My editor informs me that they are no longer there :smile:

dmsnell:
> Whoops, just noticed something. The tag state isn't working properly for a few things:
> 
> If you select a tag, it should appear in the search bar. It is empty in this branch.
> When you select a tag, it should highlight as blue in the tag drawer. On this branch it never highlights.

These will have to wait on #398 to fix because it does a bunch of work that I don't want to duplicate here and then deal with the conflicts.

dmsnell:
@roundhill I believe that I have fixed the issues you raised with the tag behavior in the search bar and in the tag list. Would you mind reviewing again?

Maybe we're ready to merge! 🤞

dmsnell:
Too old to maintain.

PR #403, opened 2016-10-11
fix access to wrong variable name

nfcampos:
`id` is not defined, I guess `noteId` is what was meant


nfcampos:
@dmsnell this fixes https://github.com/Automattic/simplenote-electron/pull/391/files#r82788408

dmsnell:
Thanks @nfcampos! Good catch

nfcampos:
@dmsnell should really add that eslint rule, this is the kind of thing linters are good at ;)

PR #406, opened 2016-10-17
Fixes search bug and refactors `filterNotes()`

dmsnell:
Previously in #382 we replaced the `RegExp`-based note search with a
simple string-based search.

In that PR we apparently missed the fact that we were rejecting notes
whose content matched the search query instead of rejecting the ones
whose content _didn't_ match the search query.

This has been fixed.

Alongside the fix I have refactored the function to be clearer in hopes
of making future bugs like this less likely.

**Testing**
Perform smoke-tests on the search dialog.

 - [ ] Do notes appear properly when they contain the search string in their content?
 - [ ] Do the searches limit results to the selected tag if a tag is selected?
 - [ ] Do the searches limit results to trashed notes if trash is selected?

cc: @roundhill @rodrigoi

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -23,11 +23,12 @@ import {
 	noop,
 	get,
 	has,
+	includes,
 	isObject,
 	map,
 	overEvery,
 	pick,
-	values
+	values,
 } from 'lodash';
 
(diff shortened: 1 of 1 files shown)


dmsnell:
Thanks @rodrigoi !

Issue #309, opened 2016-05-03
Enable Github-flavoured Markdown


eblade:
Given how I assume this app is built (not too familiar with Electron, but it looks web-ish to me), would it be possible to use for instance this renderer [marked](https://github.com/chjj/marked) to enable Github-flavoured Markdown (GFM). The main feature for me would be syntax highlighting for a number of languages.

roundhill:
We're already using `marked`, so I think we just need to enable the gfm option 👍

MJIO:
+1 on syntax highlighting support.

Are there plans to also enable Markdown for the OSX application?

rnewton:
According to the readme and the code, marked has gfm enabled by default. https://github.com/chjj/marked/blob/master/lib/marked.js#L1242

To get syntax highlighting, you need to add another library like https://github.com/isagalaev/highlight.js

dmsnell:
Will be resolved when #309 merges.

eblade:
Looking forward to it!

geekpete:
GFM seems to work for me by wrapping code in ``` backticks.

This displays fine in the electron based client preview but breaks in the web version preview.

Can we get the web version to have parity with this feature?

rachelsquirrel:
A request in #3241877-t for Fountain script (https://fountain.io/), which is a markup language for writers.

dmsnell:
This should have been fixed in the rollout of the `showdown` renderer in #577.

PR #407, opened 2016-10-19
Enable syntax highlighting (#309)

rnewton:
I decided to take a swing at adding syntax highlighting for the markdown preview. It uses highlight.js and the "hybrid" theme that seems to work well for both light and dark modes. Everything passes eslint and I think I followed the style guides, but let me know if you want any tweaks. 

screenshot from 2016-10-18 21-10-45
screenshot from 2016-10-18 21-11-02

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -7,6 +7,7 @@ import { viewExternalUrl } from './utils/url-utils';
 import NoteContentEditor from './note-content-editor';
 
 const saveDelay = 2000;
+const highlighter = code => highlight.highlightAuto( code ).value;
 
 export default React.createClass( {
 
@@ -91,11 +92,7 @@ export default React.createClass( {
 
 	renderMarkdown( divStyle ) {
 		const { content = '' } = this.props.note.data;
-		const markdownHTML = marked( content, {
-			highlight: function( code ) {
-				return highlight.highlightAuto( code ).value;
-			}
-		} );
+		const markdownHTML = marked( content, { highlight: highlighter } );
 
 		return (
 			<div className="note-detail-markdown theme-color-bg theme-color-fg"
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react';
+import highlight from 'highlight.js';
 import marked from 'marked';
 import { get, debounce, invoke } from 'lodash';
 import analytics from './analytics';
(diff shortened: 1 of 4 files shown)


lib/note-detail.jsx
```diff
@@ -90,7 +91,11 @@ export default React.createClass( {
 
 	renderMarkdown( divStyle ) {
 		const { content = '' } = this.props.note.data;
-		const markdownHTML = marked( content );
+		const markdownHTML = marked( content, {
+			highlight: function( code ) {
+				return highlight.highlightAuto( code ).value;
+			}
+		} );
```
dmsnell: might start out by pulling this function into the module scope instead of creating a new closure on each call. could also use fat-arrow syntax for functions.

```js
const highlighter = code => highlight.highlightAuto( code ).value;

…

const markdownHTML = marked( content, { highlight: highlighter } );
```

dmsnell:
Thanks for the contribution @rnewton! I'll try and take a deeper look into this in the next day or two.

rnewton:
Good call. Thanks!

package.json
```diff
@@ -55,6 +55,7 @@
     "cookie": "0.2.3",
     "create-hash": "1.1.2",
     "draft-js": "0.9.1",
+    "highlight.js": "^9.7.0",
```
dmsnell: Let's pin this to its specific version. "Compatible" updates too often aren't.

dmsnell:
@rnewton are there better choices of style for the dark theme vs. the light theme? I appreciate the balance you have picked, but I want to make sure we don't unnecessarily tie ourselves to one when we could switch between the two if that were better.

rnewton:
I'm definitely open to suggestions on the theme(s) to pick. highlight.js includes [a bunch](https://highlightjs.org/static/demo/). I'm basically trying to avoid picking things that are divisive. Some people like monokai and some people hate it. 

I'll admit that I didn't investigate light/dark as separate themes, but "atom one light/dark" might be a good fit.

dmsnell:
@rnewton I noticed an opportunity to do the rendering in a web worker. probably this isn't the greatest fit at this time, but it would be nice to have a quick assessment if that's a good path to explore

rnewton:
I'm actually quite fond of solarized light/dark as an option instead. A little easier to read than the yellow on the light theme of hybrid. 

screenshot from 2016-10-18 21-41-34
screenshot from 2016-10-18 21-41-45

rnewton:
Having never worked with web workers, I hesitate to jump on that right now...

dmsnell:
@rnewton how hard would it be to swap styles at runtime?

dmsnell:
@roundhill @drw158 and @rodrigoi this seems like a good future to put out there for `1.1.0` maybe? I think it's ready to ship, what do you all think?

roundhill:
@dmsnell sure, let's :shipit:

PR #409, opened 2016-10-21
Redux: Move authentication state into simple Redux state

dmsnell:
Precursor to #402 

This branch pulls down the first couple commits from #402 where I was moving state out of our strange mutation of Redux into a simpler plain Redux reducer/store.

This not only adds the auth code but rearranges some organization of the Redux state and supporting files.

> There should be no functional or visible changes in this PR

**Testing**

Please smoke-test authentication. Logging in and logging out should work normally. Opening the app should behave normally as far as reauthentication is concerned, etc…


lib/app.jsx
```diff
@@ -2,6 +2,15 @@ import React, { PropTypes } from 'react';
 import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
 import appState from './flux/app-state'
+import {
```
roundhill: Nit: My editor informs me that there's some extra spaces here.

lib/app.jsx
```diff
@@ -2,6 +2,15 @@ import React, { PropTypes } from 'react';
...
 import appState from './flux/app-state'
+import { 
+	reset as resetAuth, 
+	setAuthorized,
+} from './state/auth/actions';
+import {
+	authIsPending,
+	isAuthorized,
+} from './state/auth/selectors';
+import { getSelectedCollection } from './state/ui/selectors';
```
roundhill: Did you forget to commit this folder/file?

lib/app.jsx
dmsnell: removed in c6efda7

dmsnell:
Thanks for the 🐔 @roundhill - those were good catches. Somehow I had seen them and fixed them but not committed or pushed back up.

Should be ready now.

roundhill:
I logged out and saw this error:

screen shot 2016-10-24 at 1 49 59 pm

roundhill:
> I logged out and saw this error:

Also, that was after checking out the branch, running `webpack` and `electron .` so maybe something was stuck after the update?

dmsnell:
Good catch @roundhill! I will check in on that.

roundhill:
Also, seeing this warning in the console:

```
simplenote-electron/lib/app.jsx
  207:4  warning  "resetAuth" is already declared in the upper scope      no-shadow
  208:4  warning  "setAuthorized" is already declared in the upper scope  no-shadow
  211:9  warning  "isAuthorized" is already declared in the upper scope   no-shadow
  476:4  warning  "authIsPending" is already declared in the upper scope  no-shadow
  477:4  warning  "isAuthorized" is already declared in the upper scope   no-shadow
```

dmsnell:
@roundhill I have updated this PR in 8fef338 to account for logged-out and login failures. in this regard, I think it may be more robust than it was before since it makes a distinction between invalid credentials and other failures. the PR is much bigger though

roundhill:
Seems good now, feel free to :shipit:. It'd be nice if those warnings went away but not critical.

dmsnell:
> It'd be nice if those warnings went away but not critical

we have three options:
 1. Change the linter to allow it
 2. Add a comment to ignore it
 3. Change the code

I'm reluctant to change the code because I find it most readable the way it is but I'm open to changing it if we want. My preference is to keep this in for now while thinking about a solution, so I will take you up on that.

The nice thing about this is that we will have an easy time finding these to come back and refactor because the linter finds them :wink: I have posted the question in a couple places and am waiting to hear back. We _do_ have a similar dilemma in Calypso

PR #413, opened 2016-10-23
Maintain window size/position state (fix #305)

### LINKED ISSUE #305: Remember session settings when app is closed
When the app is closed and reopened it would be good if the window size and the note which was previously open are returned by default.

arshsingh:


dmsnell:
Thanks for putting this together @arshsingh! It looks like this library is writing to and reading from a special file storing the window position. However, we are already using things like browser `localStorage` and `IndexedDB`. Do you think we could get by without the dependency and simply store the settings in the browser?

I like the idea here though and hope we can get this added in.

Please feel free to ping me directly here if you need anything.

dmsnell:
@arshsingh are you still interested in working on this? we'd be happy to help if you have any questions!

arshsingh:
Hi @dmsnell. I'm still interested and agree with what you proposed. I will implement it by the end of the coming week

dmsnell:
> I will implement it by the end of the coming week

cool, but no worries if it takes more time!

arshsingh:
Hi @dmsnell. Apologies for the delay. I tried using localStorage to save the state for this, but the problem is that we need the data before initializing the window:

```javascript
mainWindow = new BrowserWindow( {
	width: mainWindowState.width,
	height: mainWindowState.height,
} )
```

I'm not sure about IndexedDB but from what I can see, the api for that is also available only after window init (?)

dmsnell:
That's an interesting problem I didn't anticipate @arshsingh! Did you find any ideas how other Electron apps are solving this? I don't mind adding that dependency if that's what we need to do

arshsingh:
I did some research, it seems like everyone is storing this data in json/config files. We will probably have to do the same. In that case, my preference would be to use this dependency instead of just replicating the logic in simplenote's codebase.

dmsnell:
@arshsingh there are some things I am reluctant about with that library but maybe we can contribute back to it. I think this PR is ready and I will try and get some final testing in. Nice work and thanks for the PR! :smile:

dmsnell:
I tested this and it looks great. I would really appreciate it if @roundhill or @rodrigoi could also give it a review pass - they should also be aware of what's happening and also get the chance to test.

After that let's :shipit:

roundhill:
Awesome, this works great. Thanks again for the PR @arshsingh

PR #414, opened 2016-10-23
Adding repository field in package.json

Gzing:
The field is used for informational purposes.
Read more about the repository field,
– http://stackoverflow.com/a/16827990/1129927

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,10 @@
   "version": "1.0.5",
   "description": "The Simplenote app running as an electron app.",
   "main": "desktop/index.js",
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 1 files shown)


dmsnell:
@Gzing thanks for adding this! I'm not sure how we managed so long without it. must have originally been a copy/paste thing because I think `npm init` asks for it and will default to the right thing if run inside a clone of a git repo.

I don't see any reason this shouldn't be ready to go - we value your contribution making Simplenote better and more accessible!

cc: @roundhill and @rodrigoi I'm marking this and other things as `v1.1` since they aren't simple bug fixes. Wha'd'ya think?

Issue #411, opened 2016-10-23
X to clear search (feature request)


owcz:
The feature I miss most from other versions of SN is the one-click ability to clear the search field. Perhaps it's just the spoils of Mac but the white X in a gray circle in the right corner of the search box was a really easy way to clear it out while refocusing the app. Now it takes several clicks to select all text and begin to type.

dmsnell:
@owcz this sounds like a great idea - thanks for reporting! I'm sure this wasn't a decision we made to not include the `x` - more like it was something we weren't able to polish in time for the earlier releases. I'll take a look and see how much work it will take to get in. If not much, we can probably sneak it in soon :smile:

drw158:
#481

PR #416, opened 2016-10-24
Clear search field on ESC keypress

dmsnell:
This patch adds the behaviors where pressing the _escape_ key while in
the note search box will clear out any text already filled in.

This is a precursor to adding the clear button as requested in #411.

In order to make this work properly I had to make the search field
itself a controlled component whereas before the `value` was free as any
given HTML input would be. Now it's tied to the `filter` app state
property.

I have also taken the opportunity in this patch to refactor the
`<SearchField />` component, converting it to a `class`-based React
component and wrapping it in `connect()`, providing whichever props I
could without major app refactoring.

**Testing**

Type in search queries into the search field. Hitting <kbd>ESC</kbd> should both clear out the search field _and_ clear out the search itself, _i.e._ all the notes should become visible again.

We should make sure that this works when searching and changing the selected tag.

cc: @roundhill @drw158

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -505,8 +505,6 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 									</button>
 									<SearchField
 										onSearch={this.onSearch}
-										placeholder={state.listTitle}
-										searchFocus={state.searchFocus}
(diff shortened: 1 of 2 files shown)


roundhill:
Nice: :shipit:

PR #418, opened 2016-10-25
Parse and format markdown code in editor

dmsnell:
Links are automagically highlighted/converted to links in the Markdown
preview because `marked` parses them properly, but previously none of
this functionality existed in the note editor itself.

This PR follows some work @nfcampos provided to highlight the links in
the editor view. It's an experimental PR attempting to provide a link detector 
based on formal grammar of a note and a parser generated by PEG.js.

Parsing URLs can be tricky and error-prone. There are many available RegExps
that would probably serve us just fine, but I wanted to explore creating a note
grammar and parsing from that. Such a grammar/parser would make further
refinements beyond linkifying URLs easy.

This grammar/parser is incomplete. Currently it demonstrates a proof-of-concept
that works fairly well. Actually, it was created in a somewhat ad-hoc way and could use lots of cleaning up.
## Related Work
- https://github.com/facebook/draft-js/issues/47
## Update
- I have added in the ability to stylize markdown header lines: see the video
- I have added in the ability to format random stuff
- The code has now been refactored to move the parser/decorator into the `util` directory and the decoration proper now happens by assigning a CSS class name to the parsed content. Now the styling can be controlled in the SCSS file instead of by modifying the code that runs after the parser.
- The parser now lives in a separate repository and npm package [automattic/simplenote-grammar-pegjs](http://github.com/automattic/simplenote-grammar-pegjs)

 
## Testing

Play around with markdown enabled for a note. Switching between markdown on and markdown off should correspondingly enable or disable the syntax formatting for the editor view.

> Note that in the case of an HTML anchor tag, the `href` will linkify but a URL inside the `<a>` tags _won't_ highlight. The native apps highlights both but I chose not to linkify the inner URL.

parserdecorator


brandonpayton:
@dmsnell I love where this is headed. Being able to work from a grammar is powerful, and I'd enjoy these features as a user. Thanks for sharing and working on this!

drw158:
Cool!

I did get some errors in my terminal, so sorry if that is related to the bugs I found!

Module not found: Error: Cannot resolve module 'draft-js-simpledecorator' in /Users/Dave/simplenote-electron/lib

--------

I noticed that it doesn't seem to work if:

- already existing note
- not MD enabled 

When I enabled MD on the existing note, the highlighting doesn't seem to work.

Another bug:

- make a new note, make sure it has MD already enabled
- turn off MD
- highlighting still works, even though MD is off

Not sure if this is a regression, but the edit/preview button should not be visible in the Trash. You shouldn't be able to edit the note while it is in the trash. No preview, just read only edit view.

I don't think we should use the monospace font quite yet. Not until we make font selection a setting. It is kinda cool that it makes MD mode more clear though.

Also, our MD parser does not render `#header` (no space) in the preview, so I don't think we should highlight it.

dmsnell:
Thanks @drw158 - I _did_ see some of those bugs and I'm pretty sure I know how to fix them already - was not super motivated to do so at 6am. Also, I probably forgot to save the simple decorator dependency since I wasn't sure if I'd go with it or not.

The main goal I hope to get out of this PR is design-related. Is this something we think would bring value to the app? How much/how little should we style?

The bulk of the work is actually writing the parser. After that I think we could do well by simply wrapping the matched pieces in CSS classnames and style them that way.

Thanks again; I welcome all of your feedback!

nfcampos:
@dmsnell this looks great! ill try to leave more feedback one of these days when i get around to it but overall big 👍 from me

dmsnell:
> @dmsnell this looks great! ill try to leave more feedback one of these days when i get around to it but overall big 👍 from me

@nfcampos this is only really possible because you put in the draft-js editor. Not sure if I ever mentioned it, but you and I were doing draft-js work at the same time (I was bringing it in to WordPress.com). I love its data model and super-appreciate your contributions bringing it to Simplenote.

nfcampos:
@dmsnell yeah draft-js is great, I really like it.
it would be awesome if this same parser could be used for the preview tab, instead of `marked`, that way it'd be completely consistent

dmsnell:
> Module not found: Error: Cannot resolve module 'draft-js-simpledecorator' in /Users/Dave/simplenote-electron/lib

@drw158 did you do an `npm install` before running this? I _did_ put it in the `package.json` file so that should install if you run that command.

dmsnell:
@drw158 I updated so that we're now styling only via CSS and not through special HTML elements. It should be much easier to grok and modify now I hope.

roundhill:
My initial thought after trying it out is that I think we should match the styles for the editor to how it looks in the markdown preview. That way when you move back and forth between the editor and preview, the transition won't be as jarring. You should just see the markdown syntax disappear.

dmsnell:
@roundhill nice suggestions! I updated in 	9f06033

drw158:
Unable to boot the branch at the moment, but I'm more in favor of the way Ulysses handles it. Editing and Preview are 2 different writing environments, and if they look exactly the same then there will be confusion. By that I mean the user will be in Preview, but think they are in Edit. They'll try to click, and then realize what mode they are in. 

We do use the same fonts in both modes, but hopefully that will change. The user could change to a monospace font, and the Preview would remain as the system font.

Also, it's jarring if the text suddenly gets really large while writing (# headings). Basically any sort of text size change would be problematic I think.

The screenshot looks good except I would prefer

- bold headings, no size change
- blue instead of red for the checkboxes

roundhill:
> The user could change to a monospace font, and the Preview would remain as the system font.

Woah: 
screen shot 2016-12-19 at 4 15 12 pm

dmsnell:
> Woah:

@roundhill is that a good woah or a bad woah?

roundhill:
> @roundhill is that a good woah or a bad woah?

I'm not sure, I think I kind of like it with the monospace font. What do you think @drw158, should we add it as an option?

drw158:
We could add it as an option for all notes, not just MD notes. The setting would need to live in the view options. We have been meaning to do that anyway. Should we go ahead and implement it? :)

dmsnell:
> The setting would need to live in the view options. We have been meaning to do that anyway. Should we go ahead and implement it? :)

Yeah let's take care of the font thing first. I would like to get a more robust parser before merging this anyway (for example, we're not handling code blocks at this point or many nested things).

@drw158 one concern I have with the font setting is that I probably want all of my Markdown notes to show in a monospaced font in the editor but I don't want non-Markdown notes to use that monospaced font. Do you think it would be bad to have this kind of checkmark or distinction?

drw158:
> @drw158 one concern I have with the font setting is that I probably want all of my Markdown notes to show in a monospaced font in the editor but I don't want non-Markdown notes to use that monospaced font. Do you think it would be bad to have this kind of checkmark or distinction?

I'd rather keep the font choice consistent for all notes. Per note font distinction is a bit too complex imo.

dmsnell:
closing for now as not relevant. further work would be good but resources aren't available at the moment. also I have discovered some new markdown parsers that may serve the needs without incorporating a custom one, notably GitHub's own parser.

PR #419, opened 2016-10-26
Fix PageUp/Down Keys from Showing Info Panel

### LINKED ISSUE #408: Pressing PageUp/PageDown Keys Causes Layout Bug
#### Steps to reproduce
1. create new note (or open existing)
2. write some text
3. push PgUp or PgDown
4. got a bug

#### What happened instead
app is open side bar and you cant close it. the left side bar is blocked.

#### OS version
Windows 7 x64

#### Screenshot / Video
default

roundhill:
Something about adding `drafts-js` was causing the note info panel to appear when you pressed the Page Up or Page Down keys when the editor had focus.

It looks to me like the form elements that are present in the info panel are getting focus when these keys are pressed. The info panel is absolutely positioned 100% to the left to sort of hide it, so the browser window actually moves over to show the element that received focus.

At first, I thought I could fix this by changing where the note info panel was rendered to be _before_ where the editor renders, but it didn't fix the issue.

The solution I came up with was to simply not render the note info panel until we want to show it. It might have a slightly longer delay when animating in, but I haven't noticed any difference.

**To Test**
1. Compile a windows or linux version, and sign in.
2. Select a long note in the editor, and press the page up or page down keys.
3. The editor should scroll down or up, and the note info panel should remain hidden.

 Fixes #408

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -559,11 +559,13 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 								onNoteInfo={() => this.props.actions.toggleNoteInfo()}
 								shouldPrint={state.shouldPrint}
 								onNotePrinted={this.onNotePrinted} />
-							<NoteInfo
-								note={selectedNote}
-								onPinNote={this.onPinNote}
-								onMarkdownNote={this.onMarkdownNote}
-								onOutsideClick={this.onToolbarOutsideClick} />
+							{ state.showNoteInfo &&
+								<NoteInfo
+									note={selectedNote}
+									onPinNote={this.onPinNote}
+									onMarkdownNote={this.onMarkdownNote}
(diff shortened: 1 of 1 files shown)


dmsnell:
@roundhill awesome job! probably only showing the info panel on demand will be _better_ for performance since we don't have it sitting there in memory and in reflow.

roundhill:
@dmsnell We could probably do the same for the navigation drawer too

roundhill:
> We could probably do the same for the navigation drawer too

I added that in 63192bb

dmsnell:
@roundhill I can't test because I don't have a PageUp/PageDown key, but the code looks good!

🚢 at your leisure!

dmsnell:
@roundhill we should probably get a patch-release out - 1.0.6 or so with this fix.

PR #424, opened 2016-11-03
Highlight search result matches

### LINKED ISSUE #52: Add highlighting to search
While typing, matching strings should be highlighted. Here is a design:

searching

dmsnell:
Resolves #52

Previously when searching through the notes there was no indication
where inside a note the searched-for text could be found.

This patch introduces a new decorator that will match the searched
filter and highlight those results by means of an added CSS class.

> <del>Note that this isn't yet finished and ready. There is currently no highlighting in the note list. I am getting on a plane and have to go nonetheless!</del>

This PR is ready for final testing and review. I have created a new global CSS class `.search-match` which can be used to style matches in the note title, preview, and content.

**Testing**

Simply play with the search. Try to break it with internationalized text input or with Emoji. Test the performance.

searching in simplenote

cc: @drw158 @roundhill @mnmami

### DIFF
--- a/lib/editor/matching-text-decorator.js
+++ b/lib/editor/matching-text-decorator.js
@@ -1,0 +1,48 @@
+import SimpleDecorator from 'draft-js-simpledecorator';
+
+import CssClassWrapper from './css-class-wrapper';
+
+const matchIndices = ( text, matcher ) => {
+	const chunks = [];
+	let match;
+
+	while ( ( match = matcher.exec( text ) ) !== null ) {
+		chunks.push( [ match.index, matcher.lastIndex ] );
+	}
+
+	return chunks;
+};
+
+const dispatch = callback => ( [ start, end ] ) =>
+	callback( start, end, { className: 'search-match' } );
+
+export const findMatchingText = matcher => ( contentBlock, callback ) => {
+	const text = contentBlock
+		.getText()
+		.toLocaleLowerCase();
+
+	if ( ! matcher || ! text ) {
+		return;
+	}
+
+	matchIndices( text, matcher ).forEach( dispatch( callback ) );
+};
+
+/**
+ * Decorator to highlight text matching given search string
+ *
+ * @param {string} search text to search for
+ * @returns {null|Object}
+ */
+export const matchingTextDecorator = search => {
+	if ( ! search ) {
+		return null;
+	}
+
+	const query = search.toLocaleLowerCase();
+	const matcher = new RegExp( query, 'g' );
+
+	return new SimpleDecorator( findMatchingText( matcher ), CssClassWrapper );
+};
+
+export default matchingTextDecorator;
--- a/lib/editor/matching-text-decorator.js
+++ b/lib/editor/matching-text-decorator.js
@@ -1,4 +1,5 @@
 import SimpleDecorator from 'draft-js-simpledecorator';
+import { escapeRegExp } from 'lodash';
 
 import CssClassWrapper from './css-class-wrapper';
 
@@ -17,9 +18,7 @@ const dispatch = callback => ( [ start, end ] ) =>
 	callback( start, end, { className: 'search-match' } );
 
 export const findMatchingText = matcher => ( contentBlock, callback ) => {
-	const text = contentBlock
-		.getText()
-		.toLocaleLowerCase();
+	const text = contentBlock.getText()
 
 	if ( ! matcher || ! text ) {
 		return;
@@ -39,8 +38,7 @@ export const matchingTextDecorator = search => {
 		return null;
 	}
 
-	const query = search.toLocaleLowerCase();
-	const matcher = new RegExp( query, 'g' );
+	const matcher = new RegExp( escapeRegExp( search ), 'gi' );
 
 	return new SimpleDecorator( findMatchingText( matcher ), CssClassWrapper );
 };
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -7,6 +7,8 @@ import {
 } from 'draft-js';
 import { includes, invoke, noop } from 'lodash';
 
+import matchingTextDecorator from './editor/matching-text-decorator';
+
 function plainTextContent( editorState ) {
 	return editorState.getCurrentContent().getPlainText( '\n' )
 }
@@ -148,7 +150,8 @@ export default class NoteContentEditor extends React.Component {
 
 	state = {
 		editorState: EditorState.createWithContent(
-			ContentState.createFromText( this.props.content, '\n' )
+			ContentState.createFromText( this.props.content, '\n' ),
+			matchingTextDecorator( this.props.filter ),
 		)
 	}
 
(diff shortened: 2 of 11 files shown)


lib/editor/matching-text-decorator.js
```diff
@@ -0,0 +1,42 @@
...
+
+	const query = search
+		.toLocaleLowerCase();
+
+	matchIndices( text, query )
+		.forEach( dispatch( callback ) );
+};
+
+export const MatchingTextDecorator = search =>
+	new SimpleDecorator(
```
nfcampos: creating the RegExp
```js
const query = search.toLocaleLowerCase();
const matcher = new RegExp( query, 'g' );
```
should be moved here for performance reasons (current code creates a new, identical one for each ContentBlock)

lib/note-content-editor.jsx
```diff
@@ -120,20 +123,20 @@ export default class NoteContentEditor extends React.Component {
...
+		if (
+			( newContent === plainTextContent( oldEditorState ) ) &&
+			( nextFilter === prevFilter )
+		) {
 			return; // identical to rendered content
 		}
 
 		let newEditorState = EditorState.createWithContent(
-			ContentState.createFromText( newContent, '\n' )
+			ContentState.createFromText( newContent, '\n' ),
```
dmsnell: @nfcampos we already have a few issues with the cursor. thanks for pointing this out. I've heard some discussions lately about saving the cursor position on each note, which would definitely help here. I'll take a look and test with this.

when searching, I think it's possible that losing the cursor could be okay if we have to but I will try and rearrange so we don't

lib/editor/matching-text-decorator.js
dmsnell: > this would be faster if here you added

sure. I will do this with a ternary when I get a chance. thanks for the review!

nfcampos: I know it's mutated but I believe `exec` will reset `lastIndex` to 0 after the last match (to be more correct after the first non-match) see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex so I do believe you can indeed re-use it

dmsnell: Nice!

dmsnell: shuffled around in recent update/rebase

davewhitley (changes requested):
After playing with a few colors, I think we should use $blue for the highlight background color. White text.

Not sure if this is possible, but It would be great if the editor scrolled down to the first found word. Meaning, if I have a really long note and my search term is at the bottom of a note, I have to scroll a long ways down to find the highlighted term.

Also, I expected to see the search highlighting in the preview pane on MD documents.

Bug:

1. Go to trash
2. Search for anything
3. Go back to All Notes
4. Info panel is empty, and Share does nothing.

dmsnell:
> we should use $blue for the highlight background color

done in f835815. please note that I added padding around the sides of the search result because the white text was fading into the background of the note. this causes a slight visual twitch when starting to search because before the highlight the padding wasn't there.

given that it only happens when going from no search query to some search query I think the glitch is understandable. it's most noticeable in the note list.

jumping padding

dmsnell:
> It would be great if the editor scrolled down to the first found word

this has already been requested in another issue and I think it's good, but work for a different PR in my opinion. that's a technically different challenge than marking up the changes

dmsnell:
> I expected to see the search highlighting in the preview pane on MD documents.

This is an interesting question @drw158 that probably needs another PR; highlight in edit mode for now as good enough.

Actually highlighting in Markdown Preview mode is a complicated question: what _should_ the search match and highlight?

Since we can legitimately embed HTML inside of Markdown we could have something like this in a note:

```html
This is <strong>strange</strong>
```

In edit mode if we search for `strong` we will find the match in those HTML tags. what should happen in the rendered preview? do we somehow pull out the HTML tags and highlight them? do we simply not highlight them since they are _structure_ and not _content_?

if we highlight the structure then we either will break the document layout by tearing apart HTML tags and attributes or we will need to build a full HTML parser into the process and do cute things like hover a box over locations where those bits of markup are working behind the scenes.

if we don't highlight the structure we hit the issue where a note appears in the search results in the note list but no obvious highlighting is there in preview pane, probably a jarring experience.

strangehtmlhighlighting

drw158:
>  please note that I added padding around the sides of the search result because the white text was fading into the background of the note.

We could use box-shadow to make a border around the span. Then there would be no jumping. We should check performance though if we do that. Lots of box-shadows.

If that doesn't work, I'd rather remove the padding and eliminate jumping; it's pretty jarring.

drw158:
As discussed in Slack, it can be very confusing if you are searching with the preview pane on:

rp4p53a9s8knfhdaaaaaelftksuqmcc

Looks broken because my search term is not being highlighted.

One solution is to switch to the editor during searching and disabling the buttons. I think this would be a better experience. Big changes in the UI is pretty common while searching a page in other apps.

dmsnell:
@drw158 I added the editor mode swapping behavior in a5e46c8. Please take it for a spin and see how you like the behavior.

 - Open a note in preview mode then start typing in the search box
 - With text in the search box, find a note in the results with Markdown enabled then click preview

drw158:
I think that works well. Seeing the editor switch to Edit while typing in the search box felt ok to me. Only missing is the disabled Preview button. Clearing the search box after clicking on Preview definitely felt odd.

screen shot 2017-01-24 at 3 00 09 pm

dmsnell:
> Only thing missing is the disabled Preview button. Automatically clearing the search box after clicking on Preview definitely felt odd.

I don't know how odd I am but this honestly feels really terrible to me. it's definitely not obvious how one is supposed to flip back to the preview mode with search text in there. I'm not sure I know of any other apps which have this kind of behavior

all of this preview/search business makes me less and less comfortable with the search highlighting. I don't like where we've landed on the questions and wonder if we should just table this PR and release without it.

drw158:
Yep, this is a very specific situation, and I can't think of a similar problem in another app. Although, apps commonly alter their UI while in search mode. It's a matter of changing it the right way in an expected manner.

 I think we could research more apps that have an editor/preview component, ideally ones that don't show them at the same time. 

Atom is a nice example:

screen shot 2017-01-25 at 12 46 57 pm

But they don't have our issue, since both modes are shown at the same time.

Another option would be to completely hide the edit/preview button while searching (and auto switch to edit of course). This would be less confusing than showing a disabled preview button ("why is this button disabled?").

As far as shipping this feature, I think this is an important one, I'd rather see it shipped. I think users will inevitably be confused if they search with the preview mode on, but it only affects Markdown users. I think let's ship without addressing that issue and figure out a solution later.

dmsnell:
@drw158 how bad would it be to simply leave the preview mode open for now? that way it might be a little confusing to not see highlights but it would be less jarring than have the UI jump around to different things.

We might also be able to get in there and highlight rendered text too, so that while `strong` wouldn't appear as an HTML tag it _would_ appear if it were written out in the note content.

drw158:
> how bad would it be to simply leave the preview mode open for now? 

Yep that sounds good. Basically revert back to before I made this comment about the Preview pane: https://github.com/Automattic/simplenote-electron/pull/424#pullrequestreview-17318181

We can iterate later!

roundhill:
:shipit:

mnmami:
Thanks a lot for making this finally happen.

lib/editor/matching-text-decorator.js
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	matchIndices( text, matcher ).forEach( dispatch( callback ) );
+};
+
+/**
+ * Decorator to highlight text matching given search string
+ *
+ * @param {string} search text to search for
+ * @returns {null|Object}
```
owcz: valid-jsdoc wants a return description for this line

BobTB:
OK, sorry to bother you all, but is this now implemented in the windows app or online webapp?

dmsnell:
@BobTB this should already be available in the 1.0.8 release on Window and Linux.

PR #428, opened 2016-11-05
Making the source list resizable

dmsnell:
Rebased and squashed version of #276 from @beaucollins.

There were some oddities in the branch `feature/resizeable-sidebar` and in order not to risk messing it up I have pushed here.

### DIFF
--- a/lib/state/settings/reducer.js
+++ b/lib/state/settings/reducer.js
@@ -57,6 +57,19 @@ const markdownEnabled = ( state = false, action ) => {
 	return action.markdownEnabled;
 };
 
+const sourceList = ( state = { width: 330, isResizing: false, canResize: false }, { type, width, canResize, isResizing } ) => {
+	switch ( type ) {
+		case 'setSourceListCanResize':
+			return Object.assign( {}, state, { canResize } )
+		case 'setSourceListResizing':
+			return Object.assign( {}, state, { isResizing } )
+		case 'setSourceListWidth':
+			return Object.assign( {}, state, { width } );
+		default:
+			return state;
+	}
+};
+
 export default combineReducers( {
 	accountName,
 	fontSize,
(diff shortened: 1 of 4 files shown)


lib/state/settings/reducer.js
```diff
@@ -57,12 +57,26 @@ const markdownEnabled = ( state = false, action ) => {
...
+		case 'setSourceListCanResize':
+			return Object.assign( {}, state, { canResize } )
+		case 'setSourceListResizing':
+			return Object.assign( {}, state, { isResizing } )
+		case 'setSourceListWidth':
+			return Object.assign( {}, state, { width } );
+		default:
+			return state;
+	}
+};
```
dmsnell: @beaucollins it would be cool here if we split this up into separate reducers each of which operates on its own simple data. Also, can we move these type constants into the `state/action-types.js` file? (please ignore the `Symbol()` in there - I put that in as an experiment and I hate it so I'm following Calypso's trend with named string constants)

```js
// state/settings/source-list/reducer.js

const width = ( state = 330, { type, width } ) =>
	SOURCE_LIST_SET_WIDTH === type
		? width
		: state;

const isResizing = ( state = false, { type, isResizing } ) =>
	SOURCE_LIST_SET_RESIZING === type
		? isResizing
		: state;

const canResize = ( state = false, { type, canResize } ) =>
	SOURCE_LIST_SET_CAN_RESIZE === type
		? canResize
		: state;

export const sourceList = combineReducers( {
	canResize,
	isResizing,
	width,
} );
```

that structure seems more fitting for the Redux pattern and it's certainly easier to test

dmsnell:
This is good work but I think it's possibly worth closing due to changes in the app which took place in the meantime. We have big work to do after v1.1.0 so that the window panes are actually separate and visible only when "visible" which could impact the way this task is accomplished. Feel free to overwrite my decision @beaucollins

PR #436, opened 2016-11-09
add define plugin to webpack config

nfcampos:
currently NODE_ENV wasn't being passed to the files being compiled, the `-p` flag doesn't do that (see http://webpack.github.io/docs/cli.html#production-shortcut-p). With this change we get faster production builds of react and redux.


dmsnell:
Another brilliant contribution, thanks @nfcampos!

PR #437, opened 2016-11-12
ES6ify BucketStore

dmsnell:
- convert constructor functions into `class`es
 - remove Promise library and use native Promises
 - utilize fat-arrow functions for lexical scoping

cc: @beaucollins

### DIFF
--- a/lib/simperium/bucket-store.js
+++ b/lib/simperium/bucket-store.js
@@ -1,113 +1,84 @@
-var Promise = require('promise');
+class StoreProvider {
+	constructor( config ) {
+		this.setup = new Promise( ( resolve, reject ) => config( resolve, reject ) );
+	}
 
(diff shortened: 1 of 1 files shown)


lib/simperium/bucket-store.js
```diff
@@ -1,113 +1,84 @@
-var Promise = require('promise');
+class StoreProvider {
+	constructor( config ) {
+		this.setup = new Promise( ( resolve, reject ) => config( resolve, reject ) );
```
beaucollins: Looking a this now, would it make more sense to move the construction of the `Promise` external to `StoreProvider` and simply pass in the promise itself? I think this is originally my weird code.

dmsnell: sure, though I didn't want to change any functionality in this PR. I think that would require making changes at the calling sites - a good change for a future PR

PR #439, opened 2016-11-19
Replaces devServer.js by using webpack-dev-server

dmsnell:
`webpack-dev-server` provides a development server _and_ watches the
source code to hot-update. it does everything that `devServer.js` did
except that it requires fewer configurations and respects the original
`webpack.config.js`

cc: @roundhill @rodrigoi


roundhill:
Tried it out, everything seems hunky-dory. :shipit:

PR #440, opened 2016-11-19
Start building a webpack dll to decrease build repeated times

dmsnell:
> Depends on #439 because it's easier to run this from webpack dev server than to deal with `devServer.js`. That custom dev server injects and loads almost 100 additional webpack modules into the bundle in order to support itself. This is not needed.
> Merge _after_ #439

Previously we were building around 900 different webpack modules on every change. This took more time than necessary when rapidly iterating on a design.

This PR introduces a new webpack DLL so that the relatively static dependencies can be build and persist through the iteration. This leaves only around 100 webpack modules that need to be built on every change and those are mostly related to the app itself and not the dependencies.

The new process will build the DLL whenever starting the development server, but on incremental changes it does not.

cc: @roundhill @rodrigoi @nfcampos

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,10 +10,12 @@
   },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "build": "NODE_ENV=development `npm bin`/webpack --config ./webpack.config.js",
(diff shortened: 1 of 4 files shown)


roundhill:
LGTM :shipit:

Issue #410, opened 2016-10-22
Request: Select edit tab when adding a new note


arshsingh:
#### Steps to reproduce
1. Select preview tab in any note
2. Add a new note
#### What I expected

When adding a new note, I expect the editor to be in edit mode, so that I can start typing
#### What happened instead

New note opens in preview mode which is not useful since there is no content
#### OS version

Ubuntu 16.04

dmsnell:
Thanks for the report @arshsingh. This behavior doesn't make sense and we will put it on the queue for updating!

PR #441, opened 2016-11-22
Open new notes in edit mode

Copons:
Fixes: https://github.com/Automattic/simplenote-electron/issues/410

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -238,6 +238,10 @@ var actionMap = new ActionMap( {
 						dispatch( this.action( 'selectAllNotes' ) );
 					}
 
+					if ( settings.markdownEnabled ) {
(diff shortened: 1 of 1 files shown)


roundhill:
Nice, :shipit:

lib/flux/app-state.js
```diff
@@ -238,6 +238,10 @@ var actionMap = new ActionMap( {
 						dispatch( this.action( 'selectAllNotes' ) );
 					}
 
+					if ( settings.markdownEnabled ) {
+						dispatch( this.action( 'setEditorMode', { mode: 'edit' } ) );
+					}
```
dmsnell: @copons do we need the `if`? should this not be a safe operation to dispatch on every call?

it is not problematic the way it's written. in fact, it is faster this way, though I'm sure this performance impact will never matter.

I like exploring always-safe process and then removing branching where possible

Issue #433, opened 2016-11-07
Tag list sidebar should be a toggle


davewhitley:
Sounds easy, but actually quite hard to design for. I sketched a lot of different ideas, and researched a lot of apps. Here is what I've found for us.

First, the current design is great for mobile, and aligns the design with iOS/Android. Having a responsive layout for Simplenote Electron actually has nice benefits, one being the ability to set a very small width and still be functional. Most native apps have a large minimum window size because they are not responsive. 

I think whatever kind of layout changes we do to solve this tag sidebar issue will need to stay on desktop only. At small widths, the design stays the same as it is now.

![screen shot 2016-11-07 at 4 00 00 pm](https://cloud.githubusercontent.com/assets/618551/20077748/422cef3e-a503-11e6-84a1-62c8caf48bb4.png)

---

Proposal:

Red box is the OS chrome

![screen shot 2016-11-07 at 4 02 23 pm](https://cloud.githubusercontent.com/assets/618551/20077817/9ba82eac-a503-11e6-87e2-737c0ec4bc71.png)

Instead of a tag sidebar icon, it's now a 3 state layout toggle. 3 pane, 2 pane, and editor only. Clicking this button just cycles through each of the 3 layouts.

![screen shot 2016-11-07 at 3 41 32 pm](https://cloud.githubusercontent.com/assets/618551/20077886/d7d0e9be-a503-11e6-8cd3-efa16dd68599.png)

By moving the search bar to inside the notes list, the layout is free to change, and the top icon bar is static. When you have a layout toggle button, you don't want it jumping around.

Doing this will also allow us to do things like:

![screen shot 2016-11-07 at 4 06 56 pm](https://cloud.githubusercontent.com/assets/618551/20077972/4061f9d2-a504-11e6-8e84-172ad33bce80.png)

drw158:
The Info button and sidebar also become a toggle, but that has no effect on what I mocked up.

drw158:
Also, you might notice there is a lot of space in the top bar, enough for the search bar. We could do this, but having the search bar in editor only mode would be counterintuitive. Maybe it's just hidden in that mode?

roundhill:
Interesting idea! Should be fairly easy to build this and try it out.

belcherj:
@SylvesterWilmott Take a look at this. Closing as the design has diverged since this request.

PR #442, opened 2016-11-23
Make sidebars toggleable

Copons:
See: https://github.com/Automattic/simplenote-electron/issues/433

## WIP Screenshots

New top bar outside editor context (big screens only)
screen shot 2016-11-23 at 15 38 21

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -18,6 +18,7 @@ import NoteList from './note-list'
 import NoteEditor	from './note-editor'
 import SearchField from './search-field'
 import NavigationBar from './navigation-bar'
+import TopBar from './top-bar'
 import Auth from './auth'
 import NewNoteIcon from './icons/new-note'
 import TagsIcon from './icons/tags'
@@ -486,7 +487,7 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			'touch-enabled': ( 'ontouchstart' in document.body ),
 		} );
 
-		const mainClasses = classNames( 'simplenote-app', {
+		const mainClasses = classNames( 'main-layout', {
 			'note-open': selectedNote,
 			'note-info-open': state.showNoteInfo,
 			'navigation-open': state.showNavigation,
@@ -507,6 +508,18 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 					</ContextMenu>
 				}
 				{ isAuthorized ?
+					<div className="simplenote-app">
+						<TopBar
+							note={selectedNote}
+							onNewNote={ this.onNewNote }
+							onTrashNote={this.onTrashNote}
+							onRestoreNote={this.onRestoreNote}
+							onShareNote={this.onShareNote}
+							onDeleteNoteForever={this.onDeleteNoteForever}
+							onRevisions={this.onRevisions}
+							onCloseNote={ () => this.props.actions.closeNote() }
+							onNoteInfo={ () => this.props.actions.toggleNoteInfo() }
+						/>
 						<div className={mainClasses}>
 							{ state.showNavigation &&
 								<NavigationBar
@@ -569,6 +582,7 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 									onOutsideClick={this.onToolbarOutsideClick} />
 							}
 						</div>
+					</div>
 				:
 					<Auth
 						authPending={ authIsPending }
--- a/lib/note-editor.jsx
+++ b/lib/note-editor.jsx
@@ -8,7 +8,7 @@ import RevisionSelector from './revision-selector'
 import marked from 'marked'
 import { get } from 'lodash'
 
-import { setRevisionState } from './state/settings/actions';
+import { setRevisionState } from './state/revisions/actions';
 
 export const NoteEditor = React.createClass( {
 	propTypes: {
@@ -163,10 +163,10 @@ export const NoteEditor = React.createClass( {
 	}
 } );
 
-const mapStateToProps = ( { settings } ) => ( {
+const mapStateToProps = ( { revisions, settings } ) => ( {
 	fontSize: settings.fontSize,
 	markdownEnabled: settings.markdownEnabled,
-	revisionState: settings.revisionState,
+	revisionState: revisions.revisionState,
 } );
 
 const mapDispatchToProps = dispatch => ( {
--- a/lib/note-editor.jsx
+++ b/lib/note-editor.jsx
@@ -8,6 +8,8 @@ import RevisionSelector from './revision-selector'
 import marked from 'marked'
 import { get } from 'lodash'
 
+import { setRevisionState } from './state/settings/actions';
+
 export const NoteEditor = React.createClass( {
 	propTypes: {
 		editorMode: PropTypes.oneOf( [ 'edit', 'markdown' ] ),
@@ -25,7 +27,9 @@ export const NoteEditor = React.createClass( {
 		onRevisions: PropTypes.func.isRequired,
 		onCloseNote: PropTypes.func.isRequired,
 		onNoteInfo: PropTypes.func.isRequired,
-		onPrintNote: PropTypes.func
+		onPrintNote: PropTypes.func,
+		revisionState: PropTypes.object,
+		onRevisionChange: PropTypes.func,
 	},
 
 	getDefaultProps: function() {
@@ -39,17 +43,6 @@ export const NoteEditor = React.createClass( {
 		};
 	},
 
-	componentWillReceiveProps: function() {
-		this.setState( { revision: null } );
-	},
-
-	getInitialState: function() {
-		return {
-			revision: null,
-			isViewingRevisions: false
-		}
-	},
-
 	componentDidUpdate: function() {
 		// Immediately print once `shouldPrint` has been set
 		if ( this.props.shouldPrint ) {
@@ -59,13 +52,15 @@ export const NoteEditor = React.createClass( {
 	},
 
 	onViewRevision: function( revision ) {
-		this.setState( { revision: revision } );
+		this.props.onRevisionChange( { revision } );
 	},
 
 	onSelectRevision: function( revision ) {
 		if ( ! revision ) {
 			return;
 		}
+
+		console.log( revision );
 
 		const { note, onUpdateContent } = this.props;
 		const { data: { content } } = revision;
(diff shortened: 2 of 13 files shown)


lib/app.jsx
```diff
@@ -507,6 +508,18 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
...
+						<TopBar
+							note={selectedNote}
+							onNewNote={ this.onNewNote }
+							onTrashNote={this.onTrashNote}
+							onRestoreNote={this.onRestoreNote}
+							onShareNote={this.onShareNote}
+							onDeleteNoteForever={this.onDeleteNoteForever}
+							onRevisions={this.onRevisions}
+							onCloseNote={ () => this.props.actions.closeNote() }
+							onNoteInfo={ () => this.props.actions.toggleNoteInfo() }
```
dmsnell: no need to create a new closure here - we can simply pass the function as data

```js
onCloseNote={ this.props.actions.closeNote }
onNoteInfo={ this.props.actions.toggleNoteInfo }
```

lib/note-editor.jsx
```diff
@@ -168,9 +163,16 @@ export const NoteEditor = React.createClass( {
...
-	markdownEnabled: settings.markdownEnabled
+	markdownEnabled: settings.markdownEnabled,
+	revisionState: revisions.revisionState,
+} );
+
+const mapDispatchToProps = dispatch => ( {
+	onRevisionChange: revisionState => {
+		dispatch( setRevisionState( revisionState ) );
+	},
 } );
```
dmsnell: what if we got rid of the more generic names moving forward in our state system and used names that more accurately describe what we are doing.

in this case, the component isn't actually changing anything - it's simply indicating the type of change it wants to see in the world. in fact, that action is inherently coupled to the meaning of the component, in which case there's little to no need to abstract it away (and runaway prop-passing abstraction is one of the big weaknesses for this app in my opinion)

instead, let's name it for what it is

```js
const mapDispatchToProps = dispatch => ( {
	setRevision: revision => dispatch( setRevisionState( revision ) ),
	unsetRevision: () => dispatch( setRevisionState( null ) ),
} );
```

and now the component doesn't need to know anything about the mechanics of how this works or how to clear. it API is simplified to two functions: `set( revision )` and `unset()`

if we want to think differently, we can compose…

```js
const mapDispatchToProps = dispatch => ( {
	setRevision: compose( dispatch, setRevisionState ),
	unsetRevision: partial( dispatch, partial( setRevisionState, null ) ),
} );
```

lib/note-editor.jsx
```diff
@@ -76,19 +71,19 @@ export const NoteEditor = React.createClass( {
...
-		this.setState( { revision: null } );
+		this.props.onRevisionChange( { revision: null } );
 		this.setIsViewingRevisions( false );
 	},
 
-	setIsViewingRevisions: function( isViewing ) {
-		this.setState( { isViewingRevisions: isViewing } );
+	setIsViewingRevisions: function( isViewingRevisions ) {
+		this.props.onRevisionChange( { isViewingRevisions } );
 	},
```
dmsnell: coming in here now would be a great opportunity to trim down on the noise…

```js
setIsViewingRevisions = isViewingRevisions => this.props.setRevisionChange( { isViewingRevisions } );
```

also, I'm a bit leery of using an object and key name to call a function that only takes one argument

lib/app.jsx
Copons: I'm definitely not happy of passing all those props there.
I've just copied them in order to get the top bar working without pushing a 1K LOC commit. :)

lib/state/index.js
```diff
@@ -11,11 +11,13 @@ import persistState from 'redux-localstorage';
 import appState from '../flux/app-state';
 
 import auth from './auth/reducer';
+import revisions from './revisions/reducer';
 import settings from './settings/reducer';
 
 export const reducers = combineReducers( {
 	appState: appState.reducer.bind( appState ),
 	auth,
+	revisions,
```
dmsnell: I'm so happy that you added this in here! the Redux state is a great place for something like it.

one question I do have though that maybe is worth some discussion is _where_ the revision state belongs. is it a property of the app? of a note? of a view?

it appears like we are selecting a "revision" in the sense that we are selecting something more akin to a full note object and not simply a marker like "last tuesday at 8pm."

I'll dig in more later, but I'm pretty sure we don't currently have a list of revisions on the note object itself. that could be really handy to have for times like this probably.

lib/state/revisions/actions.js
```diff
@@ -0,0 +1,3 @@
+export const setRevisionState = revisionState => ( {
+	type: 'setRevisionState', revisionState,
```
dmsnell: you might want to checkout out `state/action-types.js` where we can declare global action types.

warning though, in **master** I have code there that I put in but I don't like. instead of defining action types with `Symbol()` I think I would like to stick with uppercase snake-case strings so that tooling is easier

lib/state/revisions/reducer.js
```diff
@@ -0,0 +1,16 @@
...
+	console.log( action );
+	return {
+		...state,
+		...action.revisionState,
+	};
+};
+
+export default combineReducers( {
+	revisionState,
+} );
```
dmsnell: if only we could have had our Redux pow-wow last week! oh but this is a great little bit of code to illustrate some patterns. I'll come back in a bit with some more code to show what I mean, but `combineReducers` gives us the ease to eschew reducers with complex state

```js
export default combineReducers( {
	isViewingRevisions,
	revision,
} );
```

lib/note-editor.jsx
Copons: Definitely trimmable.
But about the one argument: that won't be an issue anymore once I separate the various functions included in `onRevisionChange`.

lib/top-bar.jsx
```diff
@@ -0,0 +1,94 @@
...
+		onCloseNote: PropTypes.func,
+		onNoteInfo: PropTypes.func,
+	}
+
+	toggleIcon() {
+		switch ( this.props.layoutType ) {
+			case '3-col' : return '3 COLUMNS';
+			case '2-col' : return '2 COLUMNS';
+			case '1-col' : return 'EDITOR';
+		}
```
dmsnell: this is another type of data to ponder. what is the essence of what we are trying to represent here?

is it really one, two, or three columns?

might it be the presence or absence of different UI components at a particular time?

while I'm not saying this is the only way to do it, at first thought I think about having a set wherein membership in that set indicates that the component should be visible. it also ties the state to the components themselves in an extendible way instead of hard-coding the meaning of columns to components.

check out `lodash.union()` and `lodash.difference()` too. we have native-JS sets now but those are unfortunately mutable objects 😢 

```js
const visiblePanels = ( state = [], action ) => get( {
	HIDE_PANEL: ( { panelName } ) => difference( state, [ panelName ] ),
	SHOW_PANEL: ( { panelName } ) => union( state, [ panelName ] ),
}, action.type, () => state )( action );
```

lib/state/index.js
Copons: It sounds like revisions are properties of notes, but in this case they are only part of the revision selector view.
While having them always available seems awesome, wouldn't it be quite heavy performance-wise?

lib/state/revisions/actions.js
Copons: Already checked that, but... I found it pretty much empty and just assumed this was the way to go. :D

And yeah, I'd go uppercase too.

lib/state/revisions/reducer.js
Copons: I've left `combineReducers` pretty much to be ready to eventually split the reducer as you're suggesting.
(Still, would definitely love to have the Redux thing! 😄 )

lib/top-bar.jsx
Copons: I kinda hate the naming I used there (and of course the returns are only placeholder for the actual icons), but at the same time I honestly feel the good old switch makes it more readable at a quick glance.

As for the extendibility: adding a new panel in the main layout is also a heavy design concern and I don't think it that adding a case would prove that hard in the off chance we need it.

lib/state/index.js
dmsnell: > wouldn't it be quite heavy performance-wise?

that depends on how they are stored. if they are full copies of the notes that we asynchronously grab from IndexedDB then yes of course. if however we can grab a list of indexes for valid revisions for the note and they are little more than just integers then probably not.

lib/state/revisions/reducer.js
dmsnell: the main thing here is to recognize what is intrinsically coupled and what isn't. the way this is written implies that the two state properties are intrinsically bound, but I don't think they are. we can do a simple extraction to split these reducers into simpler and much more _testable_ things

```js
export const revision = ( state = null, { type, revisionState } ) =>
	REVISION_STATE_SET === type
		? revisionState
		: state;

export const isViewingRevisions = ( state = false, { type, revisionState } ) =>
	REVISION_STATE_SET === type
		? revisionState === null
		: state;

export default combineReducers( {
	isViewingRevision,
	revision,
} );
```

Now we have two separate reducers each with simple data types for state and they are so stupid simple that we practically don't even need to test them properly in order to benefit form their ability to be easily tested.

lib/state/settings/actions.js
```diff
@@ -45,3 +45,7 @@ export const setMarkdown = markdownEnabled => ( {
 export const setAccountName = accountName => ( {
 	type: 'setAccountName', accountName
 } );
+
+export const toggleLayoutType = layoutType => ( {
+	type: 'toggleLayoutType', layoutType,
```
dmsnell: let's choose a little more verbosity here for clarity (the newline)…

```js
export const toggleLayoutType = layoutType => ( {
	type: LAYOUT_TOGGLE,
	layoutType,
} );
```

lib/top-bar.jsx
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	onToggleLayoutType = () => {
```
dmsnell: we should probably reserve `toggle` for operations that are boolean in nature - toggling on vs toggling off. `setLayout` is closer to what we are actually doing here.

lib/top-bar.jsx
```diff
@@ -0,0 +1,94 @@
...
+	revisionState: revisions.revisionState,
+} );
+
+const mapDispatchToProps = dispatch => ( {
+	onToggle: layoutType => {
+		dispatch( toggleLayoutType( layoutType ) );
+	},
+	onRevisionChange: revisionState => {
+		dispatch( setRevisionState( revisionState ) );
+	},
```
dmsnell: purely a stylistic decision, but I love the point-free style here with compose

```js
( {
	onToggle: compose( dispatch, toggleLayoutType ),
	onRevisionChange: compose( dispatch, setRevisionState ),
} )
```

dmsnell:
wonderful work @copons. I would really like to hit the data representation hard on this one because I'm suspicious that `ONE_COLUMN`/`TWO_COLUMNS`/`THREE_COLUMNS` is the right way to do this.

lib/app.jsx
Copons: Oh damn, I see what you meant now!
Yeah, that's a byproduct of me blindly copying the props from `<NoteEditor>` a bunch of lines below, where it creates those two closures.

Copons: Well, it simply doesn't dispatch the action anymore.

Copons:
Moved over to https://github.com/Automattic/simplenote-electron/pull/443

Issue #343, opened 2016-07-14
Tags not appearing in search results even with "tag:" prefix


alexjgustafson:
Searching for a tag is a feature in the web app that's not available on desktop.
#### Steps to reproduce
1. Add a tag to a note that is not a word found in the note's body, "example123"
2. Search for "example123" and notice no results
3. Search for "tag:example123" and notice no results
#### What I expected

When searching for "tag:example123" to see a list of notes tagged with "example123" as is the behaviour in the web app

![screen shot 2016-07-14 at 11 20 54 am](https://cloud.githubusercontent.com/assets/933450/16844916/0cb2e344-49b5-11e6-8ed1-d5c0245d5d7c.png)
#### What happened instead

No results listed when searching for "tag:example123"
![screen shot 2016-07-14 at 11 21 17 am](https://cloud.githubusercontent.com/assets/933450/16844929/1cd66c78-49b5-11e6-8608-9700c82343c9.png)
#### OS version

Win10

alexjgustafson:
Relevant to #2736296-t

roundhill:
Isn't this the same as filtering by the tag name when you click on it in the left sidebar?

alexjgustafson:
It is - but we provide both the clickable and searchable option in web app. Only the clickable on desktop.

elexpander:
In my opinion the search bar should search **content and tags** by default. This way you can find any note related to a word even if the word is not part of the note content.

The `tag:` prefix would still be useful if you want to search tags exclusively. Although this feature is already provided by the tag side bar as they said.

dmsnell:
I'm going to close this here as a duplicate of #54 more so to merge the requests. We should provide a way to indicate in a search that a given string is specifically for searching tags.

PR #443, opened 2016-11-25
Make sidebars toggleable (2nd attempt)

Copons:
See: https://github.com/Automattic/simplenote-electron/issues/433

**Still WIP**

## Screenshots
screen shot 2016-11-25 at 18 08 26

cc @dmsnell @drw158


lib/app.jsx
```diff
@@ -508,7 +509,25 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 				}
 				{ isAuthorized ?
 						<div className={mainClasses}>
-							{ state.showNavigation &&
+							<div className="sidebar">
+								<div className="top-bar theme-color-border">
+									<ChangeLayout />
```
dmsnell: I think we could improve the naming. This reads like a verb when I think it's probably more akin to a noun such as "LayoutChanger" or "LayoutSelection" or "VisibilityToggles"

lib/change-layout.jsx
```diff
@@ -0,0 +1,61 @@
...
+				{ visiblePanes.includes( 'navigation-bar' ) &&
+					<TwoPanesIcon />
+				}
+				{ ! visiblePanes.includes( 'navigation-bar' ) && visiblePanes.includes( 'note-list' ) &&
+					<EditorOnlyIcon />
+				}
+				{ ! visiblePanes.includes( 'note-list' ) && ! visiblePanes.includes( 'navigation-bar' ) &&
+					<ThreePanesIcon />
+				}
+			</button>
```
dmsnell: Are we showing a single icon and rotating the view based on repeated clicks to that icon? I'm not sure how I feel about this if it's true. Do I have to click twice to go back to the previous view?

I'm a big fan of how the Mac app handles the tag list. There's an arrow one can click on to open and close it - the drawer idiom.

lib/state/settings/reducer.js
```diff
@@ -57,12 +67,28 @@ const markdownEnabled = ( state = false, action ) => {
...
 };
 
+const visiblePanes = ( state = [
+	'navigation-bar',
+	'note-list',
+	'note-editor',
+], action ) => {
+	if ( SHOW_PANE !== action.type && HIDE_PANE !== action.type ) {
+		return state;
+	}
```
dmsnell: this is not needed since `get()` defaults to the existing state; it is redundant.

dmsnell:
@copons I'm wondering how available this approach is for changing the view. the panes appear to be treated not as separate entities but as pieces of separate layouts. if I wanted to open or close the tag list, would I be able to easily discover the icon in the top left? I wonder if there are ways to bring the actions closer to the context they control.

another question is whether or not the editor should ever be hidden. it's stored in the app state as either visible or not visible, but would we ever want it not to be?

please don't take this too critically - I'm excited about this work and think you're doing well. it might help to step back and make some visual mockups to see the behavior and play with it before we do too much coding. @drw158 has provided some interactive mockups in the past where we can click around and see the effects without having any code involved. maybe he would be willing to discuss this with us.

lib/state/settings/reducer.js
Copons: I'm afraid this is needed because there's a `Symbol` action type (`AUTH_SET`?) that breaks the `get()` with `Uncaught TypeError: Cannot convert a Symbol value to a string`.

Copons:
@dmsnell Well, this is not that much code! (the other PR, on the other hand... 😆  )
I should be able to push soon-ish some changes that actually handle the various hiding and showing, so that this very branch would be an interactive mockup.

lib/change-layout.jsx
Copons: I think this is the behavior suggested by @drw158.
I'm not a fan of it too, but it should be super easy to replace the single icon with all three of them.

Copons:
@dmsnell, @drw158: I've pushed a commit that actually enables the button (responsive ready too), so that the whole thing is actually testable and ready to discuss.

The CSS code and the whole classNames mess are definitely WIP in order to make this work in a rush. I'll get back to it and clean everything up as soon as possible.

drw158:
Just checked it out. I think it works really well!

First thing I noticed: can we limit the search bar to only appear in the notes list column?

Bug: Go to editor only mode, shrink viewport. Hit the tags button, and it's blank.

I think it'd we worthwhile to ship this at the same time as the resize column feature. Not sure how that'd work with the responsive styles.

drw158:
> Are we showing a single icon and rotating the view based on repeated clicks to that icon? I'm not sure how I feel about this if it's true. Do I have to click twice to go back to the previous view?
> 
> I'm a big fan of how the Mac app handles the tag list. There's an arrow one can click on to open and close it - the drawer idiom.

The toggle is a 3 state toggle (1 pane, 2 pane, editor only). Condensing this all to one button is an elegant way to also include another requested feature, editor only mode. I think the drawer metaphor still exists — you are basically opening and closing the 2 drawers on the left (just sequentially, not at the same time).

>  if I wanted to open or close the tag list, would I be able to easily discover the icon in the top left? I wonder if there are ways to bring the actions closer to the context they control.

screen shot 2016-11-28 at 10 54 14 am

I think they will be able to find the button easily, and also guess what the button does, based on the icon. Placing the layout controls to the far left is a common pattern, and the icon is recognizable as well. The other benefit is that the button is close to the panels that it controls, the 2 sidebars. These are all assumptions though. We can always test!

I decided against having multiple buttons for a few reasons. Here are the possibilities I considered that achieve the same functionality:

- one button to toggle the tags list, one button to go to editor only mode
- 2 buttons that toggle each sidebar (tags, notes)
- one button (that toggles tags), with a dropdown option for fullscreen

screen shot 2016-11-28 at 11 07 13 am

Having multiple buttons increases cognitive load (Figuring out which button to click. What happens if I toggle the tags on, but also click the "editor only" button?)

With a 3 state toggle you are only at most 2 clicks away from the layout you want to achieve (3 clicks to arrive at the layout you are currently on).

I didn't think having a dropdown was a good solution because we have few layout options. Also, with a dropdown, you are still at most 2 clicks away from the layout you want. The benefit of the dropdown,  I think, is that if you only **ever** want to toggle the tags list, then the UI is optimized for you (you have a single toggle to show/hide the tags).

We can test multiple ways of doing it to try it out or test. If we want to, we can user test as well.

Thanks @Copons for doing this!

drw158:
Here are the icons: https://cloudup.com/cmyvtPWvzWZ

dmsnell:
thanks @drw158 for your feedback. this arrangement still feels very awkward to me. I don't think there's any way to do this that requires more than two clicks to get to any desired state, but at least in selecting the layout specifically we have introduced the need to identify the editor as a thing on its own when it should always be open.

if we had two buttons that toggled their respective panes each, we would probably be nominally one click away from the desired layout, presuming that at least one of the panes (tag list, note list) would be relatively static in its on/off state.

here are some screenies from other apps I use

screen shot 2016-11-28 at 10 29 35 am
In the Mac file browser we choose the layout we want, but there is a separate button for each in a button group. the buttons provide one-click access to the desired end.

screen shot 2016-11-28 at 10 47 07 am
The Chrome DevTools provide a similar separate button for the layout selection and then within the layout the drawers each have their own collapsable toggle.

screen shot 2016-11-28 at 10 50 47 am
JetBrains' IDEA environment well full-drawer idiom

screen shot 2016-11-28 at 10 53 34 am
Paw, which I find very useful but whose layout selection I find very confusing, has a mixture of confusing icons in a button group to select, [ one, the other, or both in split ], then adds a toggle for a lower drawer

screen shot 2016-11-28 at 10 55 28 am
Photoshop, which diverges strongly from feeling like a native Mac app, uses the drawer idiom everywhere I think, plus it offers a menu in **Window** to toggle visibility of the different components

screen shot 2016-11-28 at 10 58 01 am
DayOne uses a drawer icon on their "note list" to toggle the notebook list to its left. it has an arrow to indicate what will disappear or where it will go. collapsing the notebook view makes the arrow on the icon disappear. Additionally, a toggle for the "Dual View" opens up a new pane or drawer to the right with a live view of the selected note. When collapsed, one has to open the "note" from the "note list" just like the Electron app currently does. With the dual view open, the note list is always visible and clicking on a note opens it in the right.

screen shot 2016-11-28 at 11 05 05 am
KeyNote has a much more complicated UI. On the left we have the dropdown button which selects major layout for the app. On the right we have three buttons in a group. Clicking a button selects that pane for the right drawer. Clicking on an active button hides the drawer and it slides to the right.

screen shot 2016-11-28 at 11 11 23 am
The Preview app uses the dropdown button. There are several layout arrangements.

**Not pictures**
Atom appears to delegate all visibility controls to the top-level view menu.

I did not have a particular point I'm trying to make by including these, but just wanted to show a selection of other apps how they handle it and see if we can learn from those examples. which app did you show, @drw158? was that Sketch?

dmsnell:
When thinking about cycling between layouts I discovered another problem that I can foresee, which maybe could be solved and I just don't know how: keyboard shortcuts.

If we are in a given view and want to provide a keyboard toggle for the tag list, how do we accomplish this? Do we simply selected the appropriate layout? It seems like we're at odds in the verb we're doing, which I think is my main hangup.

To me, cycling among a ring of layout choices is choosing one of several specific arrangements of views into the app. Toggling the tag list or note list is fundamentally different than cycling among views. It's a toggle operation that can be easily undone by redoing it. Cycling can only be undone by scanning through the full ring.

If we had toggles in some form for the two panes then keyboard shortcuts would provide no UX questions. Hitting the keypress would simply toggle the visibility for the associated pane.

If we are in the distraction-free mode with only the editor visible and then use the "Show notes list" keypress, we have two choices in a layout ring: show just the notes list or show both the notes list and tag list. If we choose to only show the notes list, the most appropriate choice, and further press the keystroke for showing the tag list, we can now cycle to the tag-list-only view and hide the notes list, or more reasonably switch to the tag-and-note-list-visible view. of course, at this point we have just reached a place of inconsistency because the same keystroke has a different meaning than it does when in distraction-free zone.

drw158:
Good research on gathering those that helps. The screenshot I posted is from Ulysses.

The Day One app looks interesting, and I've been meaning to get it. I wish I could try it out with a trial. We could do a similar thing: one toggle to turn on/off both sidebars at the same time. Then, to hide the tags list, there would be another button within that column (if I understand the UI correctly).

I'll add some more to the mix:

**Safari**

Simple on/off sidebar toggle

screen shot 2016-11-28 at 3 12 03 pm

**Safari Dev Tools**

One toggle for separate window, one toggle for side or bottom.

screen shot 2016-11-28 at 3 19 11 pm

**Bear**

Seems to have the same setup as us. There is no UI for this, just keyboard or through the menu bar.

screen shot 2016-11-28 at 3 21 45 pm

**Github**

Single sidebar toggle, but what's interesting is that hiding the sidebar also hides the "Add" plus button.

screen shot 2016-11-28 at 3 26 20 pm
screen shot 2016-11-28 at 3 26 36 pm

Numbers

Not a toggle, but just a dropdown with 3 options.

screen shot 2016-11-28 at 3 33 59 pm

drw158:
To solve the shortcut problem I think we would just do what Bear and others have done, create a shortcut for each of the 3 states.

screen shot 2016-11-28 at 3 21 45 pm

dmsnell:
> one toggle to turn on/off both sidebars at the same time. Then, to hide the tags list, there would be another button within that column

can I reword this to see if I understand you properly?
 - one button to toggle distraction-free mode (editor only)
 - the tag list adopts its drawer toggle as exists in the Mac app already

dmsnell:
> do what Bear and others have done

I'm very curious to know why they went this route. to me it seems far more difficult to remember a layout shortcuts than it is to remember toggles, though I think `^1` `^2` `^3` is reasonable. it may just be that I don't work this way in any apps that I'm aware of.

dmsnell:
> hiding the sidebar also hides the "Add" plus button

this is interesting. it's mutating the menu bar and I wouldn't anticipate that. is there any analogue in Simplenote that we might want to hide or show contextually?

drw158:
> I'm very curious to know why they went this route. to me it seems far more difficult to remember a layout shortcuts than it is to remember toggles, though I think ^1 ^2 ^3 is reasonable. 

Sketch gets around this by having toggles for each shortcut, but then adding an additional shortcut that toggles both sidebars on/off

screen shot 2016-11-28 at 5 03 51 pm

drw158:
> this is interesting. it's mutating the menu bar and I wouldn't anticipate that. is there any analogue in Simplenote that we might want to hide or show contextually?

We could also hide the New Note button in editor only mode. Different circumstances, but we already do that in the mobile view.

drw158:
I think we should try these 3 options

This one actually reverts back to the layout in master. A new button is added in the editor to enable editor only mode.

fullsizerender-3

Layout control with one button.

fullsizerender-2

Similar to Day One, one button toggles both sidebars. A button within the notes list toggles the tags.

fullsizerender-1

dmsnell:
@drw158 I definitely like the expand button in the context of the editor pane for distraction-free mode

Copons:
@drw158 while I like it per se, I'd avoid the dropdown as it's not really a great UX on a touch device.

As for the other two versions:

- Distraction free mode looks great; I'm not totally sold on the moving button that makes frustrating to cycle between the two layouts.
Maybe it could be grouped with the other notes buttons on the right? I mean, all in all it _is_ note-related...

- The hybrid one seemed a bit confusing on paper, but the prototype actually seems to work pretty well. Maybe having two different buttons that modify the layout increases too much the complexity of the app though?

Unrelated: in this PR version, the searchbar _will_ end up in the note list as soon as I remember to do it when actually working on it. :)

drw158:
**Option 1**

I like this one the best I think, but has some downsides.

Pros:
- More vertical space for the notes. Less wasted space in the app toolbar in general.
- Icons are close to the area they control.

Cons:
- None of the layout buttons stay in the same place after toggling. We could fix the tags button by keeping the search bar static. Distraction free mode button could be moved to the right, but I'd like to separate it.
- Integration of the naive traffic light wouldn't look quite as native, but the inset traffic light work Dan has done looks nice.
- Not quite as elegant. It uses 2 buttons to control layout.
- New Note button is hidden in distraction free mode. Good thing?

**Option 2**

It does use only 1 button, but it takes 2 clicks to get to any layout option. Mobile shouldn't be a concern, since this all goes away on small screens anyway. Assuming the mobile is a small screen...

**Option 3**

Unconventional, but is it intuitive? Makes the tags seem like an extension of the notes list, rather than its own separate sidebar. Makes sense I think.

Playing around with these, I'm going to try and fix the jumping button problems on option 2 and 3.

Copons:
@drw158 
> Option 2: Mobile shouldn't be a concern, since this all goes away on small screens anyway. Assuming the mobile is a small screen...

It's not mobile per se: it's touch that imho doesn't work well with dropdowns (think about native mobile dropdowns, that are pretty much full screen modals - I doubt we'd want that on Simplenote 😄  ), and touch is not mobile-only anymore so we need to consider touch behaviors on any screen size.

drw158:
Updated option 3 with a v2. Tags button now does not move when you toggle it (ignore the small jumping when clicking between screens).

dmsnell:
@drw158 I like the way the toggling happens in option 3 v2 and agree that the wide search bar is a bit strange.

tag searching is something I could imaging us improving, noting #343 as well and the request to search for notes containing multiple tags `tag:simplenote tag:awesome`. the wider search bar could help there, maybe. /random

drw158:
Here is a new idea that combines some previous ones. The editor toggle is now on the right.

After making these prototypes, I think the best option would be to "soft-launch" the editor only mode by just creating a keyboard shortcut for it, no UI. Not _every_ view option has to be in the UI. I'm not sure how high the demand is for an editor only mode compared to those who won't use it. The editor is prime real estate, and we should be very judicious when adding buttons. The keyboard shortcut will satisfy that demand I think!

dmsnell:
@drw158 I love the idea of a soft launch. in this latest prototype I like how the search bar doesn't grow, but I feel like it's unexpected how it jumps to the right. it sits above the separating line

not sure how I feel about putting the distraction-free mode button next to the other icons. those icons all pertain to the specifically-opened note while this button is a UI-centric action maybe.

you're doing an amazing job on these.

drw158:
> not sure how I feel about putting the distraction-free mode button next to the other icons. those icons all pertain to the specifically-opened note while this button is a UI-centric action maybe.

Agreed on this.

drw158:
@Copons to move this PR forward, I think let's:

- not worry about the editor only mode in this PR
- update PR to function like prototype in https://github.com/Automattic/simplenote-electron/pull/443#issuecomment-264917056
- everyone try it out to get a feel for it, make changes based on feedback

The prototype is actually not that far off from what is currently in master. Sorry if that causes any backtracking!

Copons:
@drw158 no worries for backtracking, I definitely don't mind iterating since it gives me the chance to clean up my code. :)

Copons:
I've removed the "editor only" option while keeping all the logic (and style) behind, so that we should be able to add it back with minimum efforts.
cc @drw158, @dmsnell

drw158:
Nice, I like it. There are a few things I noticed that would improve it before we'd merge.

- Decreased tag sidebar width
- Search bar has no transition, so it's a weird jumping effect.
- We'd also need to toggle the info panel

Copons:
@drw158 Finally working on it!

> Decreased tag sidebar width

What do you mean? It should have the very same width as before.

> Search bar has no transition, so it's a weird jumping effect.

Fixed in the upcoming commit.

> We'd also need to toggle the info panel

Uh! I believe I've never considered this.
Let's see what I can do!

drw158:
By that, I mean that we should decrease the default sidebar width because now it will be always visible. I think it's a bit too wide right now, but it was ok before, because it was not always visible. Ideally of course, the user should be able to resize it (in a different PR :) )

Copons:
@drw158 You can take a look, if you want.

Though, the more I work on it, the more I feel the current layout structure is not really adequate for this kind of redesign.
The component tree could use some refactor to decouple components from each other.
See for example the Note Info, that is rendered in the app root, while it would probably make more sense (for this redesign, I mean) if it was inside the Note Editor container.

I should probably try moving _all_ the various pieces around, and then write something more structured about it.

dmsnell:
@copons and @drw158 what's the status on this PR? have we figured out which design moves we want to make?

Copons:
@dmsnell My effort to reduxify and modularize most of the app was in order to give me more freedom to move around stuff without infinite props inheritance chains.
I should re-assess the current refactor state and see what's still missing.

hfactor:
A single button that shows a drop down to choose between Editor Only, Editor + Note List, Editor Note List and Tag List would be the easiest way. Or even have that feature under Menu bar item view (and have some shortcuts)

rodrigoi:
Closing this PR as stale. @Copons feel free to reopen this any time.

Gatada:
This is a very good and important discussion. Would love to see that Option 1 come to fruition.

Issue #332, opened 2016-07-03
Data loss after client failed to synchronize


marianoju:
Simplenote wrote: 

> In the vast majority of cases, Simplenote’s synchronization works great. We handle millions of synchronization requests each day without any problems. But if you do lose your data, even just a single character, and even if it’s not our fault, we want to hear about it. The integrity of your notes is our top priority. Any form of data loss is completely unacceptable.
#### What happened so far
1. I have recently changed the email I use to sign in to my account (from m...@gmx... to m...@posteo...). My hunch is that this transition didn't go well. 
2. Over the course of the last days I have been writing long notes (mostly bug reports), using [the latest Simplenote client for Ubuntu v1.0.2.](https://github.com/Automattic/simplenote-electron/releases/download/v1.0.2/simplenote-1.0.2.deb).  
3. Yesterday I realized, that these notes did not show up on my Android phone (using the Simplenote client for Android v1.4.0), and found, that the the Simplenote client for Ubuntu was not synchronizing. 
4. I logged out the Simplenote client for Ubuntu (File > Preferences > Account > Log Out) and logged back in. 
#### What I expected

I expected the Simplenote client for Ubuntu to synchronize latest modified notes to the server and retrieve changes to other notes from there. 
#### What happened instead

The Simplenote client for Ubuntu started synchronizing from the top, and all _new_ notes that were off-sync have disappeared. They are _not_ to be found with the trash. Note history of _old_ notes does not reflect recent changes. 
### Questions

Where and how are these files stored locally with Ubuntu 16.04? Is there a chance of recovering these notes? I am currently searching locally (using grep) for hidden files containing a unique string used in one of my notes ("kidle_inject"). 
### Suggestion

If the email address change should have caused this client's behaviour, users ought to be advised and/or forcefully logged out of all instances, to avoid data loss. 
#### My Configuration
- Ubuntu 16.04.1 LTS (Xenial Xerus) 64-bit
- Kernel Linux 4.4.0-38-generic
- GNOME 3.18.3
- Memory: 7.5 GiB
- Processor: Intel® Core™ i5-3320M CPU @ 2.60GHz × 4

marianoju:
Since this issue has not received any reply in 19 days I have permanently given up on recovering the lost data. I have not been able to find out where the latest Simplenote client for Ubuntu v1.0.2. stores my notes locally. Support has been unable or unwilling to answer my questions in detail. For the record: 

[...]

Issue #323, opened 2016-06-14
Export individual notes


prat0088:
I wanted to save a specific note as a text file.  I clicked the share arrow expecting to see an export option.  No luck.  Such a feature would be useful.

dmsnell:
@prat0088 can you help me understand the value of this here? I thought about individual note exports, but since the notes are plain text a quick copy-all gets the note content ready to paste anywhere. I find that more useful than exporting a single note because in most cases I can imagine, exporting to a file would precede opening that file and copying and pasting the content elsewhere.

dmsnell:
@prat0088 I'm going to close this issue as I don't see us adding an option to export individual notes. in the app (since #448) you can export the entire account and each note will appear in the generated ZIP file, however. I hope that meets your need!

davidwolfpaw:
An example of where this would be useful: If I want to write something and share it with someone else in Markdown format. The version that can be shared through the app has HTML as opposed to the Markdown. To get the Markdown file I need to copypaste the content into a text editor and save it as a new file. As the alternative is downloading every note, unzipping the archive, and discarding all other notes after finding the one that I need, if incorporating an individual export feature isn't too onerous, I think that it would be useful.

dmsnell:
@davidwolfpaw I wonder if we're crossing terms with `export` and `publish`. We can publish notes and they will render the markdown source of the note if Markdown is enabled for that note. If it's disabled for that note then you can publish the markdown source. You could copy a note's contents into a new note and publish both versions if you wanted to.

Exporting will always save the unformatted source of the note which sounds like what you want. If you copy all of the text content from the note when you are in the Simplenote app though, that too should be the unformatted source.

Sorry I'm not understanding your case here - please do show me where I'm not getting it.

PR #448, opened 2016-11-30
Add Export (to zip) to app when running in Electron

dmsnell:
Addresses #332
Addresses #323

The Simplenote app receives updates from the Simperium library
(`node-simperium`) and stores those updates inside of `IndexedDB`. This
provides us with a fairly reliable local copy of the note data, but that
data is not very accessible from outside of the app.

The web app on simplenote.com offers a `.zip` archives of an account's
notes with one file per note but the client applications do not offer
any kind of an export.

This patch introduces an export function into the Electron app which can
be called to export a single _JSON_ file or a single _zip_ file with all the relevant note
information. This export contains metadata unavailable from the existing
export on the web.

<del>There is not piece of this built into the user interface yet and so this
PR introduces a feature for testing and development only. The export
function is made available in the global scope and can be called from
the running console as `exportNotes()` or as `exportToZip()`. 
It should trigger a download of the `notes.json` or `notes.zip` file.</del>

All notes are being exported but trashed notes and normal "active" notes
are separated as two different lists.

```js
exportNotes()
exportToZip()
```

Access via **File** > **Export**

Please see the `README.md` for additional information and an example 
of exported data.

cc: @roundhill @drw158 @rodrigoi @beaucollins

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -164,7 +164,7 @@ function createMenuTemplate( settings ) {
 		}, {
 			type: 'separator'
 		}, {
-			label: '&Export',
+			label: '&Export Notes',
 			accelerator: 'CommandOrControl+Shift+E',
 			click( item, focusedWindow ) {
 				if ( focusedWindow ) {
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -3,6 +3,7 @@
 const {
 	app,
 	BrowserWindow,
+	dialog,
 	ipcMain,
 	shell,
 	Menu
@@ -153,23 +154,38 @@ function createMenuTemplate( settings ) {
 	var fileMenu = {
 		label: '&File',
 		submenu: [ {
-				label: '&New Note',
-				accelerator: 'CommandOrControl+N',
-				click( item, focusedWindow ) {
-					if ( focusedWindow ) {
-						focusedWindow.webContents.send( 'appCommand', { action: 'newNote' } );
-					}
+			label: '&New Note',
+			accelerator: 'CommandOrControl+N',
+			click( item, focusedWindow ) {
+				if ( focusedWindow ) {
+					focusedWindow.webContents.send( 'appCommand', { action: 'newNote' } );
 				}
-			}, {
-				type: 'separator'
-			}, {
-				label: '&Print',
-				accelerator: 'CommandOrControl+P',
-				click( item, focusedWindow ) {
-					if ( focusedWindow ) {
-						focusedWindow.webContents.send( 'appCommand', { action: 'setShouldPrintNote' } );
-					}
+			}
+		}, {
+			type: 'separator'
+		}, {
+			label: '&Export',
+			accelerator: 'CommandOrControl+Shift+E',
+			click( item, focusedWindow ) {
+				if ( focusedWindow ) {
+					dialog.showSaveDialog(
+						focusedWindow,
+						{
+							title: 'Save export as .zip archive',
+							defaultPath: path.join( app.getPath( 'desktop' ), 'notes.zip'),
+						},
+						filename => focusedWindow.webContents.send( 'appCommand', { action: 'exportZipArchive', filename } )
+					);
 				}
+			}
+		}, {
+			label: '&Print',
+			accelerator: 'CommandOrControl+P',
+			click( item, focusedWindow ) {
+				if ( focusedWindow ) {
+					focusedWindow.webContents.send( 'appCommand', { action: 'setShouldPrintNote' } );
+				}
+			}
 		} ]
 	};
 
--- a/lib/utils/export/README.md
+++ b/lib/utils/export/README.md
@@ -1,0 +1,62 @@
+# Simplenote Export
+
+This module provides a function which can be used to export all of the notes for a given account.
+
+The output format is a single JSON file containing a map of two arrays: the active notes in an account; the trashed notes in an account.
+
+The notes will be sorted by last modification time with the most recently-modified notes appearing first in the list.
+
+Currently the export functionality is only available through the developer console and has no corresponding component in the main app.
+
+## How to export
+
+Open the developer console. This will require either a modified version of the app enabling the console or running the app locally in a development environment inside a browser.
+
+Creating an export is as simple as running the following command from the developer console since the export function is made available in the global namespace.
+
+```js
+exportNotes()
+```
+
+After the export finishes it will trigger a download in the browser window.
+
+## Example export
+
+```json
+{
+  "activeNotes": [ {
+    "id": "some-auto-generated-id",
+    "content": "Simplenote is so simple! I love it.",
+    "creationDate": "2016-11-18T23:28:42.957Z",
+    "lastModified": "2016-12-22T21:59:36.623Z"
+  }, {
+    "id": "some-auto-generated-id-that-is-unique",
+    "content": "This note can be accessed from a secret URL",
+    "creationDate": "2016-07-02T07:24:22.699Z",
+    "lastModified": "2016-11-29T16:55:49.890Z",
+    "publicURL": "http://simp.ly/p/short-url-string"
+  }, {
+    "id": "some-auto-generated-id-again",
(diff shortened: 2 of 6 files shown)


dmsnell:
@drw158 and @roundhill this is ready to go I think. we just need to follow-up with a PR to add the UI for it

drw158:
Can we add the UI for this only in the menu bar for now? I'm thinking Simplenote > Export Notes...

Opens a finder window with the .zip file shown. Maybe just export to Desktop?

dmsnell:
> Can we add the UI for this only in the menu bar for now? I'm thinking Simplenote > Export Notes...
> 
> Opens a finder window with the .zip file shown. Maybe just export to Desktop?

seems reasonable @drw158 - that would allow us to get it out quickly.

do you have thoughts on also providing the `.JSON` export? it has considerably more information than the `.zip` export

lib/utils/export/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## How to export
+
+Open the developer console. This will require either a modified version of the app enabling the console or running the app locally in a development environment inside a browser.
+
+Creating an export is as simple as running the following command from the developer console since the export function is made available in the global namespace.
+
+```js
+exportNotes()
+```
```
roundhill: Maybe add `exportToZip()` here in the README as well.

lib/utils/export/to-zip.js
```diff
@@ -0,0 +1,21 @@
+import JSZip from 'jszip';
+
+export const noteExportToZip = notes => {
+	const zip = new JSZip();
+
+	notes
+		.activeNotes
+		.forEach(
+			( { id, content } ) => zip.file( `${ id }.txt`, content )
```
roundhill: One of the things users complain about with the current zip export is that it uses the id as the file name. If there's a way that we can use the note's first line, that'd be cool. What we need to watch out for is notes with the same first line so we don't overwrite the file!

roundhill: Also users have requested that the tags be in the exported file as well. Maybe just at the bottom of the file? 

`Tags: ToDo, Shopping`

dmsnell: > If there's a way that we can use the note's first line, that'd be cool.

how should we handle duplicates and blank note titles?

for duplicates I could imagine `document.txt` and `document (1).txt` but for blank notes?

 - grab the first N-characters of content -> `third line of note maybe.txt`
 - `untitled.txt` and `untitled (1).txt`

dmsnell: > Also users have requested that the tags be in the exported file as well

I thought about adding them as hashtag strings. do you like `Tags: ToDo, Shopping` better than `#ToDo #Shopping`

roundhill: > I thought about adding them as hashtag strings

I like that 👍

roundhill: > grab the first N-characters of content -> third line of note maybe.txt

I like that options vs. `untitled`

lib/utils/export/README.md
dmsnell: added in 20721aa

dmsnell:
> I'm thinking Simplenote > Export Notes...

to confirm @drw158 you want this in the top-level app menu and not in **File**?

this may not be a thing in Linux or Windows but at least in Mac there's that distinction between the file menu and the one with preferences.

lib/utils/export/to-zip.js
dmsnell: changed naming in e29452f using the following algorithm:

 - strip out all unsafe characters from each line of the note
 - trim each line of the note
 - filter out lines that are now blank
 - take the first line, or `untitled` if the note is blank after these transformations
 - limit line to 40 characters
 - save as `.txt` unless there are duplicates, in which case we append `-(1)` or `-(2)` etc…

dmsnell: > Also users have requested that the tags be in the exported file as well. Maybe just at the bottom of the file?

Added list of tags to end of note in 9d11ff0.

@roundhill in this iteration I have explicitly declared the list of tags at the end with a header because it was a bit confusing to me in my tests just having the space-separated list of hashed-tags.

e.g.
```
Chocolate Cake

Take the eggs and beat them with sugar.
Pour in chocolate.
Bake.

Tags:
 - #cake
 - #chocolate
 - #kurz-gesagt
 - #recipes
```

dmsnell:
@drw158 I added the `Export` item to the `File` menu in 07a267b.

We can change it but I found `Export` listed under `File` in an Apple HIG document and so that is what I chose.

This PR now exports all notes to a zip file appending the list of tags at the end. It opens a _save file_ dialog to get the filename and defaults to `~/Desktop/notes.zip`.

desktop/app.js
```diff
@@ -153,23 +154,38 @@ function createMenuTemplate( settings ) {
...
-				accelerator: 'CommandOrControl+P',
-				click( item, focusedWindow ) {
-					if ( focusedWindow ) {
-						focusedWindow.webContents.send( 'appCommand', { action: 'setShouldPrintNote' } );
-					}
+			}
+		}, {
+			type: 'separator'
+		}, {
+			label: '&Export',
```
roundhill: How about `Export Notes`?

lib/utils/export/to-zip.js
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @example
+ * // returns note content unchanged
+ * appendTags( { content: 'the fleeb has the fleeb juice' } )
+ *
+ * @param {Object} note note object
+ * @returns {Object} augmented note whose
+ */
+const appendTags = note => note.tags
+	? { ...note, content: `${ note.content }\n\nTags:\n${ note.tags.map( tag => ` - #${ tag }` ).join( '\n' ) }` }
```
roundhill: If we are using hashtags, I would prefer that they are just listed one each per line with out the ` - `:

```
Tags:
#todo
#shopping
```

Thoughts? I'm actually backtracking a bit now on using the hash because we don't use it anywhere else in the apps or website.

dmsnell: This seems reasonable @roundhill though I wonder if we should comma-separate them then. Especially for notes with lots of tags the list gets looking a bit eccentric

desktop/app.js
dmsnell: I did start with `Export All Notes` but it seemed very odd in the menu against the other shorter entries. I'll have a look at `Export Notes`

lib/utils/export/to-zip.js
roundhill: > I wonder if we should comma-separate them

Sounds good 👍

drw158:
All of this looks good to me. I don't have an opinion on including the JSON. If it's just a matter of adding a .json file to the notes folder, then I'd say go for it!

dmsnell:
> If it's just a matter of adding a .json file to the notes folder, then I'd say go for it!

wow. great idea @drw158! I hadn't thought about that

lib/utils/export/to-zip.js
dmsnell: Updated in cf7e759.

Now the tags appear as an indented list…

```
Just a note

Tags:
  one, two, three, …, end-of-line
  next-line, moarTags, what?
```

there's no comma between lines. should we include it?

dmsnell:
source JSON included in 9c0380c @drw158 - thanks again for the suggestion!

also in this commit I turned on compression and platform-specific options to the export

roundhill:
> source JSON included in 9c0380c

Nice, I like that we've included the json too. How did `Export Notes` look in the menu?

dmsnell:
> How did Export Notes look in the menu?

not terrible, but still a bit odd to me. what do you think? I think it's okay.

screen shot 2017-01-04 at 2 11 23 pm

roundhill:
> not terrible, but still a bit odd to me. what do you think? I think it's okay.

I'm cool with it 👍 

Ready to :shipit: IMO!

PR #449, opened 2016-11-30
Hide native toolbar on macOS

roundhill:
For us brave souls using the macOS build of the app, this PR removes the toolbar and instead adds some padding to the top of the app for a very slick look:

screen shot 2016-11-30 at 1 23 56 pm

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -34,6 +34,7 @@ import {
 	includes,
 	isObject,
 	map,
+	matchesProperty,
 	overEvery,
 	pick,
 	values,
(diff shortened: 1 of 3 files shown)


lib/app.jsx
```diff
@@ -99,6 +99,11 @@ const isElectron = ( () => {
 	return () => foundElectron;
 } )();
 
+const isElectronMac = () => {
+	return isElectron() &&
+		get( window, 'process.platform', '' ) === 'darwin';
+}
```
roundhill: Cheers, I didn't realize that `process` was provided by Electron. Also updated to use `matchesProperty` 👍

dmsnell: I hope that you tested this since I didn't. `process` is usually set by node and I suppose that Electron itself is the source when running there.

roundhill: > I hope that you tested this since I didn't.

Yes, I did test it. In the web version on OS X the check fails and the styles aren't applied. 👍

PR #451, opened 2016-12-02
Limit App to a Single Instance

### LINKED ISSUE #376: Can open several, blank instances
If you launch Simplenote more than once by mistake or because you forget it's already open, all subsequent instances are blank.

Reported in #2814725-t

roundhill:
Fixes #376 by enforcing the app to run in a single instance. Took inspiration from the [guide on the topic here](https://github.com/electron/electron/blob/master/docs/api/app.md#appmakesingleinstancecallback).

**To Test**
Run the app in Windows, and minimize to the tray. Now run the app again by double clicking on the app icon. The app should maximize and remain in the single instance without a second instance ever appearing.

For the tester's convenience, here's a windows build with the patch applied: https://cloudup.com/caMB3LsO4Bz

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -67,6 +67,23 @@ module.exports = function main() {
 		} );
 	};
 
+	const shouldQuit = app.makeSingleInstance( () => {
+		if ( ! mainWindow ) {
+			return;
+		}
+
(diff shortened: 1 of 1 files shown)


dmsnell:
Tested and verified. Works like a charm @roundhill

Issue #385, opened 2016-09-17
Sync: Changes not propagating from other apps to Electron app (Windows, Linux)


Decadence:
Hello. I'm facing issue when Simplenote stops sync notes on Windows 10.
It happens when I don't open application on PC for a few days, but I edit notes from other devices. Then I open app on PC and see old notes.
Account is set in Preferences. Only solution I found is relogin. Seems like authentication has expired.
It's already detected on two different PC.

Decadence:
It would be nice to indicate somehow sync status.

rustykeyctrl:
Sry I'm a beginner here but really intrigued with git and for the most part don't know these things that's why I'm watching this thread maybe I can learn it as I'm learning git.
I won't summit any other then feedback to learn but my sync status is computer in the shop.
I'd appreciate any help and if it not ok to give any feedback just reports things I'll understand.
Thank you.
Have A Great Day
               Victor Evans
On Sep 17, 2016 2:38 AM, Victor <notifications@github.com> wrote:It would be nice to indicate somehow sync status.

neelkadia:
@Decadence I think you can merge the issue with #384 this one.

Decadence:
@neelkadia Hardly. I don't use web version at all.

hedgefield:
Same problem here. Was supposed to be better in 1.0.3 I think but I'm still seeing it. Only logging out and back in again updates the list properly.

rustykeyctrl:
Thanks for the tip I'll try that if the problem persist
Have A Great Day
               Victor Evans
On Sep 20, 2016 2:09 PM, Tim Hengeveld <notifications@github.com> wrote:Same problem here. Was supposed to be better in 1.0.3 I think but I'm still seeing it. Only logging out and back in again updates the list properly.

raniesantos:
I use Windows 7.
The desktop electron app of Simplenote always fails to pull down new notes.
When I edit *already existing notes* from Android app or the web app it syncs the changes perfectly fine.
But when I create a *new note* from Android or web, the electron app NEVER SYNCS THEM.

**PLEASE ADD A MANUAL SYNC BUTTON ON THE DESKTOP APP**

dmsnell:
[...]

PR #452, opened 2016-12-02
Use Latest node-simperium and Bump App Version

roundhill:
Use the latest node-simperium which includes a few important fixes related to sync, especially that reported in #385.


dmsnell:
:shipit:

Issue #454, opened 2016-12-05
macOS: click dragging the toolbar buttons move the window


drw158:
In native apps, click dragging a button does not drag the window.

Try clicking on a button (history, share, etc.), holding, and dragging the window.

Same for the search bar.

Issue #453, opened 2016-12-05
macOS: Can't drag login window


roundhill:
#449 broke the ability to drag the login window. It will need the `-webkit-app-region: drag;` css property added.

drw158:
This is working for me on latest master.

PR #456, opened 2016-12-08
Fix draggable areas on macOS

roundhill:
Fixes for #453 and #454 on macOS.

**To test**
* Run the app in electron on macOS.
* Sign out! Try dragging the window around. There should be an absolutely positioned div there that allows you to drag.
* Sign in! Try clicking on dragging on a button in the top toolbar or in the search field. The app should not drag.

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -478,6 +478,7 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			isAuthorized,
 		} = this.props;
 		const electron = get( this.state, 'electron' );
+		const isMacApp = isElectronMac();
 		const { settings, isSmallScreen } = this.props;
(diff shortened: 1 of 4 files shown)


dmsnell:
@roundhill - nice work on this and the borderless window!

I noted however that when logged out I expected to be able to click and drag the windows around on my desktop when the point of click was basically anywhere on the window other than the form elements themselves. This could mean clicking on the Simplenote logo or anywhere in the space around it and dragging the window. That behavior was not there, however, and I'm not sure if it's supposed to be there.

Other than that I think this is good and ready to :ship:, so I'll mark it ready and leave that decision to you on the dragging. Maybe @drw158 could also weigh in.

roundhill:
OK, I'm going to merge this for now and we can do another PR for the login form full-page drag if @drw158 is interested.

PR #458, opened 2016-12-09
Update modals

drw158:
Updating the design of the modals with small improvements. Overall, buttons and other text is smaller.

I also snuck in some css that targets all links and forces them to use the pointer cursor.

To test:

- Look at all the modals
- Make sure pointer only shows for web links

screen shot 2016-12-09 at 5 00 55 pm
d0wsvrrfb0caaaaaaelftksuqmcc

### DIFF
--- a/scss/general.scss
+++ b/scss/general.scss
@@ -5,6 +5,11 @@
 	box-sizing: border-box;
 }
 
+a[href^="http://"],
+a[href^="https://"] {
+	cursor: pointer; // Make all web links use pointer cursor
+}
+
 html, body {
 	height: 100%;
 }
--- a/scss/about.scss
+++ b/scss/about.scss
@@ -23,7 +23,7 @@
 }
 
 .about-top {
-	margin-bottom: 40px;
+	margin-bottom: 20px;
 	text-align: center;
 
 	svg {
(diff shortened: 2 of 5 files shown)


scss/general.scss
```diff
@@ -5,6 +5,11 @@
 	box-sizing: border-box;
 }
 
+a[href^="http://"],
+a[href^="https://"] {
+	cursor: pointer; // Make all web links use pointer cursor
+}
```
dmsnell: Is using the patterns here an indication that we have failed to use `<a />` appropriately in our app? Is there a reason they shouldn't all use the pointer?

drw158: This addition is to override the rule set right above this change:

```
*, 
*:before, 
*:after {
	cursor: default; // Make everything use default cursor
	box-sizing: border-box;
}
```

This was originally added because of a recommendation for building Electron apps, because native apps don't have pointer cursors at all. _Except_ for web links that open in a browser, those should use the pointer.

I'm not sure what web elements use the pointer cursor, but we could just do this:

```
button {
cursor: default;
}
```

dmsnell:
@drw158 this is all looking fine, is there any reason not to merge it?

drw158:
I don't think so. We can merge it.

dmsnell:
:shipit: @drw158

PR #464, opened 2016-12-20
Navigation Bar: complete refactor (ES6, Redux, CSS)

Copons:
### ⚠️ VERY WIP ⚠️

Create new `NavigationBar`, `TagList`, `EditableList` components, all almost
completely independent from the root `App` component (except for the
buckets).

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -17,7 +17,7 @@ import NoteInfo from './note-info'
 import NoteList from './note-list'
 import NoteEditor	from './note-editor'
 import SearchField from './search-field'
-import NavigationBar from './navigation-bar'
+import NavigationBar from './blocks/navigation-bar/index';
 import Auth from './auth'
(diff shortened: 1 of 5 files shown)


dmsnell:
@copons I like what you are doing with the props. Do you think it would be possible to do _just_ the prop/action moves in another PR that logically precedes this one? I think that would be stellar and could be done in a PR which doesn't introduce any functional changes to the app.

Copons:
@dmsnell I'm not sure I got what you mean with functional changes. 
The only differences right now are the ES6ification of the functions (and the moving of the components into their own folders), but other than that it's exactly the same as before.

dmsnell:
@copons what's the status on this PR? does it need updating, does it need splitting, does it need abandoning?

rodrigoi:
Closing this as stale. @Copons, feel free to reopen this any time.

PR #465, opened 2016-12-20
Fix: Lists should be prefixable by `+` bullets

dmsnell:
In #398 @nfcampos added the ability to autocomplete lists. However, only
`-` and `*` were chosen as valid list bullets. This doesn't match the
behavior in the other Simplenote apps and it's not our desired list
behavior.

This PR adds `+` as an option and in the process extracts some of that
validation into a helper function.

Additionally it resolves a small bug where space-prefixed bullets
weren't registering as lists.

### DIFF
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -5,7 +5,7 @@ import {
 	EditorState,
 	Modifier,
 } from 'draft-js';
-import { invoke, noop } from 'lodash';
+import { includes, invoke, noop } from 'lodash';
 
 function plainTextContent( editorState ) {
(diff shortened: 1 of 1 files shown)


roundhill:
Working well. One thing I noticed: If I start a new note and type a `*` and then enter, it gets deleted. Do you think someone would ever want just a star at the top of their note?

dmsnell:
@roundhill this is probably a thing someone would want, but for what it's worth, the Mac app exhibits the same behavior. if you want that you have to hit enter first, then go back up and add the star.

roundhill:
> but for what it's worth, the Mac app exhibits the same behavior

Fair enough! :shipit:

PR #470, opened 2016-12-22
NoteInfo: Redux refactor

Copons:
First step of the complete `NoteInfo` refactor.

In this PR we aim to move props and actions away from `app.jsx` and instead `connect` them to the component.

### DIFF
--- a/lib/note-info.jsx
+++ b/lib/note-info.jsx
@@ -1,11 +1,15 @@
 import React, { PropTypes } from 'react';
-import { includes } from 'lodash';
+import { includes, isEmpty } from 'lodash';
 import ToggleControl from './controls/toggle';
 import moment from 'moment';
 import CrossIcon from './icons/cross';
-import { isEmpty } from 'lodash';
+import { connect } from 'react-redux';
+import { actionCreators } from './flux/app-state';
+import { setMarkdown } from './state/settings/actions';
+//import { tracks } from './analytics'
+import filterNotes from './utils/filter-notes';
 
-export const NoteInfo = React.createClass( {
+const NoteInfo = React.createClass( {
 
 	propTypes: {
 		note: PropTypes.object,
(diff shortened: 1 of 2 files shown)


lib/note-info.jsx
```diff
@@ -1,11 +1,15 @@
...
 import CrossIcon from './icons/cross';
-import { isEmpty } from 'lodash';
+import { connect } from 'react-redux';
+import { actionCreators } from './flux/app-state';
+import { setMarkdown } from './state/settings/actions';
+//import { tracks } from './analytics'
+import filterNotes from './utils/filter-notes';
 
-export const NoteInfo = React.createClass( {
+const NoteInfo = React.createClass( {
```
dmsnell: why did we remove the `export`?

lib/note-info.jsx
```diff
@@ -1,11 +1,15 @@
...
-import { includes } from 'lodash';
+import { includes, isEmpty } from 'lodash';
 import ToggleControl from './controls/toggle';
 import moment from 'moment';
 import CrossIcon from './icons/cross';
-import { isEmpty } from 'lodash';
+import { connect } from 'react-redux';
+import { actionCreators } from './flux/app-state';
+import { setMarkdown } from './state/settings/actions';
+//import { tracks } from './analytics'
```
dmsnell: this needs to go before merging

dmsnell:
@copons I noticed a couple issues here, questioning the removal of the named export and noticing the commented-out import. after those update then this looks good. I'm not sure if you are otherwise done in this PR so I'll wait to hear back before I green-light it

lib/note-info.jsx
Copons: Dang, I put this here at some point because I was sure I saw it used in some of the imported functions, but eventually I was wrong and forgot about it. :/

Copons:
@dmsnell Both fixed. Sorry for those, side-effect of going auto-pilot on these refactors.

dmsnell:
:shipit:

PR #471, opened 2016-12-23
Update versions to jump to latest node

dmsnell:
Several dependencies had to be updated in order to jump to the latest
node - `7.3.0`. Additionally there were a few code changes required to
either fix bugs or update style to conform to the updated library
versions

**Testing**

I tested building the app inside and outside of Docker. I tried running in Electron and also running `npm start` and testing in the browser.

I have not attempted to build the releases but I had issues building a release before this in **master** as well. Can someone please build the release versions from this before we merge?


Copons:
Since we're updating modules too, I get these warnings on `npm install`:

`npm WARN deprecated node-uuid@1.4.7: use uuid module instead
npm WARN deprecated jade@1.11.0: Jade has been renamed to pug, please install the latest version of pug instead of jade
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated transformers@2.1.0: Deprecated, use jstransformer`

(haven't checked any, though, this is just a reminder)

dmsnell:
> it might be worth to update it.

there are some other updates available as well. I was trying to keep this minimal for the node update and thought about a follow-up PR for those. I think it's a good idea.

rodrigoi:
Side note: On MacOS, if using `n`, remove npm (`rm -rf /usr/local/lib/node_modules/npm`) before switching to the new version to fix the `Cannot find module 'internal/fs'` error when running `npm install`.

dmsnell:
> Side note: On MacOS, if using n, remove npm (rm -rf /usr/local/lib/node_modules/npm) before switching to the new version to fix the Cannot find module 'internal/fs' error when running npm install.

Thanks @rodrigoi - I never got this error but it's good to know how to fix it!

rodrigoi:
Releases look fine: https://cloudup.com/cxMYncmhy6h. I think we should bump the `version` of `package.json` before merging this. Should we bump the _minor_ version?

dmsnell:
> I think we should bump the version of package.json before merging this. Should we bump the minor version?

Not sure on this one. I don't think it has to happen here since this isn't even a release yet. Probably would be a good idea to make a big bump because we have major changes since `1.0.7`

PR #472, opened 2016-12-23
Update/other version bumps

dmsnell:
Follows up on #471 including versions that are less-related to the node version update.

Moves Simplenote Electron to **v1.1.0** because of the big changes which have occurred since the 1.0.7 release.

~**Do not merge** until #471 is merged~


Copons:
This could strip out the `jade`-to-`pug` change after https://github.com/Automattic/simplenote-electron/pull/477

package.json
```diff
@@ -41,11 +41,11 @@
...
     "eslint-plugin-react": "6.8.0",
     "html-webpack-plugin": "2.24.1",
-    "jade": "1.11.0",
-    "jade-loader": "0.8.0",
     "json-loader": "0.5.4",
     "node-sass": "4.1.1",
     "postcss-loader": "1.2.1",
+    "pug": "2.0.0-beta6",
+    "pug-loader": "2.3.0",
     "react-onclickoutside": "4.5.0",
```
Copons: Personal request related to the various refactors: react-onclickoutside 4.5.0 has a bug when used as an HOC (for class components) with React 15.
The issue was [fixed on 4.8.0](https://github.com/Pomax/react-onclickoutside/issues/79); I've used 4.9.0 in several different experiments and I couldn't find any bugs or weird behaviours.
4.9.0 is the last version supporting mixins, which makes it backward compatible to our current code.
Do you think it's possible to bump it here, to help me ES6-fy all the components that depend on this?

dmsnell: absolutely, though let's make sure we test it.

Copons: I've tried it on https://github.com/Automattic/simplenote-electron/pull/464.
Using as an HOC for the NavigationBar, while keeping it as a mixin for everything else, and I haven't noticed any error or incorrect behaviour.

(Also, just for information: that module is now at version 5.8.3, but since 5.0 it doesn't support mixins anymore.)

dmsnell: cool! go ahead and commit it to this branch. this means we should lock-down now and no longer rebase without coordinating with each other first

dmsnell:
Closing as old and possibly irrelevant.

PR #473, opened 2016-12-23
Introduce `jest` for testing components

dmsnell:
This patch introduces `jest` as a test suite, which provides a nice was
of testing React components with minimal configuration. `jest` provides
support for _snapshot_ testing which we can use to catch visual
regressions in components.

As of this commit there is an issue with `enzyme` not being able to find
dependencies. Instead of adding those dependencies to the project (which
are not needed otherwise) I have worked around this by entering the
`node_modules/enzyme` directory and running `npm install`. Before this
merges into **master** we should have this issue resolved.

cc: @nfcampos


lib/components/tag-chip/test.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+import TagChip from './';
+
+describe( 'TagChip', () => {
+	it( 'should select tag when clicked', () => {
+		const selectIt = jest.fn();
+
+		const chip = mount( <TagChip tag="spline" onSelect={ selectIt } /> );
+
+		expect( selectIt ).not.toHaveBeenCalled();
```
beaucollins: where does `expect` come from?

dmsnell: `expect` is provided by `jest` - no need for `chai`

dmsnell:
Thanks @rodrigoi. I wasn't sure why I needed to install `react-dom` but I did it as well anyway in 6fcf816.

PR #475, opened 2016-12-24
Packager: fixes error during packing the app for MacOS.

rodrigoi:
During one of the version updates, the `asar` and `icns` filename changed. This is preventing the app to be built for MacOS (`make package-osx`).

From now on,`make package-osx` should produce a valid `.app` folder and `.dmg` packages.

### DIFF
--- a/builder.js
+++ b/builder.js
@@ -38,6 +38,7 @@ var opts = {
 	overwrite: true,
 	asar: false,
 	sign: false,
+	prune: true,
 	'version-string': {
 		CompanyName: config.author,
 		LegalCopyright: config.copyright,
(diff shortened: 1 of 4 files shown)


builder.js
```diff
@@ -50,21 +51,22 @@ var opts = {
...
+// 			ignore.push( path.join( directory, client[key] ) );
+// 		}
+// 	}
+//
+// 	return ignore;
+// }
+//
+// opts.ignore = opts.ignore.concat( whitelistInDirectory( './node_modules', [ 'deep-equal', 'electron-window-state', 'jsonfile', 'mkdirp' ] ) );
+// opts.ignore = opts.ignore.concat( whitelistInDirectory( './dist', [ 'index.html', 'app.js', 'manifest.appcache' ] ) );
+// opts.ignore = opts.ignore.concat( whitelistInDirectory( './', [ 'package.json', 'dist', 'desktop', 'node_modules' ] ) );
```
roundhill: Should we just delete this commented out code, or do you think it will still be needed?

dmsnell: I'm in favor of removing them, or fixing the reasons they are commented

rodrigoi:
@roundhill @dmsnell I've updated the build script to package only the needed files and dependencies. Otherwise, everything is included on the packed release (including things on the `/releases` folder), potentially making the size of the distributable grow out of control.

To test, pack versions for Windows, macOS and linux. Executable versions should weight around 130mb, and `dmg`, `tar` and `exe` should weight around 45mb.

roundhill:
> To test, pack versions for Windows, macOS and linux. Executable versions should weight around 130mb, and dmg, tar and exe should weight around 45mb.

Confirmed that this is the case :shipit:

PR #476, opened 2016-12-24
Introduce `react-virtualized` to fix performance issues in note list

dmsnell:
@see https://github.com/bvaughn/react-virtualized

The note list has been a very slow component disproportionately slowing
down the rest of the app. There were a few reasons for this:

 - Every note was being rendered in the note list on every render. This
   meant that each note was processed rendered even if it not actually
   visible to the user.

 - Updating the search field quickly sorted the list exacerbating the
   issues with re-rendering all of the notes. Deferring to re-render, on
   the other hand, would obscure the search results.

`react-virtualized` only renders the notes in the note list which would
be visible and it automatically handles things like overscroll,
scrolling in general, and deciding how to update the list.

One challenge with Simplenote is that the rows in the note list have
varying heights depending on the content of the notes themselves. In
order to make the virtual list work properly, however, we must provide
the heights of each row.

In this patch we are supplying the height of each row by estimating the
number of lines of text in the note excerpt. This calculation is
**extremely fragile** and depends on many other tuned constants such as
font size, line height, and note list width. That means **this will
likely be a source of bugs and regressions** moving forward and **we
need to pay special attention to the row height calculation**.

In several places the behavior of the virtual list is not intuitive and
therefore there are places in the code which trigger recalculating the
row heights. This is necessary because the list is a very DOM-stateful
thing and does not synchronize normally with React as it handles prop
changes.

## Performance comparison

The following graphs are CPU profiles from within the Chrome DevTools with the test500 account recording while typing `p-a-r-a-g-r-a-p-h` in the search box. The spikes shows processing time doing note filtering and re-rendering.

Note the comparison in the amount of time spent _idle_.

**Before**
screen shot 2016-12-24 at 12 03 57 am

**After**
screen shot 2016-12-24 at 12 05 53 am

**After further optimizations**
In this final round for the PR I have built in a few performance optimizations. Before these, the row height recalculation was taking an average of 12ms on each render of the note list for the test account.

The first additional optimization was a simple cache of the row height based on note id, width, and note preview excerpt. This brought the height recalculation time down to an average of around 1ms per render.

The second optimization involved a simple cache of the preview excerpt based on note id and content. This brought the height recalculation time down to an average of around 0.2-0.5ms per render.

screen shot 2017-01-17 at 1 20 16 am

The list is now sufficiently fast enough that it doesn't really interrupt the render path, practically eliminating the lag when switching between notes and when searching in the list.


lib/utils/note-title.js
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Returns the title and excerpt for a given note
+ *
+ * @param {Object} note a note object
+ * @returns {Object} title and excerpt (if available)
+ */
+export const noteTitleAndPreview = note => {
+	const content = note && note.data && note.data.content;
+	const match = noteTitleAndPreviewRegExp.exec( content || '' );
```
beaucollins: I wonder what the performance difference of a `strpos( "\n" )` would be vs a RegEx.

dmsnell: @beaucollins in the past I have tried beating the RegExps in places like this and struggled to do it. It's actually a pretty basic RegExp and my guess is that it's insignificant, though I considered it.

lib/note-list.jsx
```diff
@@ -1,17 +1,104 @@
...
+// and is open to adjustment if it doesn't find
+// the proper balance between visual updates
+// and performance impacts recomputing row heights
+const TYPING_DEBOUNCE_DELAY = 70;
+
+const noteDisplayMaxHeights = {
+	comfy: 24 + 18 + 21,
+	condensed: 24 + 18,
+	expanded: 130,
+};
```
Copons: Wouldn't it be better if we use constants for those values, like:

```es6
const heights = {
  previewComfy: 21,
  previewExpanded: 88,
  title: 24,
  verticalPadding: 18,
};

const noteDisplayMaxHeights = {
  comfy: heights.title + heights.verticalPadding + heights.previewComfy,
  ...
};
```

Or maybe a simple comment explaining what those numbers represent at a first glance, without the need of testing and inspecting?

package.json
```diff
@@ -64,9 +64,11 @@
     "moment": "2.17.1",
     "promise": "7.1.1",
     "react": "15.4.1",
+    "react-addons-shallow-compare": "15.4.1",
```
Copons: Is this explicitly stated because it's required by react-virtualized? Because I can't see it used anywhere except in `webpack.config.dll.js`.
Is there a reason why PureComponent couldn't be used instead?

lib/note-list.jsx
dmsnell: agreed that they could use some refinement. thanks for marking this!

package.json
dmsnell: when I installed `react-virtualized` it informed me that this was a required peer dependency but would not be installed automatically. we can try again without it, but I believe that the build will fail.

dmsnell:
> The height calculation is a bit scary

maybe we could get @aduth or @isuelda to have a look. they have been working recently with `react-virtualized`

aduth:
Yeah, variable heights are tricky in `react-virtualized`. While a bit cumbersome, the approach I took in Calypso's `<PostSelector />` component was to [access and store the node's `clientHeight`](https://github.com/Automattic/wp-calypso/blob/ae2c8ff2c2bf09cfbe37daa9f89be038e18860cd/client/my-sites/post-selector/selector.jsx#L146-L163) as part of a `ref` function callback. When rerendering, if the item height is available, [it's used](https://github.com/Automattic/wp-calypso/blob/ae2c8ff2c2bf09cfbe37daa9f89be038e18860cd/client/my-sites/post-selector/selector.jsx#L217-L219), otherwise falling back to a computed height based on some predetermined constant. `<PostSelector />` also handles hierarchies of items, not sure if that's needed here but available for reference if so.

dmsnell:
> store the node's clientHeight as part of a ref function callbac

thanks @aduth. I had considered storing this once on note load and then on note update so that `getHeight` would be a simple property lookup. sounds like there wasn't an obviously better solution in Calypso.

iseulde:
Yes, I think it's best to cache the calculated row height per note, to avoid recalculating all other note heights, and invalidate it when the note changes. It's [possible](https://github.com/bvaughn/react-virtualized/blob/8.8.1/docs/List.md#recomputerowheights-index-number) to recalculate from a certain index, but all the following note heights will also be recalculated. There's no way to recalculate a specific row index' height and recalculate the position of the rows after it without also recalculating their height. Not sure if it's meant to be up to the implementor to cache this, or if there might be room for such a method in React Virtualized.

lib/utils/note-title.js
dmsnell: @beaucollins the performance here has been worked-around in the note list by some imperative caching. in the future I hope we can move this operation out of the rendering path and into the note update / note add path which would eliminate repeated calls.

lib/note-list.jsx
dmsnell: done - thanks for the nudge @Copons

dmsnell:
thanks for the feedback @iseulde and @aduth - after caching a couple of properties this has dramatically improved in its performance. I'm not a huge fan of the style of imperative caching at play in `note-list.jsx` but it is an appropriate trade-off in order to make the app snappy for now. the real fix involves some deeper reworking of how we manage our data I believe.

roundhill:
Let's :shipit:

PR #477, opened 2016-12-26
Removes jade

beaucollins:
Uses `weback.DefinePlugin` to provide config values.

The template builder script is no longer necessary. `./get-config` is now used to provide the necessary config values to the app.


beaucollins:
Still need to account for `<meta name="viewport" ` an `<link ... ` for the Google font css.

webpack.config.dll.js
```diff
@@ -3,6 +3,8 @@ const webpack = require( 'webpack' );
 module.exports = {
 	context: process.cwd(),
 
+	devtool: '#source-map',
```
dmsnell: why this change?

get-config.js
```diff
@@ -0,0 +1,20 @@
...
+	try {
+		config = require( './config' );
+	} catch ( e ) {
+		console.warn( 'no configuration file present' );
+	}
+
+	return config;
+};
+
+function getConfig() {
```
dmsnell: what if we passed in the file path as an argument instead of hard-coding `./config`? would we have a good reason to develop or test with multiple config files?

get-config.js
```diff
@@ -0,0 +1,20 @@
+function readConfig() {
+	var config = {};
+
+	try {
+		config = require( './config' );
+	} catch ( e ) {
+		console.warn( 'no configuration file present' );
```
dmsnell: would this not be a reason to kill the program entirely?

package.json
```diff
@@ -24,6 +24,7 @@
     "npm": "3.10.10"
   },
   "devDependencies": {
+    "add-asset-html-webpack-plugin": "^1.0.2",
```
dmsnell: can we remove the `^` and pin this version?

webpack.config.js
```diff
@@ -34,14 +36,17 @@ module.exports = {
...
-			templateContent: require( './index-builder.js' ),
-			inject: false
+			title: 'Simplenote'
 		} ),
 		new webpack.DefinePlugin( {
 			'process.env.NODE_ENV': JSON.stringify(
 				process.env.NODE_ENV || 'development'
-			)
+			),
+			config: JSON.stringify( config() )
```
dmsnell: why not just store the config as a JSON file instead of creating a JS file which exports JS data which then gets serialized as JSON?

get-config.js
beaucollins: I was considering the same thing.

package.json
beaucollins: That or shrinkwrap?

webpack.config.js
beaucollins: I was hoping to add the ability to read in `git describe` to the config. The `require` in `get-config` would already pick up a `config.json` if one was defined.

I prefer configuration scripts as js instead of `JSON` usually.

beaucollins:
It's mostly just to remove jade.

We can easily get rid of the new dependency by using ejs which already comes along with `HtmlWebpackPlugin`.

beaucollins:
I think removing the `<link />` tag to Google is better than relying on a network resource for our UI when packaging for electron.

beaucollins:
@dmsnell do you know the reason behind using the `DLL` webpack plugins? If we used a single webpack config with multiple `entry` configs the `HtmlWebpackPlugin` would pick up the scripts altogether which removes the need for `add-asset-html-webpack-plugin`.

package.json
dmsnell: are you opposed to pinning?

I've personally found it helpful in that version updates have specific testable PRs.

I guess shrink wrapping can have the same effect. thoughts?

webpack.config.js
dmsnell: what do you mean by reading in `git describe`? I don't think I"m familiar with it in this context.

dmsnell:
> I think removing the <link /> tag to Google is better than relying on a network resource for our UI when packaging for electron.

such as for pulling in and "vendoring" the font?

dmsnell:
> @dmsnell do you know the reason behind using the DLL webpack plugins? If we used a single webpack config with multiple entry configs the HtmlWebpackPlugin would pick up all the scripts which removes the need for add-asset-html-webpack-plugin.

In #440 I added the DLL config to make incremental builds faster during development work. Since the base libraries are also relatively static compared to the app code I also considered it a benefit that we might have some separate vendor bundle which is long-term cacheable.

Besides performance issues, one of the things that bothered me was webpack showing us building close to 1000 modules on every change. It polluted the console with a bunch of messages I didn't care about hiding the ones I did.

webpack.config.dll.js
beaucollins: More pleasant debugging in the dll file.

beaucollins: Before:
screen shot 2016-12-28 at 12 18 41 am

After:
screen shot 2016-12-28 at 12 20 46 am

webpack.config.js
beaucollins: I like embedding the state of git during the build into the product. There are many ways of doing this. One way would be making a config script that shells out and runs `git describe --always --dirty` and uses that as the version.

Mine right now: `v1.0.6-26-g40890ca-dirty`

dmsnell: interesting. what's the point of this? is it simply to identify the exact commit which built the bundle?

beaucollins: Mostly I was trying to make minimal changes to the config portion while using the `DefinePlugin` which didn't allow using a function.

beaucollins: If something is released wether a custom build or production build we know the state of the source when it was built.

beaucollins:
@dmsnell at this point I'm mostly just changing from jade to ejs.

My hope was to remove the need of a custom template but the `HtmlWebpackPlugin` does a good job of working around customizations.

To keep this PR cleaner I can probably remove the config changes and have the ejs template handle them the same way as jade template did.

views/app.jade
```diff
@@ -1,15 +0,0 @@
-doctype html
-html(manifest=o.htmlWebpackPlugin.files.manifest)
-  head
-    meta(charset="utf-8")
-    title= o.htmlWebpackPlugin.options.title || 'Simplenote'
-    //- iOS webviews do not support indexedDB!
-    //- meta(name="apple-mobile-web-app-capable" content="yes")
```
dmsnell: @beaucollins I almost feel like this is an important-enough it of knowledge to keep in the template so that nobody accidentally tries to add it back in. how do you feel about it? can we even write comments in ejs?

beaucollins: You can do HTML comments. It's probably worth seeing if iOS 10 still has this limitation.

beaucollins: Might be possible with [iOS 10](http://caniuse.com/#search=indexeddb).

dmsnell: sweet. let's just remove this then; though a comment here in the PR/commit would probably be valuable

webpack.config.js
dmsnell: > I prefer configuration scripts as js instead of JSON usually.

I can accept this, though if you were to provide some rationale beyond "I like it" that would be even better :wink:

get-config.js
```diff
@@ -0,0 +1,21 @@
+function readConfig() {
+	try {
+		return require( './config' )();
+	} catch ( e ) {
+		console.error(
+			'Could not read in the required configuration file.\n' +
+			'This file should exist as `config.js` inside the project root directory.\n' +
+			'Please consult the project README.md for further information.\n'
+		);
+		process.exit( 0 );
```
dmsnell: @beaucollins I went ahead and instructed the process to exit here since the app won't boot without the config message.

get-config.js
```diff
@@ -0,0 +1,21 @@
+function readConfig() {
+	try {
+		return require( './config' )();
```
dmsnell: @beaucollins I'm not sure what kind of config file you are using, but the one I had (which matches what's in the project's README needs to be executed in order to return its data. Therefore I went ahead and turned this into a call on the import. If the file is missing or if we could not read it then this will fail and kill the build process.

lib/index.ejs
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="Content-type" content="text/html; charset=utf-8"/>
+    <title><%= htmlWebpackPlugin.options.title %></title>
+    <meta name="app-build-reference" content="<%= htmlWebpackPlugin.options[ 'build-reference' ] %>">
+    <meta name="node-version" content="<%= htmlWebpackPlugin.options[ 'node-version' ] %>">
+    <meta name="build-platform" content="<%= htmlWebpackPlugin.options[ 'build-platform' ] %>">
```
dmsnell: @beaucollins here I have added what I think are three important indicators of the build generator based on your comments in this PR. Please feel free to change them and add or take them away.

webpack.config.js
dmsnell: > I like embedding the state of git during the build into the product. There are many ways of doing this. One way would be making a config script that shells out and runs git describe --always --dirty and uses that as the version.

added in 812b7a7

dmsnell:
@beaucollins hopefully you won't mind that I added a couple commits onto this PR. I was trying to cleanup the bottom of the PR list tonight so we can get some of these out.

if you have any further changes you want or if you want to undo some of the things I did feel free. I apologize because I rebased before I thought that you might have local changes.

otherwise this one looks good to me. I'm going to mark it as ready to merge but you can make the call if you don't feel it is :smile:

get-config.js
beaucollins: Should it exit with a failure code since webpack would fail to build?

dmsnell: I should put a comment in there, or fail with an exit code.

Although I started with `process.exit( 1 );` it was spewing out a bunch of `npm` errors because of the exit code and the explanation then got lost. This was a pragmatic compromise to make sure the message was clear and visible.

Thoughts?

beaucollins: I feel we should exit with an error so it will work correctly with other build tools. For instance running the following would confusingly boot up electron with either: 1) a broken React app or 2) the last succesfully built version excluding any changes you have made.

```
npm run build && electron .
```

Where as exiting with an error code would prevent `electron` from starting up.

Yes, `npm` is noisy in failure mode, but I think `npm` users have to put up with that.

webpack.config.js
beaucollins: If `config` is not expected to export a function, we can use `js` or `json` or any other webpack resolvable file.

I prefer `js` because it allows for things like variables and functions and turing completeness.

beaucollins: I suppose we can perform a `typeof` check and execute it if it's a function to remain backwards compatible with existing config setups.

get-config.js
beaucollins: I changed mine to export an object literal.

dmsnell: thanks @beaucollins. how about I suggest that we either stick with the documented format or update the project README?

get-config.js
dmsnell: very fair. ugly but fair :wink:

get-config.js
beaucollins: @dmsnell to be backwards compatible we can use `typeof` and execute if it's a function but still update the README to show exporting of an object literal.

dmsnell: > to be backwards compatible we can use `typeof` and…

boo! :wink:

@beaucollins do we have a reason to change it? if we do, I'd rather just force everyone to change it; or, I don't see any compelling reason to be loose here with our input.

beaucollins: We can leave it as it is in `develop`.

Doing this change simply allows for `json` or `js` configs and matched how the webpack config works.

get-config.js
dmsnell: returns an exit code of `1` in b842094

get-config.js
```diff
@@ -7,7 +7,7 @@ function readConfig() {
 			'This file should exist as `config.js` inside the project root directory.\n' +
 			'Please consult the project README.md for further information.\n'
 		);
-		process.exit( 0 );
+		process.exit( 1 );
```
beaucollins: My linter doesn't like this:

```
⚠ at line 10 position 4:  Don't use process.exit(); throw an error instead. (no-process-exit)
```
http://eslint.org/docs/rules/no-process-exit.html

beaucollins: But I feel it's use here is perfectly reasonable.

dmsnell: Threw up in 80e3553

beaucollins: :barf:

lib/index.ejs
```diff
@@ -0,0 +1,15 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+    <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,300italic,400,400italic,600,600italic" />
```
beaucollins: @drw158 are we using Source Sans Pro anymore? Can we remove this reference? It's a pretty hefty network payload.

drw158: Nope, it can be removed.

PR #479, opened 2016-12-28
NoteEditor: Redux refactor

Copons:
First step of the complete `NoteEditor` refactor.

In this PR we aim to move props and actions away from `app.jsx` and instead `connect` them to the component.

_Note_: ~~to refactor `RevisionSelector` it's necessary to move the `NoteEditor` state into Redux.
This may be out of scope for this PR, though, and also could conflict with https://github.com/Automattic/simplenote-electron/pull/402.~~
Eventually I got around doing it. See https://github.com/Automattic/simplenote-electron/pull/479/commits/ec67f0f25c88776f79577d31392879e248b1ab60 for more info.

## Components refactor checklist

- [x] `NoteEditor`
- [x] `RevisionSelector`
- [x] `NoteToolbar`
- [x] `NoteDetail`
- [x] `TagField`
- [x] `ModeBar` (new)
- [x] `revision` Redux state subtree (new)

## Clean up

- [ ] Revision state subtree discussions
- [x] DRY-fy or Reduxify the note/revision selector

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -1,15 +1,20 @@
 import React, { PropTypes } from 'react';
+import { connect } from 'react-redux';
 import highlight from 'highlight.js';
 import marked from 'marked';
 import { get, debounce, invoke } from 'lodash';
 import analytics from './analytics';
 import { viewExternalUrl } from './utils/url-utils';
 import NoteContentEditor from './note-content-editor';
+import appState from './flux/app-state';
+import filterNotes from './utils/filter-notes';
+
+const { updateNoteContent } = appState.actionCreators;
 
 const saveDelay = 2000;
 const highlighter = code => highlight.highlightAuto( code ).value;
 
-export default React.createClass( {
+export const NoteDetail = React.createClass( {
 
 	propTypes: {
 		note: PropTypes.object,
@@ -74,7 +79,6 @@ export default React.createClass( {
 
 	render: function() {
 		const {
-			filter,
 			fontSize,
 			previewingMarkdown,
 		} = this.props;
@@ -101,7 +105,6 @@ export default React.createClass( {
 						<NoteContentEditor
 							ref={ this.saveEditorRef }
 							content={ content }
-							filter={ filter }
 							onChangeContent={ this.queueNoteSave }
 						/>
 					</div>
@@ -110,3 +113,25 @@ export default React.createClass( {
 		);
 	},
 } );
+
+const mapStateToProps = ( {
+	appState: state,
+	revision: { selectedRevision },
+	settings: { fontSize, markdownEnabled },
+} ) => {
+	const filteredNotes = filterNotes( state );
+	const noteIndex = Math.max( state.previousIndex, 0 );
+	const note = state.note ? state.note : filteredNotes[ noteIndex ];
+	const revision = selectedRevision || note;
+	return {
+		fontSize,
+		note: revision,
+		previewingMarkdown: markdownEnabled && state.editorMode === 'markdown',
+	};
+};
+
+const mapDispatchToProps = ( dispatch, { noteBucket } ) => ( {
+	onChangeContent: ( note, content ) =>
+		dispatch( updateNoteContent( { noteBucket, note, content } ) ),
+} );
+export default connect( mapStateToProps, mapDispatchToProps )( NoteDetail );
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -79,6 +79,7 @@ export const NoteDetail = React.createClass( {
 
 	render: function() {
 		const {
+			filter,
 			fontSize,
 			previewingMarkdown,
 		} = this.props;
@@ -103,9 +104,10 @@ export const NoteDetail = React.createClass( {
 						style={ divStyle }
 					>
 						<NoteContentEditor
-							ref={ this.saveEditorRef }
 							content={ content }
+							filter={ filter }
 							onChangeContent={ this.queueNoteSave }
+							ref={ this.saveEditorRef }
 						/>
 					</div>
 				) }
@@ -123,6 +125,7 @@ const mapStateToProps = ( {
 	const previewingMarkdown = get( revision, 'data.systemTags', '' ).indexOf( 'markdown' ) !== -1
 		&& state.editorMode === 'markdown';
 	return {
+		filter: state.filter,
 		fontSize,
 		note: revision,
 		previewingMarkdown,
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -117,16 +117,18 @@ export const NoteDetail = React.createClass( {
 const mapStateToProps = ( {
 	appState: state,
 	revision: { selectedRevision },
-	settings: { fontSize, markdownEnabled },
+	settings: { fontSize },
 } ) => {
 	const filteredNotes = filterNotes( state );
 	const noteIndex = Math.max( state.previousIndex, 0 );
 	const note = state.note ? state.note : filteredNotes[ noteIndex ];
 	const revision = selectedRevision || note;
+	const previewingMarkdown = get( revision, 'data.systemTags', '' ).indexOf( 'markdown' ) !== -1
+		&& state.editorMode === 'markdown';
 	return {
 		fontSize,
 		note: revision,
-		previewingMarkdown: markdownEnabled && state.editorMode === 'markdown',
+		previewingMarkdown,
 	};
 };
 
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -7,7 +7,7 @@ import analytics from './analytics';
 import { viewExternalUrl } from './utils/url-utils';
 import NoteContentEditor from './note-content-editor';
 import appState from './flux/app-state';
-import filterNotes from './utils/filter-notes';
+import getNote from './utils/get-note';
 
 const { updateNoteContent } = appState.actionCreators;
 
@@ -119,10 +119,7 @@ const mapStateToProps = ( {
 	revision: { selectedRevision },
 	settings: { fontSize },
 } ) => {
-	const filteredNotes = filterNotes( state );
-	const noteIndex = Math.max( state.previousIndex, 0 );
-	const note = state.note ? state.note : filteredNotes[ noteIndex ];
-	const revision = selectedRevision || note;
+	const revision = selectedRevision || getNote( state );
 	const previewingMarkdown = get( revision, 'data.systemTags', '' ).indexOf( 'markdown' ) !== -1
 		&& state.editorMode === 'markdown';
 	return {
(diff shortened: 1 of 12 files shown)


dmsnell:
This is looking good @Copons - I need to test it still but the code looks fine. Moving in the right direction.

Copons:
@dmsnell I've noticed that the revision handling impacts more than what I initially expected.
I've ended up creating a new, barebones, `revision` subtree in order to start Reduxifying the remaining children components.
Though, this might increase this PR size to the point of becoming un-reviewable.

At this point, I'd say there are two options:
1. Keep on working in this PR until everything is Reduxified.
2. Stop here, revert https://github.com/Automattic/simplenote-electron/pull/479/commits/ec67f0f25c88776f79577d31392879e248b1ab60 (the commit where I've changed the state), move it in another PR, and merge this one.

As you can guess, I'm eager to go with 1, but since I won't be the one reviewing it, well... I'll wait for suggestions!

lib/state/revision/reducer.js
```diff
@@ -0,0 +1,23 @@
...
+} from '../action-types';
+
+export const isViewingRevisions = ( state = false, { type, isViewing } ) =>
+	REVISION_IS_VIEWING === type ? isViewing : state;
+
+export const selectedRevision = ( state = null, { type, revision } ) =>
+	REVISION_SELECT === type ? revision : state;
+
+export default combineReducers( {
+	isViewingRevisions,
```
dmsnell: could not the boolean value of whether or not we are viewing a revision be captured in the presence or absence of state in the `selectedRevision`?

I'm not sure we need two values here when we could probably pull this out from a selector.

Copons: `isViewingRevisions` toggles `<RevisionSelector>`
`selectedRevision` contains the currently selected revision that is displayed in the note content, which can be either a revision or `null` for when the selected revision is the current version of the note.

This mimics the current `<NoteEditor>` [internal state](https://github.com/Automattic/simplenote-electron/blob/master/lib/note-editor.jsx#L46):
```js
{
   revision: null,
   isViewingRevisions: false
}
```

The former could be better renamed. Maybe something like:
`isRevisionSelectorOpen` / `REVISION_SELECTOR_OPEN`

dmsnell: aha! that makes sense. how do you feel about putting `isViewingRevisions` inside the `state.ui` subtree? I was using that for the tags to indicate UI-level state in contrast to data-object-level state.

Copons: I feel that's the way to go, but I didn't want to create any conflicts with #402, and still keep the behaviour identical to what it is right now but just replacing the internal state with Redux.

dmsnell: actually. isn't this kind-a related to the `isVisible` collection we have used for the other dialogs and modals?

it seems to me like an independently viewable dialog that could potentially have different visual designs while still keeping the prime notion of `visible` or `hidden`.

on the contrary, it does _not_ make sense to have it `visible` if the note editor is hidden, which would mean we have encoded an impossible state into our app.

maybe we could still go back to using one value to represent both things:

**open the revision slider**
 - issue `OPEN_REVISION_SLIDER`
 - something intercepts that action, loads in the revisions, then finds the newest one
 - that something issues a new `SELECT_REVISION` action to select the newest revision
 - reducer sets non-`null` value for `selectedRevision` indicating that slider should display

**cancel the revision selection**
 - issue `CLOSE_REVISION_SLIDER`
 - reducer sets `null` value for `selectedRevision` indicating that slider should hide

**select a revision from the slider**
 - issue `SELECT_REVISION`
 - reducer sets non-`null` value for `selectedRevision`, slider stays open

**close slider, selecting revision**
 - issue `SELECT_REVISION`
 - reducer sets non-`null` value for `selectedRevision`, slider stays open
 - issue `CLOSE_REVISION_SLIDER`
 - reducer sets `null` value for `selectedRevision` indicating that slider should hide

Copons: I guess we should start with the assumption that the newest revision is... the note itself!
In the current code the object that goes into `<NoteDetail>` is even called [revision](https://github.com/Automattic/simplenote-electron/blob/master/lib/note-editor.jsx#L90), and it's the selected revision (if it exists) or the note.
While I'm not sold on the naming confusion, I dig the simplicity of the logic behind this.

So, if we want to have only one `state.revision` reducer, we could simplify it like this (and notice that I'd keep the action names starting with the feature, to improve their searchability):

**default state**
- `selectedRevision = null`
- `<RevisionSelector>` hidden

**open the revision selector**
- dispatch `REVISION_SELECT` -> `selectedRevision = note`
- `<RevisionSelector>` visible

**close the revision selector** (cancel or click outside)
- dispatch `REVISION_SELECT` -> `selectedRevision = null`
- `<RevisionSelector>` hidden

**select a revision**
- dispatch `REVISION_SELECT` -> `selectedRevision = revision`
- `<RevisionSelector>` visible

**restore a revision** (and close the revision selector)
- `note = selectedRevision`
- dispatch `REVISION_SELECT` -> `selectedRevision = null`
- `<RevisionSelector>` hidden

**change note**
- dispatch `REVISION_SELECT` -> `selectedRevision = null`
- `<RevisionSelector>` hidden

---

This way, the pattern to toggle the `<RevisionSelector>` visibility is clear:
```jsx
{ selectedRevision && <RevisionSelector> }
```
and we'll spare the toggle visibility reducer (`isViewingRevisions`), that I feel might be a little out of place in `state.ui`, since it's so strictly tied to the note editor itself.

Copons:
Updated after https://github.com/Automattic/simplenote-electron/pull/389 and Reduxified the new additions.

lib/tag-field.jsx
```diff
@@ -194,6 +196,7 @@ const mapStateToProps = ( {
 	const note = getNote( state );
 	const revision = selectedRevision || note;
 	return {
+		allTags: state.tags.map( property( 'data.name' ) ),
```
dmsnell: Keep in mind that we are rebuilding the tag system in #402

Copons: Yeah, this last commit was to solve some conflicts and keep them in check with how the rest of the PR was written.
It didn't make sense to me to have only the `allTags` prop to be passed all the way from `app.jsx` while everything else is directly mapped from the state.

Once (or "if"? I recall some talk of closing it?) #402 is completed and merged, I'll definitely make sure to update this to reflect your changes.

dmsnell: Yep, just be careful about working too far ahead of other existing work. It tends to not save us as much time as we might think it could.

Copons:
Rebased against https://github.com/Automattic/simplenote-electron/pull/424

@dmsnell Am I correct in the assumption that search highlighting does not apply to the markdown preview?

dmsnell:
> @dmsnell Am I correct in the assumption that search highlighting does not apply to the markdown preview?

for now @Copons that is true. it's not just a technical issue but a decision to choose to make that happen so we're tabling it in the meantime

Copons:
@dmsnell Got it.
Just asked because the rebase conflicts were all pretty much related to the highlighting filter and, while by the code was pretty clear it was not applied to the preview, it also felt it could have been a regression caused by me.

dmsnell:
@Copons let's rebase this

Copons:
@dmsnell Rebased!

lib/note-detail.jsx
```diff
@@ -99,14 +104,36 @@ export default React.createClass( {
 						style={ divStyle }
 					>
 						<NoteContentEditor
-							ref={ this.saveEditorRef }
 							content={ content }
 							filter={ filter }
 							onChangeContent={ this.queueNoteSave }
+							ref={ this.saveEditorRef }
```
dmsnell: is this change necessary? I know they are out of order alphabetically but I tend to put `ref` at the top as a pattern since it's a special prop. in any case, I think this change is unnecessary in this PR

lib/note-detail.jsx
```diff
@@ -99,14 +104,36 @@ export default React.createClass( {
...
 } );
+
+const mapStateToProps = ( {
+	appState: state,
+	revision: { selectedRevision },
+	settings: { fontSize },
+} ) => {
+	const revision = selectedRevision || getNote( state );
+	const previewingMarkdown = get( revision, 'data.systemTags', '' ).indexOf( 'markdown' ) !== -1
+		&& state.editorMode === 'markdown';
```
dmsnell: this is a weird dangle, also how about using something more semantic like `includes( get( revision, 'data.systemTags', '' ), 'markdown' )`

we could extract this into bindings or use parenthesis to clean it all up…

```js
const isMarkdownNote = includes( get( revision, 'data.systemTags', '' ), 'markdown' );
const isPreviewing = isMarkdownNote && state.editorMode == 'markdown';
```

lib/utils/get-note.js
```diff
@@ -0,0 +1,13 @@
...
+import filterNotes from './filter-notes';
+
+export default function getNote( state ) {
+	if ( state.note ) {
+		return state.note;
+	}
+	const filteredNotes = filterNotes( state );
+	const noteIndex = Math.max( state.previousIndex, 0 );
+	return filteredNotes[ noteIndex ];
+}
```
dmsnell: seems like this could have a better name. is it getting _the current_ note or the _selected_ note or the _active_ note?

dmsnell:
@Copons this is all looking good but I'm getting lost in how many code changes there are. is it possible to break this down any further? things such as introducing the note selector into the existing code, extracting only a single component per-PR, moving things into `connect()` on the parent before extracting the components…?

as I know this makes the development burden higher I apologize for asking but I think it's hard to assess what all needs testing with so many different kinds of changes and operations being moved around. it would be much easier to focus on each thing one at a time and confirm the changes.

if not, c'est la vie, but it would definitely be helpful for review and deploying…

lib/utils/get-note.js
Copons: I'm a little rusty on this, but IIRC it's _all_ of them.
Like:
- if there's a selected note → return the selected note
- else if there's a note list index stored → return the note corresponding to that index (if I remember this right, this is needed when you delete a note, to send you back to a previously opened note)
- else → return the first note

Maybe `getCurrentNote` might be more apt, but it wouldn't really clear the meaning of this function.
Probably the best approach would be a good old comment.

Copons:
>is it possible to break this down any further?

@dmsnell I could give it a try, but it would lead to a decent amount of different PRs that must be written one after the previous is merged, increasing also the review burden.
I don't mind starting from scratch here, using this PR as a reference, so I can begin if/when it's ok for you.

lib/utils/get-note.js
dmsnell: how about `getActiveNote()`?

it's just that I would expect `getNote()` to take some kind of id parameter and return a specific note

dmsnell:
> it would lead to a decent amount of different PRs that must be written one after the previous is merged, increasing also the review burden.

on the contrary, I think this would dramatically lower the review burden. it's not many PRs that are hard to review, but big ones that involve lots of different types of changes and testing

if you don't mind I think it would be great to break it apart. often I do this when working on a big piece. I'll code it up in a type of prototype but then walk backwards after collecting it into smaller chunks. each chunk makes it a little better overall, plus I revise my large PR and it becomes smaller…

rodrigoi:
Closing this as `invalid`.  @Copons, feel free to reopen this in the future.

Issue #469, opened 2016-12-22
Copying and pasting a note carries dark mode background to pasted text.


chrisfromthelc:
Reported by a user via email.

When using Simplenote in dark mode, copying text from Simplenote and pasting into other applications (in this instance, web-based Gmail message windows) results in the pasted text carrying over the dark background from Simplenote.

#### Steps to reproduce
1. Switch to the dark theme.
2. Compose some text.
3. Copy from Simplenote and paste in Gmail

#### What I expected
Pasting should only paste the text, not formatting.

#### OS version
Windows 10, Chrome/Firefox

#### Screenshot / Video
![simplenote copy paste](https://cloud.githubusercontent.com/assets/8002138/21434351/2ff9203e-c831-11e6-848d-dfec9241dae8.PNG)

dmsnell:
Yowsers! This is a bad thing. Thanks for reporting it here @chrisfromthelc

j2xl:
This is still an issue with Simplenote v. 1.3.4 on Windows 10. Pasting into a rich text box in Chrome applies the dark-mode background.

mirka:
@j2xl This should be fixed in 1.4.0. Could you try the latest version?

j2xl:
Yes, this is fixed in 1.4.0. Thanks for the clarification. I had been concerned because I just downloaded 1.3.4 a week ago and this bug had been fixed about 2 years ago. All's well. :)

PR #480, opened 2017-01-03
Strip formatting from selected text on copy and cut

Copons:
Fix https://github.com/Automattic/simplenote-electron/issues/469

This only exists inside the note content editor.
Everywhere else, including the note markdown preview, the copy and cut events keep their default behaviour.

Note: I don't consider this to be a bug, since copying the formatted text is the correct behaviour and the onus of stripping the formatting usually falls on who handles the pasting event, or to the users themselves via the `cmd/ctrl+shift+v` shortcut.
Although, I've always found myself hitting `cmd+shift+v` way more often than `cmd+v`, so in my opinion this would be a good workflow improvement.

### DIFF
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -155,6 +155,25 @@ export default class NoteContentEditor extends React.Component {
 		)
 	}
 
+	componentWillMount() {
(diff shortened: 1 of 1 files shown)


dmsnell:
seems fine. 👍

PR #485, opened 2017-01-06
Replace config.js with config.json in .gitignore

Copons:
The config file was changed from JS to JSON in https://github.com/Automattic/simplenote-electron/pull/477.
This PR prevents accidentally committing it to git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ node_modules/**/*
 desktop-build/
 dist/
 release/
-/config.js
+/config.json
 *.pvk
 *.spc
 *.tgz
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ node_modules/**/*
 desktop-build/
 dist/
 release/
+/config.js
 /config.json
 *.pvk
 *.spc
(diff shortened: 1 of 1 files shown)


dmsnell:
Good call @Copons! can we leave them both in though? it's actually still valid to have a `config.js` as long as it's default export is an object.

Copons:
@dmsnell Dang! I recalled you saying it worked as JS too, but I missed the object part and my tests were failing and I thought it eventually got ditched. Updating this asap.

beaucollins:
To be completely future proof we could do `config.*`.

For instance exporting a plain object in a `config.jsx` file would also work with the [current webpack config](https://github.com/Automattic/simplenote-electron/blob/cac38b16ebec3d066739b1362f4472dada1b5513/webpack.config.js#L29).

Copons:
Three commits to replace `js` with `*`. This might be my best ever!  😆

beaucollins:
@Copons definitely don't squash this one when you merge it 😉.

Copons:
@beaucollins and risking to lose a clean history? No way!

PR #486, opened 2017-01-09
Update Makefile to allow for either style of config files

dmsnell:
We recently moved to a JSON-based config but a JavaScript config file
which exports a single object is also valid. This reflects those changes
and prevents a bug where the build was failing on account of it.


roundhill:
🚢

PR #494, opened 2017-01-19
Fixes a bug where the note list was shrinking for notes with many tags

dmsnell:
Previously when selecting a note with lots of tags the note list would
shrink to a narrow width.

This has now been prevented by disabling the auto-width setting of the
`<AutoSizer>` wrapper in the note list.

We may be able to remove the `<AutoSizer />` if we can set a fixed
height but this is sufficient for now and it doesn't appear to be
costing much in terms of performance if at all.

### DIFF
--- a/lib/note-list.jsx
+++ b/lib/note-list.jsx
@@ -326,7 +326,7 @@ const NoteList = React.createClass( {
 										: getRowHeight( this.props.notes, { noteDisplay, width } )
 								) }
 								rowRenderer={ renderNoteRow }
-								width={ 328 }
+								width={ width }
 							/>
 						}
 					</AutoSizer>
--- a/lib/note-list.jsx
+++ b/lib/note-list.jsx
@@ -36,6 +36,19 @@ import noteTitle from './utils/note-title';
  */
 const TYPING_DEBOUNCE_DELAY = 70;
 
+/**
+ * Maximum delay when debouncing the row height calculation
+ *
+ * this is used to make sure that we don't endlessly delay
+ * the row height recalculation in situations like when we
+ * are constantly typing without pause. by setting this value
+ * we can make sure that the updates don't happen
+ * less-frequently than the number of ms set here
+ *
+ * @type {Number} maximum number of ms between calls when debouncing recomputeRowHeights in virtual/list
+ */
+const TYPING_DEBOUNCE_MAX = 1000;
+
 /** @type {Number} height of title + vertical padding in list rows */
 const ROW_HEIGHT_BASE = 24 + 18;
 
@@ -269,7 +282,8 @@ const NoteList = React.createClass( {
 		 */
 		this.recomputeHeights = debounce(
 			() => this.list && this.list.recomputeRowHeights(),
-			TYPING_DEBOUNCE_DELAY
+			TYPING_DEBOUNCE_DELAY,
+			{ maxWait: TYPING_DEBOUNCE_MAX }
 		);
 
 		window.addEventListener( 'resize', this.recomputeHeights );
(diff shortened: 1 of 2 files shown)


dmsnell:
> Unfortunately this breaks on small-ish screens.

Doh! Thanks @copons.

I'll update and try to mirror our settings.

dmsnell:
thanks again @copons - I have relied on pure-css solutions in 3bc7b24 which is the way I should have done it from the get-go.

this is ready for another review

Copons:
@dmsnell this doesn't seem to solve the issue.
`.source-list` behaves correctly (and you could remove the `width`, since `min-width` is the same).
`AutoSizer` encloses `.note-list` in a new element, with a fixed width of `328px`.

It seems that reverting to `width={ width }` [here](https://github.com/Automattic/simplenote-electron/pull/494/files#diff-c4602e800d73b145b43560d7dcbc7df9R267) solves it, but I'd like to have another set of yes around, in case I might have messed up with CSS without noticing it.

Copons:
I've also noticed that, with "expanded" display (in other words: when the note list preview can span multiple lines), the list element height doesn't update on resize but only on component re-render.

So, it happens that, for example, in `$single-column` the preview needs only 2 lines, but on wide screen it needs 3 lines.
When enlarging the screen, the preview gets cut off at 2 lines; when narrowing, it has an additional empty line.

Clicking on the list, though, sends a new `selectedNoteId` connected prop, and the previews heights update to the correct size.

jan-20-2017 14-06-47

Though, this looks like an intended performance tradeoff.

dmsnell:
> It seems that reverting to width={ width } here solves it

this is odd because it was supposed to be back. I'm not sure how that happened but I suspect I missed something when committing or rebasing

dmsnell:
> So, it happens that, for example, in $single-column the preview needs only 2 lines, but on wide screen it needs 3 lines.
> When enlarging the screen, the preview gets cut off at 2 lines; when narrowing, it has an additional empty line.

thanks for the note @Copons. I'm not sure if I follow what you're saying here. To restate my assumptions, I thought that the meaning of the three display modes was line based:

 - Condensed displays only the title line (first line unless empty)
 - Comfy displays the title line and up to one line of content
 - Expanded displays the title line and up to three lines of content

and then this is all based on the actual newline character to split the lines.

Copons:
@dmsnell Let me try explaining it better:

1. Start with a large window and with a note of a couple of lines. Make sure there is no selected note, or the 3rd step won't work, since you'll have to close the note and trigger a Note List re-render.

2. Since the Note List is currently narrow, that note preview (aka the note list element) fills **3 lines (correct)**. The list element is correctly resized and all 3 lines are visible.

3. Resize the window under the small screen breakpoint. Now the Note List is much larger and the note preview only needs **2 lines**. You now can see that the note preview kept the old height, and it's **3 lines tall (incorrect)**, with the last line empty.

4. Open a note just so that you can trigger a re-render, then close it. The note preview is now **2 lines tall (correct)**, and there are no empty lines.

5. Resize the window above the large screen breakpoint. Now the Note List is again narrow and the note previews needs **3 lines**. Instead, the height is still **2 lines tall (incorrect)**.
The text is cut off, and you can even see part of the ascender strokes of the tallest letter of the third line, popping up from the lower border of the note preview.

dmsnell:
thanks @Copons - I think I understand it now; at least I was able to reproduce behavior where the preview didn't appropriately resize in height.

Copons:
@dmsnell Yup, that's it! 🙂

roundhill:
I think this is the same-ish bug? Just noticed if you add a second line for a note, the list doesn't update the height for that row:

jan-25-2017 13-31-23

dmsnell:
> just noticed if you add a second line for a note, the list doesn't update the height for that row

@roundhill I noticed this too but when I tested it the rows finally updated after we saved the content. there was a noticeable delay but it did pop in. I believe that this is the delay we find in `queueSaveNote()`

roundhill:
> @roundhill I noticed this too but when I tested it the rows finally updated after we saved the content. there was a noticeable delay but it did pop in. I believe that this is the delay we find in queueSaveNote()

Cool, that was actually from `master`. I tried it out in this branch and it is fixed 👍. Is there more to fix here?

dmsnell:
I think there's still a hard-coded with in there that needs to be replaced

dmsnell:
All updated and ready for final review!

Update-on-resize should now happen

Copons:
The height recompute on update content doesn't work for me.

It seems this is caused by the equality check `nextProps.notes !== this.props.notes` failing on https://github.com/Automattic/simplenote-electron/blob/92713e80cd4cfeed13590b2ca2b52c345c3844eb/lib/note-list.jsx#L223
Or maybe because the heights recompute debounce somehow conflicts with the save debounce?

Either way, I've tried adding a new connected prop on the `contentKey` of the `selectedNote`:
```js
const selectedNoteContent = get( selectedNote, 'contentKey' );
```
and then:

```js
componentWillReceiveProps( nextProps ) {
	if ( nextProps.selectedNoteContent !== this.props.selectedNoteContent ) {
		// not debounced since the save action is already debounced!
		this.list.recomputeRowHeights();
	}
	...
}
```

It seem to work great for me, but I'm wary of the possible performance impact on huge notes.

dmsnell:
> It seem to work great for me, but I'm wary of the possible performance impact on huge notes.

Thanks @copons! I'll try to add this in but feel free to directly commit next time.

Performance should be fine because string comparison is blazingly fast in JavaScript. Did you know that it can be faster than comparing integers? Most engines _intern_ the strings and so a string comparison for equality is a basic reference comparison while number comparison can involve unboxing and type-checking.

This should fix a sore point in the branch where the size doesn't update. I thought it was a dead-end trying to fix that.

Copons:
It looks fixed for me! Great work @dmsnell!

I'd say :shipit: but for the odd nature of this issue I'd probably wait for @roundhill to give another look.

roundhill:
Yay, working well now! :shipit:

PR #498, opened 2017-01-20
Fix dependencies which were borked in the jest addition

dmsnell:
We had a mismatch between react-dom and react-addons-test-utils and a webpack config issue

cc: @roundhill


roundhill:
Everything seems happy again. Grip it and :shipit:

PR #499, opened 2017-01-20
Fix app not dragging on macOS when viewing revisions

roundhill:
Fixes the app not being able to drag when viewing the versions slider on macOS.

To test: View the revisions slider when in macOS and click and drag it. The app should move around!


dmsnell:
works for me in testing!

PR #500, opened 2017-01-23
Editor: increase distance between note and tags

### LINKED ISSUE #492: Editor: no padding at the bottom
There use to be padding at the bottom of the editor, but a recent change has caused the editor to ignore the CSS style. The padding created a space above the tag list and below the last line of the note. Test by looking at a long note that scrolls.

Here is the padding that is being ignored:

screen shot 2017-01-18 at 12 41 00 pm

Copons:
Fixes #492 

screen shot 2017-01-23 at 17 18 52

To keep the scrollbar, we need to allow the note to overflow vertically.
Unfortunately, this means that it overflows over its own padding.
Even the top one, by the way: the note has a top padding of `24px` (same as bottom), but it doesn't respect it and, scrolling, the note touches the top bar border.

I've added a top padding to the tags section in order to simulate the note's bottom padding.

To be clear: I've added a `16px` padding since the inner tags container already has an `8px` padding.
screen shot 2017-01-23 at 17 19 10

---

I must say though that, increasing only the padding–top, the tags look a bit awkward.
It would take me no time - and it makes sense to do it in this PR - to adjust the bottom padding too.
Same goes if we want to add a top padding for the notes, to separate them from the top bar.

### DIFF
--- a/scss/editor.scss
+++ b/scss/editor.scss
@@ -143,6 +143,10 @@
     display: flex;
     flex-direction: row;
     justify-content: center;
+
+		div[data-contents] {
+			padding-bottom: 20px;
+		}
   }
 }
 
--- a/scss/tag-editor.scss
+++ b/scss/tag-editor.scss
@@ -1,5 +1,6 @@
 .tag-entry {
   border-left: 1px solid lighten($gray, 30%);
+	padding-top: 16px;
 }
 
 .tag-editor {
--- a/scss/tag-editor.scss
+++ b/scss/tag-editor.scss
@@ -1,6 +1,5 @@
 .tag-entry {
   border-left: 1px solid lighten($gray, 30%);
-	padding-top: 16px;
 }
 
 .tag-editor {
@@ -11,7 +10,7 @@
   white-space: nowrap;
   overflow: visible;
   max-height: 120px;
-  padding: 8px 48px 8px 12px;
+	padding: 8px 12px;
 
   &:focus {
     outline: none;


drw158:
I agree, adding top padding to that div makes the space uneven. Looks like this is the only div that could have the bottom padding:

screen shot 2017-01-24 at 12 01 33 pm

It's a nested div. Would it be ok to use that one?

Some things I think we should do:

- `.tag-editor`right-padding is 48px for some reason. Should be 12px to be consistent.

screen shot 2017-01-24 at 12 04 00 pm

- Maybe a box shadow to the top of the tag field `box-shadow: 0px -8px 12px 2px #fff;`?

screen shot 2017-01-24 at 12 14 43 pm

Copons:
@drw158

- I'm not overly fond of adding a class to that particular element, since it's automagically created by Draft.js, and we can't be 100% sure it won't change at some point. Not to speak, it doesn't have any classes, so we should have a weirdly specific rule.
Either way, this would not change the uneven vertical spacing of the tag chips: it just moves from an element to another, but visually is the same.

- Didn't notice the `.tag-editor` padding. Quick and easy fix coming up.

- I'm personally not into the "disappearing text" effect to hide content (don't know why, probably because I've used it on an old site I've made for a client I hated 😆 ), but if you believe it's ok and consistent with Simplenote's design, I have nothing to object.
This would also fix the aforementioned uneven spacing, so it'd be a win-win.

drw158:
Ok, let's leave out the box shadow for now. I was on the fence about it.

> Either way, this would not change the uneven vertical spacing of the tag chips: it just moves from an element to another, but visually is the same.

They are slightly different. If you add top padding to the tag field, it'll look awkward because the tags are not centered in the white space:

screen shot 2017-01-24 at 3 03 57 pm

If you add bottom padding to the bottom of the note, there will be equal space around the tags:

screen shot 2017-01-24 at 3 07 25 pm

Only at the bottom, there will be some breathing room below the note:

screen shot 2017-01-24 at 3 08 18 pm

Which is ok imo.

Copons:
Oh! I didn't notice the padding was only at the end of the note. I'll see what I can do about it.
What case scenario, we could apply it to the Draft element, knowing that, if it'll change in the future the worst thing that can happen is simply less padding.

Copons:
@drw158 call for another design review, thanks!

dmsnell:
@copons can you confirm that things are still good after my rebase? Feel free to merge at your leisure if so.

PR #506, opened 2017-01-30
Normalize how email tags are treated

### LINKED ISSUE #143: Don't show email tags
On iOS, we don't show the email tags, because there is a UI for adding and removing collaborators. It should be the same in Electron. Going forward, I think we should not make it apparent that you can add people via tags.

dmsnell:
Resolves #143

Previously there was a mismatch between the native macOS app and this
app in how email tags were handled. In the native app if a tag contained
something like an email address it is hidden from the list of tags but
in the Electron app it has been shown in the tag list.

Now the behavior has been matched and in this patch we have normalized
all the places which are checking if a tag contains an email address by
using a new utility `utils/is-email-tag#isEmailTag( tagName )`

**Testing**
Try adding email addresses to the tag field. In this branch we can still add collaborators to a note by doing this. The tag will immediately disappear though and a popup should appear notifying you to use the collaboration tab in the sharing pane/page.

Try adding collaborators through the share dialog. The email addresses should not appear in the tag list.

Try exporting the notes and make sure that the right tags or collaborator email addresses appear in the export. The `source/notes.json` file is particularly helpful for this.

addingemailtags

### DIFF
--- a/lib/utils/is-email-tag.js
+++ b/lib/utils/is-email-tag.js
@@ -1,0 +1,33 @@
+/**
+ * @module utils/is-email-tag
+ */
+
+/**
+ * Naively matches what might appear to be an email address
+ *
+ * This is not spec-compliant but it does not need to
+ * be. Users of the app should assume that any tag
+ * name which has the general appearance of a valid
+ * email address will be treated as if they are actual
+ * email addresses and will be used for adding people
+ * to a note as collaborators.
+ *
+ * Three main components are required:
+ *   1. Mailbox piece _before_ the `@`
+ *   2. `@`
+ *   3. Host and domain piece ending in `[some host].[some TLD]`
+ *
+ * @type {RegExp}
+ */
+const naiveEmailPattern = /^(?:[^@]+)@(?:.+)(?:\.[^.]{2,})$/;
+
+/**
+ * Indicates if a given tag name string
+ * represents an email for collaboration
+ *
+ * @param {String} tagName name of tag which might be an email address
+ * @returns {Boolean} whether or not the tag is considered an email address
+ */
+export const isEmailTag = tagName => naiveEmailPattern.test(tagName);
+
+export default isEmailTag;
--- a/lib/dialogs/share.jsx
+++ b/lib/dialogs/share.jsx
@@ -132,8 +132,8 @@ export default React.createClass({
             <div className="settings-group">
               <p>
                 Add an email address of another Simplenote user to share a note.
-                You'll both be able to edit and view the note. You can also add
-                the email as a tag.
+                You&apos;ll both be able to edit and view the note. You can also
+                add the email as a tag.
               </p>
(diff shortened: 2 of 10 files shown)


drw158:
One correction, we don't hide the email tags on MacOS, only on iOS. Not sure about Android.

roundhill:
> One correction, we don't hide the email tags on MacOS, only on iOS. Not sure about Android.

Android shows the emails in the tags, which I prefer. Any thoughts on what the standard should be @drw158?

dmsnell:
I made the mistake of assuming that the macOS and Android apps followed the iOS behavior. This PR was built based on what @drw158 wrote in #143 but we could drop it altogether too.

drw158:
Adding email as tags always seemed like a hack to me. Originally there was no interface for sharing. Email as tags is a novel idea, but a separate UI I think is better.

dmsnell:
@drw158 any opposition to merging this?

drw158:
No hard opposition, but I don't think it's ideal that the email tag will disappear after you enter it. Not sure what the solution is, maybe disable adding email tags altogether. But, I also want to accommodate the users that still try to type in email addresses into the tag field.

Just checked iOS: if you try to put an email into the tag field, a popup is shown, saying that collaboration has moved. Maybe that's the best way forward?

dmsnell:
> I don't think it's ideal that the email tag will disappear after you enter it

that's a really good point. I'll try and come up with some kind of dialog such as exists in the iOS app. thanks for the feedback!

dmsnell:
@drw158 don't know if you want to take another look at this while away, but I added a popup dialog similar to on iOS, though here there's a link to open the share dialog directly.

@megsfulton would you be willing to do a design review of this? namely there's junk CSS in there I'm sure is done better differently. I had trouble initially with the editor displaying over the popup and thus couldn't click on the share link. Also I would appreciate some review of the text in the popup as I just kind of doodled it down.

megsfulton:
>though here there's a link to open the share dialog directly 

👍 Good call, it's nice that this is actionable. 

>namely there's junk CSS in there I'm sure is done better differently.

😅 I'm not sure I'm the best person to de-junk CSS 

> I would appreciate some review of the text

Maybe we try the following:

Collaboration Has Moved
Press the Share icon in the toolbar to access the [Collaborate screen ](url)

lib/utils/is-email-tag.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Indicates if a given tag name string
+ * represents an email for collaboration
+ *
+ * @param {String} tagName name of tag which might be an email address
+ * @returns {Boolean} whether or not the tag is considered an email address
+ */
+export const isEmailTag = tagName => naiveEmailPattern.test(tagName);
+
+export default isEmailTag;
```
rodrigoi: I really love this pattern of moving away from large `utils` files into smaller single function exports.

PR #510, opened 2017-01-31
added regexp  to remove initial '#' characters from title and preview…

pacharanero:
… so that markdown syntax characters are not shown in the left sidebar note list

in order to keep it simple, this stripping of # characters is in operation for **all** notes (even ones that are not in Markdown) because otherwise there'd be a need for logic outside of `lib/utils/note-title.js` in order to detect whether the note was in Markdown mode or not.

``` 
note content:                                   returns:
# valid markdown h1 title                       match => valid markdown h1 title
#hashtag at start of a title                 no match => #hashtag at start of a title     
## valid markdown h2 title                      match => valid markdown h2 title
# markdown title with a #hashtag in it     part match => markdown title with a #hashtag in it
# title with a # randomly in it            part match => title with a # randomly in it
```

Hopefully this will allow Markdown users to have a nice clean view of the list of notes, without interfering with anything that non-markdown users might be doing, or preventing the use of hashtags in titles.

Marcus

### DIFF
--- a/lib/utils/note-title.js
+++ b/lib/utils/note-title.js
@@ -13,14 +13,21 @@ const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
  * Returns the title and excerpt for a given note
  *
  * @param {Object} note a note object
+ *
+ * Markdown title syntax characters (#) are stripped from
+ * the start of the lines.
+ *
  * @returns {Object} title and excerpt (if available)
  */
 export const noteTitleAndPreview = note => {
 	const content = note && note.data && note.data.content;
 	const match = noteTitleAndPreviewRegExp.exec( content || '' );
 
-	const title = match && match[1] && match[1].slice( 0, 200 ) || 'New note...';
-	const preview = match && match[2] || '';
+	const unstrippedTitle = match && match[1] && match[1].slice( 0, 200 ) || 'New note...';
+	const unstrippedPreview = match && match[2] || '';
+
+	const title = unstrippedTitle.replace( /^[#]+/, '' );
+	const preview = unstrippedPreview.replace( /^[#]+/, '' );
 
 	return { title, preview };
 };


dmsnell:
Thanks for your work on this @pacharanero. I think we need to be paying attention not only to leading `#` symbols but also other Markdown syntax. Should we show the leading `_` and `*` for instance if the title/first line has been emboldened or italicized?

One of the things I would like to see (probably not something I'll focus on for the next release) is to just render the first line completely and take the plain text output from the render for the title. Right now our data model doesn't perform this work but there are a few things I would like to see us do when notes are updated:

 - Render the title and excerpt then remove all formatting for the list
 - Store this text and precompute the height for the note list (react-virtualized) as a performance optimization

Thoughts?

pacharanero:
Hi @dmsnell. Thanks for the response.

Yes, I think it would be a good idea to remove other MD formatting from title/preview, and this would be a simple addition to the RegExp in `const title = unstrippedTitle.replace( /^[#]+/, '' );`

I opted to keep it very simple and just focus on '#' for this PR, because I didn't want to introduce unintended usability problems for people using '*' or '_' characters elsewhere in the title/preview. Maybe it would be worth seeing how '#' works with users and then add other chars later?

I agree that your proposed future solution of fully rendering the title and first line to MD, and then passing this (without formatting) to the sidebar list, is optimal.

Perhaps we could use this workaround temporarily until a more complete refactor of this feature can be planned.

Marcus

pacharanero:
Here's some screenshots of simplenote-electron, both with and without the patch installed, to help with design review.

without #510 (showing how messy the left sidebar looks when markdown syntax is shown)
image

with #510 (showing a much cleaner left sidebar, and no deleterious effects on other uses of # in titles)
image

M

drw158:
Thank you for providing screenshots for the design review!

dmsnell:
> …I didn't want to introduce unintended usability problems for people…add other chars later? …use this workaround temporarily

this makes me leery @pacharanero (and if you knew me you would also know that I don't believe that "temporary code" has ever actually existed :wink:).

we definitely _will_ be introducing changing behavior in this PR if we simply strip a character or to from the beginning of the note. this is definitely a problem if a note is marked as non-markdown.

like I said, I don't think I'll be focusing on this issue for this upcoming release but I wouldn't want to discourage you from working on it. I'd be happy to continue to review and test your code (and we'd love to be able to list you as a contributor!)

here are a few steps I think would be essential to merging this:
 - don't touch non-markdown notes
 - render the note preview (title + excerpt) and strip formatting. there may be a `marked` renderer which does this.
 - make sure this doesn't bring performance regressions for he note list. obviously it will be slower than it currently is, but our main goal is not noticeably dropping the frame-rate/lag while scrolling

would be happy to provide any help I can

pacharanero:
Thanks for the response @dmsnell. I think your points made are reasonable. It's going to need a more involved approach to implement, ie just fiddling with the regex won't be enough.

Probably best to close this PR and I'll have a look at whether or not I know enough about Simplenote internals to implement the Markdown syntax removal in the way you've suggested, but I understand and agree with the reasoning. 'Don't Break Userspace' :-)

dmsnell:
thanks for the activity @pacharanero! please don't hesitate to ask questions here either; we'd be happy to help any way we can.

it seems reasonable that a release coming up soon after 1.0.9 might involve some major data structure changes to the way notes are stored. this would be a perfect place to strip the markdown from the note title because it could be computed once and then whenever the notes actually change.

in the meantime, we would have to recalculate the transform on every render or cache the changes and keep them in sync with the note contents. you can check out the `note-list.jsx` and see some rather messy code I recently pushed in to keep the list performance acceptable. all of that I hope to eliminate in this big data structure update at some point.

Issue #512, opened 2017-02-02
Copy-Paste without line breaks (Electron 1.0.8.)


gubrist:
#### Steps to reproduce
1. Copy a note containing line breaks (carriage return) with [ Ctrl + C ]  within Electron 1.0.8
2. Paste it into Windows Text Editor

#### What I expected
As previous versions of Electron, I expect to get al legible text with line breaks.

#### What happened instead
A text without line breaks.

#### OS version
Windows 10

#### Screenshot / Video

![no line break 02 02 2017](https://cloud.githubusercontent.com/assets/22197180/22566582/d975c9ec-e98c-11e6-9e30-195b78d54e15.JPG)

dmsnell:
Thanks for the report @gubrist and I hope this didn't interrupt your flow too much!

@Copons maybe we missed the fact that Windows apps still use `\r\n` as newlines and we need to coerce this format [when we fill the paste buffer](https://github.com/Automattic/simplenote-electron/pull/480/files#diff-d65a368b0648bee7bc77b758399fb23cR171)

something like `selectedText.replace( /(?![\r])[\n]/, "\r\n" )` maybe (negative-lookahead `\r` followed by `\n` meaning "replace all newlines not preceded by a carriage return with a `\r\n`)

Copons:
Ouch @dmsnell I assumed the newline differences ware handled directly on a OS level.
Opening a fix ASAP, but I'm currently not in the conditions for testing on Win.

gubrist:
![copy-paste without line breaks 2a](https://user-images.githubusercontent.com/22197180/29677961-1e5803f8-88fd-11e7-8528-8cbc97c0e488.JPG)
![copy-paste without line breaks 2b](https://user-images.githubusercontent.com/22197180/29677964-1ff0d230-88fd-11e7-8836-da3758ea5f98.JPG)
![copy-paste without line breaks 2c](https://user-images.githubusercontent.com/22197180/29677970-22596cc6-88fd-11e7-994f-101ecefff5db.JPG)

Thanks, the line brake problem seems to be on the classic Wind Text editor 
(still running under Windows 10; see 1st image).

You can resolve it by using an comprehenisve Text Processor such as open Office.

On the other hand, there is an issue with ä, ö, ü etc. (see 3rd image).
- Which is the right charset to be applied?

dmsnell:
[...]

PR #513, opened 2017-02-03
Copy note content: replace \n with \r\n to keep line breaks on Windows

Copons:
Fixes https://github.com/Automattic/simplenote-electron/issues/512

This PR fixes a bug introduced in https://github.com/Automattic/simplenote-electron/pull/480, where artificially filling the clipboard didn't took into consideration the Windows newline quirks.

## Testing instructions

1. Open Simplenote on Windows.
2. Select and copy a note content spanning multiple line breaks.
3. Paste it on Notepad.

The pasted text should have all the line breaks were expected.

I'd appreciate a quick test on Windows, since I'm not currently able to do it.

### DIFF
--- a/lib/utils/export/index.js
+++ b/lib/utils/export/index.js
@@ -10,7 +10,6 @@ import {
 	uniqueId,
 } from 'lodash';
 
-export const CRLF_NEWLINE = '\r\n';
 export const LF_ONLY_NEWLINES = /(?!\r)\n/g;
 
 // needs three pieces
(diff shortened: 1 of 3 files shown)


dmsnell:
@roundhill and @Copons I would also like to try and reproduce this behavior from the note exports in Windows and tack on that fix (if needed) into this PR since they are so related.

Copons:
@dmsnell I've tentatively added the same regex to the export.
As far as I know, there's no way to export from the browser, and it's not clear to me how I can test it on a Win VM from macOS (except maybe setting up the environment in the VM and build the app there?)

Copons:
Alright: tested on Win10.
The line breaks issue happened on export too, and the regex fixes it.

I've noticed the exported json file too has wrong breaks. Is it worth fixing or can we assume it's only used for feeding it into other apps and not for "direct consumption"?

Copons:
@dmsnell Updated for JSON export too and tested on Win10!

lib/utils/export/index.js
```diff
@@ -10,6 +10,9 @@ import {
 	uniqueId,
 } from 'lodash';
 
+export const CRLF_NEWLINE = '\r\n';
```
dmsnell: `LF_ONLY_NEWLINES` I think is helpful. `CRLF_NEWLINE`'s value might be a little more dubious. might actually hamper readability since it's a straightforward constant. I support you either way, but I wouldn't have turned this one ^^^ into a named constant

dmsnell:
:shipit:

PR #528, opened 2017-03-17
changed README.md to reflect recent changes on npm electron installat…

Streppel:
…ion package

As for it is at this moment, running `npm install -g electron-prebuilt` will result in the following warning from NPM:

> npm WARN deprecated electron-prebuilt@1.4.13: electron-prebuilt has been renamed to electron. For more details, see http://electron.atom.io/blog/2016/08/16/npm-install-electron

If we follow the link, it informs us that

> We recommend updating your package.json files to use the new electron dependency, but we will continue releasing new versions of electron-prebuilt until the end of 2016.

As we are already in 2017, I think it's a good choice to change the package to the new official one in the installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ _Note: Simplenote API features such as sharing and publishing will not work with
 
 ## Electron
 
-Install electron globally `npm install -g electron-prebuilt`, then:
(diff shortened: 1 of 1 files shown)


roundhill:
Looks good, thanks for the PR :shipit:

PR #543, opened 2017-04-12
NoteEditor: Redux refactor [1] - Revisions state and selectors

Copons:
First step of the complete NoteEditor refactor (rework of https://github.com/Automattic/simplenote-electron/pull/479).

In this PR we create a new `revisions` Redux subtree and connect it to all the involved components.
In the same effort, we also create a new `getActiveNote` utility to DRY-fy the active note selection.

I've tried my best to keep the code as clean as possible, but still, when reviewing, please disregard the use of some odd conventions (such as not using shorthands even when possible).
Hopefully those oddities will be fixed in the following steps of the refactor or, if needed, in a single final dedicated PR.

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -389,7 +389,6 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 								allTags={ state.tags }
 								editorMode={state.editorMode}
 								filter={state.filter}
-								revisions={state.revisions}
 								onSetEditorMode={this.onSetEditorMode}
 								onUpdateContent={this.onUpdateContent}
 								onUpdateNoteTags={this.onUpdateNoteTags}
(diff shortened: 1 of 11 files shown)


dmsnell:
@Copons I'm getting an error while testing this - wasn't able to select an older revision.

screen shot 2017-04-18 at 12 15 46 pm

I think the hardest part about reviewing these changes is following both the code moves _and_ the logic changes, reorganizing and converting to Redux at the same time. are you getting this error as well?

Copons:
@dmsnell I'm able to move the slider and restore revisions without errors. I've even cleaned up everything to be sure.
Anyway, I'll go update the reducer to protect it for undefined revisions.

> I think the hardest part about reviewing these changes is following both the code moves and the logic changes, reorganizing and converting to Redux at the same time.

Yeah, unfortunately this refactor really involves moving methods from a file to the other, and then from this new file to another one, every time deeper into the component hierarchy.
I honestly don't know how to resize the whole thing even more.

rodrigoi:
@Copons would you mind rebasing this so we can take another look?

Copons:
Closing for inactivity

PR #545, opened 2017-04-27
Add snap install instructions

popey:
This pull request adds installation instructions for users of [snap enabled Linux distributions](https://snapcraft.io/docs/core/install). Snaps enable ISVs and projects to make their software available to millions of Linux users via the snap store and directly control delivery of software updates to their users. Snaps of stable desktop applications will also be available via the desktop Software Center, improving the discoverability of your software. Your software can be installed with a simple `snap install simplenote --beta` simplifying the installation instructions for your users.

Once the snap is released into the stable channel in the store, users no longer need to add the ```--beta``` and will appear in the desktop Software Center

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,13 @@ A Simplenote [React](https://facebook.github.io/react/) client packaged in [Elec
 * A Simperium account. [Sign up here](https://simperium.com/signup/)
 * A Simperium Application ID and key. [Create a new app here](https://simperium.com/app/new/)
 
+## Installing pre-built Snap
+
+On [snap supported](https://snapcraft.io/docs/core/install) systems, install the pre-built snap from the store.
+
+    snap install simplenote --beta
+
+
 ## Running
 
 1. Clone the repo: `git clone https://github.com/Automattic/simplenote-electron.git`


dmsnell:
Thanks for the contribution @popey. I'm wondering if this really belongs in our README though. We don't provide installation instructions for the compiled app for any other systems, only for setting up the development environment. For example, we don't include `apt` or `yum` or `portage` instructions just as we don't provide instructions for `brew`.

My guess is that it would be best to support installation for `snap` _in snap_. Thoughts?

cc: @drw158

drw158:
@dmsnell That makes sense, but I don't know enough about snap to make an informed decision.

daylightbrightledlight:
I like this idea.

dmsnell:
@popey I'm going to close this for now as I'm not sure what to do about it. we don't officially support snap or any other third-party distribution platform. I might consider a variation here but we would need to have a disclaimer about snap not being an official source; would also want to make sure we put some other methods in. since this is the development repo I might like the idea of creating a separate `INSTALLING.md` file or something like that with client-side installation instructions. How do you feel about that?

PR #546, opened 2017-04-27
Reset RegExp lastIndex to 0 during filter-note (to resolve #535)

### LINKED ISSUE #535: Search doesn't find all the notes with search term
#### Steps to reproduce
1. Create three notes in the Windows app with the following three lines, one line for each note:

House 10/3/17 Shane
House notes 18/3/17
House Visit 04/03/17

2. Try searching for "house". Only the first and last note is found. Simplenote doesn't find the second note.

I tested this in the Mac electron app and can reproduce.

Related: https://github.com/Automattic/simplenote-electron/issues/333

gie3d:
Resolves [Search doesn't find all the notes with search term #535](https://github.com/Automattic/simplenote-electron/issues/535)

When It comes to RegExp test with 'g' - global. If it can found a keyword inside a string, it will set the lastIndex to that position. So when we search multiple notes, if it can found a keyword in a note, then the next note RegExp won't start at the beginning but it will use lastIndex (last position in found in the previous note) to be its beginning. So I think we would need to reset the lastIndex to 0.

### DIFF
--- a/lib/utils/filter-notes.js
+++ b/lib/utils/filter-notes.js
@@ -3,10 +3,7 @@
  */
 import { escapeRegExp, get, includes, overEvery } from 'lodash';
 
-const includesSearch = ( text, search ) => {
-	search.lastIndex = 0;
-	return search.test( text || '' );
-}
+const includesSearch = ( text, search ) => ( new RegExp( escapeRegExp( search ), 'gi' ) ).test( text || '' );
 
 const matchesTrashView = isViewingTrash => note =>
 	isViewingTrash === !! get( note, 'data.deleted', false );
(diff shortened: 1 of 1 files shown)


dmsnell:
Good find @gie3d! That `g` always gets me. I tested out your code and found that it appears to solve the issue but I wanted to seek your input on a slightly different way of doing it and see how you feel about it.

```js
const includesSearch = ( text, search ) =>
       (new RegExp( escapeRegExp( search ), 'gi' )).test( text || '' );

overEvery( [
	matchesTrashView( showTrash ),
	matchesTag( tag ),
	matchesSearch( filter ),
] )
```

in this case, we're passing in the plain search string to `matchesSearch()` instead of passing in a `RegExp` object. the benefit is that we're removing the shared mutable state that was there before. I realize that I missed this because it _looked_ like we were creating a new `RegExp` object on every search but actually it was on every note within a search. if we create the `RegExp` inside of `includesSearch` we no longer have to worry about that mutability.

thoughts? thanks for the PR!

gie3d:
I agree with you as the code will look cleaner than reset the lastIndex to 0. I'll submit a new PR for this.

dmsnell:
cool. thanks again @gie3d - I'm going to merge in your changes. thanks for being part of the team!

Issue #548, opened 2017-05-08
Tag input (ContentEditable) caret position jumps


gie3d:
#### Steps to reproduce
1. Set up a simplenote-electron project (clone, prepare config file and npm start)
2. Create a new note
3. Put tags (such as 123456789 or abcdefg)
4. Whenever I type a char in tag-input, my cursor always moves back to the beginning.

#### What I expected
Whenever I type a char in tag-input, I expected that the cursor should be at the end of the string I type.

#### What happened instead
The cursor was moved back to the beginning of the text. So when I typed 1 2 and then 3, I end up with 321.. which should be 123

#### OS version
macOS 10.12.4
Chrome 58.0.3029.96
I reproduce this issue on both the latest commit 9c39a9b and at tag v1.0.8.

#### Screenshot / Video
![tag-input-problem](https://cloud.githubusercontent.com/assets/1977772/25786849/8b3ec25a-33c5-11e7-9afe-e5144aab12ab.gif)

gie3d:
Tested with the merged PR and the issue has been resolved.

PR #549, opened 2017-05-09
Set focus tag input whenever the div ContentEditable is rendered to make sure that the caret position will be at the end when typing

gie3d:
This is regarding to the issue #548 I found recently.

Did some research and found an issue on React https://github.com/facebook/react/issues/2047 which relates to this issue.

When a user types into tag-input which mutates the DOM and cause the tag-input to re render then the caret position jumps. So I think we would need to reset it to the end of the string by using focusInput

### DIFF
--- a/lib/tag-field.jsx
+++ b/lib/tag-field.jsx
@@ -65,7 +65,7 @@ export default React.createClass( {
 	},
 
 	hasSelection: function() {
-		return !! this.state.selectedTag && this.state.selectedTag.length;
+		return this.state.selectedTag && !! this.state.selectedTag.length;
 	},
 
 	deleteTag: function( tagName ) {
--- a/lib/tag-field.jsx
+++ b/lib/tag-field.jsx
@@ -65,7 +65,7 @@ export default React.createClass( {
 	},
 
 	hasSelection: function() {
-		return !! this.state.selectedTag.length;
+		return !! this.state.selectedTag && this.state.selectedTag.length;
 	},
 
 	deleteTag: function( tagName ) {
--- a/lib/tag-input.jsx
+++ b/lib/tag-input.jsx
@@ -128,13 +128,16 @@ export class TagInput extends Component {
 		invoke( event, 'stopPropagation' );
 	};
 
+	componentDidUpdate() {
+		this.focusInput();
+	}
+
 	render() {
 		const {
 			value,
 			tagNames,
 		} = this.props;
 
-		this.focusInput();
 		const suggestion = value.length && tagNames.find( startsWith( value ) );
 
 		return (
--- a/lib/tag-input.jsx
+++ b/lib/tag-input.jsx
@@ -134,6 +134,7 @@ export class TagInput extends Component {
 			tagNames,
 		} = this.props;
 
+		this.focusInput();
 		const suggestion = value.length && tagNames.find( startsWith( value ) );
 
 		return (


dmsnell:
thanks @gie3d! I hadn't gotten the chance to look over the issue yet but I'm glad you found this. do you think it would be possible to put this inside of `componentDidUpdate()` instead? inside of `render()` it could cause some more glitchy behavior, for example, when receiving externals updates from Simperium (this is my guess, I haven't verified)

thanks for taking the initiative to report and resolve - we're glad to have you contributing!

gie3d:
Thanks @dmsnell for your feedback. I'm not quite sure about the glitchy behavior you mentioned. Maybe you can share some information with me (such as how to produce external update from Simperium) and I'll find a way to verify?

In the meantime, I updated the code and put the `focusInput()` into `componentDidUpdate()`

And tested this morning I also found an issue, in `tag-field.jsx`, when you removing all the tags then the `this.state.selectedTag` becomes `undefined` and then cause an error `Cannot read property 'length' of undefined` in `hasSelection()`

I cannot add a unit test for this as I still cannot run it.

gie3d:
One of the thing that I don't understand is that I checked out `v1.0.8` in this repo and I still face this issue. 

But why I can't reproduce it in my account on simplenote.com or my macOS (which both are `v 1.0.8`)

dmsnell:
> I'm not quite sure about the glitchy behavior you mentioned. Maybe you can share some information with me

For example, if we have Simplenote open on another device it could trigger remote updates. React is free to serenader the components even if it doesn't change. I _think_ this might be the case if we add a new tag somewhere (even not on the same note) that it will serenader the tag field.

> in my account on simplenote.com or my macOS (which both are v 1.0.8)

simplenote.com isn't running this Electron/React app and neither is the macOS native app (the one which has been in the App Store).

normally I would prefer to leave the two issues to two separates PRs but I think it's okay to leave them together here if it all seems to work when we test.

> I cannot add a unit test for this as I still cannot run it.

unfortunately we don't have much in the way of testing but you are free to create some using our **Jest** test runner. See `lib/components/tag-chip` for an example.

I'd be happy to help you get it up and running. feel free to ask about any problems you are facing.

thanks again!

gie3d:
Ok. I'll try to simulate and test it this weekend.

On macOS, actually I use the electron version (not download from App Store) and when I clicked at Contribute, it leads me to this project.
screen shot 2560-05-11 at 9 34 32 am

For unit test, I created a #551 as I still cannot even start running it.

Many thanks for your help 👍

gie3d:
Hi, @dmsnell 

I tested by opened 2 incognito chrome browsers (let's say (1) and (2)) and opened my account and tried to update tags in both browser.

I observed an error when
1. Updated tags in (1)
2. Updated tags in (2)
3. Come back to (1) and you will see an error in console log

However, this happened before this PR and I didn't see any other issue. So I think this PR doesn't introduce a new issue.

dmsnell:
Thanks again @gie3d for your hard work on this! I was occupied recently and couldn't get around to performing a final review until today. We're glad to welcome your contribution!

gie3d:
Thanks @dmsnell  :) I close the #548 and #550.

dmsnell:
Seems like this caused a regression where the tag input field was being changed too often, whenever the app updated basically. I have made a change in #603 which corrects this behavior but I'm having trouble testing it to make sure I didn't bring back the original bug.

gie3d:
Basically the cursor should be at the end of the text when you type which I tested the latest master and it didn't bring the issue back.

But come to think of it, let's say you want to put a tag name called 'test' but you mistakenly typed 'tst'. So you move the cursor and type the 'e' then the cursor is going to move to the end of the text. Not sure if this is Ok?

Issue #293, opened 2016-04-11
Hide Menubar


ryanleesipes:
Maybe this is Gnome-specific (Gnome desktop, Ubuntu), but I'd really love a way to hide the menu bar.

![screenshot from 2016-04-10 21-25-11](https://cloud.githubusercontent.com/assets/2141621/14415274/c7afce5c-ff62-11e5-9b2f-a3f29501a470.png)

drw158:
The File, Edit, etc. menu can't be hidden on Windows or Linux. Not sure how to solve that though.

ebeloded:
Hey @drw158,

The menu can be hidden or auto-hidden (like in Slack or VS Code - you can press Alt to show/hide the menu).

Relevant methods:
- [win.setMenu(menu)](http://electron.atom.io/docs/api/browser-window/#winsetmenumenu-linux-windows) - Sets the menu as the window’s menu bar, setting it to null will remove the menu bar.
- [win.setAutoHideMenuBar(hide)](http://electron.atom.io/docs/api/browser-window/#winsetautohidemenubarhide) - Sets whether the window menu bar should hide itself automatically. Once set the menu bar will only show when users press the single Alt key.

Hope this helps to resolve the issue. The menu is a real eyesore on Windows and Linux.

nsrosenqvist:
The implementation of this in Atom can be followed in the following issue thread: https://github.com/atom/atom/issues/2914

The support for hiding the menu bar in electron apps was added with the following pull request: https://github.com/electron/electron/pull/558

natanielcz:
Looks like someone already upgraded electron version in Simplenote and app menu is working as intended. At least on Ubuntu 17.10. I think that we may close this issue too.

nsrosenqvist:
@natanielcz  It's not working for me, are you sure you're not using Unity and that it's just moved to the global menu? I still have no control of it in 1.1.0. This can't be that hard to implement, can it? Just a menu entry and keyboard event, right?

natanielcz:
@nsrosenqvist : you are right. I don't know why I thought that it works. I'll take a look after 8th of December when I'll have more time for that.

algiuxass:
This is taking way to long to add this feature. It shouldn't be that hard to fix it. (Using Linux)

cweagans:
PR #1207 addresses this.

PR #559, opened 2017-05-20
Feature: Toggle the menu bar by pressing the `ALT` key

klauscfhq:
Hello!
Absolutely love the app! 💓
I thought it would be a nice addition, to be able to toggle the menu bar whenever you need it, just by pressing the <kbd>ALT</kbd> key, especially on Windows and Linux where it might be a bit of a distraction for some people.

All it took was to add and set as **true** the boolean **autoHideMenuBar** option under the **BrowserWindow** class, following the [Electron API docs](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions).

screen

🚀 Tested on **Windows 10 Pro**, **Linux Ubuntu 16.04** & **Mac OS Sierra**!

🔧 It probably covers issue #293 

📷 Here is in action .gif running the windows build!

in-action

Hope it's helpful 😄

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -46,6 +46,7 @@ module.exports = function main() {
 			minWidth: 370,
 			minHeight: 520,
 			icon: iconPath,
+			autoHideMenuBar: true,
 			titleBarStyle: 'hidden'
 		} );
 


dmsnell:
Thanks for the contribution! This looks good to me. @drw158 any reason not to incorporate this?

drw158:
Nice, I don't use Windows/Linux so I'm sure about what is the expected behavior. Does this have any effect on Mac? 

Did we hide the menubar before this on Windows/Linux or was it always shown?

What is the part of the gif when the toolbar changes from white to black? That is the only part that seems weird to me.

Other than that looks good!

klauscfhq:
Nope, it has no effect over the MacOS build : ) This option targets only Windows/Linux platforms.

If not defined inside the **BrowserWindow** class, the `autoHideMenuBar` attribute is [set to it's default]() `false` value, which means that yes it was [always active](https://github.com/Automattic/simplenote-electron/issues/293) on the Windows/Linux builds and hiding it was not possible.

> autoHideMenuBar Boolean (optional) - Auto hide the menu bar unless the Alt key is pressed. Default is false. (taken from [here](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions))

The color  change of the app's window bar, from white to dark grey, was due to the app switching from unfocused to focused mode on my desktop, so it is just a "window manager thing" 😄 

When initially launching the app, the menu bar is hidden . Pressing once the <kbd>ALT</kbd> key, toggles it to visible, while a new up-following <kbd>ALT</kbd> key-press will un-toggle it, thus setting back to hiding mode.

I hope my explanation was somehow helpful! 😄
Happy to help out!

drw158:
Thank you very much, that makes sense. I think this looks good to me. 👍

mleko69:
Thanks @klauscfhq your workaround works like a charm. Why it still hasn't been added to program yet? Simplenote looks so sexy without that visible menubar <3

eurruby:
This addition is the simplest way to do. It works like Firefox then users will not be confused. https://github.com/roundhill https://github.com/drw158 don't you think it's OK?

PR #568, opened 2017-06-07
Replace draft-js with a textarea

dmsnell:
Draft-js was lovely but a performance hog on long notes (notes with
several thousand lines).

This patch brings back the `textarea` and tries to maintain some control
over the content, adding back the helpers such as list completion.

**Problems**:
 - state updates are wonky. we need to change the debounced updater because it leaves local state out of sync. this is also a problem in the production version with draft js
 - maybe tabbing with selection starting at end of line doesn't work?
 - search results are not highlighted


dmsnell:
The more I get into this the more I feel like we need a text editor that more intrinsically coupled to the data model. `draft-js` at least operates on atomic edit operations with descriptions, but the `textarea` don't care. It gives new states on update but doesn't indicate how the content changed. For large notes this too could lead to big enough performance issues to be difficult. Actually, I'm sure this `textarea` is leagues better than `draft-js` was, but it does add additional complexity to the state updates.

I also want to check out the [Monaco editor](https://github.com/Microsoft/monaco-editor) (which powers Visual Studio Code) and see if we can get it look right visually. It's already blazing fast and they have done the heavy-lifting to making large documents scroll and update in controlled manners.

There are approaches that involve splitting the input into separate `<p>` tags for each line but they additionally require using `contentEditable` in order to make scrolling work. Some issues revolve around updating the DOM nodes appropriately and not getting lost in the confusion.

dmsnell:
Thanks @copons - those references have been removed

dmsnell:
@copons I think this one is ready. Interested in giving it a go-around @kwight?

In my own tests it appears significantly faster to load very large notes. List autocomplete appears to be working relatively well - I'm not sure how to exhaustively test it. Indent and outdent also appear to be working as I expect them to.

Copons:
(Sorry for the ugly review dismissal message, didn't think it appeared like that - and it's not editable!)

Anyway, styling problem: the textarea vertical scrollbar.

| Preview | Edit | Edit Small Screen |
| --- | --- | --- |
| screen shot 2017-07-11 at 18 06 21 | screen shot 2017-07-11 at 18 06 11 | screen shot 2017-07-11 at 18 08 58 |

In Preview mode, the overflowing container is the entire "content section", so the scrollbar is correctly positioned next to the right edge of the screen.
In Edit mode, instead, it's the textarea that overflows, and the scrollbar is instead positioned inside the textarea itself, which is typically quite far from the screen edge.

Ideally, in Edit mode, we'd want for the containers to have no padding and no scrollbars; while applying them to the textarea instead.

```scss
.note-detail {
  overflow-y: hidden;
}
.note-detail-textarea {
  max-width: 100%;
  padding: 0;
}

.note-detail-textarea textarea {
  // small screens
  padding: 24px;

  // medium screens
  // up to 780px = max width of textarea (684px) + padding (48px per side)
  padding: 24px 48px;

  // large screens
  // dynamically calculate the horizontal padding
  padding: 24px calc( ( ( 100% - 780px) / 2 ) + 48px );
}
```

kwight:
Whenever I stop typing in the textarea, the cursor jumps to the tag section:

jul-11-2017 10-37-28

No other surprises though; also tested list behaviour (thought numbered lists wasn't working, but it seems only bulleted lists are supported?).

gwwar:
@dmsnell is this one ready to look at? Are the problems you listed in the summary something we're ok shipping with or did you need help tracking them down?

dmsnell:
I updated this branch to the latest master but I think I lost some important bits related to focus - stuff that was updated in *master* since I started working here. Need to resolve that before merging.

rodrigoi:
@dmsnell do you need any help closing this PR?

dmsnell:
Closed in favor of #692

PR #575, opened 2017-06-20
Add support for creating accounts directly from the app

roundhill:
This PR allows a user to create an account from directly within the app.

**To Test**
* Sign out of the app.
* Click on 'Sign Up'
* Fill in your credentials, and click 'Sign Up'.
* You should be signed in to the app.
* Also test that the regular sign in flow still works as expected.

This is dependent on https://github.com/Simperium/node-simperium/pull/48. You can link to it from the package.json or clone it locally and link to it from there.

screen shot 2017-06-20 at 4 46 32 pm

### DIFF
--- a/lib/auth.jsx
+++ b/lib/auth.jsx
@@ -188,4 +188,4 @@ const mapStateToProps = state => ( {
 	hasLoginError: hasLoginError( state ),
 } );
 
-export default connect( mapStateToProps, null, null, { pure: false } )( Auth );
+export default connect( mapStateToProps )( Auth );
--- a/lib/auth.jsx
+++ b/lib/auth.jsx
@@ -16,6 +16,14 @@ export const Auth = React.createClass( {
 		isAuthenticated: PropTypes.bool,
 		isMacApp: PropTypes.bool,
 		onAuthenticate: PropTypes.func.isRequired,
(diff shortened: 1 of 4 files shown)


lib/auth.jsx
```diff
@@ -124,4 +188,4 @@ const mapStateToProps = state => ( {
 	hasLoginError: hasLoginError( state ),
 } );
 
-export default connect( mapStateToProps )( Auth );
+export default connect( mapStateToProps, null, null, { pure: false } )( Auth );
```
dmsnell: are we sure we need this? using local state is fine. the bigger problem is when we are mutating properties on some object where the object's reference stays the same…

```js
render() {
  this.state.list.push( 'a' );

  return <div>{ this.state.list.join() }</div>
}
```

in this case, a shallow comparison won't find any change but we've appended an item to the list, confusing React/Redux.

it should be fine without `{ pure: false }` I think. it seemed to test okay

dmsnell:
Really nice PR @roundhill! This is so much nicer than signing up on the web and we needed it.

lib/auth.jsx
roundhill: Oh! I think I misunderstood what it was for. I thought you needed it if doing local state. Will update.

PR #576, opened 2017-06-23
Add todo lists in the markdown preview

dmsnell:
Formerly this was not working because `marked` didn't render todo lists.
I have switched to a newer markdown parser and added the logic to tie
the render to the underlying note.

Test by viewing the markdown preview and clicking on a todo. The syntax
follows Github Flavored Markdown.

Further, I believe that some of the code highlighting issues may have
been resolved as `highlight.js` now is trying to infer the language on a
block-by-block basis.

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -17,179 +17,181 @@ const renderToNode = (node, content) => {
   node.querySelectorAll('pre code').forEach(highlight.highlightBlock);
 };
 
-export default React.createClass( {
+export default React.createClass({
+  propTypes: {
+    note: PropTypes.object,
+    previewingMarkdown: PropTypes.bool,
+    fontSize: PropTypes.number,
+    onChangeContent: PropTypes.func.isRequired,
+  },
 
-	propTypes: {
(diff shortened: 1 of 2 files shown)


roundhill:
So cool! On Chrome I can't seem to actually click the tick box. I can click the text and then it checks it off ok.

giannissc:
When can we expect this to be released on ubuntu?

giannissc:
I have been waiting for this feature for quite some time now so thank you for implementing it :)

dmsnell:
still a while longer unfortunately @giannissc. we're trying to make sure we get something working across all apps before releasing it here. also we're trying to answer some of the basic questions like is this the right way to toggle todos and display them…

thanks for the interest! we definitely want to make this happen as soon as we can, but we just don't know yet when that will be

giannissc:
Keep up the good work then man. Your and your team's work will be well appreciated :)

dmsnell:
These 45 lines can be easily recreated later and it's not going to come in soon to the Electron app. I'm relegating this into Automattic/Simplenote-United#6

PR #577, opened 2017-06-30
Markdown: Replace `marked` with `showdown` parser/printer

dmsnell:
`marked` doesn't appear to have been maintained for a while while
`showdown` has more recent contributions. Further, `marked` was missing
some functionality that we wanted. Namely, we wanted an equivalence with
Github-flavored Markdown (gfm) and `showdown` does it better.

In this patch we've changed the way we are inserting the content.
Instead of using `dangerouslySetInnerHTML` we're statefully rendering
the content into the DOM and then manipulating it. This change provides
an easier way to interact with the rendered content because we can
operate at the DOM level instead of at the string level. For example,
this is how highlighting now gets inserted.

**Notable changes**
 - Todo items render (though they can only be changed in the plaintext
   view)
 - Tables render
 - Code blocks highlight based on their language
 - Certain dirty/malicious content gets sanitized before display

**Questions**
 - Since the todo items render but aren't editable it could be confusing
   for someone to use, misleading them into thinking it's supposed to
   work

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -1,13 +1,15 @@
 import React, { PropTypes } from 'react';
 import highlight from 'highlight.js';
-import marked from 'marked';
+import showdown from 'showdown';
 import { get, debounce, invoke } from 'lodash';
 import analytics from './analytics';
 import { viewExternalUrl } from './utils/url-utils';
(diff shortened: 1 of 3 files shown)


dmsnell:
Obviated by the merge of #579

PR #578, opened 2017-06-30
Fix: Can't navigate to note list on small widths without resize

dmsnell:
When starting the app in a narrow-width state it can be impossible to
navigate back to the list of notes unles a manual resize of the window
is performed.

The reason for this behavior appears to be improperly initialized state
from the browser window in the `browserShell` wrapper. Instead of
grabbing the width attributes from the window when initializing we were
setting default values of 1024 px and a wide screen. In turn, these
values remained until a resize corrected them, therefore prohibiting the
navigation.

The values are now initialized directly from `window` and the
`browserShell` higher-order component's style has been updated from
`React.createClass()` to return a `React.Component`.

**Testing**
Open the app with a small screen width. In **master** it should be stuck in the note view until resizing. In this branch it should work as expected.

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -349,12 +349,13 @@ export const App = connect( mapStateToProps, mapDispatchToProps )( React.createC
 			appState: state,
 			authIsPending,
 			isAuthorized,
+			isSmallScreen,
 			noteBucket,
+			settings,
 			tagBucket,
 		} = this.props;
 		const electron = get( this.state, 'electron' );
 		const isMacApp = isElectronMac();
-		const { settings, isSmallScreen } = this.props;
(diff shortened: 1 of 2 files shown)


Copons:
Changes LGTM and the issue is fixed.
Maybe `getState` could use a clearer name, like `getScreenWidthAttributes` (a mouthful, but you know I'm bad at naming 🙂).
Though, the file is short enough that is readable either way.

dmsnell:
Thanks @Copons - I deliberated over `getState()` too but decided it was clear enough due to the small size of the module and also due to the fact that the whole thing is pretty specifically limited to window-specific parameters. If you think it's worth it we can update the naming later.

PR #579, opened 2017-06-30
Add/copy rendered markdown

dmsnell:
Markdown: Add ability to copy rendered note HTML

When viewing a note's Markdown preview and with no selection, if you
press the copy shortcut then the rendered HTML for the note will fill
the clipboard buffer. If there is some selected text then the copy will
behave as normal.

**Testing**

Open a note preview and hit <kbd>⌘</kbd>+<kbd>C</kbd> or <kbd>Ctrl</kbd>+<kbd>C</kbd>. When pasting you should see HTML generated from the note. If you select some of the rendered content before copying you should find that when you paste it will be some kind of text view of the contents.

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -11,6 +11,11 @@ const saveDelay = 2000;
 const markdownConverter = new showdown.Converter();
 markdownConverter.setFlavor( 'github' );
 
+const renderToNode = ( node, content ) => {
+	node.innerHTML = markdownConverter.makeHtml( content );
+	node.querySelectorAll( 'pre code' ).forEach( highlight.highlightBlock );
+};
+
 export default React.createClass( {
 
 	propTypes: {
@@ -22,6 +27,7 @@ export default React.createClass( {
 
 	componentWillMount: function() {
 		this.queueNoteSave = debounce( this.saveNote, saveDelay );
+		document.addEventListener( 'copy', this.copyRenderedNote, false );
 	},
 
 	componentDidMount: function() {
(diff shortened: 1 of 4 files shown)


Copons:
Testing works as expected, but I'm having a weird behaviour on selection.

The attached gif should be clear, but this is what seems to happen.
In both Text and Markdown Edit modes, once I select a piece of text, the selection "origin" gets stuck. So, if I try to select another piece of text, the selection always starts from the same position of the first selection starting point.
Clicking outside the selection does not reset it. The only way to remove this behaviour is to move the cursor around with the keyboard. 

jul-03-2017 11-02-49

Though, this happens on master as well, so it's not caused by any changes made here.
Might be worth opening an issue, I guess.

dmsnell:
Thanks @Copons. I saw that behavior too and I think it could relate to draft-js. Note also that I have a PR to pull it out #568.

Did you also experience this in the preview mode? I'm not sure what "Markdown edit" mode means here.

Copons:
> Thanks @Copons. I saw that behavior too and I think it could relate to draft-js. Note also that I have a PR to pull it out #568.

Yay! I totally missed that review request, gonna give it a look asap!

> Did you also experience this in the preview mode? I'm not sure what "Markdown edit" mode means here.

Means that it only happen in Edit mode (which only exists for Markdown, therefore "Markdown Edit" 🙂), but not in Preview.
So yeah, definitely all hints point to Draft.

Copons:
As noted [here](https://github.com/Automattic/simplenote-electron/pull/568#pullrequestreview-47649346), I tried with a fresh `node_modules` to see if something breaks up after removing `marked`, since it's left in https://github.com/Automattic/simplenote-electron/blob/master/webpack.config.dll.js#L15.

I got no issues, apparently because `marked` is imported by some other module (it's there in `node_modules` anyway).
I guess it should be removed from the dll though.

dmsnell:
Updated @copons so that now the vendor bundle builds with `showdown` instead of `marked`

It appears like `jest` is actually pulling in `marked` but only for development. Shouldn't make it into the final bundle.

```bash
npm ls marked
```

taimaiduc:
Is work on window version?

PR #586, opened 2017-07-12
Flip the direction of the arrow on the revision icon

### LINKED ISSUE #584: Revisions icon arrow points in wrong direction
The icon for navigating a note's revisions is turning in the forward direction but revisions indicate history and the arrow should point in the backward direction. See the `gridicons-history` icon for reference.

We should update the icon to fix this. The iOS app already points in the proper direction.

@see https://ctrl.blog/entry/review-simplenote

@drw158 would it be much work to update this?

dmsnell:
Resolves #584 

Formerly the arrow pointed forwards in time but since this is opening up
the history revisions we want it to point backwards in time.

screen shot 2017-07-12 at 5 31 39 pm

### DIFF
--- a/lib/icons/revisions.jsx
+++ b/lib/icons/revisions.jsx
@@ -3,7 +3,7 @@ import React from 'react'
 export default function RevisionsIcon() {
 	return (
 		<svg className="icon-revisions" xmlns="http://www.w3.org/2000/svg" width="22" height="22" viewBox="0 0 22 22">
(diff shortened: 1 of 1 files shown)


dmsnell:
I can't pin it down, but for some reason the way the iOS app does it, with the arrow curling around the top and pointing to 9 feels more natural to me than this icon, pointing at 3.

drw158:
Yep, that's why I initially just settled with the arrow pointing forward in time. This icon has a mirror effect and looks "backward". But, you saying that validates my concern. I can try and alter the icon so that it looks more like the iOS icon.

dmsnell:
@drw158 it may just be the "different" effect, though I think for consistency's sake we should probably _align_ them all the same across the apps.

drw158:
Yea, iOS will be a bit more work to change since those icons have a different thin stroke style, and we would need to replace all 4 icons (Send, History, Collaborate, Trash). Not a ton of work, but does need to happen at some point.

PR #603, opened 2017-08-22
Add/keyboard shortcuts

### LINKED ISSUE #37: Keyboard bindings/shortcuts
Make the UI keyboard navigation enabled:
- [x] Navigate and select a note - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>Up</kbd>/<kbd>Down</kbd>/<kbd>j</kbd>/<kbd>k</kbd>
- [x] Open and navigate tag menu - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>t</kbd>
- [x] Global binding to focus search field - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>f</kbd>
- [x] Toggle markdown preview - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>p</kbd>
- [x] Open/focus list view - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>[</kbd>
- [ ] ~Trash current note~

The app should be fully usable without a mouse.

dmsnell:
Resolves #37 

Adds several global keyboard shortcuts for navigating around Simplenote

There's a pattern here :wink: - we're adopting <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> as the sort of global escape sequence. It was already being used for exporting a note, with <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>e</kbd>. The one exception is toggling between the editor and tag editor and that one should only work when the editor area is focused.

- [x] Navigate and select a note - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>Up</kbd>/<kbd>Down</kbd>/<kbd>j</kbd>/<kbd>k</kbd>
- [x] Open and navigate tag menu - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>t</kbd>
- [x] Global binding to focus search field - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>f</kbd>
- [x] Toggle markdown preview - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>p</kbd>
- [x] Open/focus list view - <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>n</kbd>
- [x] Toggle between editor and tag editor - <kbd>CmdOrCtrl</kbd> + <kbd>t</kbd>

There are some oddities in these: because we don't track whether or not the different window panes are open (they are always open and hidden through crafty CSS) it is hard to selectively enable the hotkeys. That means interactions like selecting the prev/next note work regardless of app state; if you open the list view and navigate to the next note, for example, it will jump back to the note detail view instead of merely highlighting the next note in the list. Similarly, the markdown toggler will always work even if no note is selected or if a note without Markdown support is selected.

**Implementation Note**
You'll notice that this is a fairly "brute-force" PR where I've just gone in imperatively and used global state (window listeners) to trap the hotkeys on individual components and that I have similarly passed up focus functions through reverse-prop-passing. While _dirty_ in most senses of the word the solution I chose was a tradeoff between rewriting the entire way the app works and doing something we could manage. To that end I have tried to be _consistent_ in how the implementations in the various components are done. This should make it reasonably easy to do a search and find all places implementing hotkeys.

**Testing**

I would appreciate considerable manual testing of these shortcuts to see if they make sense and feel right, also that nothing is broken or too strange with the way they work. Did it not work when you expected it to? Did the action seem broken to you? etc…

**Futher work**

This needs a corresponding update in the application documentation and a nice modal overlay showing the shortcuts would be essential in my opinion. Stay tuned!

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -96,7 +96,7 @@ const isElectron = (() => {
   const foundElectron = has(window, 'process.type');
 
   if (foundElectron) {
-    fs = __non_webpack_require__('fs');
+    fs = __non_webpack_require__('fs'); // eslint-disable-line no-undef
   }
 
   return () => foundElectron;
@@ -212,7 +212,7 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
             })
           )
           .then(blob => fs.writeFile(command.filename, blob, 'base64'))
-          .catch(console.log);
+          .catch(console.log); // eslint-disable-line no-console
       }
 
       const canRun = overEvery(
@@ -291,7 +291,7 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
     },
 
     initializeElectron() {
-      const remote = __non_webpack_require__('electron').remote;
+      const remote = __non_webpack_require__('electron').remote; // eslint-disable-line no-undef
 
       this.setState({
         electron: {
@@ -495,7 +495,7 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
     renderDialog({ params, ...dialog }, key) {
       var DialogComponent = Dialogs[dialog.type];
 
-      if (DialogComponent == null) {
+      if (DialogComponent === null) {
         throw new Error('Unknown dialog type.');
       }
 
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -86,6 +86,8 @@ function mapDispatchToProps(dispatch, { noteBucket }) {
     openTagList: () => dispatch(actionCreators.toggleNavigation()),
     resetAuth: () => dispatch(reduxActions.auth.reset()),
     setAuthorized: () => dispatch(reduxActions.auth.setAuthorized()),
+    setSearchFocus: () =>
+      dispatch(actionCreators.setSearchFocus({ searchFocus: true })),
   };
 }
 
@@ -180,6 +182,15 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
       // open tag list
       if (cmdOrCtrl && 'T' === key && !this.state.showNavigation) {
         this.props.openTagList();
+
+        event.stopPropagation();
+        event.preventDefault();
+        return false;
+      }
+
+      // focus search field
+      if (cmdOrCtrl && 'F' === key) {
+        this.props.setSearchFocus();
 
         event.stopPropagation();
         event.preventDefault();
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -83,6 +83,7 @@ function mapDispatchToProps(dispatch, { noteBucket }) {
     setSortType: thenReloadNotes(settingsActions.setSortType),
     toggleSortOrder: thenReloadNotes(settingsActions.toggleSortOrder),
 
+    openTagList: () => dispatch(actionCreators.toggleNavigation()),
     resetAuth: () => dispatch(reduxActions.auth.reset()),
     setAuthorized: () => dispatch(reduxActions.auth.setAuthorized()),
   };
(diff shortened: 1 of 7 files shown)


rodrigoi:
These are the things I've found: 

When using <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>p</kbd> on MacOS (<kbd>Ctrl</kbd>, not <kbd>CmdOrCtrl</kbd>... <kbd>Cmd</kbd> worked fine), on a note without markdown enabled, the editor selects the text from the caret position to the end of the line.

aug-22-2017 13-44-16

Also, <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>[</kbd> may be confusing on non-EN keyboards. Maybe we can use <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>l</kbd>. Maybe adding it to the _View_ menu is a good idea.

We have duplicated shortcuts for the search field: <kbd>CmdOrCtrl</kbd> + <kbd>Shift</kbd> + <kbd>f</kbd> and <kbd>CmdOrCtrl</kbd> + <kbd>f</kbd>.

Otherwise, the code looks good. Is managing window state via redux actions in our future plans?

gwwar:
Which version of electron do we ship with? dev dependencies look pretty out of date, and install instructions note to install electron globally which will pick up the latest published package

rodrigoi:
I usually use the local electron from `node_modules`:

```
> ./node_modules/electron/dist/Electron.app/Contents/MacOS/Electron .
```

gwwar:
Just noting that we should pin versions if we haven't yet for the build. For example, I notice we're using a non LTS of node in Docker that isn't latest either.

gwwar:
I wasn't able to get the docker container going, but testing this local dev seems to work well. There was one I wasn't able to figure out if anything happened for `Open/focus list view`

dmsnell:
The list view one only really applies in narrow screen environments because it gets hidden. Again, some major app architectural pain is behind these quirks

dmsnell:
> We have duplicated shortcuts for the search field: CmdOrCtrl + Shift + f and CmdOrCtrl + f.

Good catch here. I believe that our intention was to make <kbd>CmdOrCtrl</kbd> + <kbd>f</kbd> search within a note and be repeatable to scan through the results whereas the global key search all notes.

dmsnell:
> When using Ctrl + Shift + p on MacOS (Ctrl, not CmdOrCtrl... Cmd worked fine), on a note without markdown enabled, the editor selects the text from the caret position to the end of the line.

How odd! I will look into this

dmsnell:
> Also, CmdOrCtrl + Shift + [ may be confusing on non-EN keyboards. Maybe we can use CmdOrCtrl + Shift + l. Maybe adding it to the View menu is a good idea.

That's a good point. I'll use `l`

Thanks by the way for the great testing!

dmsnell:
@rodrigoi I adjusted the note list view to use <kbd>n</kdb> instead of <kbd>[</kbd> - that was a good suggestion!

kwight:
I had a similar experience to @gwwar , I couldn't tell what difference `focus list view` made. Then I noticed that it always mad the top note active, but didn't necessarily give the list focus, eg. with the tags column open, `cmd+shift+n` makes the top note active (and updates the preview), but my focus was still in the Tags column (I think?).

While the other shortcuts worked as described, I had one unexpected behaviour. Once having opened the tags with `cmd+shift+t`, I wanted navigation keys to move up and down the list of tags so I could select one. I understand that breaks behaviour how things work now, but thought I'd mention it (I did it many times, and it always jarred me to suddenly close the Tags column.

kwight:
This is fine for me – my above comments could be later hopeful improvements, but don't block a release at all.

ParhamP:
@dmsnell `Ctrl + Shift + p` or `cmd + shift + p` don't work in MacOS Yosemite to enable/disable the markdown preview.

PR #604, opened 2017-08-23
Search: Add tag search, multiple term search

### LINKED ISSUE #596: Allow searching for more than one tag at a time
In the web app you can type _tag:tag1 tag:tag2_ which will return all notes with either tag1 or tag2. It would be good to have the same functionality in the electron app.

dmsnell:
Resolves #596

The search has been limited by a couple factors:
 - one could only search for a single tag by
   clicking on it in the tag list
 - one could only search for a full phrase

In this patch the `filter-notes` module gets additional
code to extend the search behavior: namely it allows for
searching by one or more tags and it treats each separate
term in the search input field as part of the query.

What does that mean?

`tag:fruit fresh farm` - searches through notes with the "fruit" tag
which also have the words "fresh" and "farm" in them

This means that each term is joined in an `AND` like expression.

Because someone may want to search for text with spaces in it which
would normally be treated as separate terms, one may surround a text
with double-quotes to search for that as a single piece or phrase.

`"2 + j3"` would search for a note having the exact string `2 + j3` in
it instead of the three terms `2`, `+`, and `j3`.

### DIFF
--- a/lib/editor/matching-text-decorator.js
+++ b/lib/editor/matching-text-decorator.js
@@ -1,5 +1,4 @@
 import SimpleDecorator from 'draft-js-simpledecorator';
-import { escapeRegExp } from 'lodash';
 
 import CssClassWrapper from './css-class-wrapper';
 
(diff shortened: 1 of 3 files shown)


dmsnell:
> moving matchesSearch closer to matchesTag on lib/utils/filter-notes.js.

what do you mean @rodrigoi ?

PR #606, opened 2017-08-23
Update: General version-bumps

dmsnell:
Bumps up to `node@8.4.0`
Updates several dependencies, nothing major


kwight:
Came across #613 , but it's in `master` also; while other spelunking around, things seem good 👍

PR #607, opened 2017-08-23
Fix: Remove reference to `marked` in `note-editor.jsx`

dmsnell:
We removed `marked` and replaced it with `showdown`, but I forgot to
replace this usage (printing notes).

This patch udpates that last remaining reference.


dmsnell:
Merging as this was a silly mistake not to have fixed. We can revert if we need to :smile:

PR #609, opened 2017-08-23
Search: Create new note contents from search field

### LINKED ISSUE #581: Use search text for new note (web app parity)
On app.simplenote.com,
1. Type a phrase in the web app's main search bar
2. Click the plus icon ("+") to create a new note
3. Text from the search bar is duplicated to the first line of the note

Whereas (after clicking the plus icon)...
* the macOS native app clears the search bar and creates an empty note.
* the electron app does NOT clear the search bar and creates an empty note.

The expected action, I think, would be a combination: create a new note using the search bar's text and then clear the search bar.

Tagged for parity: #445

dmsnell:
Resolves #581

Previously when searching and clicking on the new note icon a blank note
would be inserted. Now the contents of the search field will be inserted
into the new note and the search field will clear.

**Testing**

Create new notes. Play around with the search field before doing so.

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -228,7 +228,7 @@ var actionMap = new ActionMap( {
 		},
 
 		newNote: {
-			creator( { noteBucket } ) {
+			creator( { noteBucket, content = '' } ) {
 				return ( dispatch, getState ) => {
 					const state = getState().appState;
 					const settings = getState().settings;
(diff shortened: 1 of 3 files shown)


dmsnell:
Do we think there could be places where this isn't wanted? I think I saw a comment somewhere about only creating the new note content when there were no existing search results.

That is, if we typed in a search and notes returned then *New Note* creates a blank note.
If no notes match the search we inject the content of the search field.

Would that flow make any sense to y'all?

rodrigoi:
I personally find this feature odd. I wasn't aware the web app had this behaviour, and this may be a case when we'll have to remove something instead of adding.
I think that clearing the search field and adding an empty note makes more sense, but feel free to merge this for consistency with the web app.

kwight:
I'd vote to just fix it now for parity with the web app. Intentionally leaving it out feels like a bigger conversation. (And yes, it works well.)

dmsnell:
the macOS app doesn't do it - I'll dig in and compare platforms

PR #617, opened 2017-08-24
Packager: downgrades node engine requirements to LTS

rodrigoi:
In order to prevent prune error when electron packager runs `npm prune --production` on npm versions `> 5.2`. This has been documented on https://github.com/npm/npm/issues/17781.

The PR also includes a small change in light of a packager deprecation warning and an error log fix.

To test, downgrade to node `6.11.2` and npm `3.10.10` and

```
> make package-linux
```

Packing should complete and the folder `release/Simplenote-linux-x64/resources/app/node_modules` should contain only production dependencies.

### DIFF
--- a/builder.js
+++ b/builder.js
@@ -24,7 +24,7 @@ var opts = {
 	author: config.author,
 	platform: builder.getPlatform( process.argv ),
 	arch: builder.getArch( process.argv ),
-	version: electronVersion,
+	electronVersion,
 	appVersion: config.version,
 	appSign: 'Developer ID Application: ' + config.author,
 	out: './release',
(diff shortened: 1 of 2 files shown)


gwwar:
@spen should we also be using v1.6.11 electron and node v7.4.0 here? I forgot about the vulnerability details.

https://github.com/Automattic/wp-desktop/pull/322

spen:
Great point - I would say so. 
I'd have to refresh myself on the details but I'm fairly sure this vuln exists for any usage of node v6.9.0 or lower. This won't be an issue for the projects version of node, rather just the version of node we ship in the app (determined by Electron & the version of Electron we build with).

In researching for Automattic/wp-desktop#322  I found that the node versions found in Electron's history jumps past v6.9.0, landing at v7.4.0 in Electron v1.6.11.

It's a little late so I'll refresh myself on those details tomorrow to be sure, but I'm at least fairly confident in recommending Electron v1.6.11 to avoid that issue.

Also, to reiterate - the project wide version of node won't have any play in this vulnerability so that's safe to be pre v6.9.0.

PR #620, opened 2017-08-29
Redux: Introduce state.ui.visiblePanes

Copons:
Groundwork for the refactor of the panes into independent entities.

Introduce a new Redux state tree, `ui`, with a new property `visiblePanes`.
`visiblePanes` contains an array of currently visible panes.

The available panes are:
- `editor`
- `noteList`
- `tagDrawer`

Provided here is also a set of action that should fit into the upcoming UI refactor.

---

I've opted for more generic action types (e.g. `VISIBLE_PANES_ADD`) instead of more natural ones (e.g. `TAG_DRAWER_SHOW`) in order to keep this flexible enough, should we want to add new panes, or bring existing UI components into `ui.visiblePanes`.

For the same reason, I've chosen to use arrays as `panes` arguments, in order to keep all the functions consistent between each other, and limit unnecessary code.


lib/state/ui/actions.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const enterDistractionFree = () => ( {
+	type: VISIBLE_PANES_SET,
+	panes: [ 'editor' ],
+} );
+
+export const leaveDistractionFree = ( previousPanes = DEFAULT_VISIBLE_PANES ) => ( {
+	type: VISIBLE_PANES_SET,
+	panes: previousPanes,
+} );
```
dmsnell: For now why don't we limit this to existing behaviors?

Copons:
> Why did you see it better to go the other way?

@dmsnell The idea was to hide the implementation as much as possible, without going over-specific.

`TAG_DRAWER_SHOW` is only good for `showTagDrawer`, even though it's implementation is exactly the same as the one for an hypothetical `showSomeOtherDrawer`.

`VISIBLE_PANES_ADD` is both good for `showTagDrawer` and it's future proof.
It's implementation is simple enough that `showSomeOtherDrawer` involves pretty much just a copy-paste.

In other words, I'm going with over-generic action types, hidden behind over-specific (and readable, and obvious) actions.

I'm also wary of `_TOGGLE` actions. They are very handy, until they are not anymore and you have to replace them with `_ADD`/`_REMOVE` everywhere (e.g. to add side effects to just one).

> Also I think the selectors are problematic here because they will lie until we couple the existing behavior into state. Maybe we could just not create them now or instead try to return the correct values based on the imperative state of the app.

I'd definitely avoid wiring up the current state into this.
I'd add a comment saying like "Do not use these selectors just yet!", or probably just outright drop them for now.

lib/state/ui/actions.js
Copons: 👍 
You're right, I was experimenting on this and then decided to drop it and forgot to remove here.

Copons: We do not currently have a distraction free mode though.

If we want to toggle between "editor + panes" / "just editor", we need to know which panes were visible before hiding them.

See for example @drw158 mockup:
- `[ notes, editor ]` -> enter DF -> `[ editor ]` -> leave DF -> `[ notes, editor ]`
- `[ tags, notes, editor ]` -> enter DF -> `[ editor ]` -> leave DF -> `[ tags, notes, editor ]`

My idea is to store the `previousPanes` list into the component internal state, and dispatch it accordingly.
If we don't dispatch a `previousPanes`, no harm done, we just return to the default state.

Copons:
Here you go @dmsnell!

In c2c9e27 I've stripped this down of everything except the toggle tag drawer logic.

In 8c0d4f2 I've wired it in the app, removing all "deprecated" mentions of `appState.showNavigation` and `toggleNavigation`.
Where it made sense, for actions to have the `hideTagDrawer` as side-effect, I've added it directly in the component.
Where it didn't make sense (e.g. no need to explicitly hide the tag drawer when opening the note info, as the tag drawer's outside click took care of that already), I've just dropped the previous logic.

Copons:
Closed in favour of #621 

This might still come in handy for when we'll replace the existing patterns with the new Redux approach.

PR #622, opened 2017-08-30
Update build environment versions - webpack1 > webpack3

dmsnell:
This change is mainly about updating `webpack` from the 1.x versions to
3.x versions. Several secondary changes are the result of that.

The app is now being built with es2015 modules instead of CommonJS
modules. Theoretically this should shrink the build size and maybe speed
things up on account of webpack's tree-shaking algorithm and having
fewer closures in the final code. I haven't tested yet and I don't
anticipate this being significant as an Electron release.

Several small updates are going on too:
  - The node version has been bumped to 7.9.0 to match that running
    inside of Electron.
  - `package.json` doesn't need `npm bin` to prefix commands
  - I'm trying to let Babel focus on the version of Chrome that Electron
    is running and ignore older browsers.
  - We weren't using `AppCache` and didn't need it. It's a deprecated
    API anyway.

@rodrigoi this is updating the node version again. I don't mean to start a version war :wink: but 7.9.0 I think is what Electron is running internally so I figured that would be the best place to settle. If we have build problems we can go back again or fix them with other updates.

**Testing**

Hopefully things should smoke-test safely here.

 - `npm run build`
 - `npm start`
 - `npm run build:prod`
 - make a release and test running the binaries
 - the app loads
 - can logout and login
 - `npm test` (there's a failing test but that's for different reasons)

**Why?**

It's been too long since we updated our core bundler: webpack. This update is timely before the next release and should help us keep our config simple as we move forward with new features.


rodrigoi:
Let the version wars begin! `7.9.0` is safe. The real restriction is npm `> 5.2.0`.

Issue #633, opened 2017-09-21
New Account: UI messed up with no notes available


dmsnell:
When I created a new account from within the app and there were no notes the UI was incomplete.
We need to make sure that it either appears properly or pre-fills an empty note.

<img width="780" alt="screen shot 2017-10-01 at 1 33 34 pm" src="https://user-images.githubusercontent.com/5431237/31057219-27ae7b8a-a6ad-11e7-8bc9-19dcd7960264.png">

This apparently appeared in #603 when I added keyboard shortcuts and conditionally rendered the note editor. This leaves the question: what _should_ happen when we have no notes?

Maybe it's not a problem: if we add the welcome note then the likelihood of having no notes is low so the problem is effectively mitigated?

roundhill:
Resolved by #640

dmsnell:
Kind of, but people can delete that note and get back to an empty list...

PR #640, opened 2017-10-01
Add welcome note when creating a new account

dmsnell:
In the other Simplenote platforms, when someone creates a new account a
welcome message appears in the note list. The Electron app has been
missing this.

This change adds that new message in on account creation.

Addresses #633 because new accounts don't start with an empty note list

Could use some review over the welcome message text.

Eventually it would be cool if instead we had a Redux action with username and password for new account creation and let middleware do all the work to make that account. For the time being, I'm adding this into `boot.js` where it has been handled so far - an attempt at making a limited and pointed PR.

### DIFF
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -18,6 +18,11 @@ import { render } from 'react-dom';
 import { Provider } from 'react-redux';
 import { some } from 'lodash';
 
+import { content as welcomeMessage } from './welcome-message';
+
+import appState from './flux/app-state';
+const { newNote } = appState.actionCreators;
+
 const config = getConfig();
 
 const cookie = parse(document.cookie);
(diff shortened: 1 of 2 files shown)


roundhill:
Awesome stuff @dmsnell. Some copy edits:

```
Welcome to Simplenote!
 
To create a new note, click on the new note button.
 
To search your notes, type in the search bar and enter any text. Simplenote will show you matching results instantly.
 
Got a really important note? Open the note info bar and press the 'Pin to top' button to pin it to the top of the list.
 
Use tags to help organize your notes. Look for the tag bar at the bottom of the note editor.
 
Deleted notes go in the trash. You can restore them if you want, or empty the trash to get rid of them forever.
 
You can access your notes on the web and your other devices. Go to https://app.simplenote.com to get started.
 
We hope you enjoy using Simplenote!
```

dmsnell:
Thanks @roundhill - I updated the text in 1b01f20 and removed some newline issues I thought would occur from the way I had it originally.

Issue #645, opened 2017-10-15
Delay showing window until app is ready (reduce flicker)


secretfork:
#### Steps to reproduce
1. launch app
2. observe empty window loading
3. observe content loading (causing flicker)

#### OS version
Windows 7, SP1 / simplenote 1.0.8

#### Suggestion
```javascript
mainWindow.once( 'ready-to-show', function() {
		mainWindow.show()
} );
```
Added to app.js  in function `activateWindow` will make window stay hidden until presentable.
Read more about `'ready-to-show'` in the electron documentation [here](https://electron.atom.io/docs/all/#using-ready-to-show-event)

dmsnell:
nice suggestion @secretfork - we'll have to try this out if you don't beat us to it with a pull request :smile:

makes me wonder why it's taking so long to get ready, also if we should toss up an indicator that it's loading at least

PR #646, opened 2017-10-16
Delay showing window until app is ready (reduce flicker) #645

secretfork:
This will make window stay hidden until presentable, as discussed in https://github.com/Automattic/simplenote-electron/issues/645
However, i couldn't test it against the current HEAD revision, since it won't compile cleanly on windows.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -86,9 +86,7 @@ module.exports = function main() {
     });
 
     // wait until window is presentable
-    mainWindow.once('ready-to-show', function() {
-        mainWindow.show();
-    });
+    mainWindow.once('ready-to-show', mainWindow.show);
   };
 
   const shouldQuit = app.makeSingleInstance(() => {
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -50,6 +50,7 @@ module.exports = function main() {
       icon: iconPath,
       autoHideMenuBar: true,
       titleBarStyle: 'hidden',
+      show: false,
     });
 
     // and load the index of the app.
@@ -82,6 +83,11 @@ module.exports = function main() {
       // in an array if your app supports multi windows, this is the time
       // when you should delete the corresponding element.
       mainWindow = null;
+    });
+
+    // wait until window is presentable
+    mainWindow.once('ready-to-show', function() {
+        mainWindow.show();
     });
   };
 


dmsnell:
Nice work!

@roundhill I had wondered about the UX implications here - is it ridiculous to think about a splash screen? should we spend the time to figure out why the app takes so long to load?

@secretfork thanks for this effort - by any chance do you have insight into how long it's taking to be ready and what it's doing until then?

secretfork:
@dmsnell now worries. It's actually not that long. Maybe about 0.5 second or so. But it's the small things that give the impression of "flicker". To be totally honest, many people probably wont notice (or care). It's actually quite common that developers fail to adress this issue, and it's also a common issue since many ui frameworks has some kind of hook/callback like 'ready-to-show' to remedy just this.

I don't think you need to think about a splash screen. Now investigating where the delay is is a better idea, however, i don't think 0,5 seconds is very long. Delaying showing the window until it is presentable is what people do. Not just with electron, but with native toolkits too. It's just that sometimes you don't notice the lack of it (that said, if you present the window before it's ready, it might look ok on a fast desktop machine because it's speedy, but it will degrade on slower hardware, so it's *always*a good idea to wait til it's ready).

I did two short mp4s to illustrate the issue:
before - https://www.dropbox.com/s/46zjrc1wjj2rji4/before.mp4?dl=0
after - https://www.dropbox.com/s/ne6g7u8kmuoghln/after.mp4?dl=0

roundhill:
I'm all or this 👍, creates a nicer UX.

desktop/app.js
```diff
@@ -83,6 +84,11 @@ module.exports = function main() {
       // when you should delete the corresponding element.
       mainWindow = null;
     });
+
+    // wait until window is presentable
+    mainWindow.once('ready-to-show', function() {
+        mainWindow.show()
```
roundhill: Would you mind throwing a semicolon at the end here? Just want to try to match the coding style with the rest of the file.

secretfork:
@roundhill Sorry about the missing semi-colon. Amended.

desktop/app.js
```diff
@@ -83,6 +84,11 @@ module.exports = function main() {
       // when you should delete the corresponding element.
       mainWindow = null;
     });
+
+    // wait until window is presentable
+    mainWindow.once('ready-to-show', function() {
+        mainWindow.show();
+    });
```
dmsnell: do we need to be passing the `this` context here? if not we can simplify

```js
mainWindow.once('ready-to-show', mainWindow.show);
```

desktop/app.js
dmsnell: On that note a run through of `prettier` should take care of any and all styling issues.

desktop/app.js
secretfork: That would work just as well. Let me know if you want me to update the pull request.

roundhill:
I tested this on macOS, and it does help the launch experience to look/feel better. There's not much of a delay difference with this patch, it feels like the same launch time for any native macOS app.

desktop/app.js
roundhill: @secretfork cool if you don't mind updating to that one last time I will merge this!

secretfork: @roundhill pushed the change dmsnell suggested (i assume that is what you meant), so feel free to review and merge :+1:

roundhill:
Thanks again @secretfork!

Issue #599, opened 2017-08-08
Display date/time of creation and/or last modification in note list


Topy44:
It would be useful to see the date and time of note creation and/or the last time it was modified in the note list. This could possibly be part of the "Expanded" view, or a new option entirely.

natanielcz:
I took care of this one. If I should change something drop me a line.

thesacredpath:
I see this is still open. Is it still being considered?

iamgabrielma:
Requested on 1869427-zen

iamgabrielma:
Requested on 1942434-zen

belcherj:
Thanks! I’ve added this to our feature requests!

cbcaron15:
is this feature no longer being considered? I can't seem to find the page for this feature request

dmsnell:
@cbcaron15 our feature list is currently private. we consider all feature requests that people make though we don't prioritize them the same way everyone else does (which of course would be impossible because what's a top priority to someone contradicts what someone else consider a top priority).

You can follow the repository or our release notes to discover which features we're working on and put into the app.

PR #648, opened 2017-10-27
#599 Added modification date to note list

natanielcz:


roundhill:
Thanks for the PR @natanielcz! This is one of those "If we add it here, we should add it to the rest of the apps" things. We do get requests to show the date from users every now and again, so might be worth revisiting. Marking for design review.

rodrigoi:
@drw158 would you mind taking a look at this PR?

drw158:
@natanielcz Can you please add a before and after screenshot of the changes you made? Thanks!

natanielcz:

natanielcz:
## Before changes
### Comfy
screenshot from 2018-01-20 12-47-48
### Condensed
screenshot from 2018-01-20 12-48-36
### Expanded
screenshot from 2018-01-20 12-48-53
## After changes
### Comfy
screenshot from 2018-01-20 12-50-17
### Condensed
screenshot from 2018-01-20 12-50-34
### Expanded
screenshot from 2018-01-20 12-50-49

dmsnell:
1. Would we want the date on all views or limit it to comfy and expanded, or just expanded?

2. Further, would we want to take the same kind of idea present in the WordPress.com notifications to split notes into time groupings with headers for each grouping like, "Today", "Yesterday", "Older than two days", "Older than a week" where we could provide more or less in the date string based on the context of the group?

3. Do we need to display `Modified` or is it clear enough as the chosen sort value?

Thanks for the screenshots @natanielcz!

natanielcz:
1. I thought about that and I think that expanded view is too high right now. So I think that point 2 can help to reduce used space in each view type. If we decide to group them like in Wordpress.com notifications we could avoid the `Modified` text.

drw158:
Sorry for the late response. A few things to think about:

- The sorting options might complicate the idea of using time headers to group notes (last modified, last created, alphabetical, reverse).
- We've never had the date in the note list (if we have it's been a very long time), so this is a pretty big change we should think about across platforms if we proceed.
- The date as it is right now is way too long. It should be something no more than 5 or so characters, like `12:38 AM`, `Jan 12`, or `8/12/17`.

natanielcz:
Maybe we should display modification date as a group title when sorting by modification date is active.
For other sorting types, we could display colored bar on the right 
- green, when a note was modified in last 3 days
- orange, when a note was modified in last 7 days
- red otherwise 

screenshot from 2018-02-26 19-52-03

On hover, we could expand this bar to the right and show modification date more precisely

drw158:
I love the idea of adding the modification date when the date sorting is active!

roundhill:
Closing due to inactivity.

Issue #555, opened 2017-05-17
Title in Markdown rendering is no larger than standard text


gerv:
In the "Edit" screen, the note title on the first line is several points bigger than the standard text. If I switch to "Preview", it's the same size as everything else. Adding "##" before and after the title fixed it, but means all my titles in the title list have hashes in and are really ugly. Adding "----" underneath it also fixed it, but that's also ugly because you don't offer a monospace font option, and because it then takes space in the preview :-)

What I think we need is an implicit "#" before and after the title when rendering a note as Markdown.

jakobdkp:
Or have the # in the note title automatically removed from the notes list.

gerv:
Well, maybe. That would require adding similar hash-removal code to every client. But the title is already semantically distinct; making it so that it's styled to match seems like a cleaner solution.

blukas:
I agree that it would be great to remove the # from the title in the note list. I have all my notes in markdown, so the first line always has a #. This makes for a lot of visible clutter and hurts scannability of the note title list.
![example of visual clutter when not masking the # in title](https://user-images.githubusercontent.com/1406519/29850305-cb7f24c2-8ce9-11e7-95ca-670a27c174ee.png)

natanielcz:
I've fixed issue with # tags on the left sidebar but couldn't figure out how to modify font size for the first row of the editor. Looks like it's hidden somewhere in draft-js editor.
![image](https://user-images.githubusercontent.com/1231703/32113624-227fb68a-bb41-11e7-80c1-07ba37b5d3bc.png)

PR #649, opened 2017-10-27
#555 Remove # from title on left panel

natanielcz:
**Testing**

Add notes with a leading `#` in the title (the first line in the note)

Make sure there is a mixture of markdown-enabled notes and plain notes.
Make sure this works as expected when there is a single `#` as well as a string of them, such as `### header`

Make sure that it leaves _wanted_ `#` in there, such as `# # # # Page # # # #`

Note any cases not mentioned here that may be significant to pay attention to.

### DIFF
--- a/lib/utils/note-title.js
+++ b/lib/utils/note-title.js
@@ -28,7 +28,7 @@ const noteTitleAndPreviewMdRegExp = /[# ]*(\S[^\n]*)\s*(\S[^\n]*)?/;
 export const noteTitleAndPreview = note => {
   const content = note && note.data && note.data.content;
   const isMarkdown =
-    note && note.data && note.data.systemTags.some(x => x === 'markdown');
+    note && note.data && note.data.systemTags.some(tag => tag === 'markdown');
 
   const match = isMarkdown
     ? noteTitleAndPreviewMdRegExp.exec(content || '')
--- a/lib/utils/note-title.js
+++ b/lib/utils/note-title.js
@@ -10,6 +10,16 @@
 const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
 
 /**
+ * Matches the title and excerpt in a note's content skipping opening # from title
+ *
+ * Both the title and the excerpt are determined as
+ * content starting with something that isn't
+ * whitespace and leads up to a newline or line end
+ * @type {RegExp} matches a title and excerpt in note content skipping opening # from title
+ */
+const noteTitleAndPreviewMdRegExp = /[# ]*(\S[^\n]*)\s*(\S[^\n]*)?/;
+
+/**
  * Returns the title and excerpt for a given note
  *
  * @param {Object} note a note object
@@ -17,7 +27,12 @@ const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
  */
 export const noteTitleAndPreview = note => {
   const content = note && note.data && note.data.content;
-  const match = noteTitleAndPreviewRegExp.exec(content || '');
+  const isMarkdown =
+    note && note.data && note.data.systemTags.some(x => x === 'markdown');
+
+  const match = isMarkdown
+    ? noteTitleAndPreviewMdRegExp.exec(content || '')
+    : noteTitleAndPreviewRegExp.exec(content || '');
 
   const title = (match && match[1] && match[1].slice(0, 200)) || 'New note...';
   const preview = (match && match[2]) || '';
--- a/lib/utils/note-utils.js
+++ b/lib/utils/note-utils.js
@@ -7,8 +7,7 @@
  *
  * @type {RegExp} matches a title and excerpt in note content
  */
-//empty group at the beginning is intended to match group indexes with regex for Markdown notes
-const noteTitleAndPreviewRegExp = /()(\S[^\n]*)(\s*)(\S[^\n]*)?/;
+const noteTitleAndPreviewRegExp = /\s*(\S[^\n]*)\s*(\S[^\n]*)?/;
 
 /**
  * Matches the title and excerpt in a note's content skipping opening # from title
@@ -16,9 +15,10 @@ const noteTitleAndPreviewRegExp = /()(\S[^\n]*)(\s*)(\S[^\n]*)?/;
  * Both the title and the excerpt are determined as
  * content starting with something that isn't
  * whitespace and leads up to a newline or line end
+ *
  * @type {RegExp} matches a title and excerpt in note content skipping opening #'s from title and preview
  */
-const noteTitleAndPreviewMdRegExp = /(#{1,6}\s+)?(\S[^\n]*)(#{1,6}\s*)?(\S[^\n]*)?/;
+const noteTitleAndPreviewMdRegExp = /(?:#{0,6}\s+)?(\S[^\n]*)\s*(?:#{0,6}\s+)?(\S[^\n]*)?/;
 
 /**
  * Returns the title and excerpt for a given note
@@ -33,10 +33,21 @@ export const noteTitleAndPreview = note => {
     ? noteTitleAndPreviewMdRegExp.exec(content || '')
     : noteTitleAndPreviewRegExp.exec(content || '');
 
-  const title = (match && match[2] && match[2].slice(0, 200)) || 'New note...';
-  const preview = (match && match[4]) || '';
+  const defaults = {
+    title: 'New Note...',
+    preview: '',
+  };
 
-  return { title, preview };
+  if (!match) {
+    return defaults;
+  }
+
+  const [, title, preview] = match;
+
+  return {
+    title: title.slice(0, 200) || defaults.title,
+    preview: preview || defaults.preview,
+  };
 };
 
 export const isMarkdown = note => {
--- a/lib/utils/note-utils.js
+++ b/lib/utils/note-utils.js
@@ -7,7 +7,7 @@
  *
  * @type {RegExp} matches a title and excerpt in note content
  */
-const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
+const noteTitleAndPreviewRegExp = /()(\S[^\n]*)\s*(\S[^\n]*)?/;
 
 /**
  * Matches the title and excerpt in a note's content skipping opening # from title
@@ -17,7 +17,7 @@ const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
  * whitespace and leads up to a newline or line end
  * @type {RegExp} matches a title and excerpt in note content skipping opening # from title
  */
-const noteTitleAndPreviewMdRegExp = /[# ]*(\S[^\n]*)\s*(\S[^\n]*)?/;
+const noteTitleAndPreviewMdRegExp = /(#+ +)?(\S[^\n]*)\s*(\S[^\n]*)?/;
 
 /**
  * Returns the title and excerpt for a given note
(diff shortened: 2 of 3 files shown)


lib/utils/note-title.js
```diff
@@ -10,14 +10,29 @@
...
  * Returns the title and excerpt for a given note
  *
  * @param {Object} note a note object
  * @returns {Object} title and excerpt (if available)
  */
 export const noteTitleAndPreview = note => {
   const content = note && note.data && note.data.content;
-  const match = noteTitleAndPreviewRegExp.exec(content || '');
+  const isMarkdown =
+    note && note.data && note.data.systemTags.some(x => x === 'markdown');
```
roundhill: Might be nice to use `tag` instead of `x` here for easier code readability?

lib/utils/note-title.js
```diff
@@ -10,14 +10,29 @@
...
 
 /**
+ * Matches the title and excerpt in a note's content skipping opening # from title
+ *
+ * Both the title and the excerpt are determined as
+ * content starting with something that isn't
+ * whitespace and leads up to a newline or line end
+ * @type {RegExp} matches a title and excerpt in note content skipping opening # from title
+ */
+const noteTitleAndPreviewMdRegExp = /[# ]*(\S[^\n]*)\s*(\S[^\n]*)?/;
```
roundhill: For reference, here's how we do it in the macOS app with a string replace:

https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/NSString%2BCondensing.m#L60

But I think the regex is fine too.

dmsnell: Just a note that this works exactly the same as `noteTitleAndPreviewRegExp` because of the _kleene star_ at the beginning `[# ]*` - in other words, this RegExp will _always_ match when the other one matches

natanielcz: Exactly. This one is just a little bit extended to catch `#` at the beginning. We can't join them to one because we don't want to omit `#` for not markdown notes.

lib/utils/note-title.js
natanielcz: fixed

roundhill:
Just tried this out, very cool! Can we update the regex to match for a space character after `#`? That would match up the behavior with the native Mac app. So `#hashtag` would still show the hash.

lib/utils/note-title.js
dmsnell: Thanks for clarifying @natanielcz - I had been _assuming_ that pulling it out on normal notes is also okay.

lib/utils/note-title.js
```diff
@@ -10,14 +10,29 @@
...
  * Returns the title and excerpt for a given note
  *
  * @param {Object} note a note object
  * @returns {Object} title and excerpt (if available)
  */
 export const noteTitleAndPreview = note => {
   const content = note && note.data && note.data.content;
-  const match = noteTitleAndPreviewRegExp.exec(content || '');
+  const isMarkdown =
+    note && note.data && note.data.systemTags.some(tag => tag === 'markdown');
```
dmsnell: totally fine here as-is, but this makes me think we need to get together a function like `isMarkdown( note )` since that's becoming an important test condition.

natanielcz:
I've modified regex and extracted method isMarkdown as you suggested. I also renamed file note-title.js to note-utils.js

lib/utils/note-utils.js
```diff
@@ -7,7 +7,7 @@
  *
  * @type {RegExp} matches a title and excerpt in note content
  */
-const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
+const noteTitleAndPreviewRegExp = /()(\S[^\n]*)\s*(\S[^\n]*)?/;
```
dmsnell: is this change a typo here?

lib/utils/note-utils.js
```diff
@@ -17,7 +17,7 @@ const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
  * whitespace and leads up to a newline or line end
  * @type {RegExp} matches a title and excerpt in note content skipping opening # from title
  */
-const noteTitleAndPreviewMdRegExp = /[# ]*(\S[^\n]*)\s*(\S[^\n]*)?/;
+const noteTitleAndPreviewMdRegExp = /(#+ +)?(\S[^\n]*)\s*(\S[^\n]*)?/;
```
dmsnell: couple questions:
 - do we want to trap on `# # # # ` and friends?
 - valid headers are one through six long, should we trap that?
 - is a tab or quad-space different than the ASCII space?

```js
/(#{1,6}\s+)…
```

lib/utils/note-utils.js
```diff
@@ -27,17 +27,23 @@ const noteTitleAndPreviewMdRegExp = /[# ]*(\S[^\n]*)\s*(\S[^\n]*)?/;
...
 };
 
+export const isMarkdown = note => {
+  return (
+    note &&
+    note.data &&
+    note.data.systemTags.includes(tag => tag === 'markdown')
+  );
+};
+
```
dmsnell: oh thanks! sorry I didn't mean to imply this had to be done here in this PR, but I guess it's in a reasonable spot for it

lib/utils/note-utils.js
natanielcz: No. This is intended. Standard library for RegEx doesn't have named groups so I had to add a fake empty group at the beginning so in further part we can refer to the same indexes for notes in markdown and not. Few lines below I've changed indexes to match new regex [1] -> [2], [2] -> [3]

lib/utils/note-utils.js
natanielcz: good questions. 
- I think that `# # # #` should be displayed on the list as `# # #` because this is a proper behavior of markdown. 
- nice catch. I'll change that
- probably yes. \s+ is a better solution

lib/utils/note-utils.js
natanielcz: That was only moving few lines to the separate method. I did while making other changes.

rodrigoi:
@natanielcz would you mind adding some testing instructions?

dmsnell:
> @natanielcz would you mind adding some testing instructions?

@rodrigoi I updated.

natanielcz:

natanielcz:
I've fixed this issue. I've also removed `#`'s from the notes second line.

rodrigoi:
Thanks @natanielcz !!

roundhill:
New bug :)

It seems to not show the second preview line now, when a note is markdown:

screen shot 2018-01-23 at 11 18 00 am

natanielcz:
Sorry, I've missed your comment. I'll investigate this issue.

natanielcz:
Should work with recent changes

lib/utils/note-utils.js
```diff
@@ -17,12 +28,19 @@ const noteTitleAndPreviewRegExp = /(\S[^\n]*)\s*(\S[^\n]*)?/;
...
-  const match = noteTitleAndPreviewRegExp.exec(content || '');
+
+  const match = isMarkdown(note)
+    ? noteTitleAndPreviewMdRegExp.exec(content || '')
+    : noteTitleAndPreviewRegExp.exec(content || '');
 
   const title = (match && match[1] && match[1].slice(0, 200)) || 'New note...';
   const preview = (match && match[2]) || '';
 
   return { title, preview };
```
dmsnell: so this would end up being "above and beyond" this PR, but maybe it would be a good time to add it in. how would you feel about making these result indices clearer by eliminating them?

```js
const match = …;

if ( ! match ) {
	return { title: 'New note…', preview: '' };
}

const [ /* full match */, title, preview ] = match;

return {
	title: title.slice( 0, 200 ) || 'New note…',
	preview: preview || '',
};
```

I can see value in doing this and also in leaving it the same way.

natanielcz: Looks clearer. I'm on it

dmsnell: Thanks! this shouldn't have been your burden but I noted it simply as a result of the fact that the indices ended up being important in this PR

lib/utils/note-utils.js
```diff
@@ -7,8 +7,19 @@
  *
  * @type {RegExp} matches a title and excerpt in note content
  */
+//empty group at the beginning is intended to match group indexes with regex for Markdown notes
```
dmsnell: let's move this into the docblock already present.

```js
/**
 * Matches the title…
 *
 * …
 *
 * Note! there is an "extra" capture group of whitespace at the
 * beginning of the pattern. This is there so that we can maintain
 * parity among the capture group indices with the pattern for
 * notes where Markdown is active.
 *
 * @type {RegExp} …
 */
const noteTitleAndPreviewRegExp = …
```

lib/utils/note-utils.js
dmsnell: actually we might as well also skip the pattern matching when the content is empty. no need. I like the cleanliness of `( content || '' )` but actually that's also needless.

would you be interested in catching empty-content cases above the match and return the defaults before then?

```js
if ( ! content ) {
	return {
		title: 'New note…',
		preview: '',
	}
}
```

lib/utils/note-utils.js
natanielcz: this isn't valid since my previous commit. I've removed this note.

lib/utils/note-utils.js
```diff
@@ -0,0 +1,56 @@
...
+ * @type {RegExp} matches a title and excerpt in note content
+ */
+const noteTitleAndPreviewRegExp = /\s*(\S[^\n]*)\s*(\S[^\n]*)?/;
+
+/**
+ * Matches the title and excerpt in a note's content skipping opening # from title
+ *
+ * Both the title and the excerpt are determined as
+ * content starting with something that isn't
+ * whitespace and leads up to a newline or line end
```
dmsnell: nitpick: we should have a newline here. I wouldn't worry about it though - if you want to you can add it in but otherwise we'll come back later and add it. I don't want this to hold-up the PR

lib/utils/note-utils.js
```diff
@@ -0,0 +1,56 @@
...
+ * @type {RegExp} matches a title and excerpt in note content skipping opening #'s from title and preview
+ */
+const noteTitleAndPreviewMdRegExp = /[#]{0,6}\s+?(\S[^\n]*)[\s]*[#]{0,6}\s*?(\S[^\n]*)?/;
```
dmsnell: this comment is more for my own note-taking as I review the RegExp

```js
const noteTitleAndPreviewMdRegExp = new RegExp(
	'[#]{0,6}\s+?' + // start matching when we have a header with space after it (ungreedy)
	'(' + // match group 1 - title
		'\S[^\n]*' + // title is some non-whitespace character and then everything until the end of the line
	')' +
	'[\s]*' + // followed by some amount of whitespace including the newline
	'[#]{0,6}\s*?' + // the next bit might be prefixed by a header, and if it is we won't use it for the excerpt
	'(' + // match group 2 - preview
		'\S[^\n]*' + // preview is the first non-whitespace character and then everything until the end of the line
	')?' // there may or may not be an excerpt
);
```

actually now I'm wondering if this is really what we want. it may also match on `content: '$# not a title'`

**questions**
 - is there a reason we have the `[]` where they aren't required, like around the `#` and `\s`?
 - should we be processing the first line in two stages for Markdown notes? It seems like this RegExp pattern isn't handling Markdown matches properly when they aren't prefixed by the hash.

Markdown Title:
 - Skip all whitespace prefixing content
 - Grab first line
 - If prefixed by hashes that would indicate a header then strip them

Markdown Preview:
 - Skip whitespace after title
 - Grab until next newline
 - If prefixed by hashes that would indicate a header then strip them

dmsnell:
```js
const match = /\s*(\S[^\n]*)\s*(\S[^\n]*)?/.exec( content || '' );
if ( ! match ) {
	return defaults;
}

const [ /* full match */, first, second ] = match;

const title = isMarkdown
	? first.replace( /^#{1,6}\s+/, '' )
	: first;

const preview = isMarkdown
	? second.replace( /^#{1,6}\s+/, '' )
	: second;

return {
	title: title.slice( 0, 200 ) || default.title,
	preview: preview || default.preview
}
```

thoughts?

natanielcz: `[]` are not required. You're right that this one won't work. Here is form that will work: `/(?>#{0,6}\s+)?(\S[^\n]*)\s*(?>#{0,6}\s+)?(\S[^\n]*)?/`. Please take a look on that @dmsnell 

I think that doing one regex is better than making 3 but I'm not an expert.

dmsnell: Sadly I don't think that `(?>…)` exists in Javascript.

One vs. three RegExps isn't usually clear; it's not always better to have fewer and it depends on the characteristics of the expression _and_ the document.

In this case yours would probably be fine and faster though I just end up questioning the complexity of the source code. I don't know that this is a performance problem and I find the logic simpler when operating on the line-by-line basis - I almost also wrote `content.split('\n').slice(0,2)` up there too!

Actually I'm pretty sure that performance shouldn't be an issue here. After the initial note list generation it should cache the previews for the notes and the only one being recalculated should be the one that's in the editor changing or coming in from the server in updates.

We can replace the "atomic groups" in this new RegExp with non-capture groups which are similar… `(?:#{0,6}\s+)…` They might get the job done.

natanielcz: You're right that with caching any approach should be enough fast for this simple task. But those regex's are not so complicated so, for now, I'll live it as it is. And I'll use non-capture groups instead of atomic groups which don't exist in js.

dmsnell:
Thanks again @natanielcz!

Issue #422, opened 2016-11-02
AppImage for Linux


probonopd:
* Operating system: Linux
* Simplenote version: Latest

Please provide an [AppImage](http://appimage.org/) for Linux. This will allow Simplenote to run on many Linux distributions at once. `electron-builder` has support for it, so should be really easy.

swalladge:
@probonopd while this would be a nice idea, I also suggest reaching out to package maintainers for the distro you are wanting to run Simplenote on to see if someone will be willing to create distro-specific packages for it (if it doesn't already exist). For example, I am currently maintaining a [pkgbuild on the aur](https://aur.archlinux.org/packages/simplenote-electron-bin/) for easy installation on Archlinux.

dmsnell:
@probonopd that sounds like a great idea. I looked into this a little bit while on a flight yesterday but was limited due to lack of reliable internet. To my knowledge it doesn't look like there is anything to prevent us from doing it. Of course, we might get there faster if someone offers a PR for it :wink:. We'll update here as we learn more and if we put out some code to this end. Thanks for filing the issue and pointing us at AppImage!

@swalladge it's great to hear about awesome work like what you are doing. thanks for making it possible to more easily spread Simplenote ❤️ !

probonopd:
@dmsnell it's so trivially easy that I don't think its worthwhile to do a PR:

1. Edit package.json to use the latest electron-builder.

2. Edit package.json — add after "build": { ... }:

```
    , "linux": {
      "target": ["AppImage"]
    },

```
Then, `npm install` or `npm run dist or npm run dist -- --linux=AppImage` if you want to explicitly set target and do not edit package.json.

natanielcz:
I let myself to take care of this one because it wasn't modified for a year. I didn't create AppImage installer because I think that still there aren't many people using this format. Instead I've added support for RPM. @probonopd if you think that I'm wrong and there is a lot of people using AppImage and FlatPak I can take care of this two installers too. 

I'd be grateful If someone could test changes from my recent PR because I don't have RPM based distro.

swalladge:
@natanielcz btw it doesn't matter how many people use the appimage format, because it's simply an executable file with no dependencies. Others like snap and (afaik) flatpak require installing a manager software first though.

probonopd:
[...]

Issue #530, opened 2017-03-18
Request: Official RPM for Fedora


eduardohki:
I would love to have an Official RPM package for Fedora, just like the DEB package.

Thanks!

natanielcz:
@eduardohki: I did it in #422 because it was just one additional line. After merging my pull request we may close this request.

yoyoyojomo:
FYI #422 has been merged

dmsnell:
Thanks @yoyoyojomo

yoyoyojomo:
It looks like something still needs to be tweaked so the rpm shows up at https://github.com/Automattic/simplenote-electron/releases

PR #650, opened 2017-10-27
#422 Added support for building rpm and AppImage installers. Moved generating deb and tar.gz from fpm to eletron-builder

natanielcz:


dmsnell:
This looks like some great stuff @natanielcz - can you help us by providing some background on the changes? It looks like there's far more in play than just adding RPM support.

natanielcz:
I've updated PR title. Generally, I added support for two new formats: rpm and AppImage. And move deb and tar.gz packaging from fpm to electron-builder. To make it easier for you during releasing a new version I updated makefile and added building deb, rpm, tar.gz, and AppImage to target `linux`.

This pull request refers also to #530.

I think that we should also consider moving Windows and Mac installers to electron-builder to have it all in one place. 

However, there is still a bug in electron-builder electron-userland/electron-builder#1907 and application icon isn't visible in the Software / Ubuntu Store. Electron builder will allow us to make also Snap installer for Linux, DMG, MAS (Mac Application Store) and PKG for Mac, MSI installer, NSIS Installer and Web Installer but also AppX (for Windows Store) for Windows.

natanielcz:
One more fix. We don't need copy icons manually so I've deleted that for Linux build.

electron-builder.json
```diff
@@ -0,0 +1,37 @@
+{
+  productName: "Simplenote",
```
rodrigoi: would you mind adding quotes to the keys on this file?

rodrigoi:
@natanielcz would you mind rebasing this PR? Thanks!

natanielcz:

natanielcz:
@rodrigoi: Rebased.

rodrigoi:
thanks @natanielcz !! I have a couple comments:

I don't have `electron-builder` as a global. To `make package-linux` work I've renamed `-PACKAGE_WIN32` to `ELECTRON_BUILDER` and used it instead of `@electron-builder`. It was also complaining about `projectDir`, so the command looked like:

```make
package-linux: linux
	@$(ELECTRON_BUILDER) --linux --x64
```

We usually build on macOS, and `rpm` fails with a warning of a missing dependency. Maybe we should add `brew install rpm` to the documentation.

And finally, `yarn.lock` is creating an artificially large diff, and at this moment, this project does not use `yarn`. You may want to create a new PR to add `yarn`.

Besides that, this tests very well. All packages work (except for the image on Software/Ubuntu Store you've [mentioned before](https://github.com/Automattic/simplenote-electron/pull/650#issuecomment-341082624))

I'll try to get another set of 👀 to review this, but this looks really promising. Thanks for the hard work!

natanielcz:
@rodrigoi You're right that electron-builder installed globally is currently required but I'm working on it. I removed electron-builder from global dependencies and will test a little bit different approach.

natanielcz:
@rodrigo, @roundhill @dmsnell : I've also created `.snap` the installer which could be put in the [snapcraft store](www.snapcraft.io). Give me your emails and I'll add you to collaborators to my test Simplenote app on snapcraft so You can see and test installation from there.

roundhill:
Seems to be working quite well 👍 I did get this warning when attempting to install the .deb:

screen shot 2018-03-06 at 4 04 48 pm

natanielcz:
Lintian returns more thanks that one error. I'll compare my deb with deb installed from the website to check what else should be done

natanielcz:
@roundhill which app did you use to install this deb? I can't reproduce that issue.

roundhill:
> @roundhill which app did you use to install this deb? I can't reproduce that issue.

I used whatever the default Ubuntu installer is, is that Ubuntu Software Center? Sorry, I'm not much of a linux person :)

roundhill:
My VM is actually on an older version of Ubuntu (15), maybe it's not an issue in newer releases?

natanielcz:
@roundhill You may be right about that. I have Ubuntu 17.10 and I don't see anything like that but I'll try to fix at least critical linter problems. It may take a while due to some electron limitations and my lack of knowledge about loading system libraries.

roundhill:
OK, I will update it and let you know if I see the same error.

roundhill:
> OK, I will update it and let you know if I see the same error.

No error on 17.10 👍 

Can you remove the `projectDir` argument that @rodrigoi mentioned? It doesn't appear to be needed, unless I'm missing something. It prevents me from building on macOS without removing it.

natanielcz:
I removed that already

roundhill:
I think this is good to go. I've tested the builds and they work great.

Issue #508, opened 2017-01-30
Disable/Hide the tags' Edit button when there are no tags


hoverduck:
#### Steps to reproduce
1. Open Simplenote
2. Click the Tags icon at the upper left
3. Ensure there are no existing tags

#### What I expected
To either be unable to see or click the Edit button, or otherwise have some indication that it's not going to work without any existing tags

#### What happened instead
Clicking the button cycles the label between Edit and Done with no visual indication of what that means

#### OS version
Tested on Simplenote v1.0.8 on Windows 10 and Linux Mint 18

natanielcz:
I've created PR for this feature. #652

natanielcz:
We should probably close this one. Fixed in #652

PR #652, opened 2017-10-31
#508 Added hiding Edit button on tag list when there is no tags

natanielcz:


dmsnell:
@natanielcz thanks again for the contribution - can you provide some context in the description here? why you want this change, screenshots or at least descriptions of the behavior you want to change?

also a quick note, we will probably not want to introduce the nested ternary, preferring instead something like explicit `if` structures or better a separate function

natanielcz:
@dmsnell: Before my changes when we had no tags we were able to open left pane and click on `edit` button in the tags section. Then we were able to click again this time on `Done` button but actually, nothing was happening. Screenshots:
image
image

After my changes, this panel looks like this below. Edit button disappeared.
image

lib/tag-list.jsx
```diff
@@ -53,22 +53,31 @@ export class TagList extends Component {
...
+          className="tag-list-edit-toggle button button-borderless"
+          tabIndex="0"
+          onClick={this.props.onEditTags}
+        >
+          {this.props.editingTags ? 'Done' : 'Edit'}
+        </button>
+      );
+    }
+    return null;
+  }
```
roundhill: I hate to be the 'semicolon person', but could you please add one here ;)

dmsnell: no need to @roundhill - that's what `prettier` is for :wink:

dmsnell:
@natanielcz I was going to leave a comment then I felt like it would be annoying for you - so I hope you don't mind but I just inlined that function. if you do, please just raise it here.

dmsnell:
@roundhill looks like this is ready to go to me!

roundhill:
Thanks all!

PR #653, opened 2017-11-01
Fixes issues when switching accounts (#634 and #641)

### LINKED ISSUE #634: New Account: Note from previous account cross-over
When I logged out and then created a new account from inside the app I got the notes from the previous account. They should have cleared out of state.

What appears to have happened is not that the notes carried over but rather the note that was currently in the editor. Once I started operating inside the note editor or added a note the old data disappeared.

### LINKED ISSUE #641: Logout/Login: Notes don't load until other app action
When logging out of the app and then logging in again I'm noticing that the notes don't load. Once I try to create a new note then all of the other notes load as they should have when I first logged in.

This is an issue that only happens when logging out and logging in during the same app session.

It appears related to #634 and I believe that what is happening is that for some reason we're not reestablishing a proper connection with Simperium. I'm actually seeing a valid `notesLoaded` Redux action fly by after switching accounts, but (unlike the `tagsLoaded` message) contains `notes: []` regardless of the account.

Once I create a new note then I see another `notesLoaded` message dispatch with the actual `notes` in its payload. Maybe this is a timing issue? Maybe the code flows aren't clear. I'm not sure and I wasn't able to trace the flow to resolve this problem.

roundhill:
* Fixed an issue with the `isIndexing` flag in Simperium, fixes #641.
* Clear note editor upon unauthentication (fixes #634)

**To Test**
1. Sign in to the app, observe that your notes are loaded.
2. Sign out, and sign in with another account.
3. Your notes should load and everything should work as normal!

To test the fix for #634, you need to switch to an account that has *no notes at all* from an account that *has notes*. The editor should be blank after you switch.

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -250,6 +250,7 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
       actions.authChanged();
 
       if (!client.isAuthorized()) {
+        actions.closeNote();
         return resetAuth();
       }
 
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "serve-favicon": "2.4.3",
     "showdown": "1.7.2",
     "showdown-xss-filter": "0.2.0",
-    "simperium": "0.2.6"
+    "simperium": "0.2.7"
   },
   "optionalDependencies": {
     "appdmg": "0.4.5"


roundhill:
> Probably something deeper, but you may have seen the behavior where after logging in no notes appear until the new note button is clicked. That's something I think probably needs to be addressed before the release.

That's what I fixed in `node-simperium`, are you still seeing that behavior on this branch?

dmsnell:
> are you still seeing that behavior on this branch?

yeah @roundhill I tested a few times to make sure, but I didn't `rm -rf node_modules && npm install` so maybe I had the old `node-simperium` library in there

roundhill:
> but I didn't rm -rf node_modules && npm install so maybe I had the old node-simperium library in there

Ah yes, I should have put that in the instructions. Should work better after you `npm install`

dmsnell:
@roundhill I just did a full clean build on this branch and reinstalled the `npm` packages and still had the problem. After logging back into my test account it appeared as though there were no notes. The tag list was filled out and I could select tags, but no actual notes appeared in the note list until I created a new note, at which time all of the notes magically appeared.

roundhill:
Sort of got it resolved in Slack, I'm going to merge this and keep testing :)

PR #654, opened 2017-11-06
Fixes '0' from displaying near the tags edit button

roundhill:
I noticed that a `0` was being displayed for the tags edit button after you signed out/in again:

screen shot 2017-11-06 at 3 37 13 pm

This fixes that by checking for a zero `tags` array length. Note: The `tags` array is always created in the state so I don't think we need to worry about it being `undefined` here.

### DIFF
--- a/lib/tag-list.jsx
+++ b/lib/tag-list.jsx
@@ -64,7 +64,7 @@ export class TagList extends Component {
       <div className={classes}>
         <div className="tag-list-title">
           <h2 className="panel-title theme-color-fg-dim">Tags</h2>
-          {tags.length && (
+          {0 !== tags.length && (
             <button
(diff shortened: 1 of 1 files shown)


lib/tag-list.jsx
```diff
@@ -64,7 +64,7 @@ export class TagList extends Component {
       <div className={classes}>
         <div className="tag-list-title">
           <h2 className="panel-title theme-color-fg-dim">Tags</h2>
-          {tags.length && (
+          {0 !== tags.length && (
```
dmsnell: good catch! either this or maybe `tags.length > 0 &&` might be more common. this is good either way

roundhill: cool, I'll correct that.

PR #655, opened 2017-11-07
Fix Building for Windows

roundhill:
We updated the `electron-builder` version in https://github.com/Automattic/simplenote-electron/commit/fe4117546a7e9fcb6242a1a276bee9c6816c0b2c which caused some issues with building the app for Windows. Some options were deprecated (many of the strings such as app title are now pulled from the main `package.json` file, and I needed to move where the output directory is defined. I also removed the NSI config file, since that doesn't appear to be needed any longer.

**To Test**
 * Run `make package-win32` (instructions in the field guide for getting the signing certs)
 * You should end up with an .exe installer in the `release` directory that successfully installs on windows.
 * Also try `make package-osx` and `make package-linux`, they should still output binaries as well.


dmsnell:
did you consider the work in #650 for this @roundhill?

rodrigoi:
@roundhill this is testing well. It'll be interesting to evaluate the approach on https://github.com/Automattic/simplenote-electron/pull/650/ (which at the moment is not working for me) in the future, but for the moment this works fine.

roundhill:
> did you consider the work in #650 for this @roundhill?

I did not, but it can still be applicable. My main goal was fixing the Windows build.

PR #660, opened 2017-11-08
Remove autoHideMenuBar so we get the menu back on windows/linux.

### LINKED ISSUE #659: Where'd the Menu Go?
In 1.1.0 RC 2 the menu has disappeared in Windows/Linux. You can open it by pressing the `alt` key on windows. I think we want to show the menu all of the time.

Reported by @hoverduck

roundhill:
The update to electron must have made `autoHideMenuBar` actually work :) I removed it so we get the menu back on Windows/Linux:

screen shot 2017-11-08 at 11 18 33 am

Fixes #659

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -48,7 +48,6 @@ module.exports = function main() {
       minWidth: 370,
       minHeight: 520,
       icon: iconPath,
-      autoHideMenuBar: true,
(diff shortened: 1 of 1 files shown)


dmsnell:
Does this leave it open on macOS? If so, :shipit:

PR #662, opened 2017-11-09
Fix external links in Electron 

### LINKED ISSUE #656: 'Forgot Password' link does nothing
The forgot password link on the signup form doesn't seem to do anything now.

roundhill:
Links with `window.open` and `target=_blank` weren't opening properly in Electron.

I found a solution at https://github.com/electron/electron/issues/1344#issuecomment-171516261 that will intercept the links and open them in an external browser.

We have some code in `url-utils.js` to open a link externally as well, but looks like we don't use it anywhere now.

**To Test**
* Click on the 'forgot password' link in the auth view, and links in the about screen. They should open in an external browser window when in electron (and in the browser as well)

Fixes #656

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -76,7 +76,7 @@ module.exports = function main() {
 
     mainWindowState.manage(mainWindow);
 
-    mainWindow.webContents.on('new-window', function(event, linkUrl){
+    mainWindow.webContents.on('new-window', function(event, linkUrl) {
       event.preventDefault();
       shell.openExternal(linkUrl);
     });
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -76,6 +76,11 @@ module.exports = function main() {
 
     mainWindowState.manage(mainWindow);
 
+    mainWindow.webContents.on('new-window', function(event, linkUrl){
+      event.preventDefault();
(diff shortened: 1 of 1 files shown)


dmsnell:
Code looks fine. I don't think we have to guard against user-input links but it's worth pondering a bit, if we can open external URLs that came from untrusted sources (shared notes, for example).

It didn't seem obvious that `new-window` could come from user-input code though, rather only from our own code. Is that right?

roundhill:
> It didn't seem obvious that new-window could come from user-input code though, rather only from our own code. Is that right?

Hmm I can't quite tell either. Do you know how the markdown parser handles links? I would guess that it would pick up on those links as well and run them through this code.

dmsnell:
> Do you know how the markdown parser handles links?

we should be able to do whatever we want with links from the markdown.

https://github.com/Automattic/simplenote-electron/blob/c829a2f4404b4676084a4f54ae84b65f1c4d61a5/lib/note-detail.jsx#L16

roundhill:
I created https://github.com/Automattic/simplenote-electron/issues/665 to have a look at the markdown links.

Going to merge this one for now so we can get the app-created links working again.

PR #664, opened 2017-11-09
Fix Windows Build for 32 Bit

roundhill:
The `electron-builder` update was building 64 bit by default and installing to a different directory than the app previously went.

Now we:
* Force 32 bit build only
* Use `perMachine` flag to install the app to the `program files (x86)` folder, so updates work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ package: build-if-changed
 	@cp -R $(THIS_DIR)/dist $(DESKTOP_BUILD_DIR)
 
 package-win32: win32
-	@$(PACKAGE_WIN32) ./release/Simplenote-win32-ia32 --platform=win --config=./resources/build-config/win32.json
+	@$(PACKAGE_WIN32) ./release/Simplenote-win32-ia32 --win --ia32 --config=./resources/build-config/win32.json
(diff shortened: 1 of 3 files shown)


dmsnell:
Why was this problematic, @roundhill?

roundhill:
> Why was this problematic, @roundhill?

The app would install to a different folder, and you'd have two instances of the app on windoze.

roundhill:
And also testers on 32 bit windows would get an error in the installer that the app required 64 bit!

dmsnell:
Might it be possible to…

 - check if the machine supports 32bit and only install that one if so
 - check if the app already exists at the location and choose based off of that?

Or, is there any reason that 64bit is advantageous? If we had no history here, would we prefer 32bit over 64bit? Have we only so far produced 32bit builds?

roundhill:
> Have we only so far produced 32bit builds?

That's my understanding, yup. Not sure about advantages we'd get with 64 bit, other than that I really liked Super Mario 64.

dmsnell:
Thanks @roundhill - I don't have any problem keeping up with our existing behaviors. :shipit:

PR #666, opened 2017-11-09
Show placeholders when user has no notes

roundhill:
I had a go at adding placeholders to the notes list and note editor when there are no notes to show. I used the native macOS app as inspiration. The editor toolbar buttons are also disabled.

@drw158 have a look! I used `$gray` for the color, but maybe we should add a bit of opacity to it so it's not quite so bold?

screen shot 2017-11-09 at 3 24 29 pm
screen shot 2017-11-09 at 3 23 37 pm


roundhill:
> easy time splitting off early (up in or around app.jsx)

I could probably fake it with a few divs. 🤔 I thought it would be nice to keep the editor toolbar at the top with the dividers, but I can certainly fake that too :)

dmsnell:
> I thought it would be nice to keep the editor toolbar at the top with the dividers

personally I wasn't sure how hard or easy it would be. would probably be nice to at least explore the idea and see because if it could be done I think it might end up as something we would rather have. if it's too much of a headache (and this would definitely seem plausible) then please do what is pragmatic.

roundhill:
@dmsnell I got it working mainly in `app.jsx`. The only tradeoff was that you don't see the toolbar buttons at all now, but I think it works.

dmsnell:
> The only tradeoff was that you don't see the toolbar buttons at all now, but I think it works.

To me that seems like a boon - the toolbar buttons are useless without a note to operate on :wink:

roundhill:
- [ ] Bug: You can't drag the placeholder toolbar to reposition the app when in electron.

roundhill:
> The alignment of the header is off when running inside the electron chrome.

Thanks @rodrigoi I updated it for the macOS version in 
31bbdb2

rodrigoi:
it also looks awesome on Windows and Ubuntu. :shipit: !

PR #667, opened 2017-11-10
About View: Added Dynamic Year

roundhill:
A tester noticed that the copyright date was behind in the about screen. How about we make it dynamic?

screen shot 2017-11-09 at 4 01 49 pm

### DIFF
--- a/lib/dialogs/about.jsx
+++ b/lib/dialogs/about.jsx
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react';
+import moment from 'moment';
 import SimplenoteLogo from '../icons/simplenote';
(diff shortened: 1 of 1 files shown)


lib/dialogs/about.jsx
```diff
@@ -110,7 +112,7 @@ export default React.createClass({
               href="https://automattic.com/"
               rel="noopener noreferrer"
             >
-              &copy; 2016 Automattic, Inc.
+              &copy; {thisYear} Automattic, Inc.
```
dmsnell: also available without any external libraries as `new Date().getFullYear()` (already in local time that one)

roundhill: WFM!

PR #668, opened 2017-11-10
Fix context menu for markdown preview

roundhill:
@designsimply reported that the right-click menu was not working for the markdown preview.

This updates the `context-menu` code so that it will allow the menu to show from the markdown div:

screen shot 2017-11-09 at 4 19 10 pm

p.s. `prettier` wanted to format the code that way :)


dmsnell:
> p.s. prettier wanted to format the code that way :)

😆

PR #677, opened 2017-12-01
Markdown: Interactive preview mode switching implicity

dmsnell:
In all of the simplenote apps there has been an inherent distinction
between edit mode and preview mode when interacting with
Markdown-enabled notes.

In this patch we're exploring the implicit transitions, such that the
default view for a note is the rendered markdown "preview" but as soon
as you click to edit the note if flips back into the edit view.

This was inspired by prior work and for now is an experimental change.

clickmarkdown mov

### DIFF
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -156,13 +156,19 @@ export const NoteDetail = React.createClass({
   },
 
   render: function() {
-    const { filter, fontSize, previewingMarkdown } = this.props;
+    const { filter, fontSize, previewingMarkdown, setEditorMode } = this.props;
 
     const content = get(this.props, 'note.data.content', '');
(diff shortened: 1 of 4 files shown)


drw158:
This is cool, hopefully I'll have time to test it out. Judging from the screenshot, I think the biggest pain point will be users not knowing how to exit 'edit' mode. ESC key is nice, but not obvious in the UI. 

I could imagine something like the top bar changing color with an X button to escape edit mode. The bar would go back to normal in preview mode.

I still like the idea of something in-between, like highlighting the syntax, but not displaying it exactly how it will look in the preview. I do like the idea of simplifying things and getting rid of the idea of a preview though.

dmsnell:
Closing because it was an experiment but we don't have the time to dedicate to figuring out if this is a good way to go.

PR #682, opened 2017-12-06
Fix Improper Tag Creation from Duplicating Tags

roundhill:
https://github.com/Automattic/simplenote-android/issues/474 led me back to an issue in this app.

Our tag model is a little funny, it has an `id` which _should be_ a lower-case and percent escaped version of the tag name the user enters. So a tag named `Woot` should have an id of `woot`, `W@@t` should have an id of `w%40%40t`, etc.

In this app we were letting Simperium [assign a random id](https://github.com/Simperium/node-simperium/blob/master/src/simperium/bucket.js#L18) whenever we added a new tag:
screen shot 2017-12-06 at 3 22 07 pm

This was causing the issue in other apps that were checking for an existing id before adding a new tag. I fixed it by using `bucket.update` instead of `bucket.add`, which allows a custom id to be passed and still creates a new object.

**To Test**
* Create a new tag in this app.
* Head over to another Simplenote client like iOS or android and add the new tag to another note.
* It should have added the same tag you originally created and not make a duplicate.

This PR does not attempt to fix existing tags in the app that were created improperly, as I couldn't quite wrap my head around how to do that :)

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -136,7 +136,8 @@ export const actionMap = new ActionMap({
     newTag: {
       creator({ tagBucket, name }) {
         return () => {
-          tagBucket.add({ name }, () => {
+          // tag.id should be a lower case version of the tag name
+          tagBucket.update(name.toLowerCase(), { name }, () => {
             // dispatch( this.action( 'loadTags', { tagBucket } ) );
           });
         };
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -136,8 +136,9 @@ export const actionMap = new ActionMap({
     newTag: {
       creator({ tagBucket, name }) {
         return () => {
-          // tag.id should be a lower case version of the tag name
-          tagBucket.update(name.toLowerCase(), { name }, () => {
+          // tag.id must be the tag name in lower case and percent escaped
+          const tagId = encodeURIComponent(name.toLowerCase());
+          tagBucket.update(tagId, { name }, () => {
             // dispatch( this.action( 'loadTags', { tagBucket } ) );
           });
         };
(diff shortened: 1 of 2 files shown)


lib/flux/app-state.js
```diff
@@ -136,7 +136,9 @@ export const actionMap = new ActionMap({
     newTag: {
       creator({ tagBucket, name }) {
         return () => {
-          tagBucket.add({ name }, () => {
+          // tag.id must be the tag name in lower case and percent escaped
+          const tagId = encodeURIComponent(name.toLowerCase());
+          tagBucket.update(tagId, { name }, () => {
             // dispatch( this.action( 'loadTags', { tagBucket } ) );
```
dmsnell: what's the point of this empty function? should we just go ahead and remove it now? should we reload the tags here?

lib/flux/app-state.js
```diff
@@ -136,7 +136,9 @@ export const actionMap = new ActionMap({
     newTag: {
       creator({ tagBucket, name }) {
         return () => {
-          tagBucket.add({ name }, () => {
+          // tag.id must be the tag name in lower case and percent escaped
+          const tagId = encodeURIComponent(name.toLowerCase());
```
dmsnell: is this how the mobile clients create them or are they using a kind of UUID? not sure if we have reason to match them up with mobile clients but I'm pretty sure that the tag id is intentionally unrelated to the label

someone creates a label, adds it to a bunch of notes, decides the label should be something else (e.g. "work" to "naperville file"), renames the label, now that tag has nothing to do with the text "work"

roundhill: Yeah it doesn't make much sense after you rename a tag. But it is indeed how the other apps are doing it: https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/TagListViewController.m#L237

lib/flux/app-state.js
roundhill: I removed it in c6b0f8e. The tags list appears to update fine after adding the tag so I don't think it's needed.

PR #686, opened 2017-12-29
Limit protocols allowed for external links

roundhill:
We received a report that the app is happy to open any url provided to it in the markdown preview. This PR locks it down to only open `http`, `https`, and `mailto` links.

**To Test**
 * Create a markdown note with some links in it, some of them with http, https, mailto and some other naughty ones such as `file://`.
 * Only the allowed protocols should open when clicked.

### DIFF
--- a/lib/utils/url-utils.js
+++ b/lib/utils/url-utils.js
@@ -1,5 +1,7 @@
 import { has } from 'lodash';
 
+const allowedProtocols = ['http://', 'https://', 'mailto:'];
+
 let state = {
   isRunningElectron: null,
 };
@@ -19,4 +21,15 @@ function init() {
   state = { ...state, isRunningElectron, openExternalUrl };
 }
 
-export const viewExternalUrl = url => state.openExternalUrl(url);
+export const viewExternalUrl = url => {
+  const lowerCaseUrl = url.toLowerCase().trim();
+  const urlStartsWithProtocol = function(protocol) {
+    return this.startsWith(protocol);
+  };
+
+  if (!allowedProtocols.some(urlStartsWithProtocol, lowerCaseUrl)) {
+    return;
+  }
+
+  state.openExternalUrl(lowerCaseUrl);
+};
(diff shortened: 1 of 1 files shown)


lib/utils/url-utils.js
```diff
@@ -19,4 +21,15 @@ function init() {
...
-export const viewExternalUrl = url => state.openExternalUrl(url);
+export const viewExternalUrl = url => {
+  const lowerCaseUrl = url.toLowerCase().trim();
+  const urlStartsWithProtocol = function(protocol) {
+    return this.startsWith(protocol);
+  };
+
+  if (!allowedProtocols.some(urlStartsWithProtocol, lowerCaseUrl)) {
+    return;
+  }
```
dmsnell: we may want to use something like `URL` instead which does actual URL parsing…

```js
const protocol = new URL( url ).protocol;
if ( ! allowedProtocols.some( allowed => allowed === protocol ) ) {
	return;
}
```

roundhill: Nice! Updated in b3e2d2bfbac66fdbc93f3256d213d333db8aa923. Unfortunately it doesn't look like you can use fat arrow functions to support `this`.

dmsnell: you're right, but we don't even need `this`…

```js
if ( ! allowedProtocols.some( allowed => allowed === protocol ) ) {
	return;
}
```

this is because `Array.prototyp.some()` takes a callback and it passes in as the first argument of that callback the current value of the array being iterated over.

```js
allowedProtocols.some( ( currentValue, indexOfArray, originalArray ) => someTruthyOrFalsyReturnValue() )
```

roundhill: I see! Could have sworn I tried that... anyways it works great.

tmchartrand:
This is definitely the right decision for a default, but it would be great to have the whitelist be user-modifiable to allow for other use cases (file and other protocols)

PR #692, opened 2018-01-15
Editor: Replace draft-js with Monaco

dmsnell:
We've had a few issues with draft-js including performance conderns.
Recently we've also been wanting to explore Markdown syntax highlighting
and while draft-js is ammenable to that it leaves considerable
implementation details up to us.

In #568 I proposed a change to replace draft-js with a plain old
textarea which greatly improved the performance for large notes but also
had the major drawback that we lost our ability to easily decorate text.

Microsoft's Monaco editor powers Visual Studio code and is incredibly
performant thanks to the fact that they have rebuild all the editor
behaviors in custom ways to handle big files and syntax highlighting.

In this patch we're taking advantage of the work that Microsoft has done
and I'm proposing a switch to Monaco. It includes Markdown syntax
highlighting for free. This comes through a React wrapper around the
editor.

The only real drawback I see here is that Monaco is much larger than the
textarea in terms of raw KB though I'm not sure how it compares to
draft-js. In Electron this isn't too much of a concern but if/when we
push this version of the app out to the web it could come with other
problems. Monaco also doesn't officially support mobile devices either I
think.

simplenotemonaco mov

**Remaining work**

Monaco already has great auto-indent support and provides many of the
behaviors for us which we were implementing by hand with draft-js or in
the textarea. However, it doesn't provide all such as auto-lists.

There remains some dataflow issues that exist already in the app too but
which aren't solved here. When making changes it's possible to queue up
changes but get updates from the server in the meantime and mess up the
editor.

For the purposes of continuing to develop this branch I haven't removed
the draft-js-based editor module. It still provides all the examples of which
editor behaviors we need to implement (auto lists etc…)

The editor isn't forced to use a monospace font but it does use one by
default. As we talk about monospace fonts we should make sure this
respects the app's font choices.

> For now I have added a fixed/variable-width font switch based on whether
> Markdown support is enabled for a given note. It needs finessing but I
> think that it conveys a reasonable behavior for us to explore.

We're not highlighting search results here. We _can_ do it but it requires
writing up some code to do so and I didn't have the time to do it when
putting together this branch.

**Notes**

 - performance is stellar
 - markdown highlighting is 💯


rodrigoi:
This works great. Monaco looks like a great choice. Tried it with really large notes and I've noticed no performance degradation.
The only thing I've found is purely aesthetic. There's a double scrollbar (at least on Mac OS) and the editor does not fill the screen.

screen shot 2018-01-14 at 22 45 28

dmsnell:
> There's a double scrollbar (at least on Mac OS) and the editor does not fill the screen.

Thanks @rodrigoi - I believe that these are trivial styling adjustments that need to be made.
The big deal is that we have to specify pixel values for the editor and that's why there's a
resize handler in there. We should probably rearrange a few things in there…

 - Use `requestAnimationFrame()` for resizing adjustments instead of a throttle
 - Figure out the right pixel values to use instead of the naïve styling I have in here

Of course I would welcome help as pixels are my weak point.

roundhill:
This would be a great place to sync up on the styling of the editor for the live highlighting. We should standardize the coloring and styles for things like headers and block quotes. All for @drw158 I suppose :)

I have a similar branch for the native Mac app here: https://github.com/Automattic/simplenote-macos/tree/add/markdown

dmsnell:
> This would be a great place to sync up on the styling of the editor for the live highlighting.

@roundhill it's probably obvious but we do have full control over the colors in Monaco. In this PR all I needed to overwrite was the background color, but we have options for the foreground color, for the highlighting, selection, caret?, and all the syntax elements

natanielcz:
Those gifs looks nice. One day I had an idea to replace Simplenote editor and create something like [Typora](https://typora.io/) has. I mean that they created live markdown editor which doesn't require switching modes. What do you think about that? I think that more people without knowledge of markdown could use this advanced mode. Then we could extend it to have inline formatting options like [Medium Editor](https://github.com/IonicaBizau/medium-editor-markdown).

dmsnell:
@natanielcz Typora is definitely a nice product! We've been working on some kind of syntax highlighting for Markdown for a while. Maybe the first actual code was in #418. Also played with a change of the edit/preview idiom in #677.

While I doubt we'll have inline formatted text in edit mode any time soon, I think syntax highlighting on the Markdown itself is going to happen. Just lots of figure out on how to try and make sure the experience is harmonized among the various Simplenote apps (web, this one, mobile, and macOS native).

Thanks for the input though! We're anxious to get this in ourselves.

roundhill:
Here's a little styling proposal for the live markdown editor (this is from the native app):

screen shot 2018-01-23 at 9 25 53 am

I was thinking it'd be good to not do any coloring at all, except for some grays for quotes and inline code. Keep it Simple :) Let us know what you think @drw158

dmsnell:
Fine @roundhill but I think when we brought this up before we were also opposed to changing the size of the headers and sizes in general. (In this case with Monaco it's harder to change sizes but in draft-js it was actually trivial)

Let's see those CSS color values!

roundhill:
> opposed to changing the size of the headers and sizes in general

That's ok with me, but IMHO the large headers feel weird when you are editing.

> Let's see those CSS color values!

Link/Image: https://github.com/Automattic/simplenote-electron/blob/master/scss/variables.scss#L2
Quote/code (light theme): https://github.com/Automattic/simplenote-electron/blob/master/scss/variables.scss#L9

roundhill:
> opposed to changing the size of the headers and sizes in general

Oh, do you mean to not have them be a different size at all? Keep them the same size as the body? I'm cool with that. Can we still do the trick where the first line is larger though?

drw158:
I like the idea of just using blues and grays, and not changing the font size for the # Headers.

roundhill:
> not changing the font size for the # Headers.

What if we made the headers **Bold**?

dmsnell:
@roundhill and @rodrigoi I have updated this. I think I've taken care of your concerns. it has some differences from the existing solution but I don't think they are necessarily better or worse (many things are better)

fun fact, you can use <kbd>F1</kbd> to open up the Monaco command input and there are some fun tools included.

monacoeditor mov

lib/note-detail.jsx
```diff
@@ -156,7 +156,12 @@ export const NoteDetail = React.createClass({
   },
 
   render: function() {
-    const { filter, fontSize, previewingMarkdown } = this.props;
+    const {
+      filter,
```
roundhill: This is now unused and can be deleted.

roundhill:
@dmsnell is it cool if I try add some CSS updates? I'd like to:

* Style the scrollbar to match the notes list scrollbar
* Fix scrollbar positioning/padding
* Make the first line of text always larger

dmsnell:
> is it cool if I add some CSS updates?

you would be my hero, not like you aren't already, but you'd be my HERO!

roundhill:
I've added a few fixes, but man it's a bit hacky to get it all working right.

There's still a few issues:

- [ ] Monaco rearranges the divs for the content, so I think it might be impossible to make the first line always larger?
- [ ] I can't seem to click to put the cursor anywhere but the very end of the last line.
- [ ] The scrollbar margins seem difficult to move around to match the notes list.
- [ ] There is vertical scrolling for notes that are small and shouldn't show any scrollbars.

dmsnell:
> so I think it might be impossible to make the first line always larger?

maybe through forcing the CSS but we can add a custom range and have more control. things like spacing are awkward but font size adjusts fairly well

```js
var decorations = editor.deltaDecorations([], [
    { range: new monaco.Range(1,1,1,1), options: { isWholeLine: true, inlineClassName:'first-line'}},
]);
```

```
.first-line {
    font-size: 180%;
}
```

dmsnell:
>  I can't seem to click to put the cursor anywhere but the very end of the last line.

I'm not sure I follow - what are you trying to do? Set a specific cursor position programmatically? If so, check out [modifyPosition](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.imodel.html#modifyposition) on the editor model

roundhill:
> I'm not sure I follow - what are you trying to do? Set a specific cursor position programmatically? If so, check out modifyPosition on the editor model

In Safari I can't click to place the cursor anywhere in the last line of a note, unless it is the very end of the line. I think it may have been because of some CSS margin hacks I added...

roundhill:
> but we can add a custom range and have more control

I tried adding the range w/o much luck, could you have a go at adding it @dmsnell? I think we're pretty close to getting this looking good.

rodrigoi:
@roundhill @dmsnell I just gave this a try and still works great. I did notice the same things @roundhill reported here: https://github.com/Automattic/simplenote-electron/pull/692#issuecomment-367490754
Should we feature flag this?

dmsnell:
@rodrigoi I don't see much use in feature-flagging it because then it'd just be maintaining things without the benefit of the merge tools. I'd personally rather see us keep the branch up to date or merge it entirely. I simply don't have the time available to do this now.

dmsnell:
Sadly I don't have enough time to give this the attention it needs. I'm closing the issue and I welcome someone to reopen it as they see fit.

There are two or three separate changes being made in this PR and some of them would be good to do separately.
 - Track changes to the notes vs. just update notes on changes
 - Update the cursor position when changes come in
 - Replace the editor

Updating the cursor is a big one that would be a good step towards this goal.
Thanks to all who provided feedback and review.

PR #698, opened 2018-01-23
Disable app window drag and drop

roundhill:
Electron doesn't disable drag and drop by default, so we'll explicitly stop navigating when the `will-navigate` event fires.

See: https://discuss.atom.io/t/prevent-window-navigation-when-dropping-a-link/24365

**To Test**
* Try and drag anything into the app. The only thing that should work is when you drag text into the editor.
* Click on external links, they should still open in an external browser.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -81,6 +81,11 @@ module.exports = function main() {
       shell.openExternal(linkUrl);
     });
 
+    // Disables navigation for app window drag and drop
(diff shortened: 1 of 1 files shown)


roundhill:
@dmsnell would you mind 👀 this one?

roundhill:
Thanks, I simplified it like you suggested 🌮

PR #699, opened 2018-01-24
Move markdown toggle to the toolbar

roundhill:
While working on markdown in the native app, @drw158 and I discussed the current `Edit/Preview` toggle that is displayed in the editor. It takes up quite a bit of vertical space in the editor so we decided to change it to a preview button in the toolbar instead. I'm making that change to this app as well in this PR:

screen shot 2018-01-24 at 12 40 31 pm

screen shot 2018-01-24 at 12 40 16 pm

**To Test**
* Enable markdown for a note, the preview icon should appear.
* Click the icon, it should show the markdown preview and change to a cancel icon
* View the trash, the preview icon should disappear.
* Return to the note, and disable markdown, the icon should disappear.


dmsnell:
I like the way this opens up screen real-estate but I know we need to think about harmony between the apps, also consider #677 for a paradigm-shift in the preview.

drw158:
Only hesitation I have is that it bumps up the icon count to 6 for small viewport widths. 5 is the ideal maximum. I wouldn't say it's a blocker though, since this is a desktop app.

roundhill:
I'd like to move forward with merging this, as we're going to ship the same icon/UX in the macOS app. If we decide to change how the preview is displayed/dismissed like in #677, we can then do it in both apps for consistency 👍

PR #702, opened 2018-01-30
Warn users when before signing out with unsynced notes

roundhill:
Some users are losing notes when they sign out because for one reason or another they have unsynced notes. This is a quick solution to that which searches for unsynced notes upon signout and warns the user if so. An option is presented to view the web app to reconcile their unsynced notes:

Windows/Linux:
screen shot 2018-01-29 at 4 01 46 pm
screen shot 2018-01-29 at 3 53 33 pm

On the web (currently not used in production), a humble confirm dialog is displayed:
screen shot 2018-01-29 at 4 04 50 pm

One nice thing that we could also do is offer to export the notes for the user. Maybe in place of the 'Visit Web App' button?

**To Test**
* Take the app offline, and create a note.
* Click Simplenote menu and then `Sign Out`. You should see the alert dialog.
* Verify all of the buttons work as expected.
* Also verify that you don't see the prompt if all of your notes have synced.

Refs: https://github.com/Automattic/Simplenote-United/issues/7


roundhill:
One other thing to note, the iOS and macOS PRs _also_ check for notes with queued changes that haven't synced yet. I haven't dug into whether or not we can do that with `node-simperium` for this PR .

roundhill:
> One other thing to note, the iOS and macOS PRs also check for notes with queued changes that haven't synced yet.

@beaucollins helped me get a check for this added to `node-simperium` as well 👍 

@dmsnell could you have a look a this one at your convenience? 🙏

lib/dialogs/settings.jsx
```diff
@@ -37,6 +38,74 @@ export const SettingsDialog = React.createClass({
...
+          new Promise((resolve, reject) =>
+            getVersion(note.id, (e, v) => (e || v === 0 ? reject() : resolve()))
+          );
+
+        Promise.race(notes.map(noteHasSynced)).then(
+          () => onSignOut(), // All good, sign out now!
+          () => this.showUnsyncedWarning() // Show a warning to the user
+        );
+      })
+    );
```
dmsnell: what's going on here? is there a missing `if` statement body? something is wrong.

roundhill: Whoops, I had an `if` there that didn't need to be there. Callbacks are silly.

dmsnell: :smile: looks better now

lib/dialogs/settings.jsx
```diff
@@ -37,6 +38,74 @@ export const SettingsDialog = React.createClass({
...
+          return;
+        }
+
+        // Also check persisted store for any notes with version 0
+        const noteHasSynced = note =>
+          new Promise((resolve, reject) =>
+            getVersion(note.id, (e, v) => (e || v === 0 ? reject() : resolve()))
+          );
+
+        Promise.race(notes.map(noteHasSynced)).then(
```
beaucollins: Should this be `Promise.all` not `Promise.race`? You want to make sure that none of  the notes are unsynced right?

lib/dialogs/settings.jsx
```diff
@@ -37,6 +38,72 @@ export const SettingsDialog = React.createClass({
...
+        return;
+      }
+
+      // Also check persisted store for any notes with version 0
+      const noteHasSynced = note =>
+        new Promise((resolve, reject) =>
+          getVersion(note.id, (e, v) => (e || v === 0 ? reject() : resolve()))
+        );
+
+      Promise.race(notes.map(noteHasSynced)).then(
```
beaucollins: Looks like I commented on an outdated change but the `Promise.race` remains.

It seems you want to check if any of the notes have a version of `0` and if any of them do have a version of `0` then you will want to show `showUnsyncedWarning`.

`Promise.race` will return the first promise that resolves or rejects and ignore all others where as `Promise.all` will wait until all Promises resolve or fail early if any of them reject.

roundhill: Hmm ok, I think I indeed want Promise.all then. If we ever see a version of zero we should stop and show the warning. Otherwise, keep checking.

roundhill: Updated in 5324686. Nice catch!

dmsnell: good catch @beaucollins - the `race` was from my original idea where only the unsync'd notes would resolve and so we wouldn't have needed to wait for all Promises - an early abort. this was a wrong idea. :smile:

roundhill: I've learned a lot about `Promise` via this PR! :)

dmsnell: @roundhill you don't need to learn that much about promises. just let your "yes" be "yes" and your "no" be "no" :wink:

Issue #602, opened 2017-08-17
Offline changes from desktop can be overwritten on next sync.


zedgoat:
#### Steps to reproduce
1. Desktop Client 1.0.8, Windows 10. Working as expected for a period of months.
2. Experienced network issue for approx. 2 days, continued to use desktop client exclusively during the window.
3. Then, when using iOS client after this window, noticed that existing notes had not been updated, new notes had not been synced.
4. When desktop regained connection, notes that had been opened on iOS w/ new update time pushed to client, overwriting changes.
5. Desktop history wasn't present in version history. New notes created on desktop remained there and didn't sync to other clients.

Note, didn't use other clients (web, mac) in the interim - just the W7 desktop and iOS client.

#### What I expected
Desktop changes would have taken priority over older notes and published as the current version.

#### What happened instead
Desktop changes were destroyed across ~6 notes by older content from the first mobile client to connect. Notes were not recoverable in the note history.

#### OS version
iOS - current release as of 18.07.2017
Windows 7 - 1.0.8

dmsnell:
Thanks for finding this and getting some steps in! We'll try and see if we can replicate.

I'm confused some on step 4

> When desktop regained connection, notes that had been opened on iOS w/ new update time pushed to client, overwriting changes.

Were the notes new on the desktop and unsynchronized in iOS? What exactly was overwritten? New notes on iOS should nominally be different entities altogether than new notes on the desktop.

zedgoat:
Thanks for the reply, to clarify:
- The notes that were overwritten were existing notes that were on both Desktop and iOS.
- The offline desktop client had new content, that had not been synced.
- The online mobile client (iOS) had the old versions of the notes, that I opened to check.
- When the desktop client came online, the new desktop content was removed, and the old content from the mobile client was there.

New notes created on the desktop didn't appear to sync into the iOS client at the time, but were still there.

dmsnell:
This seems related to an issue we had in the past. We might need some time to dig in but we take sync bugs seriously. Thanks again for the report and debugging!

zedgoat:
All good! Good luck hunting it down. Put the suggestion forward via twitter / support that it might be worth having a diff interface for this stuff?

Breaks the simplicity mantra, but aside from if there are sync issues it would also be useful where multiple clients have been offline for a while and the notes have diverged quite a bit.

gerv:
[...]

PR #706, opened 2018-02-02
Fix offline/reconnect note sync with active note in the editor 

roundhill:
I believe I've discovered some bugs that can help with #602, but only for notes that are currently active in the editor.

**Bugs fixed**
 * We were applying duplicate changes when the app came online again. I added a check for an equal diff, and we don't do any patching if so. You can reproduce in `master` by going offline, making a change in electron, then going online again. You'll see the change apply twice.
 * We weren't updating the editor content properly when the note updated after coming online again. You can reproduce this in `master` by applying a change in another app while electron is offline. When you reconnect, you'll see the content update in the notes list but not in the editor! This could be why a lot of users lost content. I fixed this by using the `loadAndSelectNote` action vs. the `selectNote` action.

You can test this in general by taking the app offline and making changes. Also try making changes to a note from different clients at the exact same time. The diffs should merge properly (as best as it can).

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -1,4 +1,4 @@
-import { partition } from 'lodash';
+import { get, partition } from 'lodash';
 import update from 'react-addons-update';
 import ActionMap from './action-map';
 import isEmailTag from '../utils/is-email-tag';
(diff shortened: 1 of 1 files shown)


dmsnell:
For reference: p2XLIi-1i0-p2

roundhill:
Here's what I think is going on:

When we reach this code, Simperium has already merged whatever it got from the server with whatever was _saved locally_ (so basically anything but the current 2 second autosave content). So the `loadNotes` call at line 440 should be ok, I think. 

It then moves on to check if there was anything in the currently active editor that may need to be merged, which is why it uses the object in the state because that gets updated on every keypress in the editor, I believe?

> if this patch is helping then I'm all for it, though I think it's just brushing the surface.

Yes, this is a band-aid, but I think could help a scenario we recently received where two users working on a shared note would often lose content. It fixes two things:

1. Changes applied twice to an active note in the editor when coming back online and receiving changes from the server.
2. The editor was not updating with the new note content after receiving changes in the same scenario as #1.

dmsnell:
> It then moves on to check if there was anything in the currently active editor that may need to be merged, which is why it uses the object in the state because that gets updated on every keypress in the editor, I believe?

@roundhill I think we should merge this as it seems to fix _a problem_ we are experiencing, but this data flow is what scares me and I think it's good to be working in these parts. we need some terms for the different places we're storing note data.


anyway, thanks for talking through this. hard stuff!

roundhill:
OK, sounds good. The conversation doesn't have to stop here! It'll be interesting to see how this impacts the number of reports we get about sync issues.

PR #709, opened 2018-02-08
Updates babel, adds async/await

beaucollins:
Uses `babel-preset-env` in the `.babelrc` config and targets electron browser.

Also adds `async/await` via `babel-plugin-transform-async-to-generator` and shows example use in  settings dialog.

While this is perfectly good:

```js
const noteHasSynced = note =>
  new Promise((resolve, reject) =>
  getVersion(note.id, (e, v) => (e || v === 0 ? reject() : resolve()))
);

Promise.all(notes.map(noteHasSynced)).then(
  () => onSignOut(), // All good, sign out now!
  () => this.showUnsyncedWarning() // Show a warning to the user
);
```

Using `async/await` allows for a more straight forward approach:

```js
for (const note of notes) {
  if ((await getVersion(note)) === 0) {
    // Show a warning to the user
    this.showUnsyncedWarning();
    return;
  }
}
onSignOut();
```

Also, simperium/node-simperium#56 adds a `Promise` based interface to all bucket methods so it could potentially improve code readability by changing callback code to `async/await` code.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,7 @@
 
   },
   "presets": [
-    [ "env", { "targets": { "electron": "1.7" } } ],
+    [ "env", { "targets": { "electron": "1.7", "uglify": true } } ],
     "react"
   ],
   "plugins": [
--- a/.babelrc
+++ b/.babelrc
@@ -1,15 +1,14 @@
 {
   "env": {
-    "test": {
-      "presets": [ "es2015", "react" ]
-    }
+
   },
   "presets": [
-    [ "es2015", { "modules": false } ],
+    [ "env", { "targets": { "electron": "1.7" } } ],
     "react"
   ],
   "plugins": [
     "babel-plugin-transform-class-properties",
-    "babel-plugin-transform-object-rest-spread"
+    "babel-plugin-transform-object-rest-spread",
+    "babel-plugin-transform-async-to-generator"
   ]
 }
--- a/lib/dialogs/settings.jsx
+++ b/lib/dialogs/settings.jsx
@@ -38,29 +38,25 @@ export const SettingsDialog = React.createClass({
     });
   },
 
-  onSignOutRequested() {
+  async onSignOutRequested() {
     // Safety first! Check for any unsynced notes before signing out.
     const { onSignOut, noteBucket } = this.props;
     const { notes } = this.props.appState;
     const { getVersion } = noteBucket;
 
-    noteBucket.hasLocalChanges((error, hasChanges) => {
-      if (hasChanges) {
+    if (await noteBucket.hasLocalChanges()) {
+      this.showUnsyncedWarning();
+      return;
+    }
+
+    for (const note of notes) {
+      if ((await getVersion(note)) === 0) {
         this.showUnsyncedWarning();
         return;
       }
+    }
 
-      // Also check persisted store for any notes with version 0
-      const noteHasSynced = note =>
-        new Promise((resolve, reject) =>
-          getVersion(note.id, (e, v) => (e || v === 0 ? reject() : resolve()))
-        );
-
-      Promise.all(notes.map(noteHasSynced)).then(
-        () => onSignOut(), // All good, sign out now!
-        () => this.showUnsyncedWarning() // Show a warning to the user
-      );
-    });
+    onSignOut();
   },
 
   showUnsyncedWarning() {
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "babel-eslint": "7.2.3",
     "babel-jest": "20.0.3",
     "babel-loader": "7.1.2",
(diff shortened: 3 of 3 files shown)


lib/dialogs/settings.jsx
```diff
@@ -38,29 +38,25 @@ export const SettingsDialog = React.createClass({
...
+      this.showUnsyncedWarning();
+      return;
+    }
+
+    for (const note of notes) {
+      if ((await getVersion(note)) === 0) {
         this.showUnsyncedWarning();
         return;
       }
+    }
```
dmsnell: is this going to change the semantics? will this spread out each async action and then stop getting the versions of others once one is found?

beaucollins: Yes, it will do an early abort if it enters the true branch here.

beaucollins: Right, this will only do a single async action at a time so presumably it good be less performant in that it won't make the requests in parallel.

beaucollins: Another rewrite that preserves the potential parallelization (but throwing an Error is kind of gross to me):

```js
try {
  await Promise.all(notes.map( async note => {
    if ( await getVersion( note ) === 0 ) {
       throw new Error( 'not synced' );
    }
  } ) );
} catch ( error ) {
    this.showUnsyncedWarning();
    return;
}
onSignOut();
```

beaucollins:
Do we know why `npm test` isn't working? Would be nice to get CI going.

lib/dialogs/settings.jsx
dmsnell:
```js
const noteVersions = await Promise.all( notes.map( async note => await getVersion( note ) ) )
const needsSync = noteVersions.some( v => v === 0 )

return needsSync
	? this.showUnsyncedWarning()
	: onSignOut()
```

???

beaucollins: The beauty of the `reject`in the original implementation is that it _will_ short circuit any pending async operations.

beaucollins: You can get rid of the `async/await` in the closure since `getVersion` returns a `Promise` and it becomes:
```js
const noteVersions = await Promise.all( notes.map( getVersion ) )
```

dmsnell: so we could still reject here too, no? I was overlooking that the first `reject()` rejects and abort the entire `Promise.all()`

```js
const noteHasSynced = note => /* identical to before */

try {
  await Promise.all( notes.map( noteHasSynced ) )
  onSignOut()
} catch ( e ) {
  this.showUnsynchedWarning()
}
```

beaucollins: If JavaScript catch statements had pattern matching we could check for the expected error type.

The `reject` in `noteHasSynced` should be providing an `Error` instance to `reject`.

roundhill:
Pretty awesome, although I just learned `Promise` and now need to learn `await`? I can't keep up with javascript! :)

.babelrc
```diff
@@ -1,15 +1,14 @@
...
   "presets": [
-    [ "es2015", { "modules": false } ],
+    [ "env", { "targets": { "electron": "1.7" } } ],
     "react"
   ],
   "plugins": [
     "babel-plugin-transform-class-properties",
-    "babel-plugin-transform-object-rest-spread"
+    "babel-plugin-transform-object-rest-spread",
+    "babel-plugin-transform-async-to-generator"
```
dmsnell: is this needed since Chrome 55+ supports `async` functions?

dmsnell:
`npm run build:prod` was producing this error

```
ERROR in app.js from UglifyJs
Unexpected token: punc (() [./flux/app-state.js:33,4][app.js:915,15]
```

It relates to the fact that `uglifyjs` doesn't understand any ES2015+ syntax. We had a couple options here and one was stop minifying the files altogether (a viable option since this is built for Electron) or tell Babel to transform everything so that `uglifyjs` doesn't freak out.

For the time being I felt like adding the `uglify` flag would be the best option because it's more obvious why we're performing work we shouldn't need to and hopefully also easier to remove when `uglify`/`webpack` are able to handle the syntax.

Added the flag in 933b64d

dmsnell:
> Do we know why npm test isn't working? Would be nice to get CI going.

@beaucollins it was working for me.

```
> simplenote@1.1.2 test ./simplenote-electron
> jest

 PASS  lib/components/tag-chip/test.jsx
 PASS  lib/utils/test/is-email-tag.js

Test Suites: 2 passed, 2 total
Tests:       10 passed, 10 total
Snapshots:   1 passed, 1 total
Time:        1.049s
Ran all test suites.
```

beaucollins:
@dmsnell I haven't been able to get `jest` to run. I'm probably doing something wrong.

belcherj:
Closing as this is pretty out of date.  Will revisit async/await as time permits

PR #710, opened 2018-02-16
Simple tray icon implementation

trustlix:
- [x] Simple tray icon implementation following
https://github.com/electron/electron/blob/master/docs/api/tray.md
- [x] Command line argument to hide to tray on startup (`--hide-to-tray`)

Tested on Linux MInt 18.3, with awesome wm 4.2.2

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -7,6 +7,7 @@ const {
   ipcMain,
   shell,
   Menu,
+  Tray
 } = require('electron');
 
 const path = require('path');
@@ -51,6 +52,19 @@ module.exports = function main() {
       titleBarStyle: 'hidden',
       show: false,
     });
+
+    // Tray
+    const tray = new Tray(iconPath)
+
+    tray.on('click', () => {
+      mainWindow.isVisible() ? mainWindow.hide() : mainWindow.show()
+    })
+    mainWindow.on('show', () => {
+      tray.setHighlightMode('always')
+    })
+    mainWindow.on('hide', () => {
+      tray.setHighlightMode('never')
+    })
 
     // and load the index of the app.
     if (typeof mainWindow.loadURL === 'function') {
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -54,18 +54,18 @@ module.exports = function main() {
     });
 
     // Tray
-    const tray = new Tray(iconPath)
+    const tray = new Tray(iconPath);
 
     tray.on('click', () => {
-      mainWindow.isVisible() ? mainWindow.hide() : mainWindow.show()
+      mainWindow.isVisible() ? mainWindow.hide() : mainWindow.show();
     })
     // Visual effects for macos
     // https://github.com/electron/electron/blob/master/docs/api/tray.md#traysethighlightmodemode-macos
     mainWindow.on('show', () => {
-      tray.setHighlightMode('always')
+      tray.setHighlightMode('always');
     })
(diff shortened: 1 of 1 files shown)


dmsnell:
Thanks for the patch @trustlix! I'll try and get on it this weekend :+1:

desktop/app.js
```diff
@@ -52,6 +53,19 @@ module.exports = function main() {
...
     });
 
+    // Tray
+    const tray = new Tray(iconPath)
+
+    tray.on('click', () => {
+      mainWindow.isVisible() ? mainWindow.hide() : mainWindow.show()
+    })
+    mainWindow.on('show', () => {
+      tray.setHighlightMode('always')
```
dmsnell: what are these highlight modes for? can we add a comment to explain since thy are a bit surprising here in the code?

trustlix: They are visual effects available on macos. Doesn't make a diference on Linux.
More details here: https://github.com/electron/electron/blob/master/docs/api/tray.md#traysethighlightmodemode-macos
I can either remove it or put the comment, what you prefer?

desktop/app.js
```diff
@@ -52,6 +53,19 @@ module.exports = function main() {
       show: false,
     });
 
+    // Tray
+    const tray = new Tray(iconPath)
```
trustlix: @dmsnell I created a deb package and found out that this icon path doesn't exist in the final installation dir (`/usr/share/simplenote`). The icon is actually placed at `/usr/share/pixmaps/simplenote.png`. Should I reference this path here instead?

desktop/app.js
dmsnell: Please leave a comment. Oftentimes I raise the question just to hint at the need for a comment in the code so that people will have the context in the future when they read it :wink:

dmsnell:
Wow - this is fun, I tried this out on my Mac and although the tray icon worked it was a bit surprising 😄 maybe an icon issue?

screen shot 2018-02-24 at 12 07 30 pm

That's the Simplenote icon in the tray in case it wasn't obvious 😉

natanielcz:
@trustlix: Maybe we should add an option to disable closing to tray?

trustlix:
@dmsnell yeah, quite surprising. Unfortunately I don't have a mac here to test :-/
@natanielcz makes a lot of sense. Will try to squeeze some time to work on that... things are a bit crazy here lol

dmsnell:
@natanielcz do other apps typically provide a disable on this behavior? maybe I'm just not looking, but I feel like I've never seen such an option among many apps that close to tray.

natanielcz:
@dmsnell Look at skype if you're looking for example.

rodrigoi:
@trustlix thanks for this contribution!

I'm also running into the image size problem with macOS. Apparently, the icon should be 16x16 or 32x32 (@2x). My recommendation would be to have different assets for each target platform.

Also, it won't change status if I close the window, and will throw this exception if I click it to open the window:
screen shot 2018-04-17 at 18 25 27

Also for macOS, it'll be nice to have a monochromatic icon.

roundhill:
Closing due to inactivity.

aderchox:
I searched the Issues and it seems that no one has suggested the tray icon capability for Windows, so I just wanted to express this need here :) (without creating a separate issue). So only a suggestion. Thanks

thryl:
Yeah, having the app running while minimized to Windows tray would be a great feature!

PR #711, opened 2018-02-17
Update small versions and reapply formatting

dmsnell:
I updated the package versions that I though should be safe to update.
We should still verify that there aren't unexpected failures at runtime.

I'm not sure if the formatting slipped through in the past or if the
update to `prettier` and `eslint` triggered these changes but I applied
them with `eslint --fix`


roundhill:
Thanks @dmsnell! :shipit:

PR #712, opened 2018-02-17
Update to React 16

dmsnell:
This major upgrade involves only a few basic changes:

 - No more `React.createClass()`
 - `PropTypes` from separate npm package
 - `onClickOutside` is now a higher-order component instead of a mixin

In this patch I've updated the necessary dependencies and made the
necessary code changes in order to bump our React version.

Performance _should_ be better but probably not noticeably so.
This update is more to keep in line with React updates.

**Testing**

Tests should involve the basic operations throughout the app.
Particular attention should be given to dialogs where a click outside
of them should close the dialog as we have bumped the `react-onclickoutside`
dependency as well.

Please scrutinize the code changes and look for omissions!

These changes were made by hand and not by a codemod.


roundhill:
This is awesome! I did find one `createClass` remaining here: https://github.com/Automattic/simplenote-electron/blob/6b8dc3c2d97c56473454bd67c38691cdeb74e7a0/lib/list.jsx#L3

Also, there's some conflicts now 🤷‍♂️

dmsnell:
@roundhill I don't think that `list.jsx` was in use so I deleted it 🙃

roundhill:
OK awesome! :shipit:

Issue #694, opened 2018-01-16
checklist with markdown not working properly


gocarlos:
#### Steps to reproduce
1. start a new note
2. type 
```
- [ ] asdfasdf
```
3. expected: 

4. result:

#### What I expected
- [x] asdfasfd

#### What happened instead
```
<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> asdfasdf
``` 

#### OS version
Windows 7 64 bit, using version 1.1.2

#### Screenshot / Video
![2018-01-16 21_13_23-simplenote](https://user-images.githubusercontent.com/1331762/35010311-c18cca82-fb02-11e7-861f-e0e390631b0e.png)
![2018-01-16 21_14_04-simplenote](https://user-images.githubusercontent.com/1331762/35010313-c2b6c3f4-fb02-11e7-829a-9230365a6159.png)

gocarlos:
https://app.simplenote.com/ also does not seem to work correctly. 
http://demo.showdownjs.com/ does work correctly

roundhill:
@dmsnell this used to work, but appears to have broken. Mind taking a peek?

roundhill:
Oh, I bet this broke because we don't allow html tags in the preview now, they get escaped.

mirka:
Closing as this was fixed in #724.
I've added two minor issues as enhancements: #874 #875

Issue #681, opened 2017-12-06
Checkboxes support


giannissc:
I was under the impression that this was already submitted in the repository and would be included in the 1.1.0 update yet when i try to show a check box i only get this (image attached)
#### Steps to reproduce
1. Create a check box list
2. Press preview

#### What I expected
When writing check box list when pressing preview to render them properly

#### What happened instead
The check boxes were rendered as the equivalent HTML text for check boxes (see picture) 

#### OS version
Ubuntu 17.10

#### Screenshot / Video
![image](https://user-images.githubusercontent.com/20277283/33660651-7efa3ade-da7c-11e7-9a65-cd73e0abe10d.png)

laskin82:
same bug on windows 7

giosoft:
how can you create checkboxes? I'm typing [_] but it doesn't work

laskin82:
new note
- [] test1
- [] test1
- [] test1
- [] test1

giosoft:
thanks, but it doesn't work. shows me this in preview: `<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1
<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1
<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1
<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1`

giosoft:
`<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1
<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1
<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1
<input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> test1
`

giannissc:
It actually does work but it has a bug. Let me elaborate:
1. You create a checkbox in markdown like this:
- [ ] Blah blah
- [x] Blah blah
2. The markdown texted is parsed into the equivalent html
3. The html text is parsed and rendered into the checkbox

Somehow step 3 does not happen and instead you only see the html text and not the checkbox that you are supposed to

[...]

Issue #314, opened 2016-05-12
Add Task List (checkbox) Markdown support


valdotdev:

Trollwut:
This would make my household (4 persons) gladly switch to simplenote. :>

Any word on considering this request?

rafaelpimpa:
Would love to see this too!

dmsnell:
This is something we definitely plan on implementing at some point but haven't yet and probably won't be able to get in for a couple of releases. I think that somehow it should already be working with the Markdown rendering but that obviously isn't. We would absolutely welcome some help on this! :smile:

johnkjoseph:
In the meantime is it somehow possible according to #309 using GFM? Couldn't get it to work though.

dmsnell:
@johnkjoseph this is on my own personal backlog to try and track down. I'm simply not sure why it's not working is all.

rik-shaw:
Yes, checklists would be great.  Unfortunately, the `[` and `]` are hard to get to on Android if it follows this syntax.  But I suppose a hard to use standard is better than no standard at all.  Maybe I need to find a markdown friendly Android keyboard? :-)

Topy44:
I agree that checklists would be great. Especially if they can be activated/deactivated by clicking on the box.

pedrojuk:
+1 for Markdown checkbox support. Would be very useful

greatkithain:
Yeah, checkbox in Simplenote would be awesome!

giannissc:
I was under the impression that this was already submited in the repository and would be inlcuded in the 1.1.0 update yet when i try to show a checkbox i only get this 
![screenshot from 2017-11-21 12-13-01](https://user-images.githubusercontent.com/20277283/33071918-57f38a94-ceb5-11e7-8d50-c065a86e91a9.png)

MorgothSauron:
+1 for checkbox support in Simplenote :-)

giosoft:
This is a must, pls

toast018:
+1 for checkbox support in Simplenote. Made a Github account just to come here and +1 this.

Laurent001:
+1 for checkbox support

Robertght:
+1 #1236175-zen

sinisavukovic:
+1 for checkbox support

virtadpt:
+1 for checkbox support, also.

BandonRandon:
Twitter request https://twitter.com/TEDxOdessa/status/1029813699949867009

[...]

PR #714, opened 2018-02-22
Fix for checklist rendering problem

Jackymancs4:
Trying to address https://github.com/Automattic/simplenote-electron/issues/694, https://github.com/Automattic/simplenote-electron/issues/314, https://github.com/Automattic/simplenote-electron/issues/681

Hello,
after some investigation I found that the problem was indeed the sanitization of html performed by https://github.com/leizongmin/js-xss through https://github.com/VisionistInc/showdown-xss-filter.

```js-xss``` provides powerful filtering API, but ```showdown-xss-filter``` forces only the default behavior that kills every type of input and form.

Since ```showdown-xss-filter``` has pretty old dependencies, I'm going to maintain (mainly for personal use) an updated [fork](https://github.com/jackymancs4/showdown-xss-config) that enables all of js-xss configurations.

This PR includes a whitelisting configuration that allows only checklist rendering, while performing xss validation on everything else.

## Test:

```
Test1

- [ ] Test
- [ ] Test
- [ ]Test
- [x] Test

[ ] Test
[x] Test

[ ]
[x]

<script>alert('xss!')</script>
```

| Before | After |
| -- | -- |
| schermata 2018-02-22 alle 17 31 11 2 | schermata 2018-02-22 alle 17 24 48 2 |

### DIFF
--- a/lib/note-editor.jsx
+++ b/lib/note-editor.jsx
@@ -2,14 +2,28 @@ import React, { PropTypes } from 'react';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
 import showdown from 'showdown';
-import xssFilter from 'showdown-xss-filter';
+import xssfilter from 'showdown-xss-config';
 import NoteDetail from './note-detail';
 import TagField from './tag-field';
 import NoteToolbar from './note-toolbar';
 import RevisionSelector from './revision-selector';
 import { get, property } from 'lodash';
 
-const markdownConverter = new showdown.Converter({ extensions: [xssFilter] });
+// whitelist <input> (only checkbox type) and <li> tags
+const xssConfig = {
+  onTag(tag, html) {
+    if (tag === 'input') {
+      if (html.includes('type="checkbox"')) return html;
+    }
+    if (tag === 'li') {
+      return html;
+    }
+  },
+};
+
+const markdownConverter = new showdown.Converter({
+  extensions: [xssfilter(xssConfig)],
+});
 markdownConverter.setFlavor('github');
 
 export const NoteEditor = React.createClass({
--- a/lib/note-detail.jsx
+++ b/lib/note-detail.jsx
@@ -2,7 +2,7 @@ import React, { PropTypes } from 'react';
 import { connect } from 'react-redux';
 import highlight from 'highlight.js';
 import showdown from 'showdown';
-import xssFilter from 'showdown-xss-filter';
+import xssfilter from 'showdown-xss-config';
 import { get, debounce, invoke, noop } from 'lodash';
 import analytics from './analytics';
 import { viewExternalUrl } from './utils/url-utils';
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by dmsnell** [comment]
Thanks for the patch @Jackymancs4!

If there another way we can accomplish this with the existing `showdown-xss-filter`? Should we be looking at `js-xss` instead? Could your forked-patch apply to the upstream project?

File: lib/note-editor.jsx [hunk 1]
```diff
@@ -2,14 +2,28 @@ import React, { PropTypes } from 'react';
...
 import NoteToolbar from './note-toolbar';
 import RevisionSelector from './revision-selector';
 import { get, property } from 'lodash';
 
-const markdownConverter = new showdown.Converter({ extensions: [xssFilter] });
+// whitelist <input> (only checkbox type) and <li> tags
+const xssConfig = {
+  onTag(tag, html) {
+    if (tag === 'input') {
+      if (html.includes('type="checkbox"')) return html;
```
dmsnell: we would definitely want to lock this down. right now it appears like if the input is a checkbox then we'll also allow XSS attacks in other unrelated fields of the same HTML

Jackymancs4:
> If there another way we can accomplish this with the existing showdown-xss-filter?

Nope, I already extensively studied the code and it seems impossible to me, because of this line
https://github.com/VisionistInc/showdown-xss-filter/blob/e26046384e266be5646ef0eb58816ff4029da565/showdown-xss-filter.js#L20

> Should we be looking at js-xss instead? 

Well, I think it's a possible way, but I'm not able to see any real gain doing so.
Showdown is working on expanding is extension capability, and I think it's pretty cool.
showdown-xss-filter and showdown-xss-config simply pipe showdown output in a js-xss instance, but more can be done in the future.

> Could your forked-patch apply to the upstream project?

Actually, my fork is not a patch, but a complete rewrite of the code according to the more recent showdown extension interface. And I'm planning to improve testing coverage. Also, the original repo has been unmantained for almost three years. I can try, of course.

lib/note-editor.jsx
Jackymancs4: Can you articulate more? I'm very interested if I missed something.
Without whitelisting this, rendering checkboxes is near impossible.
Also, stricly allowing only checkbox type input seems pretty harmless..

Jackymancs4: I can actually improve sanitization of <input>, stripping it from all the non-type attributes, if that was the problem.

Like, enforcing that only ```<input type="checkbox disabled>``` are returned.

roundhill:
Just tried this out, thanks for the fix @Jackymancs4. We are indeed allowing any input with a `type="checkbox"` now. I'm wondering if we could limit it to only allow if it is within a `<li>` element? Otherwise I can add it in note content and it gets rendered:

screen shot 2018-03-07 at 1 38 42 pm

There's also a small conflict with `package.json` now. If we can fix those two things then I think we can :shipit:.

roundhill:
> Could your forked-patch apply to the upstream project?

We should also see about this, or maybe put the fork in one of our repos.

dmsnell:
Closing this due to #721 and #724

Issue #707, opened 2018-02-05
Toggle for startup update checker in Linux


hfel:
Hi!

It would be nice if there was an option to turn the automatic update check on the app startup off for those who've installed Simplenote from a repo and would prefer to wait until it's updated there.

dmsnell:
@hfel from which repository did you install Simplenote? Usually these kinds of things are performed by the repository maintainers. Maybe we could make it easier for them to perform this behavior.

hfel:
The stable branch (shannon) of Solus.

mirka:
It would be nice to have a toggle to disable update checking, regardless of this specific use case here. A "Check for updates" checkbox is pretty much standard (on Mac apps, at least).

I'm going to close this in favor of the more generic #842

PR #722, opened 2018-03-08
Add option to disable app updater

roundhill:
In order to submit the app to the Windows Store, we need a way to disable the update checker since all updates will just come through the store.

I've added a config option to handle this, it defaults to having the updater enabled.

**To Test**
* Set the version in `package.json` to `1.1.2`.
* Build the linux app, and run it. You should see the update prompt.
* Rebuild the app, now with `updaterEnabled` set to false.
* The update prompt should now not show.

This could help for #707 as well.

### DIFF
--- a/desktop/config-updater.json
+++ b/desktop/config-updater.json
@@ -1,4 +1,5 @@
 {
+  "updaterEnabled": true,
   "forceUpdate": false,
   "updater": {
     "url":
--- a/desktop/updater/index.js
+++ b/desktop/updater/index.js
@@ -37,6 +37,10 @@ function urlBuilder(version) {
 
 module.exports = function() {
   app.on('will-finish-launching', function() {
+    if (!config.updaterEnabled) {
+      return;
+    }
+    
     let url = urlBuilder(app.getVersion());
 
     if (platform.isOSX()) {


roundhill:
I wonder if we actually would want to disable it by default 🤔. That way if the app is built for some other purpose it wouldn't bother with the prompts (like mentioned in #707)

natanielcz:
Also with this approach, we will have to set this `flag` before each build. Maybe we should consider to have separate config for this flag for each build target?

roundhill:
We'd still want to build a Windows app with the updater enabled to be offered as a download for Windows 7/8 on Simplenote.com. But not for the .appx, which I think we need to either add support for in the build process or just run it manually when we want to submit the update to the Windows Store.

natanielcz:
I suppose that we'd like to submit the update to the Windows Store after each version so It's a lot of work and changing values after each released version. We could specify many `env` configs such as:
- `appx.env` 
- `msi.env`
- `deb.env` 
etc. and then just link specific `env` file in the `npm` scripts.

rodrigoi:
@roundhill do you still need this reviewed (since the Windows Store App was released a few weeks ago)?
Instructions work as expected, but 👍 to having `ENV` configs, so let me know if you need help moving that forward.

roundhill:
Sure @rodrigoi if you want to have a go at that I'd appreciate it! I'm thinking of adding an additional step when building the win32 app that will also create an .appx package.

adlk:
@roundhill just for checking the Windows Store there is actually a pretty simple solution by using `process.windowsStore`. 

```js
if (!process.windowsStore) {
  // let's rock the auto updater
}
```

PR #721, opened 2018-03-08
Final fix for checklist rendering problem 

Jackymancs4:
Hello @roundhill, @dmsnell 
Still trying to address https://github.com/Automattic/simplenote-electron/issues/694, https://github.com/Automattic/simplenote-electron/issues/314, https://github.com/Automattic/simplenote-electron/issues/681

After some thought about the concerns emerged in #714 , I think this is the best solution. This PR simply change the filtering pipeline that sanitize the input.
From:
Potentially malicious input -> ```showdown``` --(```showdown-extension```)--> ```js-xss``` -> Renderable HTML (without any input tags)
To:
Potentially malicious input -> ```js-xss``` -> ```showdown``` -> Renderable HTML (with trustworthy input tags generated by showdown)

For testing:

```
Test1

- [ ] Test
- [ ] Test
- [ ]Test
- [x] Test

<input type="checkbox"> test
<input type="checkbox">
<input xtype="checkbox" type='text'> test

[ ] Test
[x] Test

[ ]
[x]

<script>alert('xss!')</script>
```

that render:

| Before | With #714 | Now |
| -- | -- | -- |
| final3 | final1 | final2 |

The last one being the more desirable.

PR #724, opened 2018-03-10
XSS Refactor: Replace `showdown-xss` with in-house sanitizer

dmsnell:
See #681
See #694
See #721

The `showdown-xss` library wasn't really doing for us what we wanted.
It was transforming our checkbox inputs and making them display as plain
text instead of as actual HTML.

In this patch I've removed `showdown-xss` and created a new centralized
function to render HTML which calls our custom sanitization.

Why do it ourselves? We don't need the kind of sanitiation which only
removes the malicious code and leaves as much as it can untouched.
Instead we are free here to strip out basically everything except for
a few white-listed tags and attributes since we ourselves are the ones
producing the output; we don't have to support full HTML in the notes.

This patch should guard against everything on the OWASP list of XSS
attacks. It will remove significant styling and custom HTML but when a
tag is removed it will usually just take out the tag itself and leave
the inner content as plain text. Some tags are "forbidden" and all their
children are removed with them.

cc: @Jackymancs4 

With the contributions and discussion about the checkboxes I really wanted
to get that bug fixed but I was also uneasy about the proposed solutions in
#721 and #714. I was uneasy because our XSS libraries were already opaque
and less-maintained than we would want. Digging into libraries also revealed
custom HTML parsing code that I wasn't sure I trusted.

In this proposal we're relying on the browser to parse the HTML and walk
through it node-by-node giving us the best possible parse with the protection
the browser already has coded into it against attacks that target the parser…

```html
<
script
>…<
/script>
```

…since we're generating markup and we don't promise "everything Markdown
can do is available here" then we can produce whatever subset of Markdown
we want to support. I have chosen to try and mimic the allowed tags and attributes
as is currently supported in the web version of Simplenote. We can add more
tags such as `<abbr>` and `<acronym>` and I think if we wanted to we could
open up some styling or class attributes but for now I'm reaching for harmony
across the platforms more than one being better than others.

**Testing**

Try to break rendering with malicious attacks and also verify that checkboxes
and expected output appears.

### DIFF
--- a/lib/note-editor.jsx
+++ b/lib/note-editor.jsx
@@ -2,16 +2,13 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
-import showdown from 'showdown';
-import xssFilter from 'showdown-xss-filter';
 import NoteDetail from './note-detail';
 import TagField from './tag-field';
 import NoteToolbar from './note-toolbar';
 import RevisionSelector from './revision-selector';
 import { get, property } from 'lodash';
 
-const markdownConverter = new showdown.Converter({ extensions: [xssFilter] });
-markdownConverter.setFlavor('github');
+import { renderNoteToHtml } from './utils/render-note-to-html';
 
 export class NoteEditor extends Component {
   static propTypes = {
@@ -158,7 +155,6 @@ export class NoteEditor extends Component {
   };
 
   render() {
-    let noteContent = '';
     const { editorMode, note, revisions, fontSize, shouldPrint } = this.props;
     const revision = this.state.revision || note;
     const isViewingRevisions = this.state.isViewingRevisions;
@@ -181,12 +177,7 @@ export class NoteEditor extends Component {
       }
     );
 
-    if (shouldPrint) {
-      const content = get(revision, 'data.content', '');
-      noteContent = markdownEnabled
-        ? markdownConverter.makeHtml(content)
-        : content;
-    }
+    const content = get(revision, 'data.content', '');
 
     const printStyle = {
       fontSize: fontSize + 'px',
@@ -233,7 +224,9 @@ export class NoteEditor extends Component {
           <div
             style={printStyle}
             className="note-print note-detail-markdown"
-            dangerouslySetInnerHTML={{ __html: noteContent }}
+            dangerouslySetInnerHTML={{
+              __html: markdownEnabled ? renderNoteToHtml(content) : content,
+            }}
           />
         )}
         {!isTrashed && (
--- a/lib/utils/sanitize-html.js
+++ b/lib/utils/sanitize-html.js
@@ -1,0 +1,246 @@
+import validUrl from 'valid-url';
+import { filter } from 'lodash';
+
+/**
+ * Determine if a given tag is allowed
+ *
+ * @param {Node} node node being examined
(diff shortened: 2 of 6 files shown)


lib/note-editor.jsx
```diff
@@ -251,7 +242,9 @@ export const NoteEditor = React.createClass({
           <div
             style={printStyle}
             className="note-print note-detail-markdown"
-            dangerouslySetInnerHTML={{ __html: noteContent }}
+            dangerouslySetInnerHTML={{
+              __html: markdownEnabled ? renderNoteToHtml(content) : content,
```
roundhill: I noticed that my malicious note would be rendered quickly when printing (if markdown was disabled for the note), maybe we should always pass this through `renderNoteToHtml` for safety's sake?

lib/utils/sanitize-html.js
```diff
@@ -0,0 +1,246 @@
...
+    // a live collection and we will introduce bugs if we remove
+    // as we go on the first pass.
+    //
+    // @see https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes
+    filter(
+      node.attributes,
+      ({ name, value }) =>
+        !isAllowedAttr(tagName, name) ||
+        // only valid http(s) URLs are allowed
+        (('href' === name || 'src' === name) && !validUrl.isWebUri(value))
```
roundhill: How about `mailto` links? The markdown for that is typically:

```
[emailme](mailto://dennis@raddude.com)
```

lib/utils/sanitize-html.js
```diff
@@ -0,0 +1,246 @@
...
+ * Determine if a given tag is allowed
+ *
+ * @param {Node} node node being examined
+ * @returns {Boolean} whether the tag is allowed
+ */
+const isAllowedTag = node => {
+  const tagName = node.nodeName.toLowerCase();
+
+  if ('input' === tagName) {
+    return 'checkbox' === node.getAttribute('type');
```
roundhill: I noticed that checkboxes render with bullets, I think we should remove those so they look more like GitHub:

screen shot 2018-03-15 at 3 19 05 pm

lib/note-editor.jsx
dmsnell: oh you know what…we shouldn't be setting `__html` to `content` if Markdown is disabled. it's not HTML so it shouldn't be treated that way - this is a separate issue so let's do it in another PR (because we'll have to make sure the styling all works out

```js
{ markdownEnabled ? (
	<div
		style={printStyle}
		className="note-print note-detail-markdown"
		dangerouslySetInnerHTML={{ __html: renderNoteToHtml( content ) }}
	/>
) : (
	<div style={printStyle} className="note-print note-detail">{ content }</div>
) }
```

lib/utils/sanitize-html.js
dmsnell: I'd be more keen to fix this in a separate PR since this one right now is solely there for security. The rendering of bullets is going to be more related to how `showdown` generates the Markdown than how we sanitize it.

dmsnell:
Updated @roundhill to allow email links but I left the other comments for separate issues. Thoughts?

lib/utils/sanitize-html.js
roundhill: ok 👍

roundhill:
Nice, I think we're good to go here. :shipit:

PR #743, opened 2018-04-04
Update Electron to 1.8.4

roundhill:
The current version of electron has a vulnerability, although from the report I think Simplenote is safe from it: https://www.electronjs.org/blog/webview-fix. But, better safe than sorry.

**To Test**
* `npm install`
* `electron .`
* Verify electron-y things work correctly (app menus, printing, etc)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "babel-preset-react": "6.24.1",
     "classnames": "2.2.5",
     "css-loader": "0.28.9",
-    "electron": "1.7.11",
+    "electron": "1.8.4",
     "electron-builder": "20.2.0",
     "electron-packager": "9.1.0",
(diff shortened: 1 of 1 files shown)


roundhill:
> Did you create actual release builds after this update from a fresh clone of the repo?

Good call to test that. It is probably unrelated to the electron update but there's an error with the new `isemail` lib we are using:

```
ERROR in app.js from UglifyJs
Unexpected token name «i», expected punc «;» [../node_modules/isemail/lib/index.js:121,0][app.js:21300,13]
```

roundhill:
> It is probably unrelated to the electron update but there's an error with the new isemail lib we are using:

Looks like we need to update webpack to fix that. I'll do that in a follow-up PR.

PR #748, opened 2018-04-14
Add the ability to toggle menubar

### LINKED ISSUE #735: Possibility to hide menu bar on linux app
Please, add possibility to hide top menu bar on linux app.

Version: 1.1.3
Os: Ubuntu 14.04.3

I'm using deb package.

ugly_menubar

Thanx :+1:

mariosant:
Hi, this PR adds the ability to toggle menubar, via the alt key. 

Closes #735.

**How to test**

_The following instructions apply to Windows and Linux users._

Launch the desktop app.
Menubar should be visible.
Hitting Alt key should hide it.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -70,6 +70,8 @@ module.exports = function main() {
     Menu.setApplicationMenu(appMenu);
 
     ipcMain.on('settingsUpdate', function(event, settings) {
+      mainWindow.setMenuBarVisibility(settings.menubarVisible);
+
       Menu.setApplicationMenu(
         Menu.buildFromTemplate(createMenuTemplate(settings))
       );
--- a/desktop/menus/view-menu.js
+++ b/desktop/menus/view-menu.js
@@ -140,6 +140,15 @@ var buildViewMenu = function(settings) {
           }
         },
       },
+      {
+        label: 'T&oggle Menubar Visibility',
+        accelerator: 'Ctrl+Shift+M',
+        click(item, focusedWindow) {
+          focusedWindow.webContents.send('appCommand', {
+            action: 'toggleMenubarVisibility',
+          });
+        },
+      },
     ],
   };
 };
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -79,6 +79,7 @@ function mapDispatchToProps(dispatch, { noteBucket }) {
         'setNoteDisplay',
         'setMarkdown',
         'setAccountName',
+        'toggleMenubarVisibility',
       ]),
       dispatch
     ),
@@ -386,6 +387,8 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
       const isMacApp = isElectronMac();
       const filteredNotes = filterNotes(state);
       const hasNotes = filteredNotes.length > 0;
+
+      console.log('menubar: ', this.state.menubarVisible); // eslint-disable-line
 
       const selectedNote =
         state.note || (!isSmallScreen && hasNotes ? filteredNotes[0] : null);
--- a/lib/state/settings/actions.js
+++ b/lib/state/settings/actions.js
@@ -57,3 +57,7 @@ export const setWPToken = token => ({
   type: 'setWPToken',
   token,
 });
+
+export const toggleMenubarVisibility = () => ({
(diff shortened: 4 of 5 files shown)


rodrigoi:
@mariosant thanks for your contribution!

Would you mind add some testing instructions and the effect this has on the different platforms? I've tested on macOS and I can confirm that there are no regressions. Will test on Windows and Linux as soon as possible.

Also, how is this solution different than using the [Frameless Window](https://electronjs.org/docs/api/frameless-window) approach?

roundhill:
Is there a way to have the menu show up if you hover near it? I tested on windows and there was no menu at all, the only way to show it was by pressing the `alt` key which I think the average user won't be able to discover easily.

mariosant:
Hello @rodrigoi and @roundhill! :wave:

@rodrigoi, to be honest, I was hoping you guys have a solution about testing it. I can add instructions on how to test this feature manually, I would prefer automating the test though. My experience with Electron is kinda limited so, if it's not too much to ask, can you guide me please?

This is different from Frameless Window approach, in my understanding, since the app still utilizes OS' native components, such as window borders and menus. I understand that using the Frameless Window approach, we will have to implement the menu (and the window borders) ourselves.

@roundhill, sure we can do that. I am not very sure how this would work under macOS (since the menubar will be visible at all times in my understanding).
I have an idea, how about displaying it by default and add an option within settings to auto-hide it?

roundhill:
Instead of a setting, can we show the menu at launch and then you can press `alt` to hide it if desired?

mariosant:
That is a great idea @roundhill! I will see how to get it done.

mariosant:
Here you are @roundhill! When `mainWindow` initializes, it makes menubar visible. Works like a charm under Linux, still curious on how to test this.

natanielcz:
I would sugest to save previous app menu state which should be used. For example when I hide app menu I'd like to keep it hidden on the next launch

mariosant:
@natanielcz I understand that. Perhaps this should be added to settings.

mariosant:
So guys what do you think? Should we continue this? If yes, how?

roundhill:
@mariosant I think we're close. Instead of adding a setting, I think @natanielcz was suggesting that we could just store a boolean of the last state of the menu, hidden or not.

It can be added to `settings` state so that it persists between launches:

https://github.com/Automattic/simplenote-electron/tree/master/lib/state/settings

Let me know if you need any help with that.

rodrigoi:
I don't think it'll make much sense to have this setting present on Windows or macOS, unless we unify all the platforms on a Frameless Window. We could use `process.platform` to show this only on Linux.

mariosant:
Sure, we can do that @rodrigoi. I would like to ask you though, why do you believe it doesn't make sense for Windows users? Keep in mind that I haven't touchd Windows for about 8 years. :rofl:

mariosant:
@roundhill, it seems that Electron does not provide an event  to handle menubar visibility changes. Perhaps might be a better option to add this as a menu entry? Visual Studio Code has a view menu with the option to "Toggle Menu Bar", for example.

roundhill:
@mariosant any chance you can finish this PR up? It's gone stale.

mariosant:
Apologies @roundhill, things got a bit busy. I will try to give it a shot this weekend.

mariosant:
@roundhill it took a bit more, but here it is!
This PR now uses `Ctrl-Shift-M` to toggle menubar visibility, persists preference via application settings state and adds a menu -> view -> Toggle menubar visibility entry.

Again, my apologies for taking so long, things are a bit busier than it should :)

roundhill:
I just noticed that this was attempted before: https://github.com/Automattic/simplenote-electron/pull/559. It was removed because setting that flag hid the menu by default.

Should we just remove the menu item and map the toggle to the `alt` key to fit in better across the platforms?

mariosant:
I haven't really found a way to listen to menubar toggles, so I am not sure how to persist the state. Any ideas?

roundhill:
We could have the `MenuItem` there still, but tied to the `Alt` key with a `visible` property set to `false`?

https://github.com/electron/electron/blob/master/docs/api/menu-item.md

roundhill:
Closing due to inactivity.

Issue #430, opened 2016-11-07
Published note does not sync & I can't generate public URL for note


marianoju:
#### What I did 
1. Checking on a previously published note I realized it is not synchronized with current note edited with Simplenote app 
2. Tried to generate a public URL for the current note and failed (“Publishing note“ does not produce an URL, and mouse-click on “Copy” in sidebar does not copy a public link to clipboard, see screenshots). 

#### What I expected
1. I expected published notes to reflect edits after the publication. 
2. I expected to be able to generate public URL within the Simplenote client. 

#### What happened instead
None of my expectations was met, see above. 

#### My Configuration
- Ubuntu 16.04.1 LTS (Xenial Xerus) 64-bit
- Kernel Linux 4.4.0-45-generic
- GNOME 3.18.3
- Memory: 7.5 GiB
- Processor: Intel® Core™ i5-3320M CPU @ 2.60GHz × 4
- Simplenote v1.0.6

#### Screenshots

![screenshot from 2016-11-07 14-13-29](https://cloud.githubusercontent.com/assets/6254428/20059337/11a74632-a4f6-11e6-8692-0ae5d4a6bb25.png)

![screenshot from 2016-11-07 14-13-49](https://cloud.githubusercontent.com/assets/6254428/20059339/15de2c16-a4f6-11e6-90c2-96cac0693df3.png)

roundhill:
Sometimes it can take a while for the note to publish. If you return to this note now is it published?

dmsnell:
@roundhill I just [noticed](https://github.com/Automattic/simplenote-electron/pull/228#issuecomment-258653775) this issue when addressing #228. I'm not sure when it creeped in or why.

roundhill:
> @roundhill I just noticed this issue when addressing #228. I'm not sure when it creeped in or why.

oh, I see. Thanks for the additional info!

marianoju:
@roundhill Checking 19 hours later on my note it is *still* public, but
the content is garbled (mixing old and new edits, including text that
has been deleted).

Republishing of the note i.e. creating a (new?) public URL is still not
possible.

I have toggled “Make public link” but note remains public when it should
not be, and resetting note to public does not help republishing.

I have created a new note with the content of the latest version, and it
publishes within seconds. I have made edits and they are visible on
reloading the public URL.

chaitanyamsv:
Reported by the user in #947090-zen.

marianoju:
[...]

PR #749, opened 2018-04-20
Fixing publish status not updating 

roundhill:
Fixing #430 seemed trivial until I discovered that we weren't really setting the `note` state properly when the app would run for the first time. I fixed this by moving some logic that used to live in `render()` of `app.jsx` to the app state.

The main fix for the bug reported in #430 was to properly access the note object from the state in `share.jsx`.

**To Test**
* Launch the app, the first note should be selected.
* Click on the share button, the dialog should show properly.
* Select a note further down the list, and delete it. The previous note in the list should now be selected.
* Publish a note. When it is published, the UI should properly update with the short Url.


rodrigoi:
I've run a quick test and looks like it fixes #430. But the change from `params: { note },` to `appState` on `share.jsx` props leaves a couple call to `showDialog` with the wrong parameters:

https://github.com/Automattic/simplenote-electron/blob/7c4a1abb5eef13062dee97aab934857f5f5cc9d5/lib/app.jsx#L342

https://github.com/Automattic/simplenote-electron/blob/7c4a1abb5eef13062dee97aab934857f5f5cc9d5/lib/email-tooltip.jsx#L22

we no longer need to pass `params: { note },`. Otherwise, it appears to work fine 😉

roundhill:
> we no longer need to pass params: { note },

Good call @rodrigoi! Removed those params in 91d6afb

roundhill:
Thanks @rodrigoi!

Issue #312, opened 2016-05-09
Spellcheck isn't active


rachelsquirrel:
Spellcheck _should_ be on in the app but it apparently isn't. This has been reported in [other](https://github.com/Automattic/simplenote-electron/issues/517) issues (#3153710-t [@Hywan]) here and with [draft-js](https://github.com/facebook/draft-js/issues/744) as well.

Spellcheck is activated in the code but not the app.

Some testing revealed that this works when running the app locally in a browser but the Electron app appears to not support it.

This issue is here to figure out _why_ spellcheck isn't on when it should be.

drw158:
Note: We should allow the user to turn off/on spellcheck like in any textarea in the browser. If that's not possible, then this will require developing a feature and adding a settings to the app.

dmsnell:
@drw158 and @rachelsquirrel this came for free when we merged in the draftjs editor (thanks @nfcampos) in #370. While it's now on-by-default, we could add an option to disable it.

@drw158 do you think an item in the menu would be best for this? add an option in both the menu and in the app settings pane? under the _Display_ settings?

drw158:
Somehow I missed this.

I think the spellcheck setting should be in the Edit menu. Nothing in the app settings pane.

@dmsnell Is spellcheck now enabled via draft.js?

We have a recent issue for spellcheck #517

drw158:
Library for adding spellcheck in the duplicate issue:

https://github.com/electron-userland/electron-spellchecker

martynchamberlin:
I'm noticing that in SimpleNote 1.0.8, we've lost spellcheck. Would love to have this back.

Hywan:
@martynchamberlin :+1:

One of our user is asking to get spellcheck back on Windows, #3153710-t.

martynchamberlin:
@Hywan Yeah. I am sending out emails with typos in them because the hassle of manually running a spellcheck is 2007. Nobody wants to have to do that. It should be automatic.

belenos:
Any news on that? I'm using the Linux version and Simplenote still doesn't have a working spell check. That's a must-have feature for any decent text editor nowadays.

martynchamberlin:
I personally switched to [Bear](bear-writer.com). An editor that doesn't have spellcheck is ridiculous.

belenos:
Bear is nice on iOS, I have tested it. I'm using Ghostwriter on Ubuntu, but it lacks a mobile app or sync between devices.

[...]

PR #753, opened 2018-04-24
Add Spellchk Supprt

roundhill:
It's a real shame that the app doesn't support spellcheck (#312). I've got a good start to getting it working by utilizing Electron's `spellCheckProvider`. It now shows spelling errors in the editor:

screen shot 2018-04-24 at 1 35 04 pm

~But, you can't take any action on it. We need to get the spelling suggestions and pass them into our `context-menu` component and make them actionable. @dmsnell if you're ever twiddling your thumbs maybe you could take this on? Next 🍺 on me.~
Did this ^


dmsnell:
Tough one @roundhill - this will come with much acclaim but not having suggestions and corrections is a big deal I think. I'm not sure how this will tie in to the editor work we've looked at (#692)

Maybe over the weekend I can try and put some time on this.

roundhill:
@dmsnell I got it working 🎉

screen shot 2018-05-15 at 2 12 21 pm

Turns out a `context-menu` in electron provides a misspelled work if there is one beneath where it was called. 

I am having a problem with the `node-spellchecker` lib though, it is a native module and I'm not able to figure out how to package the app while building the .node file for the windows or linux platforms from macOS. We might have to build and package from windows and linux itself?

roundhill:
@dmsnell Ready for a review here as well. I was 🤔, should we also add a menu option to disable spellcheck? I'd imagine some users would not want it.

roundhill:
@adlk mentioned that this library may be better to use: https://github.com/electron-userland/electron-spellchecker

We could also just use that lib for copy/paste/undo commands in the right-click menu, but we'd want to ensure that you can still copy the markdown preview content via right-click as well.

roundhill:
Closing in favor of #821

PR #764, opened 2018-05-25
Add WP Sign In

roundhill:
This PR adds support for signing to Simplenote via your WordPress.com account. It supports creating a new Simplenote account if one didn't exist yet, or alternatively it will sign the user to an existing account.

wp-sign-in

**To Test**
* From the login screen tap 'Sign in with WordPress.com'
* Authorize a WordPress.com account
* 3 things should happen:
 ** If a Simplenote account already exists for the account, it should be signed in successfully.
 ** If a Simplenote account didn't exist for the account, a new one should be created and you'll see the welcome note.
 ** If the WordPress.com account hasn't been verified by email, you should get an error message.

We're storing the WP.com token in the state for future integrations. If a user signs out, it will be deleted.

_Note: The reviewer will need some new keys to add to their config file. Ping me on slack to get those._


lib/boot.js
```diff
@@ -180,6 +180,16 @@ let props = {
     }
     analytics.tracks.recordEvent('user_signed_out');
   },
+  authorizeUserWithToken: (accountName, userToken) => {
+    localStorage.access_token = userToken;
```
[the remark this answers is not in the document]
roundhill: I think the dot notation access works in addition to `setItem` for `localStorage`. This `localStorage` object is actually how `node-imperium` keeps the user token.

I copy pasta'd from this line: https://github.com/Automattic/simplenote-electron/pull/764/files/f5283df47cd25ef89b9e35816787df0b8372b38f#diff-f041a427c603cb2a78907c18cf90d341L156. We can use setItem for both lines if you'd prefer?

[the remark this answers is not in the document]
roundhill: Ah, for sure let's do the recommended way. Updated in 
bde3059

roundhill:
@dmsnell any other thoughts on this one? 😬

lib/auth.jsx
```diff
@@ -159,6 +178,118 @@ export class Auth extends Component {
...
+      if (errorCode === '1') {
+        this.authError(
+          'Please activate your WordPress.com account via email and try again.'
+        );
+      } else {
+        this.authError('An error was encountered while signing in.');
+      }
+
+      return;
+    }
```
dmsnell: this is more or less about style and so not a blocker but I think in cases where we are checking a list of possible values it can be helpful to use a `switch` consistently to follow the logic…

```js
switch ( params.get( 'code' ) ) {
  case '1':
    return this.authError( 'Please activate…' );

  default:
    return this.authError( 'An error was encountered…' );
}
```

anyway I don't feel too strongly about this but I see the pattern more often in several different ways and the switch I think is a nice way to keep the intention clear while allowing for more linear expansion (such as when we recognize a second or third distinct error code).

lib/auth.jsx
```diff
@@ -159,6 +178,118 @@ export class Auth extends Component {
...
+        return;
+      }
+
+      this.processUrl(parsedUrl);
+    } catch (error) {
+      return;
+    }
+  };
+
+  processUrl = url => {
```
dmsnell: `processUrl` is what it _does_ but not really why it exists. could we think of a more intentional name?

`authenticateFromUrl(url)` for example?

dmsnell: taken further we can simplify this `onBrowserNavigate` function if we say "`authenticateFromUrl` exists to handle URL-based authentication reponders…" and then move the try/catch logic in there…

```js
onBrowserNavigate = url => {
  try {
    this.authenticateFromUrl( new Url( url ) );
  } catch ( e ) {
    // failure to authenticate likely means
    // it wasn't an authentication URL
    // so do nothing
  }
}

authenticateFromUrl = url => {
    if ( parsedUrl.protocol !== 'simplenote:' ) {
      return;
    }

    …
}
```

roundhill:
> Why are we storing the WP token? We don't seem to be using here in this PR at least…can we take that out? is it used for any purpose somewhere else?

The plan is to use it for integrations in the near future. This is the first step to get us there.

lib/auth.jsx
roundhill: > this is more or less about style and so not a blocker but I think in cases where we are checking a list of possible values it can be helpful to use a switch consistently to follow the logic…

I'm cool with that, and would look nicer if we want to check additional error codes in the future.

roundhill:
Thanks for the suggestions @dmsnell! I committed some updates that I hope you'll find satisfactory.

roundhill:
Thanks @dmsnell!

> Any final concerns you want me to review or look for?

Feeling good about this one 👍

PR #770, opened 2018-06-19
Fix Remote Code Execution in WordPress Auth Window

roundhill:
I friendly hacker reported that the WordPress sign-in auth window allows remote code execution via drag and drop operations. We fixed a similar issue in the main app window a while back, but I didn't think about applying it to the new window used for the WP auth.

As I was looking into fixing this, I noticed that if a link with a "_blank" target was clicked it would open _another_ in app `BrowserWindow` and it would also be vulnerable to the same RCE. To fix this I've moved the code that creates the window and handles the auth operation to the main electron process in `app.js`  in order to utilize `preventDefault()` when the browser tries to open a new window. Details: https://github.com/electron/electron/issues/4473#issuecomment-184571997

**To Test**
* Follow the same instructions as #764, you should be able to sign in with WordPress.com and/or see appropriate error messages.
* Drag anything into the WP auth window, nothing should happen.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -1,5 +1,9 @@
 'use strict';
 
+let authState,
+  authWindow,
+  mainWindow = null;
+
 const {
   app,
   BrowserWindow,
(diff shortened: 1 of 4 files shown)


desktop/app.js
```diff
@@ -1,28 +1,32 @@
 'use strict';
 
+let authState,
+  authWindow,
+  mainWindow = null;
```
dmsnell: not important, just a note of style, we have put `let`s one per line in Calypso. `var`s are still combined to match the semantic of hosting, but not `let`

roundhill:
Thanks for looking @dmsnell. This was still buggy after some testing, I have found a _much better_ solution which is just a few lines of code. Going to abandon this one and create a new one.

PR #771, opened 2018-06-19
Fix javascript execution in print view on Linux

roundhill:
A HackerOne report stated that you could execute code in the print view, but only on linux and if you used a `javascript:` uri in a link. I've fixed this by adding an early check in our HTML sanitizer for the javascript uri. And for additional safety I've disabled clicks on the print div.

**To Test**
* You must test this on Linux, as macOS and Windows don't allow you to click outside of a print dialog.
* Add something like `<a href=" javasCript:alert('awesome');">CLICK ME</a>` to a note, and enable markdown and turn the preview on.
* Print the note, and attempt to click outside of the print dialog on the link. Nothing should happen.

### DIFF
--- a/lib/note-editor.jsx
+++ b/lib/note-editor.jsx
@@ -232,6 +232,7 @@ export class NoteEditor extends Component {
               dangerouslySetInnerHTML={{
                 __html: renderNoteToHtml(content),
               }}
+              onClick={event => event.preventDefault()}
             />
           )}
         {shouldPrint &&
--- a/lib/note-editor.jsx
+++ b/lib/note-editor.jsx
@@ -232,7 +232,6 @@ export class NoteEditor extends Component {
               dangerouslySetInnerHTML={{
                 __html: renderNoteToHtml(content),
               }}
-              onClick={event => event.preventDefault()}
             />
           )}
         {shouldPrint &&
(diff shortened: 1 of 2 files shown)


lib/note-editor.jsx
```diff
@@ -232,6 +232,7 @@ export class NoteEditor extends Component {
               dangerouslySetInnerHTML={{
                 __html: renderNoteToHtml(content),
               }}
+              onClick={event => event.preventDefault()}
```
dmsnell: what's the purpose here? will this not prevent clicking on links?

lib/utils/sanitize-html.js
```diff
@@ -195,6 +195,16 @@ export const sanitizeHtml = content => {
...
     filter(node.attributes, ({ name, value }) => {
+      // No javascript Uris allowed
+      if (
+        value
+          .toLowerCase()
+          .trim()
+          .startsWith('javascript')
+      ) {
+        return true;
+      }
```
dmsnell: might be best to move this into `isAllowedAttr()` and pass the value in there

```js
function isAllowedAttr( name, value, tagName, tagValue ) { … }
```

this way we can be explicit in our rules. I wouldn't want to eliminate a valid attribute just because it starts with Javascript

lib/note-editor.jsx
roundhill: Yeah I guess I double-fixed the issue. I can probably just remove this, but I was trying to disable clicks on the print div.

lib/utils/sanitize-html.js
roundhill: That sounds better, I also wasn't a fan of returning `true` early there.

roundhill: All set via cdd0d06

PR #773, opened 2018-06-19
Clear all cookies upon app logout

roundhill:
The WordPress cookie was being stored by the app, so you'd always be signed in to WordPress.com even if you signed out of Simplenote. This PR ensures the WordPress.com account is signed out when the user signs out of the app by clearing the cookies.

**To Test**
* Sign in to the app with a WordPress.com account.
* Sign out of the app.
* Sign in with WordPress.com again, you should not be signed in to WordPress.com.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -7,6 +7,7 @@ const {
   ipcMain,
   shell,
   Menu,
+  session,
 } = require('electron');
 
 const path = require('path');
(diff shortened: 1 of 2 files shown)


desktop/app.js
```diff
@@ -74,6 +75,24 @@ module.exports = function main() {
...
+          cookieUrl += cookie.path;
+
+          session.defaultSession.cookies.remove(
+            cookieUrl,
+            cookie.name,
+            () => {} // Ignore callback
+          );
+        });
+      });
+    });
```
dmsnell: I think this one is going to need some explanation here - what we are doing, why we are building this URL, why we are looping through `cookies`, etc… can you provide some explanation or a link to an explanatory post?

roundhill: Added some comments in 92b1fde. Thanks @dmsnell!

PR #784, opened 2018-07-12
About View: Add new row about Simperium

roundhill:
We're adding a blurb to all of the apps in the about screen for Simperium to help promote the service.

screen shot 2018-07-12 at 10 04 16 am

I'm wondering if this is too many rows now 🤔. If the reviewer agrees we can maybe replace the `Apps` row with this new Simperium one.


roundhill:
Thanks @dmsnell, went ahead and removed the Apps row since we were in agreement on that.

Issue #782, opened 2018-07-12
Change shortcut key for close/exit/quit in Linux app


mdrovdahl:
#### Steps to reproduce
1. Open app
2. Press `Ctrl+Q` 
3. Nothing happens
4. Mouse up to the File menu and see that `Exit` is `Alt+F4`

#### What I expected
The app to close/exit/quit on `Ctrl+Q`  which is fairly standard, at least w/Gnome apps per https://developer.gnome.org/hig/stable/keyboard-input.html.en

#### What happened instead
Deep sorrow and grief

#### OS version
markd@xps:~$ cat /proc/version 
Linux version 4.9.0-6-amd64 (debian-kernel@lists.debian.org) (gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1) ) #1 SMP Debian 4.9.88-1+deb9u1 (2018-05-07)

PR #785, opened 2018-07-12
split non-osx menu item adjustments into distinct linux & win32 block

mdrovdahl:
Addresses #782 by splitting non-osx menu item adjustments into specific "linux" and "win32" blocks. 

"linux" block then modifies "File > Exit" keyboard shortcut (from Alt+F4 to Ctrl+Q) and adds "Help > Help & Support" keyboard shortcut (F1) following the HIG at https://developer.gnome.org/hig/stable/keyboard-input.html.en

Tested on Debian 9, should still be tested on OSX and Windows...

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -259,8 +259,39 @@ function createMenuTemplate(settings) {
     ],
   };
 
-  // non-osx menu item adjustments
-  if (process.platform !== 'darwin') {
+  // linux menu item adjustments
+  if (process.platform === 'linux') {
+    // add about menu item to Help
+    helpMenu['submenu'].push({
+      type: 'separator',
+    });
+    helpMenu['submenu'].push(aboutMenuItem);
+    helpMenu['submenu'].push({
+      label: 'Help',
+      accelerator: 'F1',
+      click: function() {
+        shell.openExternal('http://simplenote.com/help');
+      },
+    });
+    // add exit and settings items to File
+    fileMenu['submenu'].push({
+      type: 'separator',
+    });
+    fileMenu['submenu'].push(settingsMenuItem);
+    fileMenu['submenu'].push({
+      type: 'separator',
+    });
+    fileMenu['submenu'].push({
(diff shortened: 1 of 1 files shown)


mdrovdahl:
Just realized that by changing the logic from non-osx to linux and windows, I've inadvertently dropped other non-linux and non-windows platforms, such as 'freebsd' and 'sunos'.

Update forthcoming...

mdrovdahl:
Ok, now we have a menu blocks for 1) linux and 2) not-linux AND not-osx

There may be a better way to do that comparison operator...I am not a JS developer =)

desktop/app.js
```diff
@@ -259,8 +259,39 @@ function createMenuTemplate(settings) {
...
+      label: 'E&xit',
+      accelerator: 'Ctrl+Q',
+      click: function() {
+        app.quit();
+      },
+    });
+  }
+
+  // non-osx and non-linux (eg:windows, freebsd, sunos) menu item adjustments
+  if (process.platform !='darwin' && process.platform !='linux') {
```
roundhill: The linter picked up a few things: you should use a strict not equal here, and add a space to match the style guidelines:

```
process.platform !== 'darwin' && process.platform !== 'linux'
```

desktop/app.js
```diff
@@ -259,8 +259,39 @@ function createMenuTemplate(settings) {
...
+    // add about menu item to Help
+    helpMenu['submenu'].push({
+      type: 'separator',
+    });
+    helpMenu['submenu'].push(aboutMenuItem);
+    helpMenu['submenu'].push({
+      label: 'Help',
+      accelerator: 'F1',
+      click: function() {
+        shell.openExternal('http://simplenote.com/help');
```
roundhill: Would you mind updating these Urls to use `https`?

mdrovdahl:
Thanks @roundhill. I pushed the fixes and removed "Needs Author Reply" label. Ready for review again.

PR #801, opened 2018-08-28
Update npm version

mirka:
`package-lock.json` was out of sync with `package.json`, probably because people were using different versions of npm (pre- and post-5.0).

Bumped to version 6 so we can also take advantage of `npm audit`.


roundhill:
Tested and working fine. (ignoring the CI stuff for now) :shipit:

PR #806, opened 2018-08-29
Purge unused react packages

mirka:
Removes two packages that were unused and deprecated:

## [react-addons-test-utils](https://www.npmjs.com/package/react-addons-test-utils)

This was probably [needed in the past](https://github.com/Automattic/simplenote-electron/pull/473#pullrequestreview-17549884) as a peer dependency of something, but not with the package versions we have now. Removing this will also remove a peer dependency warning (`react-addons-test-utils@15.6.2 requires a peer of react-dom@^15.4.2 but none is installed`).

## [react-addons-shallow-compare](https://www.npmjs.com/package/react-addons-shallow-compare)

This was [needed in the past](https://github.com/Automattic/simplenote-electron/pull/476#pullrequestreview-14551333) as a peer dependency of `react-virtualized`, but has been [unnecessary since v9.0](https://github.com/bvaughn/react-virtualized/pull/577).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16381,12 +16381,6 @@
         "object-assign": "^4.1.0"
       }
     },
-    "react-addons-test-utils": {
-      "version": "15.6.2",
-      "resolved": "https://registry.npmjs.org/react-addons-test-utils/-/react-addons-test-utils-15.6.2.tgz",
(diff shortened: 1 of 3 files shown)


roundhill:
:shipit:

PR #813, opened 2018-08-30
[macOS] Put “Bring All to Front” in proper menu

mirka:
There was an off-by-one error causing the “Bring All to Front” menu item to be in the wrong menu.

This fixes the misplacement, and removes the usage of array indices to avoid similar bugs.

### Before:
screen shot 2018-08-30 at 22 22 14 screen shot 2018-08-30 at 22 22 18

### After:

screen shot 2018-08-30 at 22 21 32 screen shot 2018-08-30 at 22 21 40


roundhill:
:shipit: Nice fix.

PR #814, opened 2018-08-30
Add Line Length Setting (Part 1)

mirka:
Closes #797 

(Part 1 of a stacked pull request. Part 2: #815)

These changes require Electron v2+ (dependent on #812).

This first part focuses on refactoring and reordering the View menu to prepare for the addition of the Line Length feature.

### ToDo

- [x] Reorder app menu to match simplenote-macos
- [x] Refactor view-menu and add tests
- [x] Prepare CSS styles
- [x] Implement line length action
- [x] Add Line Length to Settings Dialog

PR #815, opened 2018-08-31
Add Line Length Setting (Part 2)

mirka:
(Part 2 of a stacked pull request. Part 1: #814)

This implements the Line Length feature, adding controls to both the app menu and the settings dialog. The default setting is "Narrow". When set to "Full", the `max-width` rule on the note body is removed so that the text can wrap at window width.

line-length

## Settings dialog
screen shot 2018-08-31 at 20 21 33

### DIFF
--- a/lib/state/settings/reducer.js
+++ b/lib/state/settings/reducer.js
@@ -41,6 +41,14 @@ const noteDisplay = (state = 'comfy', action) => {
   return action.noteDisplay;
 };
 
+const lineLength = (state = 'narrow', action) => {
+  if ('setLineLength' !== action.type) {
+    return state;
+  }
+
+  return action.lineLength;
+};
+
 const accountName = (state = null, action) => {
   if ('setAccountName' !== action.type) {
     return state;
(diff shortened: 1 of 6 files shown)


lib/state/settings/reducer.js
```diff
@@ -41,6 +41,14 @@ const noteDisplay = (state = 'comfy', action) => {
   return action.noteDisplay;
 };
 
+const lineLength = (state = 'narrow', action) => {
```
roundhill: Don't think it really matters, but you could have made this a boolean settings since it's sort of a toggle vs. storing a string. But this will future-proof it in case we add another option in the future!

mirka: Yes, I was wondering about that. I kind of assumed that was where we were going when it changed from a single “Full Width” toggle in your first iteration to a multiple choice. I do anticipate some kind of View Options dialog in the future 🙂

screen shot 2018-09-04 at 21 30 43

mirka:
Re: [horizontal padding](#pullrequestreview-151875126), there was a responsive rule that increased the horizontal padding above a certain screen width. I've removed it altogether, since it doesn't really make sense anymore with this new feature.

Now it is a consistent `padding: 24px` for all line length modes and screen widths. Thanks @roundhill !

PR #821, opened 2018-09-03
Add spellchecker (electron-spellchecker) 

adlk:
This is an alternative approach to #753 that is using electron-spellchecker.

I'm extending some electron-spellchecker classes to provide more flexibility and to mimic the setup of #753 

This PR introduces the new make targets:
- `make dev`: starting the webpack dev server
- `make start`: start electron and check if a native dependency rebuild is necessary
- `make rebuild-deps`: rebuilds native dependencies. This is a dependency of `make start`


lib/app.jsx
```diff
@@ -6,7 +6,7 @@ import appState from './flux/app-state';
 import reduxActions from './state/actions';
 import selectors from './state/selectors';
 import browserShell from './browser-shell';
-import { ContextMenu, MenuItem, Separator } from './context-menu';
+// import { ContextMenu, MenuItem, Separator } from './context-menu';
```
roundhill: Let's delete this!

lib/app.jsx
```diff
@@ -37,6 +37,7 @@ import * as settingsActions from './state/settings/actions';
 
 import filterNotes from './utils/filter-notes';
 import SearchBar from './search-bar';
+// import '../spellchecker';
```
roundhill: And delete this! 👍

desktop/contextMenu/index.js
```diff
@@ -0,0 +1,21 @@
...
+// Avoid an error on Windows and Linx when body is not yet available
+window.onload = () => {
+  window.spellCheckHandler.attachToInput();
+};
+window.spellCheckHandler.switchLanguage(navigator.language);
+
+let contextMenuBuilder = new ContextMenuBuilder(window.spellCheckHandler);
+// eslint-disable-next-line no-unused-vars
+let contextMenuListener = new ContextMenuListener(info => {
+  if (info.inputFieldType === 'none' && !info.isEditable) return;
```
roundhill: I think we may want to consider only showing the context menu for the editor div. I'm seeing spell check on the login, tag, search, etc inputs.

mirka:
Can we keep all the file/folder names spinal-cased to match the rest of the codebase?

mirka:
Also, I initially had some trouble getting this to work in my environment. Turns out it was a NODE_MODULE_VERSION issue like the one [described here](https://github.com/onmyway133/blog/issues/69), but it was hard to notice there was even such an error because it was only thrown in the devtools console.

I'm afraid `electron-spellchecker` will break rather silently like this whenever it's run in a node version it doesn't expect. (Is that a valid concern?) Should we be adding an `electron-rebuild` step somewhere?

desktop/contextMenu/index.js
adlk: This can be easily controlled with `spellchecker="false"` on the DOM element. I'm going to add it to the obvious fields as a demo.

adlk:
@mirka @roundhill the required changes were made. Plus,

1. I've added the make targets:
   - `make dev`: starting the webpack dev server
   - `make start`: start electron and check if a native dependency rebuild is necessary
   - `make rebuild-deps`: rebuilds native dependencies. This is a dependency of `make start`

1. Spellchecker is now just used in fields where we really want spell checking to happen.

I'm going to test everything once more and then we should be done here.

Makefile
```diff
@@ -21,62 +31,84 @@ SIMPLENOTE_JS := $(THIS_DIR)/dist/app.js
...
 
+SKIP_BUILD = false
+
+.PHONY: start
+start: rebuild-deps
+	@npx electron .
+
+.PHONY: rebuild-deps
+rebuild-deps:
+	@npx electron-rebuild
```
mirka: Worked great 👍

PR #828, opened 2018-09-05
Update and simplify Babel config

mirka:
This will give us two improvements:

1. We will target Electron in `browserslist` so we won't have to think about converting to Chrome versions anymore.
2. Babel transpilation settings will be just right for `browserslist` targets, so it will result in a smaller build. `npm run build:prod` will give us a 340 KB app.js, down from 368 KB.


package.json
```diff
@@ -45,17 +45,16 @@
     "trailingComma": "es5"
   },
   "devDependencies": {
+    "@babel/core": "^7.0.0",
```
roundhill: Here's some wildcarded dependencies again. Maybe ok for a dev dependency?

mirka: TBH I'm kind of torn on this though. With lockfiles in place, there's no real reason to worry about these unless we decide to adopt an automated dependency updating system like Renovate. The package-lock will ensure that all versions stay the same until explicitly changed... It seems like an unnecessary fight against the default `npm install -D` behavior at this point.

I guess some people like to refer to the package.json for exact versions rather than doing `npm list <package>` ?

PR #832, opened 2018-09-05
Organize scss files

mirka:
This moves the scss files to live with their components.

This is a first step towards containing the CSS rules to each component. The end goal is to have each component be responsible for their own styles, and for them to not have side effects on other components. There is still a lot of renaming and refactoring to be done, but we will do that gradually (or as the need arises).


roundhill:
Tested this, didn't see any issues with the CSS. :shipit:

PR #841, opened 2018-09-07
Makefile improvements

adlk:
With this PR I tried to make the various tasks that were split between npm scripts and make targets more consistent. I'm also moving the config closer to the config of [Automattic/wp-desktop](https://github.com/Automattic/wp-desktop).

The npm scripts are now pointing to the make equivalent. 

### Available targets in `npm` and `make`
- **`make start`**: starts electron
- **`make dev`**: starts development server
- **`make build`**: creates bundle for web-app
- **`make test`**: runs jest
- **`make lint`**: runs linter
- **`make format`**: runs prettier

### Make only targets
- **`make osx`** builds the osx binary
- **`make win32`** builds the windows binary
- **`make linux`** builds the linux binary

- **`make package-osx`** builds the defined osx targets 
- **`make package-win32`** builds the defined windows targets 
- **`make package-linux`** builds the defined linux targets


mirka:
Thank you! I was able to confirm the spellchecker working in Mac dev, Mac packaged app, and Win dev. The npm/make scripts are working on my Mac.

A few questions:

- The spellcheck context menu doesn't seem to be working in the packaged apps for Windows and Linux. Is this expected?
- Is the makefile supposed to work on Windows, too? I was able to install `make` through something called MinGW, and `make start` worked (despite giving a "Path must be a string. Received undefined" error). `make dev` didn't work — it can't find `webpack` or `webpack-dev-server`. Do you recommend that we do all dev work on Mac, and just copy over a packaged app for testing?
- Can we run `npm audit fix` on this? It's reporting a lot of vulns.
- `make package-win32` (on Mac) is ending in an error:
    ```
      • building block map blockMapFile=release/Simplenote-win-1.1.7.exe.blockmap
    Error: Exit code: ENOENT. spawn prlctl ENOENT
    (...long stack trace continues)
    ```
    It results in a working .exe file though... 🤔

roundhill:
> make package-win32 (on Mac) is ending in an error:

I see this as well, but it looks like it is when it is attempting to make the .appx, which only works from Windows. Maybe we could skip the .appx build if not running the script from Windows @adlk?

adlk:
> The spellcheck context menu doesn't seem to be working in the packaged apps for Windows and Linux. Is this expected?

Yes, this is expected. Spellchecker is a native dependency that must be compiled on the target OS. The context menu module and spellchecker are closely tied together. Therefore, if spellchecker isn't working (because of a wrongly built binary) the context menu isn't working either. 

There is still the option to switch to `electron-spellchecker-prebuilt` to skip the platform specific build in favor of using prebuilt binaries. (We are currently doing this with wp-desktop). But before moving more projects to this dependency, I'd suggest hosting our own fork of `cld`, `keyboard-layout` and `@paulcbetts/spellchecker` and building the binaries there.

With wp-desktop I prefer  like to move 

> make package-win32 (on Mac) is ending in an error:

mirka:
Got it. Thank you for clarifying! I understand it better now.

> Maybe try it with classic npm debugging (rm -rf node_modules && npm install) 😁

I tried this and it's now throwing different errors 🙄 Anyway, I think I'll be fine since `make start` is working. As per your recommendations, I won't be doing in-browser dev on Windows.

PR #843, opened 2018-09-07
[WIP] Decompose app.jsx (Part 1)

mirka:
Moves out the dialog rendering code from app.jsx.

### DIFF
--- a/lib/dialog-renderer/index.jsx
+++ b/lib/dialog-renderer/index.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import * as Dialogs from './dialogs/index';
+import * as Dialogs from '../dialogs';
 import { compact, concat, flowRight, map } from 'lodash';
 
 export const DialogRenderer = props => {
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -7,7 +7,6 @@ import reduxActions from './state/actions';
 import selectors from './state/selectors';
 import browserShell from './browser-shell';
 import { ContextMenu, MenuItem, Separator } from './context-menu';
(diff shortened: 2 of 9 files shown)


lib/dialog-renderer/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+    const { dialogs } = props;
+
+    const makeDialog = (dialog, key) => [
+      dialog.modal && (
+        <div key="overlay" className="dialogs-overlay" onClick={null} />
+      ),
+      renderDialog(dialog, key),
+    ];
+
+    return flowRight(compact, concat, map)(dialogs, makeDialog);
```
roundhill: I can learn something here :) I'm not too familiar with all of the lodash functions, but I can see that it is taking the `dialogs` and running them all through `makeDialog()`, and then what happens after that?

roundhill:
Oops missed that this was `WIP`!

lib/dialog-renderer/index.jsx
mirka: Yeah, this is some seriously terse magic! (By Dennis, I think?)

After running each dialog through makeDialog, it then runs the resulting collection through the other arg functions right to left — so that's `concat` and then `compact` (remove falsey elements).

I'm not immediately sure why the concat and compact is necessary here... the intention might have been to flatten the array elements with `concat` and then remove the false elements where `dialog.modal === false`, but AFAICT the lodash concat doesn't really do anything when the args is a single array like in this case. It still seemingly works when I remove both `concat` and `compact`.

roundhill: Ahhh yes I see that this was existing code now that was copied ever. Yeah Dennis does some cool stuff 😎

PR #846, opened 2018-09-10
Decompose app.jsx and NoteEditor (Part 2) 

mirka:
Moves the placeholder code into their respective components, and organizes some of the CSS to be with their components.


lib/note-detail/index.jsx
```diff
@@ -153,35 +154,43 @@ export class NoteDetail extends Component {
...
-            <NoteContentEditor
-              ref={this.saveEditorRef}
-              storeFocusEditor={this.storeFocusContentEditor}
-              storeHasFocus={this.storeEditorHasFocus}
-              content={content}
-              filter={filter}
-              onChangeContent={this.queueNoteSave}
-            />
+      <div className="note-detail-wrapper theme-color-border">
+        {!note ? (
```
roundhill: Good idea to move the placeholder logic here 👍

lib/note-list/index.jsx
```diff
@@ -413,43 +413,51 @@ export class NoteList extends Component {
       isSmallScreen,
     });
 
+    const isEmptyList = notes.length === 0;
+
+    const emptyTrashButton = (
```
roundhill: This change probably caused this issue: https://github.com/Automattic/simplenote-electron/pull/857#issuecomment-421426835

PR #852, opened 2018-09-12
Decompose app.jsx and NoteEditor (Part 4)

mirka:
Extracts and reduxifies RevisionSelector. (CSS for it is broken at this point — will fix at the end)

### DIFF
--- a/lib/app-layout/index.js
+++ b/lib/app-layout/index.js
@@ -1,19 +1,35 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, cloneElement } from 'react';
 import PropTypes from 'prop-types';
+import RevisionSelector from '../revision-selector';
 
-export const AppLayout = ({ searchBar, noteList, noteEditor }) => {
+export const AppLayout = ({
+  note,
+  revisions,
+  onUpdateContent,
+  searchBar,
+  noteList,
+  noteEditor,
+}) => {
   return (
     <Fragment>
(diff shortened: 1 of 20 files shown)


lib/app-layout/index.js
```diff
@@ -1,19 +1,35 @@
...
+}) => {
   return (
     <Fragment>
       <div className="app-layout-source-list theme-color-bg theme-color-fg">
         {searchBar}
         {noteList}
       </div>
-      {noteEditor}
+      <RevisionSelector
+        note={note}
```
roundhill: I think you can simply pass `note` and `onUpdateContent` here since they have the same name.

lib/revision-selector/index.jsx
```diff
@@ -126,8 +144,19 @@ export class RevisionSelector extends Component {
...
-  revisions: PropTypes.array.isRequired,
-};
-
-export default onClickOutside(RevisionSelector);
+const { setRevision, setIsViewingRevisions } = appState.actionCreators;
+
+const mapDispatchToProps = dispatch => ({
+  setRevision: revision => dispatch(setRevision({ revision })),
+  resetIsViewingRevisions: () => {
+    dispatch(setIsViewingRevisions({ isViewingRevisions: false }));
```
roundhill: Bug: When I click `Restore Note` it doesn't actually restore the revision. It should restore the note and dismiss the revisions selector.

PR #853, opened 2018-09-12
Decompose app.jsx and NoteEditor (Part 5)

mirka:
Wraps NoteToolbar in a redux-connected container, and extracts the whole thing out of NoteEditor and into AppLayout. (CSS for it is broken at this point — will fix at the end)

### DIFF
--- a/lib/note-toolbar-container.js
+++ b/lib/note-toolbar-container.js
@@ -8,6 +8,7 @@ import filterNotes from './utils/filter-notes';
 
 export class NoteToolbarContainer extends Component {
   static propTypes = {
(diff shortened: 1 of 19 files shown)


lib/note-toolbar-container.js
```diff
@@ -0,0 +1,133 @@
...
+  noteRevisions,
+  restoreNote,
+  setEditorMode,
+  setIsViewingRevisions,
+  showDialog,
+  toggleNoteInfo,
+  trashNote,
+} = appState.actionCreators;
+
+const mapDispatchToProps = dispatch => ({
```
roundhill: Yay!

PR #859, opened 2018-09-13
Improve DX

adlk:
This PR adds and improves the way we currently do electron development.

### New
- Add menu entries in dev mode
   - Reload <kbd>Cmd</kbd>+<kbd>R</kbd>
   - View Dev Tools <kbd>Cmd</kbd>+<kbd>Alt</kbd>+<kbd>I</kbd>
- Add `./env.js` helper 
   To use `isDev` simply do the following:
   ```js
   const { isDev } = require('./env');
   ```
   `isDev` checks on various conditions like `NODE_ENV=development` or if the app was started from the default electron binary (electron from `node_modules`)
   **Note:** Please keep in mind that electron files aren't compiled, so you can't use `import { isDev } from...`

### Improvements

- Replaced `make dev` logic with [`Misty`](https://github.com/adlk/misty)
   We can now launch the dev server and the electron app just with `make dev`. Whenever a file changes in `./desktop` electron will be automatically restarted (necessary for main renderer development)
- When `isDev` and the env var `DEV_SERVER` are true (default when running `make dev`) electron loads the simplenote dev server on http://localhost:4000 instead of the local files. This enables easy frontend development from electron.
- Moved all npm scripts to make
`npm start` now executes `make start` instead of `make dev` (ensure consistency)
-  `npm electron` now executes `make start` with `NODE_ENV=development`

### How to develop with this PR
Simply run
```
$ make dev
```
The electron app will automatically launch as soon as the dev server has been started on http://localhost:4000 and loads this url instead of the local files. This allows simultaneous debugging in the browser as well as electron.

That's it! :tada:

... oh and please pardon the weird commit history 🙄


Makefile
```diff
@@ -21,23 +21,40 @@ SIMPLENOTE_CHANGES_STD := `find "$(THIS_DIR)" -newer "$(SIMPLENOTE_JS)" \( -name
 SIMPLENOTE_BRANCH = $(shell git --git-dir .git branch | sed -n -e 's/^\* \(.*\)/\1/p')
 
 
-# Build configurations
+##############
+# SIMPLENOTE #
+##############
+
+# Node environment
 NODE_ENV = production
```
dmsnell: Would we want this to assign only if we haven't overwritten it from the command line?

```Makefile
NODE_ENV ?= production
```

Makefile
```diff
@@ -21,23 +21,40 @@ SIMPLENOTE_CHANGES_STD := `find "$(THIS_DIR)" -newer "$(SIMPLENOTE_JS)" \( -name
...
+
 # Defines if we should compile web app via `build` target
 SKIP_BUILD = false
 
+# Host for dev server
+HOST = 0.0.0.0
+
+# Port for dev server
+PORT = 4000
```
dmsnell: same as above for these. what is the suggested way to choose a different port, or does it have to be hard-coded?

desktop/env.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  isDev: process.env.NODE_ENV === 'development' || process.defaultApp || /node_modules[\\/]electron[\\/]/.test(process.execPath)
+}
```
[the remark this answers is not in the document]
adlk: I'm with you but believe it wouldn't be easily possible with our current setup as we don't use webpack for the desktop files.

In case of using webpack we should also think of a solution for building _shared_ modules that can be imported in the `main` as well as the `render` process without adding every module in both bundles. 
From what I know, [electron-webpack](https://github.com/electron-userland/electron-webpack) is doing something like this with the `common` directory.

Checking things like`process.defaultApp` wouldn't be possible in webpack as it's just available in electron – this however shouldn't be a biggy as we can still decide to only rely on `NODE_ENV`.

Makefile
adlk: This is actually brilliant as it also enables us to call:
```
NODE_ENV=development make dev
```
in addition to
```
make dev NODE_ENV=development
```

Dennis, 💯 points to you! 😄

desktop/env.js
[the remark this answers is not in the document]
dmsnell: I guess what I'm most curious about is whether this impacts the browser. this app needs to continue working in both environments ; how is not as important 😉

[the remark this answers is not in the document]
adlk: The web app and the code base for the electron app don‘t have any overlap atm.

desktop/app.js
```diff
@@ -22,7 +22,7 @@ module.exports = function main() {
   require('./updater')();
   const url =
     isDev && process.env.DEV_SERVER
-      ? 'http://localhost:4000'
+      ? 'http://localhost:4000' // TODO: find a solution to use host and port based on make config.
```
dmsnell: could we not set the environment variables and use them here?

```Makefile
dev-server:
  @NODE_ENV=$(NODE_ENV) HOST=$(HOST) PORT=$(PORT) npx webpack…
```

```js
const url = isDev && process.env.DEV_SERVER
	? `${ process.env.HOST }:${ process.env.PORT }`
	: `file://${ path.join( __dirname, '..', 'dist', 'index.html' );
```

PR #857, opened 2018-09-13
Decompose app.jsx and NoteEditor (Part 7)

mirka:
Turns state-dependent `.revisions` styles into clearly-scoped modifiers.

This is **the end** of the stacked PR. Everything should be working properly, including CSS.

Please use this branch for testing, and refer to the others for reviewing code changes in smaller, related chunks: #843 #846 #851 #852 #853 #856

PR #860, opened 2018-09-13
Focus Mode

### LINKED ISSUE #796: Add Focus Mode
The macOS Desktop app now has Focus Mode which makes the note editor full-width: https://github.com/Automattic/simplenote-macos/pull/229

It should have a menu item to toggle Focus Mode, as well as a keyboard shortcut. macOS uses `cmd+shift+f`.

We should add it to this app as well! I think we may need to rearrange the divs of the app in order to support easily hiding the notes list and expanding the note editor div to full width.

mirka:
Depends on #857 
Closes #796 

This is a minimum viable Focus Mode. The feature can only be toggled from the app menu (or by shortcut).

focus-mode

There are a few possible ways we can enhance this later:
- Keep the Tags button in the top left corner, which will jump straight to the tag drawer
- Add a Sidebar button (something like [this](https://thenounproject.com/search/?q=sidebar&i=1926136)) in the left corner of the Note Toolbar, which will act as a Focus Mode toggle (related: #433)
- Add a nice transition

### DIFF
--- a/desktop/menus/view-menu.js
+++ b/desktop/menus/view-menu.js
@@ -133,6 +133,16 @@ const buildViewMenu = settings => {
         type: 'separator',
       },
       {
+        label: 'Focus Mode',
+        accelerator: 'CommandOrControl+Shift+F',
+        type: 'checkbox',
+        checked: settings.focusModeEnabled,
+        click: appCommandSender({ action: 'toggleFocusMode' }),
+      },
+      {
+        type: 'separator',
+      },
+      {
         label: 'T&oggle Full Screen',
         accelerator: (function() {
           if (process.platform === 'darwin') {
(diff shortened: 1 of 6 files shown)


mirka:
> 2. Resize to the compact size.
> 3. Enable focus mode (with the list view showing)
> 4. The screen goes blank:

Fixed, thanks!

mirka:
> I do find it a bit odd that there is no UI element to leave focus mode

Yes, I'm starting to worry about the users who invoke the Focus Mode keyboard shortcut by accident and can't figure out how to exit. In the end we'll want a Sidebar toggle button, but in the meantime I think I'll keep the Tags button there as an emergency exit. I'll do another PR for that.

Issue #818, opened 2018-08-31
Font Size menu items should be labelled as "Zoom"


mirka:
The "Font Size" menu items (Bigger/Smaller/Reset) do not follow the standard convention used in most apps. This is especially jarring here because Font Size often lives in the Format menu, and is a rich text setting. Rich text editing apps usually have separate settings for Font Size and Zoom.

<img width="388" alt="screen shot 2018-09-01 at 0 56 04" src="https://user-images.githubusercontent.com/555336/44923835-68c79a00-ad84-11e8-928b-07dbe0cd040b.png">

We should follow the convention and label them:

- Zoom In
- Zoom Out
- Actual Size

They could probably live at the root level in the View menu, fenced by separators.

gie3d:
I think it looks nice to put those Zoom In, Zoom Out and Actual Size into the root level. And doing this will left Note Editor a single submenu which is Line Length, so should we pull it out into root level as well?.

I tried to hack it and it looks like this. What do you think?
![screen shot 2018-09-15 at 09 30 27](https://user-images.githubusercontent.com/1977772/45581535-1c26a780-b8ca-11e8-979f-71cc49f05980.png)

One more thing, I think it's more intuitive to label the Zoom In short key using ⌘+ rather than ⌘= but I still cannot find the way to do so. What do you think?

mirka:
@gie3d Thanks for trying this out! I think it looks great 👍 

> I think it's more intuitive to label the Zoom In short key using ⌘+ rather than ⌘= but I still cannot find the way to do so.

Yes, I was thinking about fixing that, too. It should work if you write it as `CommandOrControl+Plus`.

gie3d:
Ok, I'll try and will submit a pull request.

PR #863, opened 2018-09-17
adjusted menu to display zoom at view root level

gie3d:
Based on #818 

- I changed Font Size Bigger, Smaller, to be Zoom In and Zoom Out and moved it to view root level. 
- Changed the short key for Zoom in from ⌘= to ⌘+

### DIFF
--- a/desktop/menus/view-menu.js
+++ b/desktop/menus/view-menu.js
@@ -106,12 +106,12 @@ const buildViewMenu = settings => {
         type: 'separator',
       },
       {
-        label: '&Zoom In',
+        label: 'Zoom &In',
         accelerator: 'CommandOrControl+Plus',
         click: appCommandSender({ action: 'increaseFontSize' }),
       },
       {
-        label: '&Zoom Out',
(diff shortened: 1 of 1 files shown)


desktop/menus/view-menu.js
```diff
@@ -132,6 +105,24 @@ const buildViewMenu = settings => {
       {
         type: 'separator',
       },
+      {
+        label: '&Zoom In',
+        accelerator: 'CommandOrControl+Plus',
+        click: appCommandSender({ action: 'increaseFontSize' }),
+      },
+      {
+        label: '&Zoom Out',
```
mirka: The `&` is for the [access key feature on Windows](https://github.com/electron/electron/issues/1698), and they should be set on unique characters for each menu level. (Sorry for the lack of documentation there!) I think we can go with `Zoom &Out` here.

gie3d:
Saw that the CI checks have fail. Looking through the errors and I think it's about deployment. Not sure that it is something that my code causes this trouble?

PR #865, opened 2018-09-17
Move and restyle WordPress.com login button

mirka:
This moves the WordPress.com login button to the bottom of the login screen, and restyles it so it's not a primary button. I also changed the wording from "Sign in" to "Log in", for consistency across our products.

Janitorial changes:
- Rename CSS classes to match convention
- Change the `.login__auth-failure` class to a modifier

## Before
screen shot 2018-09-18 at 3 18 28

## After
screen shot 2018-09-18 at 3 14 10

### DIFF
--- a/lib/auth/index.jsx
+++ b/lib/auth/index.jsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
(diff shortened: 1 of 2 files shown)


mirka:
I don't have a strong opinion either way — I just though it might be better to follow our internal style guide (which says to use Log In and Sign Up) for consistency across the board, including documentation and support interactions. This is very low priority, so I will create issues on the other apps, and perhaps some first time contributors could pick them up 🙂

`Connect with WordPress.com` seems a little confusing in this particular context, so maybe not here. We could say `Connect` for the upload to WordPress feature.

PR #867, opened 2018-09-19
Tracks: Opt Out

### LINKED ISSUE #404: How do I disable Developer Telemetry?
It appears to me that every time I search, on top of the API call, there is an additional query to pixel.wp.com. I understand that developer telemetry is normal practice, but most applications allow users to opt out.
1. Can anyone verify the same behaviour _(i.e. to confirm my install is not broken or something)_
2. Is there a privacy setting somewhere that I've missed?
#### Steps to reproduce
1. Install Simplenote and log-in, then exit
2. Open Fiddler or Wireshark, start network capture
3. Open Simplenote
4. Search for something
5. Observe/Capture requests to pixel.wp.com
#### OS version
- Windows 10 (Enterprise) - 64bit - Build 14393.2222 
- Simplenote 1.0.3
#### Fiddler Screenshot

simple_note_telemetry

### LINKED ISSUE #837: Tracks: Opt Out
### Details:
- [ ] Add a new boolean field to the`Settings` entity (Simperium backed) to persist if the user opted out from Tracks
- [ ] Add a new Switch in the Settings UI

Ref. [iOS Issue](https://github.com/Automattic/simplenote-ios/issues/222)

mirka:
Closes #837
Closes #404 

- [x] Add toggle in settings dialog
- [x] Set up the preferences bucket
- [x] Reflect settings in analytics
- [x] Tweak dialog UI

settings dialog ui

### DIFF
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -64,6 +64,9 @@ const client = simperium(
         objectStore.createIndex('alphabetical', 'contentKey');
       },
     },
+    preference: function(objectStore) {
+      console.log('Configure preference', objectStore); // eslint-disable-line no-console
+    },
     tag: function(objectStore) {
       console.log('Configure tag', objectStore); // eslint-disable-line no-console
     },
@@ -105,6 +108,7 @@ document.body.appendChild(app);
 let props = {
   client: client,
   noteBucket: client.bucket('note'),
+  preferenceBucket: client.bucket('preference'),
   tagBucket: client.bucket('tag'),
   onAuthenticate: (username, password) => {
     if (!(username && password)) {
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -64,8 +64,8 @@ const client = simperium(
         objectStore.createIndex('alphabetical', 'contentKey');
       },
     },
-    preference: function(objectStore) {
-      console.log('Configure preference', objectStore); // eslint-disable-line no-console
+    preferences: function(objectStore) {
+      console.log('Configure preferences', objectStore); // eslint-disable-line no-console
     },
     tag: function(objectStore) {
       console.log('Configure tag', objectStore); // eslint-disable-line no-console
@@ -108,7 +108,7 @@ document.body.appendChild(app);
 let props = {
   client: client,
   noteBucket: client.bucket('note'),
-  preferenceBucket: client.bucket('preference'),
+  preferencesBucket: client.bucket('preferences'),
   tagBucket: client.bucket('tag'),
   onAuthenticate: (username, password) => {
     if (!(username && password)) {
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -72,7 +72,7 @@ const client = simperium(
     },
   },
   'simplenote',
-  40
+  41
 );
 
 const l = msg => {
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -32,17 +32,17 @@ const auth = new Auth(config.app_id, config.app_key);
 const appProvider = 'simplenote.com';
 
 const token = cookie.token || localStorage.access_token;
-const appId = config.app_id;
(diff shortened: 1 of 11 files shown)


lib/boot.js
```diff
@@ -64,12 +64,15 @@ const client = simperium(
...
+    preferences: function(objectStore) {
+      console.log('Configure preferences', objectStore); // eslint-disable-line no-console
+    },
     tag: function(objectStore) {
       console.log('Configure tag', objectStore); // eslint-disable-line no-console
     },
   },
   'simplenote',
-  40
+  41
```
roundhill: Can we make this a constant like `dbVersion` or something so that future devs can grok this better?

roundhill:
Looking pretty good, I sent a few fixes via Slack. Also, we need to see if the iOS/macOS apps can get fixed from sending `1/0` instead of `true/false`. If it can't get fixed we'll need to set 0/1 from all of the other clients instead of a boolean :(

Ref: https://github.com/Automattic/simplenote-ios/pull/228#issuecomment-422970461

lib/boot.js
mirka: Good point. There are a lot of arguments here, so I changed it into an object to make it clearer all around. 95f94cd

lib/flux/app-state.js
```diff
@@ -700,19 +700,32 @@ export const actionMap = new ActionMap({
...
-              );
+          preferencesBucket.get(objectKey, (e, preferences) => {
+            const preferencesData = preferences ? preferences.data : {};
+            const {
+              analytics_enabled: analyticsEnabled = true,
+            } = preferencesData;
+
+            // Create empty preferences object in the bucket if not yet there
+            if (preferences === undefined) {
+              preferencesBucket.update(objectKey, preferencesData);
```
mirka: I added a few steps in here to make sure a `preferences-key` object is created if it doesn't exist yet.

lib/flux/app-state.js
```diff
@@ -696,6 +696,78 @@ export const actionMap = new ActionMap({
...
+
+    loadPreferences: {
+      creator({ callback, preferencesBucket }) {
+        return dispatch => {
+          const objectKey = 'preferences-key';
+
+          preferencesBucket.get(objectKey, (e, preferences) => {
+            const preferencesData = preferences ? preferences.data : {};
+            const {
+              analytics_enabled: analyticsEnabled = true,
```
roundhill: Unfortunately we need to check if this is either a boolean or a int value of `0` or `1` because of how iOS saves the boolean value (it won't be fixed in the near future). We can still save it as a boolean because iOS can read the value properly still.

mirka: Added a step to cast it to a boolean 👍

roundhill: Javascript can be so magical sometimes ❤️

lib/flux/app-state.js
```diff
@@ -696,6 +696,79 @@ export const actionMap = new ActionMap({
...
       });
     },
+
+    loadPreferences: {
+      creator({ callback, preferencesBucket }) {
+        return dispatch => {
+          const objectKey = 'preferences-key';
+
+          preferencesBucket.get(objectKey, (e, preferences) => {
+            const preferencesData = preferences ? preferences.data : {};
```
roundhill: Suggestion: This might be a nice place to use `lodash`'s `get` method: https://lodash.com/docs/4.17.10#get

something like `get(preferences, 'data.analytics_enabled', true)`?

mirka: Thanks! I'm not yet familiar with all the lodash functions, so please keep these suggestions coming 🙂

roundhill: Nice! Change looks good 👍

PR #869, opened 2018-09-20
Improve auto updater

adlk:
This PR replaces the built in electron auto updater with`electron-updater`. One of the major benefits is added support for windows updates (nsis) and works perfectly in sync with the `electron-builder` binaries.

We also don't longer need to maintain a custom API to manage releases. Releases are now handled directly via Github releases that are automatically created whenever a new tag is pushed.

### How to do dev/ux-testing
1. Make sure your package.json has a version lower than `1.2.0`
1. In project root, create the file `dev-app-update.yml` with the following content:
   ```yml
   # dev-app-update.yml
   owner: adlk
   repo: simplenote-test
   provider: github
   ```
1. Run `make start` – assuming you have already built the web app once. Otherwise, run `make dev`
1. Watch the console output of the main process to see something like.
   ```console
   Checking for update
   Found version 1.2.0 (url: Simplenote-macOS-1.2.0.zip, Simplenote-macOS-1.2.0.dmg)
   Downloading update from Simplenote-macOS-1.2.0.zip, Simplenote-macOS-1.2.0.dmg
   ```

Please keep in mind that you can't actually update the app as you are running a local and unsigned binary of electron. So in any case, you'll most likely get an error when closing the update dialog. Even though we are in dev mode, all status events will be triggered.

### How to do "production" testing
1. Grab a built binary for Mac or Windows from `CircleCI > Workflows >  Automattic > simplenote-electron > test/auto-updater (latest build) > artifacts (job) > artifacts`
1. Install the app to e.g. `/Applications`
1. Launch the app and wait for the auto updater to prompt
1. Update and check that you are on Simplenote 1.2.0 (via menu: Simplenote > About Simplenote)

More infos about electron-updater can be found [here](https://www.electron.build/auto-update).


desktop/updater/auto-updater/index.js
```diff
@@ -5,58 +5,64 @@
...
-			}
-		} );
-	}
+AutoUpdater.prototype.formatGithubReleaseNotes = function(releaseNotes) {
+  return sanitizeHtml(releaseNotes, {
+    allowedTags: ['br', 'li'],
+    allowedAttributes: [],
+  })
+    .replace(/<br \/>/g, '\n')
+    .replace(/<li>(.*)<\/li>/g, '- $1');
```
dmsnell:
```html
<ul>
  <li> Major Changes
    <ul>
      <li>New AutoUpdater</li>
      <li>RegExps parsing HTML</li>
    </ul>
  </li>
  <li>Minor Changes
    <ul>
      <li>Updated code</li>
    </ul>
  </li>
</ul>
```

```html
<ul>
  <li> Major Changes
    <ul>
       - New AutoUpdater
       - RegExps parsing HTML
    </ul>
  </li>
  <li>Minor Changes
    <ul>
       - Updated code
    </ul>
  </li>
</ul>
```

If we want to perform this replacement we should parse the HTML and replace those nodes otherwise we'll get unexpected results.

adlk: In the beginning, I thought so too as I wanted to be able to display more complex text structures by e.g. converting the HTML to a cleaned up form of markdown (removing urls, text styles, ...). 

By looking at the way release notes are currently done, I'm just not sure if we wouldn't over engineer this by starting to parse the HTML – instead I went for the IMO simplest solution – but I'm definitely not holding on to this if I should do it differently!

Semi off topic: On the long run, I'd vote for a solution like using `conventional-changelog` and webview enhanced dialogs (once available in electron) to display changelogs.

dmsnell: Wouldn't necessarily call a two-level list as a more complex structure, or even this…

```html
<li><p>This spans<br />
two lines</p></li>
```

Also, HTML parsing isn't that hard - the browser does it for us - do we have the browser available here?

In other words, we can use a `TreeWalker` to iterate over the `li` nodes and replace them _and_ their containing `ul` or `ol` nodes with the proper list symbol.

Alternatively we could strip all tags or show the webview as you mention. If we go half-way I'm guessing we'll get crazy unexpected displays.

roundhill:
Everything seemed to work OK for me until I tried the production testing. The app never updated on macOS.

@adlk How can I get access to an AppVeyor build to test on Windows?

desktop/updater/auto-updater/index.js
```diff
@@ -5,58 +5,68 @@
  */
 const EventEmitter = require('events').EventEmitter;
 const util = require('util');
-const electron = require( 'electron' );
-const autoUpdater = electron.autoUpdater;
-const dialog = electron.dialog;
+const { dialog } = require('electron');
+const { autoUpdater } = require('electron-updater');
+const sanitizeHtml = require('sanitize-html');
+const TurndownService = require('turndown');
```
mirka: I wondered if we could do the HTML to MD conversion with `showdown` (the MD converter we're already using), but it looks like the feature [hasn't shipped yet](https://github.com/showdownjs/showdown/issues/582) 🤷‍♀️ We'll replace it when it's released.

roundhill:
> * Works as long as the app is in the `Applications` folder.

Ah this must have been my issue, I had it in a subfolder in `Applications` because I already had the native macOS app there.

adlk:
@mirka @roundhill thanks for your review. I'll check your issues to figure out what's going on.

In the meantime, here are a few answers:
> Windows (10)... Apps installed with both the .exe and .appx from the artifacts above did not trigger an update dialog. 

At least for appx, the updater is disabled as the Windows Store has its own update mechanism. 

> ...there is quite a long period of no feedback during the download, which will worry the user.

👍 In general, I'm a big fan of giving the user more feedback of what's happening – even if it's just a simple spinner. IMO, Github for Desktop does a great job (even though they just inform the user about the update once it's already downloaded)

> Linux (Ubuntu Bionic)... I could not get it to trigger an update dialog.

🤦‍♂️ This is actually a stupid mistake I've made. As the electron-updater checks Github for new releases. Everything that is using the manual updater (linux) is still accessing the old API.

adlk:
The white screen of death is fixed in #885

adlk:
@mirka @roundhill I've now added support for the manual updater and did a refactor of the updater code base as this will be used for `wp-desktop` as well.

Together with Lena, we've decided to move away from parsing the changelog and showing it in the popup. Instead, we'll show a button to open the changelog.  

screen shot 2018-09-28 at 14 50 12

mirka:
Thank you @adlk !

### Production build tests

#### Updated artifacts for testing
https://circleci.com/gh/Automattic/simplenote-electron/703#artifacts/containers/0
https://ci.appveyor.com/project/Automattic/simplenote-electron/builds/19476778/artifacts

- Works on Mac
- Works on Windows, however there seems to be an escaping issue with the `Update & Restart` button label:
    screen shot 2018-10-13 at 8 55 17

- Still could not get it to trigger on Ubuntu... Can you reproduce?

mirka:
@adlk I was able to confirm it working on Ubuntu! Thanks.

adlk:
@mirka the `&` issue on Windows is fixed. We now just need to wait for #971 to be resolved to get this landed 🎉

desktop/config-updater.json
```diff
@@ -1,8 +1,8 @@
 {
-  "forceUpdate": false,
   "updater": {
-    "url":
-      "https://app.simplenote.com/desktop/{platform}/version?compare={version}",
+    "downloadUrl": "https://github.com/Automattic/simplenote-electron/releases/latest",
+    "changelogUrl": "https://github.com/Automattic/simplenote-electron/blob/master/changelog.md",
```
mirka:
```suggestion
    "changelogUrl": "https://github.com/Automattic/simplenote-electron/blob/master/CHANGELOG.md",
```

Just a small tweak to the file name casing.

PR #870, opened 2018-09-20
Add mode-based config in webpack

### LINKED ISSUE #839: CSS sourcemaps causing FOUC
The CSS sourcemaps introduced in #833 are causing a FOUC ([known issue](https://github.com/webpack-contrib/css-loader/issues/613)).

Possible fixes:

- Just disable in production
- Use a plugin called [css-visor](https://github.com/NeekSandhu/css-visor) (risky since it is not widely adopted)

mirka:
This adds conditional settings based on whether the `mode` is production or development, so we aren't doing sourcemaps, minification, etc. when unnecessary.

Developing in `webpack-dev-server` should feel a lot snappier. This will also fix the FOUC in production builds (closes #839).

### DIFF
--- a/webpack.config.dll.js
+++ b/webpack.config.dll.js
@@ -1,7 +1,7 @@
 const webpack = require('webpack');
 
-module.exports = (env, argv) => {
-  const isDevMode = argv.mode === 'development';
+module.exports = () => {
+  const isDevMode = process.env.NODE_ENV === 'development';
 
   return {
     context: process.cwd(),
--- a/webpack.config.dll.js
+++ b/webpack.config.dll.js
@@ -1,42 +1,46 @@
 const webpack = require('webpack');
 
-module.exports = {
-  context: process.cwd(),
+module.exports = (env, argv) => {
+  const isDevMode = argv.mode === 'development';
 
-  devtool: 'source-map',
+  return {
+    context: process.cwd(),
+    mode: isDevMode ? 'development' : 'production',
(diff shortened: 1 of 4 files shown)


webpack.config.dll.js
```diff
@@ -1,42 +1,46 @@
 const webpack = require('webpack');
 
-module.exports = {
-  context: process.cwd(),
+module.exports = (env, argv) => {
+  const isDevMode = argv.mode === 'development';
```
dmsnell: why not base this off of `NODE_ENV`?

mirka: Good point. Thanks!

webpack.config.dll.js
```diff
@@ -1,42 +1,46 @@
...
-module.exports = {
-  context: process.cwd(),
+module.exports = () => {
+  const isDevMode = process.env.NODE_ENV === 'development';
 
-  devtool: 'source-map',
+  return {
+    context: process.cwd(),
+    mode: isDevMode ? 'development' : 'production',
+    devtool: isDevMode && 'cheap-module-eval-source-map',
```
dmsnell: one thing we did with Calypso was to let someone specify a sourcemap from the CLI as well since we sometimes want to make sure we get a full sourcemap to debug

```js
const devTool = process.env.SOURCEMAP || ( isDevMode && 'cheap-module-eval-source-map' )
```

https://github.com/Automattic/wp-calypso/blob/e46363e640727564c989bcd33c9407674bc4623d/webpack.config.js#L124

PR #877, opened 2018-09-22
Close sidebar when Settings dialog is opened

### LINKED ISSUE #146: Land on notes list after logging in
To reproduce:

1) Click Settings
2) Log out
3) Log in
4) The sidebar is open with the content faded out

I think you should land on note list with the sidebar closed.

To solve this, I've been thinking that we should close the sidebar when the Settings button is clicked. I think it's a better flow — after fiddling with any of the settings, you want to go straight back to the notes list.

mirka:
Closes #146 

This is a small enhancement to close the tag drawer when the Settings dialog is opened. It fixes two usability issues in the web version:

- Logging out and logging back in would result in an initial screen where the tag drawer was open.
- The tag drawer had to be closed manually after accessing the Settings dialog.

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -112,10 +112,16 @@ export const actionMap = new ActionMap({
         }
       }
 
-      return update(state, {
+      const updateCommands = {
         dialogs: { $push: [{ ...dialog, params, key: state.nextDialogKey }] },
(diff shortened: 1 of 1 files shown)


roundhill (approved):
Looks good @mirka! :shipit:

PR #881, opened 2018-09-25
Add Sidebar toggle button

mirka:
Depends on #880 

This adds a Sidebar toggle button that is effectively the Focus Mode toggle.

note-list-anim

The button is hidden when the window is narrow, since the view will already be in single-column:
screen shot 2018-09-26 at 1 53 25

UX concerns to think about:

- Is the icon intuitive?
- Is it confusing that we're invoking two different concepts to describe what is effectively the same thing? (i.e. Toggle Sidebar vs. Focus Mode)
- Does its utility justify the added clutter?

We might think of something better if/when we redesign the toolbars, but I think this will suffice for now.

### DIFF
--- a/lib/app-layout/style.scss
+++ b/lib/app-layout/style.scss
@@ -7,7 +7,6 @@
       overflow: hidden;
       opacity: $fade-alpha;
       width: 0;
-      transition: width .2s ease-in-out, opacity .2s ease-in-out;
     }
   }
 
@@ -38,7 +37,7 @@
 
   &.is-note-open {
     .app-layout__source-column {
-      transition: opacity .2s ease-in-out;
+      transition: width .2s ease-in-out, opacity .2s ease-in-out;
     }
 
     @media only screen and (max-width: $single-column) {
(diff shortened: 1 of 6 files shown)


roundhill:
I think I like it. Do you think it'd be too weird to only show the icon when focus mode is enabled? There's a lot of buttons at the top with this addition and I'm wondering if it is too much clutter.

mirka:
> Do you think it'd be too weird to only show the icon when focus mode is enabled?

I considered that for a while, but the thing is, that button has to always be there in the web app anyway 😅

My impression is that the clutter problem should be dealt with from the note toolbar side first. I don't have an accurate picture of usage patterns yet, but if the Revisions and Share buttons are used infrequently enough they could be moved to the Info pane. (Just a thought!)

We could also consider adding some hierarchy to the button designs. For example, the New Note button should be the primary call-to-action, the Markdown Preview button should be a different class of button, etc.

roundhill:
OK, how about we roll with this for now and we'll iterate on it later?

mirka:
@roundhill 

> One bug is that it shows the animation on startup of the app

I've resolved this by preventing the Focus Mode setting from being persisted across launches. I think this will be better UX because there are situations in which launching straight into Focus Mode would be disorienting (like when the note is empty). Also the state management will be much simpler this way.

Issue #883, opened 2018-09-26
Remove `selectedNoteId` from appState


mirka:
`appState.selectedNoteId` seems redundant given that there is an `appState.note` object with an `id` property.

belcherj:
Closing, we will visit this as we clean up app state.

PR #884, opened 2018-09-26
Keep selected note in React in sync with Redux state

### LINKED ISSUE #838: Crash When Pressing Share Button
You can "white screen of death" the app by following these steps:

1. Add a new tag to a note.
2. Open the tag drawer and select the tag.
3. Press the share button, the app will go blank.

Exception is:
```
Uncaught TypeError: Cannot read property 'data' of null
    at ShareDialog._this.collaborators (index.jsx?9832:78)
    at ShareDialog._this.renderTabContent (index.jsx?9832:155)
    at TabbedDialog.render (tabbed-dialog.jsx?6a46:48)
    at s (react-dom.production.min.js:130)
    at beginWork (react-dom.production.min.js:133)
    at r (react-dom.production.min.js:161)
    at a (react-dom.production.min.js:161)
    at o (react-dom.production.min.js:162)
    at A (react-dom.production.min.js:169)
    at C (react-dom.production.min.js:168)
```

mirka:
Closes #838 

## Problem

The Redux `appState.note` could be out of sync with the actual displayed note in React. This was causing crashes because some parts of the app (e.g. ShareDialog) were relying on `appState.note` to determine state, while other parts (e.g. NoteDetail) were using a `selectedNote` computed independently by app.jsx.

## Fix

In this fix, app.jsx will no longer attempt to compute its own "selected note", and will always rely on the Redux `appState.note`. This should prevent the kind of bugs/crashes that are caused by the two being out of sync.

There are non-trivial changes to the state-handling logic, so this should be tested thoroughly before release.

### DIFF
--- a/lib/app-layout/index.js
+++ b/lib/app-layout/index.js
@@ -13,6 +13,7 @@ export const AppLayout = ({
   note,
   noteBucket,
   revisions,
+  onNoteClosed,
   onUpdateContent,
   searchBar,
   noteList,
@@ -38,6 +39,7 @@ export const AppLayout = ({
           onUpdateContent={onUpdateContent}
         />
         <NoteToolbarContainer
+          onNoteClosed={onNoteClosed}
           noteBucket={noteBucket}
           toolbar={<NoteToolbar note={note} />}
         />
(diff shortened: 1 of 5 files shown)


mirka:
@roundhill Oh yes I forgot to mention, I've opted to deal with the `selectedNoteId` redundancy as a separate issue (#883), if that's ok. It's being used in several places, and I was afraid there might be ramifications that are not immediately apparent 😬

roundhill:
> @roundhill Oh yes I forgot to mention, I've opted to deal with the `selectedNoteId` redundancy as a separate issue (#883), if that's ok. It's being used in several places, and I was afraid there might be ramifications that are not immediately apparent 😬

Ah, ok. :shipit: then!

PR #888, opened 2018-09-26
Fix note tag add/remove re-render bug

mirka:
#834 introduced a bug in which adding/removing a note tag would update the data, but the screen wouldn't re-render.

This fixes the issue.

tag-update


lib/app-layout/index.js
```diff
@@ -1,6 +1,7 @@
 import React, { cloneElement } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { get } from 'lodash';
```
roundhill: 👍 😄

PR #892, opened 2018-10-06
Linting updates to auto-updater and manual-updater

acrosman:


acrosman:
I built the application successfully (MacOS and node 10.7.x) before opening the PR.  The check appears to fail because a dependency didn't load currently for the test runner, not because of anything triggered by the linting process.

acrosman:
Yeah, I'll cherry pick the README comment into a separate PR.

PR #896, opened 2018-10-09
Update secrets for new appveyor org

adlk:
We are currently moving to the official A8C appveyor organization. Therefore, the secrets need to be re-encrypted for the new account.

### DIFF
--- a/dev-app-update.yml
+++ b/dev-app-update.yml
@@ -1,0 +1,3 @@
+owner: adlk
+repo: simplenote-test
+provider: github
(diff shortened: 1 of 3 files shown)


dev-app-update.yml
```diff
@@ -0,0 +1,3 @@
+owner: adlk
+repo: simplenote-test
```
mirka: Are you adding this file temporarily for a build test?

adlk: Oh, good catch! This file should actually be ignored in .gitignore but this isn't merged yet 🤦‍♂️

PR #907, opened 2018-10-11
Send app version with Tracks events

mirka:
Our Tracks events weren't sending the app version with them, which is necessary for analyzing the data accurately.

I sent one test event to check that it is being received in Tracks:

screen shot 2018-10-12 at 3 24 41

Also, I've added some logic to disable event sending when in development.

### DIFF
--- a/lib/analytics/index.js
+++ b/lib/analytics/index.js
@@ -30,7 +30,10 @@ const analytics = {
         return;
       }
 
-      eventProperties = eventProperties || {};
+      eventProperties = {
+        ...eventProperties,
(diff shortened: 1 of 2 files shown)


lib/analytics/index.js
```diff
@@ -26,11 +26,14 @@ const analytics = {
 
   tracks: {
     recordEvent: function(eventName, eventProperties) {
-      if (!window.analyticsEnabled) {
+      if (!window.analyticsEnabled || process.env.NODE_ENV === 'development') {
```
roundhill: Nice 👍

PR #911, opened 2018-10-12
corrected shortcuts in menus

### LINKED ISSUE #816: Labeling and shortcut errors in menus
There are some errors in the menus that make the app feel less native.

Most of these can be fixed by adding the correct `role` and/or removing the `label` where unnecessary, as Electron has [a lot of the logic built in](https://github.com/electron/electron/blob/master/lib/browser/api/menu-item-roles.js).

We should probably do some refactoring to allow easier per-platform logic based on composition, since we will be adding more menu items. The current system is quite error-prone and is bloating the app.js file.

### all
- [x] `accelerator` for “Font Size ▸ Bigger” should be `CommandOrControl+Plus` instead of `CommandOrControl+=` ([Quirk](https://github.com/atom/electron/issues/1507) seems to be fixed now) (#863)
- [ ] Multi-step items (“Preferences”, “Export Notes”, “Print”, “Search Notes”) should have an ellipsis `…` appended

### darwin
- [ ] Shortcut for “Hide Others” should be Cmd+Opt+H
- [ ] “Hide” and “Quit” should be suffixed by app name
- [ ] Help menu should include searchable field (can be achieved by adding `role: 'help'`)
- [x] “Bring All to Front” is in the wrong menu (#813)

### linux
- [ ] “Exit” should be “Quit”
- [ ] Window menu should not exist
- [ ] Help / Help & Support is duplicated

### win32
- [ ] Shortcut for “Redo” should be Ctrl+Y
- [ ] Window menu should not exist

rakhi2104:
Resolves #816 
I don't have access to a Mac so, couldn't test the functionality there.
Someone can give me a hand there ... :handshake:

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -191,7 +191,7 @@ function createMenuTemplate(settings) {
   };
 
   const settingsMenuItem = {
-    label: 'P&references',
+    label: 'P&references ...',
     accelerator: 'CommandOrControl+,',
     click: function(item, focusedWindow) {
       if (focusedWindow) {
@@ -235,7 +235,7 @@ function createMenuTemplate(settings) {
         type: 'separator',
       },
       {
-        label: '&Export Notes',
+        label: '&Export Notes ...',
         accelerator: 'CommandOrControl+Shift+E',
         click(item, focusedWindow) {
           if (focusedWindow) {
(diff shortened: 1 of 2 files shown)


mirka:
Sorry, I'm going to put this on hold and try to refactor the menu stuff first 🙇🏻‍♀️ The platform-specific logic is fragile and there are pre-existing mistakes still remaining.

rakhi2104:
I'll definitely update them. :v:

mirka:
Thank you for your work on this. I refactored the menu code (#1004) so it will be easier to make changes like this in the future 🙏

Issue #915, opened 2018-10-14
Inconsistent user interface concerning Share/Collaborate


marianoju:
#### Steps to reproduce
1. Open a note in Simplenote 
2. Select “Share” icon (in the toolbar in the upper right corner) 
3. Add Email 
4. Select “Done” (and thereby close the “Share” window (aka “Collaborate screen”) 
5. Select “Add a tag” (bottom), add an email and press Enter 

#### What I expected
I expected both approaches to work. This assumption was encouraged by the following wording: “You can also add the email as a tag.”, see screenshot. 

![image](https://user-images.githubusercontent.com/6254428/46916118-82c8df00-cfb6-11e8-9854-e6ffd66f9ac3.png)

#### What happened instead
It is not possible to add an email as a tag. A notification or error message is displayed: “Collaboration has moved. Press the Share icon in the toolbar to access the _Collaborate screen_”. 

![image](https://user-images.githubusercontent.com/6254428/46916142-dfc49500-cfb6-11e8-8740-33aa81fd3ea4.png)

#### Simplenote version

Simplenote 1.1.7

#### OS version
Ubuntu 16.04 

#### Screenshot / Video
See above

mirka:
Thank you for this report!

The sentence “You can also add the email as a tag.” should be removed from the Share dialog.

PR #919, opened 2018-10-16
update message in Share Dialog

rakhi2104:
Resolves issue #915


rakhi2104:
My pleasure, @mirka  :smile:

PR #920, opened 2018-10-16
Fix appx builds

adlk:
This PR splits the Appveyor CI builds into two separate steps as the binaries that are being used to generate the appx package needs to be unsigned. 

- `Default` generates the signed binaries and the signed NSIS .exe installer
- `AppX` generates an unsigned binary and appx package

If the binary is signed prior to creating the appx – the publisher info for the appx is being pulled from the certificate instead of the provided `publisher` config from `electron-builder-appx.json`


adlk:
@roundhill you now have 3 build jobs for
- signed production build (.exe)
- unsigned Windows Store build (.appx)
- signed Windows Store test build (.appx)

PR #921, opened 2018-10-16
Only decrypt secrets when CSC_ENC_KEY is set

adlk:
For forked (external) PR's, the `CSC_ENC_KEY` environment is not defined which causes the CI builds to fail during secret decryption. With this PR we only decrypt the secrets when `CSC_ENC_KEY` is set.


roundhill:
@adlk do we still need this?

belcherj:
Closing as everything seems to work without it.

Issue #916, opened 2018-10-15
Evernote importer


mirka:
Carrying on from #365 

## TODO
- [x] design importer dialog ([Sketch prototype](https://sketch.cloud/s/aGbWP/n7DZdz/play)) @mirka 
- [x] implement importer dialog #957 @mirka 
- [x] add `Import From ▶` `Evernote...` item in the File menu to trigger the dialog @mirka 
- [x] write the parser #940 @roundhill 
- [x] write the importer #922 @roundhill 

I think we should keep it as simple as possible for the first iteration, and we can add enhancements later.

SeanKilleen:
In terms of the parsing, this project may help: https://github.com/claytron/ever2simple

PR #922, opened 2018-10-16
Importer: Add core Simplenote format support

roundhill:
Here's the first step towards getting an importer going. This module expects to receive a JSON object that contains note data in the same format as when you export your notes from the app (the `notes.json` file).

I'm slightly concerned about performance here with large imports, perhaps we could use a [web worker](https://www.w3schools.com/html/html5_webworkers.asp)? The exporter is crazy fast with thousands of notes, however, so maybe we don't need to worry about it.

**To Test**
* Verify `npm run test` passes (I added some very basic tests)
* Export some notes from the app, and grab the `notes.json` file and put it in the source code in the same folder as `app.jsx`.
* Add `import exportedJSON from './notes.json'` and `import importNotes from './utils/import';` to the top of app.jsx.
* In `render()` of `app.jsx` (or anywhere you'd like) run `importNotes(exportedJSON, this.props.noteBucket);`
* The notes should import to the app and sync!

Refs: #916 

cc @dmsnell!

### DIFF
--- a/lib/utils/import/README.MD
+++ b/lib/utils/import/README.MD
@@ -2,13 +2,12 @@
 
 This is the core importer for Simplenote. It expects a JSON object of notes in the same format as the `import` module. It is aggressive about only allowing certain properties, and converts dates into timestamps where appropriate. Importing specific export files (Such as Evernote .enex) should use a separate conversion script and then pass the resulting JSON object to `importNotes()`.
 
-Example JSON:
+Example JSON with allowed properties:
 
 ```
 {
   "activeNotes": [
     {
-      "id": "279d0a2a9cbf4107830175fd6b562f62",
       "content": "Random thought: How much wood could a woodchuck chuck if a woodchuck could chuck wood?",
       "creationDate": "2013-10-16T20:17:41.760Z",
       "lastModified": "2018-10-10T22:07:54.128Z",
@@ -20,8 +19,7 @@ Example JSON:
   ],
   "trashedNotes": [
     {
-      "id": "80461f52-2c63-4f3f-825e-69a0236da327",
-      "content": "Hello?",
+      "content": "Hello, world!",
       "creationDate": "2016-02-29T23:01:03.115Z",
       "lastModified": "2016-02-29T23:01:08.000Z"
     }
--- a/lib/utils/import/README.MD
+++ b/lib/utils/import/README.MD
@@ -1,0 +1,30 @@
+# Simplenote Import
+
+This is the core importer for Simplenote. It expects a JSON object of notes in the same format as the `import` module. It is aggressive about only allowing certain properties, and converts dates into timestamps where appropriate. Importing specific export files (Such as Evernote .enex) should use a separate conversion script and then pass the resulting JSON object to `importNotes()`.
+
+Example JSON:
+
+```
+{
+  "activeNotes": [
+    {
+      "id": "279d0a2a9cbf4107830175fd6b562f62",
+      "content": "Random thought: How much wood could a woodchuck chuck if a woodchuck could chuck wood?",
+      "creationDate": "2013-10-16T20:17:41.760Z",
+      "lastModified": "2018-10-10T22:07:54.128Z",
+      "pinned": true,
+      "tags": [
+        "Reminders"
+      ]
+    },
+  ],
+  "trashedNotes": [
+    {
+      "id": "80461f52-2c63-4f3f-825e-69a0236da327",
+      "content": "Hello?",
+      "creationDate": "2016-02-29T23:01:03.115Z",
+      "lastModified": "2016-02-29T23:01:08.000Z"
+    }
+  ]
+}
+```
--- a/lib/utils/import/README.MD
+++ b/lib/utils/import/README.MD
@@ -1,8 +1,8 @@
 # Simplenote Import
 
-This is the core importer for Simplenote. It expects a JSON object of notes in the same format as the `import` module. It is aggressive about only allowing certain properties, and converts dates into timestamps where appropriate. Importing specific export files (Such as Evernote .enex) should use a separate conversion script and then pass the resulting JSON object to `importNotes()`.
+This is the core importer for Simplenote. It expects an object of notes in the same format as the `import` module. It is aggressive about only allowing certain properties, and converts dates into timestamps where appropriate. Importing specific export files (Such as Evernote .enex) should use a separate conversion script and then pass the resulting object to `importNotes()`.
 
-Example JSON with allowed properties:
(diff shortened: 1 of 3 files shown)


lib/utils/import/index.js
```diff
@@ -0,0 +1,49 @@
...
+  'tags',
+  'deleted',
+];
+
+const importNotes = (notes = {}, noteBucket) => {
+  if (isEmpty(notes)) {
+    throw new Error('No notes to import.');
+  }
+
+  if (!get(notes, 'activeNotes') || !get(notes, 'trashedNotes')) {
```
mirka:
```suggestion
  if (!notes.activeNotes || !notes.trashedNotes) {
```

I’m wondering if this would be more readable and functionally equivalent in this case.

lib/utils/import/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+    if (importedNote.creationDate) {
+      importedNote.creationDate = Date.parse(importedNote.creationDate);
+    }
+
+    noteBucket.add(importedNote);
+  };
+
+  get(notes, 'activeNotes', []).map(note => importNote(note));
+  get(notes, 'trashedNotes', []).map(note => importNote(note));
```
mirka: I think `.map(importNote)` would be equivalent here.

lib/utils/import/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const importNote = note => {
+    const importedNote = pick(note, propertyWhitelist);
+
+    // We don't want to allow these properties to be imported, but they need to be set
+    importedNote.publishURL = '';
+    importedNote.shareURL = '';
+    importedNote.systemTags = [];
+
+    importedNote.deleted = get(importedNote, 'deleted', false);
```
mirka: The `deleted` prop isn’t set on trashed notes in the export json, so nothing is being set to `true` here. We will have to explicitly set this prop to `true` with each note in trashedNotes.

lib/utils/import/README.MD
```diff
@@ -0,0 +1,30 @@
+# Simplenote Import
+
+This is the core importer for Simplenote. It expects a JSON object of notes in the same format as the `import` module. It is aggressive about only allowing certain properties, and converts dates into timestamps where appropriate. Importing specific export files (Such as Evernote .enex) should use a separate conversion script and then pass the resulting JSON object to `importNotes()`.
+
+Example JSON:
```
mirka: Maybe the “Example JSON” should be “Example JSON with allowed properties”, and then remove the `id` prop?

lib/utils/import/test.js
```diff
@@ -0,0 +1,14 @@
+import importNotes from './';
+
+describe('Importer', () => {
+  it('should throw when no notes are passed', () => {
+    expect(importNotes.bind(importNotes, null)).toThrow(/No notes to import./);
+  });
+
+  it('should throw when invalid json is passed', () => {
+    const bogusNotes = { actveNotes: [] };
+    expect(importNotes.bind(importNotes, bogusNotes)).toThrow(
```
mirka: Out of curiosity, are these `bind` statements intended to do anything more than a simple `expect(() => importNotes(null))`? It had me wondering why a bind was necessary here.

dmsnell:
How about we add some tag to indicate which import the notes came from? even if we leave it as a user-facing tag the customer could remove them if they wanted.

could make mess-ups significantly easier to deal with, like when dealing with duplicate notes or partial imports.

roundhill:
> * We also need to add the tags to the tagBucket.

Good call!

lib/utils/import/index.js
roundhill: Should we add `deleted` prop to the export?

roundhill: The `deleted` property is required, or else Simperium will reject it.

lib/utils/import/test.js
roundhill: I think it was executing the function during the test instead of evaluating it for the result. Hope that makes sense :)

roundhill:
> How about we add some tag to indicate which import the notes came from? even if we leave it as a user-facing tag the customer could remove them if they wanted.

Hmm. How about we add a `systemTag` for that? Then it wouldn't be user facing.

dmsnell:
> How about we add a systemTag for that? Then it wouldn't be user facing.

Seems just as good. I bet some people want to see it, others don't. System tag seems fine. `importedFrom: [ 'some-export-filename.json', '2018-10-17T05:13:31-07:00Z' ]` or something like that

lib/utils/import/index.js
mirka: I'm guessing that the `deleted` prop was not in the export because having that and also keeping a trashedNotes array would be redundant. We could flatten the export structure, but that would be disruptive for users who have export files with the old structure.

Explicitly setting back the `deleted` prop in the importer side should be ok, I think.

lib/utils/import/test.js
mirka: Ok, if it's just to avoid the immediate call, then simply wrapping the call in a function (e.g. `expect(() => importNotes(null))` ) should be sufficient.

lib/utils/import/index.js
roundhill: Also this should be an `&&` not an `||` 😬

lib/utils/import/test.js
roundhill: Nice! Thanks.

roundhill:
ToDo:
- [ ] Via #938, we should import the `markdown` property to the `systemTags` if present.

mirka:
Sorry I didn't catch this on the first pass, but this CoreImporter is actually not dealing with JSON at all (it just takes JavaScript objects as input), so we should probably remove mentions of JSON in the readme.

(So why is the `import exportedJSON from './notes.json'` working? TIL: webpack >= 2.0.0 has a json-loader enabled by default that parses the JSON string from a file into a JS object automatically 😮)

To allow users to import their JSON files, we're going to need a separate SimplenoteImporter to read from `File`s and pass the `JSON.parse`d result to the CoreImporter. (If we want to get the Evernote importer out the door, I guess we can add it later though? Doesn't seem like a must-have feature for the first iteration, and it will spread the surface area for testing.)

roundhill:
> CoreImporter is actually not dealing with JSON at all

Oh yeah, lol! I think I can change `importNotes` of CoreImporter to handle a File, in the same way the other importers do it. It'd take the file and load the file in as a string to `JSON.parse()`. Could be some memory issues with that, but Simplenote exports should be relatively small files for most users.

roundhill:
I'm good!

PR #931, opened 2018-10-18
Add PUBLISH option to make

adlk:
This PR adds the option to modify the publish flag for electron-builder which previously defaulted to `onTagOrDraft`. The new default is `onTag`.

**How to use the new option:**

```
PUBLISH=always|onTag|onTagOrDraft|never make package
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,9 @@ PORT ?= 4000
 # Access dev server or locally built web app files
 DEV_SERVER ?= false
 
+# Port for dev server
+PUBLISH ?= onTag
+
 
 # Main targets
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,6 +42,9 @@ PORT ?= 4000
 # Access dev server or locally built web app files
 DEV_SERVER ?= false
 
+# Port for dev server
```
mirka: I found a copy-paste artifact 🙂
(Btw it may be helpful to put the possible choices in the comments here)

adlk: Good idea! The comments are now updated!

PR #932, opened 2018-10-18
Add web app detection to analytics

### LINKED ISSUE #908: Add Web app name to Tracks
When we start using this on the web, we should use an appropriate Tracks name so we know which requests came from the Web app.

We _could maybe_ just [change this code](https://github.com/Automattic/simplenote-electron/blob/master/lib/analytics/index.js#L20) from returning `unknown_` to `spweb_`?

mirka:
Closes #908

This sets the prefix `spweb` to Tracks events when the app is [not running in Electron](https://github.com/electron/electron/issues/2288#issuecomment-337858978).

### DIFF
--- a/lib/analytics/index.js
+++ b/lib/analytics/index.js
@@ -1,7 +1,7 @@
 /**
  * Internal dependencies
  */
-var user, prefix;
+var user;
 require('./tracks');
 
 const analytics = {
@@ -12,12 +12,12 @@ const analytics = {
 
   getPlatformPrefix: function() {
     if (navigator.appVersion.indexOf('Win') !== -1) {
-      return 'spwindows_';
+      return 'spwindows';
     } else if (navigator.appVersion.indexOf('Linux') !== -1) {
-      return 'splinux_';
+      return 'splinux';
     }
 
-    return 'unknown_';
+    return null;
   },
 
   setUser: function(newUser) {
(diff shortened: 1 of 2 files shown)


lib/analytics/index.js
```diff
@@ -10,14 +10,20 @@ const analytics = {
...
-    if (navigator.appVersion.indexOf('Win') !== -1) {
-      return 'spwindows_';
-    } else if (navigator.appVersion.indexOf('Linux') !== -1) {
-      return 'splinux_';
+    if (!navigator.appVersion.includes('Electron')) {
+      return 'spweb';
     }
 
-    return 'unknown_';
+    switch (true) {
```
roundhill: Is this better than an `if` statement here? 🤔

lib/analytics/index.js
```diff
@@ -30,26 +36,17 @@ const analytics = {
         return;
       }
 
-      eventProperties = {
+      const prefix = analytics.getPlatformPrefix();
+
+      if (!prefix) return;
```
roundhill: We should always use [braces around if statements](https://github.com/Automattic/wp-calypso/blob/master/docs/coding-guidelines/javascript.md#spacing). (We try to follow the Calypso guidelines where appropriate)

lib/analytics/index.js
mirka: My unconscious mind was wishing for 'Mac' to join my list of valid platforms 😁

lib/analytics/index.js
mirka: Thank you! (I was thinking of updating our eslint/prettier rules... They are really not in line with the Calypso guidelines)

roundhill: Sounds good! Although I have gotten used to our prettier format 🤔

PR #940, opened 2018-10-19
Add Evernote Import Parser

roundhill:
Building on top of #922 with an Evernote importer! This only works with Evernote `.enex` files, and _does not_ work in the browser unfortunately due to the large file sizes of the Evernote exports (they contain binary data, base 64 encoded).

Evernote note data is exported in a nasty HTML format, so I've done my best to try and strip out all of the markup while conserving newlines. It also imports the tags 🏷 

It makes use of the `sax` module to stream the file while parsing to conserve memory. I've tested this with ~100MB file and it works well, but I would imaging large Evernote exports could be in the gigabytes easily so we should test that.

The caller of the import can receive emits of the progress of the import or when it fails:
```
const evernoteImporter = new EvernoteImporter({ noteBucket, tagBucket });
evernoteImporter.on('status', (eventName, details) => {
    console.log(eventName, details); // e.g. 'progress', 42
});

evernoteImporter.importNotes('/Users/user/test/evernote.enex');
```

**To Test**
* Pass `importEvernotes` a file path to an `.enex` Evernote export (full path, like `/Users/user/Desktop...`), along with the noteBucket and tagBucket.
* The notes should import and sync to your account!

**ToDo**
- [x] Passing progress or error messages back to the caller


mirka:
Thank you so much for working on this 😂

I have a few suggestions for the integration:

- Have the importer constructors take a single `buckets` object with named props `{ noteBucket: Bucket, tagBucket: Bucket }` so we never mess up the arg order.
- Make all importers expose their default import method as  `import()`, so consumers don’t have to keep track of the specific method names. That way, instead of knowing to do `.importTextFiles()` or `.importEvernotes()`, we can just call the same method name: `whateverImporter.import()`.
- Have the `import()` method take an options object as a second argument. To start, I think the `markdown` option is a must. As a future enhancement, we could allow things like “add arbitrary tag”.

roundhill:
> It looks like the errors currently emitted in the two importers are all of the former kind. Shall I just assume for this iteration that I won’t receive errors of the latter kind?

I can add some note-level errors, should I send the event name as `error-note` or something?

roundhill:
> * Have the `import()` method take an options object as a second argument. To start, I think the `markdown` option is a must

Would the `markdown` option set every imported note to have the `markdown` systemTag then?

mirka:
> I can add some note-level errors, should I send the event name as `error-note` or something?

Since I won't need note-level errors in real time, it might be cleaner to have the `complete` event send an optional error object (or array?) containing whatever note-level errors that occured.

roundhill:
```
import = filesArray => {
```

This looks pretty weird due to `import` so common in js :) How about the standard method name is `importNotes`?

lib/utils/import/evernote/index.js
```diff
@@ -0,0 +1,158 @@
+const fs = __non_webpack_require__('fs'); // eslint-disable-line no-undef
```
mirka: I realized we're going to need to do some conditional loading here so it doesn't break in browsers (or jest).

↓ Something like we're doing in app.jsx

```
let fs = null;
const foundElectron = has(window, 'process.type');

if (foundElectron) {
  fs = __non_webpack_require__('fs'); // eslint-disable-line no-undef
}
```

PR #950, opened 2018-10-23
Improve a11y for modals

mirka:
These improvements currently apply to the Share, Settings, and About dialogs.

### A11y Improvements

- <kbd>Esc</kbd> will close the dialog.
- Close dialog buttons are clearly labeled with `aria-label`.
- Keyboard focus is correctly trapped inside the dialog.
- Dialog will be correctly `aria-label`ed even if a visible heading does not appear within the dialog (Like in the About dialog).

### Technical Changes

- The `multiple` (previously `single`) property of the dialog redux data now defaults to `false` so we don't have to set it every time.
- The `modal` property of the dialog redux data has been removed.
- Dialog settings (`type` and `title`) now live in a single place (`lib/dialogs/types`).

### To Test

In `scss/buttons.scss`, remove the `outline: 0` rules from `button` and `.button` so you can see the focus outlines. (FYI: These focus outlines will be addressed in a separate PR)

Some things to check:

- Dialogs look correct in both Light and Dark mode.
- <kbd>Esc</kbd> closes a dialog.
- Keyboard focus is trapped (Test by <kbd>tab</kbd>ing around immediately after the dialog has opened).
- A dialog element (the one with `role="dialog"`) has an `aria-label`. (Check in devtools)


desktop/app.js
```diff
@@ -180,11 +182,7 @@ function createMenuTemplate(settings) {
       if (focusedWindow) {
         focusedWindow.webContents.send('appCommand', {
           action: 'showDialog',
-          dialog: {
-            type: 'About',
-            modal: true,
-            single: true,
-          },
+          dialog: DialogTypes.ABOUT,
```
roundhill: 👍

lib/dialog-renderer/style.scss
```diff
@@ -0,0 +1,23 @@
+.dialog-renderer--overlay {
```
roundhill: I think we usually use underscores to signify a structure difference in sass?

`.dialog-renderer__overlay`

lib/dialog/style.scss
```diff
@@ -1,45 +1,13 @@
...
-}
-
-.dialog-box {
-  flex: 1 1 auto;
-  margin: auto;
+  width: calc(100vw - 2rem);
   background: white;
   border-radius: $border-radius;
   border: 1px solid lighten($gray, 30%);
   box-shadow: 0 2px 4px 0 rgba($gray-darkest, 0.15);
```
roundhill:
```suggestion
  box-shadow: 0 2px 4px 2px rgba($gray-darkest, 0.15);
```

I think if we add a 2px spread here it helps the dialog stand out a little bit more along the edges of the dialog, especially in the dark theme:
screen shot 2018-10-23 at 1 49 42 pm

lib/dialog/style.scss
```diff
@@ -1,45 +1,13 @@
...
 }
 
 .theme-dark {
-  .dialog-box {
+  .dialog {
     box-shadow: 0 2px 4px 0 rgba(darken($gray, 45%), 0.5);
```
roundhill:
```suggestion
    box-shadow: 0 2px 4px 2px rgba(darken($gray, 45%), 0.5);
```

lib/dialog-renderer/style.scss
mirka: Brain fart!! 🙈 Thank you 🙏🙏🙏

lib/dialog/style.scss
mirka: Great idea!

PR #952, opened 2018-10-24
Add Text File Importer

roundhill:
Building on #922 with an importer for text files in a folder. You simply construct a `TextFileImporter` with the note and tag bucket, and then pass along an array of `File` objects to `importNotes()`:

```
const textImporter = new TextFileImporter({
        this.props.noteBucket,
        this.props.tagBucket
      });

      textImporter.on('status', (eventType, message) => {
        console.log(eventType, message);
      });

      const filesArray = [array of `File` objects]
      textImporter.importNotes(filesArray);
```

It will only pull in files that end in .txt that are less than 1mb in size.

Refs #916


lib/utils/import/text-files/index.js
```diff
@@ -0,0 +1,58 @@
...
+    this.noteBucket = noteBucket;
+    this.tagBucket = tagBucket;
+  }
+
+  importTextFiles = filesArray => {
+    const coreImporter = new CoreImporter(this.noteBucket, this.tagBucket);
+    let importedNoteCount = 0;
+
+    if (!filesArray) {
+      this.emit('stats', 'error', 'No text files to import.');
```
mirka:
```suggestion
      this.emit('status', 'error', 'No text files to import.');
```

lib/utils/import/text-files/index.js
```diff
@@ -0,0 +1,58 @@
...
+        this.emit('status', 'progress', importedNoteCount);
+      };
+
+      fileReader.readAsText(file);
+    };
+
+    for (const file of filesArray) {
+      importTextFile(file);
+    }
+  };
```
mirka: Could we emit a `'complete'` status event at the end of this, too?

lib/utils/import/text-files/index.js
roundhill: Thanks ;)

lib/utils/import/text-files/index.js
roundhill: Yeah, this is async so I need to find a way to know when we've reached the end.

roundhill: Added a solution in 2bdc4bfcd9978bb15f310f7d957cb652d4675fad. I'm not in love with it since it may fire early due to the async nature of `FileReader`, but it will probably suffice.

mirka: Btw this is exactly where Promises come in handy ✨

Simplified example:

```
const importTextFile = file => {
  return new Promise((resolve, reject) => {
    const fileReader = new FileReader();
    fileReader.onload = event => {

      // import note here...

      this.emit('status', 'progress', importedNoteCount);
      resolve();
    };

    fileReader.readAsText(file);
  });
};

Promise.all(filesArray.map(importTextFile))
  .then(() => {
    this.emit('status', 'complete', importedNoteCount);
  });
```

(Not a blocker, the lastFileName strategy is probably good enough and we can promisify it later!)

roundhill: Oh nice! I haven't worked with Promises before. 👍

roundhill:
I think we may want to see if the title of the file is contained in the text. I have an export from apple notes that puts the title of the note in the file title, but not in the content:

[😘Breakfast.txt](https://github.com/Automattic/simplenote-electron/files/2526377/Breakfast.txt)

roundhill:
> If the title isn't the first part of the content, we should use it as the title.

Added via 876c7be082edb754c21ff7dc8f358a9689d1825b

SeanKilleen:
@roundhill thanks so much for this! Is going to save me a ton of time in my life.

PR #962, opened 2018-10-29
Restore :focus outlines

### LINKED ISSUE #953: Restore :focus outlines
Most interactive elements (e.g. buttons, toggles) do not have visible `:focus` styles, making the app difficult to navigate by keyboard.

We need to restore the focus outlines, or add alternative styles where suitable.

mirka:
Closes #953 

This restores :focus outlines for buttons, radio buttons, checkbox toggles, and sliders.

Thanks to a [handy polyfill](https://github.com/WICG/focus-visible), they will only be shown when the user is interacting with a keyboard. (Normal mouse clicks will not show them)

outlines

There are several buttons that will not show a focus outline due to incorrect semantics (Will be fixed in a separate PR #961 for easier review):

- Log In button
- Buttons in RevisionSelector
- "Clear search" button in SearchField

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
+import 'focus-visible/dist/focus-visible.js';
 import appState from './flux/app-state';
 import reduxActions from './state/actions';
 import selectors from './state/selectors';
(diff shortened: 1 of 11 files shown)


roundhill:
One sort-of-related thing that's a little weird is when you tab into the note editor and it locks you in to tabbing the text inside of it, there's no way to get out of that as far as I can see.

mirka:
> when you tab into the note editor and it locks you in to tabbing the text inside of it, there's no way to get out of that as far as I can see.

Yes, that's something we should address! Thanks for bringing it up. (Issue posted at #970)

PR #986, opened 2018-11-06
Ensure that all imported notes are sent to the server

mirka:
Importing over 350-400 notes at once can cause the syncing to get stuck. When the syncing is stuck, it can be resumed by calling `client.connect()`.

How can we tell that syncing is stuck? We first need to be able to detect whether the syncing is active or idle. When syncing is idle despite there being unsynced notes left, we can tell that it is stuck. Then, we can resuscitate the transfer by calling `client.connect()`.

This workaround adds a way to detect syncing activity (active/idle), by attaching debounced listeners to the simperium client's `send` and `message` events. We can now hook arbitrary functions to these state changes. (This will probably come in handy when we add a syncing indicator in the future.)

#### Also changed

- Added the [`debug`](https://www.npmjs.com/package/debug) module for easier debugging (cc: @adlk)
- Increased the max file size for SimplenoteImporter to 5 MB

#### To test

1. To see the pertinent logs in the console, enter this into the devtools console and reload: `localStorage.debug = 'sync:*'`.
1. Try doing a 400+ note import. See that the syncing is reactivated when it goes idle with unsynced notes remaining.
1. After the sync is over, verify that you can log out without warnings.

Here are some import files of various sizes: [mock-simplenote-json.zip](https://github.com/Automattic/simplenote-electron/files/2553819/mock-simplenote-json.zip)

### DIFF
--- a/lib/utils/sync/activity-hooks.js
+++ b/lib/utils/sync/activity-hooks.js
@@ -5,7 +5,7 @@ const debug = Debug('sync:activityHooks');
 
 export const debounceWait = 500;
 
-export const onSyncActive = debounce(
+const onSyncActive = debounce(
   (data, onActive) => {
     debug('Sync active...');
 
@@ -20,7 +20,7 @@ export const onSyncActive = debounce(
   }
 );
 
-export const onSyncIdle = debounce((data, onIdle) => {
+const onSyncIdle = debounce((data, onIdle) => {
   debug('Sync idle');
 
   if (isFunction(onIdle)) {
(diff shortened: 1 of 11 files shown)


lib/utils/sync/activity-hooks.js
```diff
@@ -0,0 +1,47 @@
...
+
+export const onSyncActive = debounce(
+  (data, onActive) => {
+    debug('Sync active...');
+
+    if (isFunction(onActive)) {
+      onActive();
+    }
+  },
+  500,
```
roundhill: Maybe create a constant for these two `500` values?

roundhill:
I also tried an import while offline, and then connecting up again. It synced up in the same way and I could see the idle reviver working 👍

lib/utils/sync/activity-hooks.js
mirka: Good idea!

Issue #540, opened 2017-04-06
Large line spacing when printing markdown note


grvrulz:
#### Steps to reproduce
1. Open a note with markdown in the desktop app (on Linux)
2. Print a note by pressing Ctrl + P
3. Print to file, and save
4. Open pdf and observe

#### What I expected
PDF looks sane

#### What happened instead
PDF has insane line spacing. (Printing to pdf from the web app works fine.)

#### OS version
Tested on elementary OS 0.4 and Ubuntu 16.04

#### 
[printed_from_browser.pdf](https://github.com/Automattic/simplenote-electron/files/902487/printed_from_browser.pdf)
[printed_on_app.pdf](https://github.com/Automattic/simplenote-electron/files/902488/printed_on_app.pdf)
[Actual Note](https://app.simplenote.com/p/W5Ph7k)

grvrulz:
For easier testing, here's the markdown file that I used:
[test.md.txt](https://github.com/Automattic/simplenote-electron/files/906257/test.md.txt)

dmsnell:
Thanks for the report @grvrulz 

Looks like we might need some adjustment to the print stylesheet

grvrulz:
Looks like this also happens on Windows, and the new Mac OS beta, according to user reports and some testing I(and others) did.

arunsathiya:
Noting that this is not the case on the latest stable version of Mac and Electron apps. 

- Tested on Simplenote for Mac (Simplenote 1.3.8) and the PDF save looks as expected. [PDF save from Simplenote for Mac.pdf](https://github.com/Automattic/simplenote-electron/files/2476640/PDF.save.from.Simplenote.for.Mac.pdf)
- Tested on Simplenote Electron 1.1.7 and the PDF looks as expected as well. 
[PDF save from Simplenote Electron.pdf](https://github.com/Automattic/simplenote-electron/files/2476641/PDF.save.from.Simplenote.Electron.pdf)

Tested using the same test file mentioned earlier on https://github.com/Automattic/simplenote-electron/issues/540#issuecomment-292591886. The saves are fine for other tests I did using other content as well. Should we consider closing this issue, @mirka?

mirka:
Thanks for testing @arunsathiya !

Unfortunately I can still reproduce this issue on Simplenote Electron 1.1.7. It looks like you might have printed before enabling Markdown on the note. With Markdown enabled, the PDF looks like this: 
[...]

PR #992, opened 2018-11-06
Fix line spacing when printing markdown note

vadimnicolai:


vadimnicolai:
Intent is to adjust spacing, which I have done with additional commit. I have borrowed resets from Gutenberg: https://github.com/BafS/Gutenberg/blob/master/scss/_print-reset.scss

mirka:
Ok, thanks for clarifying! However, I was thinking the line spacing issue brought up in #540 is caused by the `white-space: pre-line`, and it should probably be fixed by removing that one rule. Did you try that?

Issue #593, opened 2017-07-27
Rendered markdown in note list?


giannissc:
#### Steps to reproduce
1. Write a markdown note and add (#) to create a heading for the title of the note will show the unrendered markdown on the side note display.

#### What I expected
I would prefer that the side pane also renders the markdown (at least the first few lines that show) to keep consistency with non markdown notes. This is just a suggestion

#### Screenshot / Video
![screenshot from 2017-07-28 00-17-38](https://user-images.githubusercontent.com/20277283/28692726-79077146-732a-11e7-8f21-7f72d897a1f2.png)

iblazhko:
Another option would be to clear Markdown markup so that the notes list would look like this:

    Random Note
    NOTE 1
    ---
    Names
    Planning ahead
    ---
    Book
    Just Listen: ...

PlugaruT:
Any updates in this one?

audiovideodisco:
I would looove it if Simplenote showed the rendered markdown in the note list, as @giannissc suggested. Clearing the Markdown markup, as @iblazhko suggested, would be a fine second choice. 

As it is currently, the markup visible in the note list is so visually janky and hard to parse that I just can't bring myself to use the Markdown in my notes at all. Instead I've searched out a collection of unusual HTML characters that I use for outlining, separating, and emphasizing sections of my notes.

rakhi2104:
I fear a problem, if the title is rendered H1 in few notes in the notes-list and not all, that would look weird.
Instead @giannissc 's hack would be much better.
- One more suggestion would be different titles for each note, so as to differentiate between content of the note and the title itself. :thinking:

PR #996, opened 2018-11-09
Strip markdown in note list previews

ksdme:
Note list previews have raw markdown text which really doesn't help in anyway. This patch drops the markdown from both previews and titles of note list items.

Example:
image

### DIFF
--- a/lib/utils/note-utils.js
+++ b/lib/utils/note-utils.js
@@ -1,3 +1,5 @@
+import removeMarkdown from 'remove-markdown';
+
 /**
  * Matches the title and excerpt in a note's content
  *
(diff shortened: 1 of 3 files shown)


lib/utils/note-utils.js
```diff
@@ -9,29 +11,17 @@
...
 /**
  * Returns the title and excerpt for a given note
  *
  * @param {Object} note a note object
  * @returns {Object} title and excerpt (if available)
  */
 export const noteTitleAndPreview = note => {
-  const content = note && note.data && note.data.content;
+  let content = (note && note.data && note.data.content) || '';
+  content = isMarkdown(note) ? removeMarkdown(content) : content;
```
mirka: What do you think about setting the [`stripListLeaders`](https://www.npmjs.com/package/remove-markdown#usage) option to false?

screen shot 2018-11-09 at 19 01 26

It might actually be confusing if we strip them out.

mirka:
Will close #593

lib/utils/note-utils.js
ksdme: Yeah, I think that is a good idea.

PR #1000, opened 2018-11-09
Make tag entry and removal smoother

mirka:
Adding and removing tags was not a smooth experience because the cursor went away after removing a tag with <kbd>Backspace</kbd>.

This fixes the problem, so that we can smoothly remove and add multiple tags without re-clicking the entry field.

### Before

tags-before

### After

tags-after


roundhill:
And congrats on #1000! 🎉

PR #1022, opened 2018-11-16
Fix password change handling

roundhill:
This is at least a partial fix for the reports we've been getting in #1007 and potentially others.

**The Issue**
We were not signing out the user when their token was no longer valid. If a user changes their password at app.simplenote.com/settings, Simperium creates a new token and invalidates all of the existing ones for the user.

**Another Issue 🤪**
I discovered that we were preserving the IndexedDB content when this invalid token scenario was reached. This is by design so that unsynced notes get preserved, but if a user signed in with a different account than the previous one, all of the content would still be in the app. We now check if the accounts changed when signing in again and clear out the db if they are different.

**To Test**
* Sign in to an account in the app.
* Head to https://app.simplenote.com/settings for the same account, and change the password.
* Make any change in the app, like editing a note. The token should now be invalid and you get signed out.
* Sign back in with the same account, all of the content should still be there.
* Repeat the same above steps, but instead sign in with a different account. You should only see the notes associated with that account.

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -287,6 +287,10 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
       setAuthorized();
       analytics.initialize(accountName);
       this.onLoadPreferences();
+
+      // 'Kick' the app to ensure content is loaded after signing in
+      this.props.actions.loadNotes({ noteBucket: this.props.noteBucket });
+      this.props.actions.loadTags({ tagBucket: this.props.tagBucket });
     };
 
     onNotePrinted = () =>
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -168,8 +168,6 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
         .on('message', this.syncActivityHooks)
         .on('send', this.syncActivityHooks);
 
-      this.onNotesIndex();
-      this.onTagsIndex();
       this.onLoadPreferences(() =>
         // Make sure that tracking starts only after preferences are loaded
         analytics.tracks.recordEvent('application_opened')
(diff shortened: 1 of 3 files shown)


lib/boot.js
```diff
@@ -33,8 +33,8 @@ const cookie = parse(document.cookie);
 const auth = new Auth(config.app_id, config.app_key);
 const appProvider = 'simplenote.com';
 
-const token = cookie.token || localStorage.access_token;
 const appID = config.app_id;
+let token = cookie.token || localStorage.access_token;
```
mirka: By the way do you know if this cookie is only here for legacy reasons? (Can we drop it...?)

lib/boot.js
```diff
@@ -118,13 +120,15 @@ let props = {
...
       .then(user => {
+        resetStorageIfAccountChanged(username);
         if (!user.access_token) {
           return store.dispatch(resetAuth);
         }
 
         store.dispatch(setAccountName(username));
         store.dispatch(setAuthorized());
         localStorage.access_token = user.access_token;
+        token = user.access_token;
```
mirka: It looks like we could probably avoid maintaining this `token` variable at file scope if we directly used `localStorage.access_token` as the single source of truth for our token.

lib/boot.js
roundhill: The cookie is set by app engine when signing in on the web. We could probably have it pass it directly instead of using a cookie? We can probably sort that out in a separate PR.

lib/app.jsx
```diff
@@ -287,6 +287,10 @@ export const App = connect(mapStateToProps, mapDispatchToProps)(
       setAuthorized();
       analytics.initialize(accountName);
       this.onLoadPreferences();
+
+      // 'Kick' the app to ensure content is loaded after signing in
+      this.props.actions.loadNotes({ noteBucket: this.props.noteBucket });
+      this.props.actions.loadTags({ tagBucket: this.props.tagBucket });
```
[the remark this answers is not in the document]
roundhill: Good call, I think we were running those calls twice in a row. I removed those and also made use of the `onIndex` methods in https://github.com/Automattic/simplenote-electron/pull/1022/commits/9d4771835e852b9b86d092de1139dc0f8cc44e45

lib/boot.js
roundhill: Created https://github.com/Automattic/simplenote-electron/issues/1026 to sort that out :)

lib/boot.js
mirka: Oh ok, thanks for clarifying 👍

PR #1025, opened 2018-11-19
Evernote import: Add better support for clipped web content

mirka:
This adds some better handling of Evernote notes clipped from a web browser.

### Enhancements

* use alt text when available
* handle nested media in blank tags
* handle `<a>` tags with missing hrefs
* handle inline svgs

### DIFF
--- a/lib/utils/import/evernote/enml-to-markdown.js
+++ b/lib/utils/import/evernote/enml-to-markdown.js
@@ -1,7 +1,11 @@
 import TurndownService from 'turndown';
-import { get, identity } from 'lodash';
+import { identity, startsWith } from 'lodash';
 
-const mediaPlaceholderFor = node => `(${node.getAttribute('type')})`;
+const mediaPlaceholderFor = node => {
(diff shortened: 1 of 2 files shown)


lib/utils/import/evernote/enml-to-markdown.js
```diff
@@ -1,7 +1,11 @@
 import TurndownService from 'turndown';
-import { get, identity } from 'lodash';
+import { identity, startsWith } from 'lodash';
 
-const mediaPlaceholderFor = node => `(${node.getAttribute('type')})`;
+const mediaPlaceholderFor = node => {
+  const alt = node.getAttribute('alt');
+  const type = node.getAttribute('type') || 'media';
+  return alt ? `${alt}` : `(${type})`;
```
roundhill: How about we return the type and the alt text if it is available, instead of one or the other?

```
(image/png)
A picture of a beautiful flower
```

mirka: You're right!

PR #1042, opened 2018-11-26
Add support for sorting the tags list

roundhill:
Adding a new setting to the app that allows you to sort the tags list alphabetically:

sort-tags

I've added a new indexed db on the tag name, which automatically returns the results sorted by the index. This makes it super fast for us to get the sorted tags list!

It also adds a new `sortTagsAlpha` boolean to the state, which we use in a few places to apply the proper sorting and to disable the 'drag' icon when editing tags when alpha sort is enabled.

**To Test**
* Launch the app and view the tags list. It should be in the custom order.
* Select View -> Tags -> Sort Alphabetically. The tags list should now be sorted alphabetically.
* Edit the tags list, you should not see the drag icon to reorder the list.
* Turn off alphabetical sort, it should go back to the previous custom order. You should be able to Edit the tags list and reorder items again.

### DIFF
--- a/lib/editable-list/index.jsx
+++ b/lib/editable-list/index.jsx
@@ -12,6 +12,7 @@ export class EditableList extends Component {
     editing: PropTypes.bool.isRequired,
     items: PropTypes.array.isRequired,
     renderItem: PropTypes.func.isRequired,
+    sortTagsAlpha: PropTypes.bool.isRequired,
     getItemKey: PropTypes.func,
     onRemove: PropTypes.func,
     onReorder: PropTypes.func,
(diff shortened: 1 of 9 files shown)


lib/editable-list/index.jsx
```diff
@@ -64,7 +65,7 @@ export class EditableList extends Component {
   };
 
   render() {
-    const { editing, onRemove, onReorder } = this.props;
+    const { editing, onRemove, onReorder, sortTagsAlpha } = this.props;
```
mirka: Let's also add a propType check for `sortTagsAlpha`.

PR #1052, opened 2018-11-29
Fix invalid property name on importer Tracks event

### LINKED ISSUE #1049: Update Tracks Import Event Naming Convention
We should use underscores here:

https://github.com/Automattic/simplenote-electron/blob/6789d9c3383f6fc820a75c6bfd0a59cb0c066c27/lib/dialogs/import/source-importer/executor/index.jsx#L75

So `note_count` instead of `noteCount`.

mirka:
Closes #1049 

Related: #1053 to avoid similar mistakes in the future


roundhill:
> but none of my tests seemed to record the event in tracks.. maybe because it isn't a production build

It eventually showed up! Confirmed fixed :shipit:

PR #1059, opened 2018-11-30
Add console warning for wrong database config

mirka:
This adds a console message to make it easier to see which Simperium database the app is configured with. The message will be a warning if the NODE_ENV is production and Simperium is not.

This would not have prevented something like #1058, but hopefully it will make it a little easier to detect problems with CI config changes in the future.


roundhill:
Great idea! Should we actually display something in the UI? A dev mode ribbon or something?

mirka:
> Should we actually display something in the UI? A dev mode ribbon or something?

screen shot 2018-12-03 at 18 20 25

Maybe something subtle in the bottom corner?

Issue #795, opened 2018-08-20
Improve color contrast


mgalbritton:
When in dark mode, the text contrast makes the text almost unreadable.

#### OS version
Windows 10; Simplenote 1.1.6

Reported at #1351437-zen

mirka:
Some improvements we should make:

- Light theme
  - Change the `$blue` from #4895D9 to #287AC3
  - Change `$gray` from #899199 to #6F7780
- Dark theme
  - Change the background from #2D3034 to #282A2E
- Toggle switches: Add cues to that don't rely on color (currently hard to differentiate with red–green color blindness)

codebykat:
There were lots of color updates in https://github.com/Automattic/simplenote-electron/pull/1565 which have hopefully addressed the worst of the contrast concerns, particularly in dark mode. I think it's on the radar to do a more comprehensive contrast audit, but for now I'm closing this issue since it's been almost a year since the last report and quite a bit has changed. Please do re-open if there are specific problem areas still.

PR #1062, opened 2018-11-30
Dark Theme Tweaks

roundhill:
I'd like to propose we change the dark them to match up with the recent changes we've made to the iOS and macOS apps. It essentially desaturates the 'blueness' of the dark theme, and increases the contrast by making the text a bit lighter and the background color darker. The border color is also way more subtle than before.

Before:
screen shot 2018-11-30 at 3 05 53 pm

After:
screen shot 2018-11-30 at 3 09 27 pm

**To Test**
* Enable the dark theme, and explore the app to ensure everything looks good!
* Also ensure that the light theme still appears correctly (It uses the `$gray*` variables which were slightly tweaked in this PR)


scss/_variables.scss
```diff
@@ -6,10 +6,10 @@ $blue-light: lighten($blue, 20%);
 $blue-active-highlight: lighten($blue, 10%);
 
 // Grays
-$gray: #899199;
+$gray: #999999;
```
mirka: `#899199` against a white background wasn't AA-compliant to begin with, so we can't make it even lighter in the Light theme.

I did a [quick contrast check](https://github.com/Automattic/simplenote-electron/issues/795#issuecomment-442735104) the other day, and to meet AA it looks like we won't be able to share the same blue and gray values across both Light/Dark themes.

roundhill: Yeah! I've updated it to `#808080`, which is AA-compliant for the white theme. The same color in the dark theme should now adjust it to be a bit lighter where it is used to `#999999` (#808080 isn't AA-compliant against the #1E1E1E bg)

I also snuck in a fix for the markdown preview with `inline code` syntax, it was changing the font size and looked a bit odd.

mirka: Dark theme — The grays used for `.dialog-tabs-button` and `.settings-group p` text need to be lightened. (Currently at contrast ratio 2.95)

Light theme — I'm not sure where you got `#808080` as AA-compliant for normal text against `#FFF`, since it is at contrast ratio 3.95. But I'm ok with this value for this PR, since it isn't worse than the current state and we need to make other tweaks to the Light theme colors anyway 👍

mirka: Interesting! It guess it's ok for Cocoa apps because macOS has an "Increase contrast" option in Accessibility settings...?

mirka:
> Dark theme — The grays used for `.dialog-tabs-button` and `.settings-group p` text need to be lightened. (Currently at contrast ratio 2.95)

I think we can merge once this is addressed ☝️

roundhill:
> Dark theme — The grays used for `.dialog-tabs-button` and `.settings-group p` text need to be lightened. (Currently at contrast ratio 2.95)

OK I've added some fixes for those in ea7d690

screen shot 2019-01-14 at 12 08 52 pm

scss/theme.scss
```diff
@@ -90,6 +93,18 @@
     }
   }
 
+  .settings-group p {
+    color: lighten($gray, 45%);
+  }
+
+  .dialog-tabs button {
```
mirka:
```suggestion
  .tab-panels__tab-list li {
```

The classes for this have changed in `master` ☝️

roundhill: Thanks, I also needed to update to use the `is-active` class in 36ee908.

randallagordon:
I finally upgraded to a version that includes this change—is it just me, or is this a bit...too contrastical? 🧐

I'm very much a 👍 for meeting AA ratios for folks that need it (as I probably will in a few years!), but I've had to switch back to light mode while I wait for other apps to add dark modes so I can drop my display brightness overall to compensate for how searing the white now is on this black. Any chance for getting a third theme option added? I'd be happy to work on a PR for it!

PR #1064, opened 2018-12-03
Cross-browser CSS fixes

mirka:
CSS adjustments for cross-browser support. Tested in Chrome, Firefox, Safari, IE11, and Edge.

The success criteria for IE11 is "is it usable", rather than "does it look the same as in other browsers".

### Things that will be addressed in a separate PR
- The tag input field in IE11 (Broken due to `contenteditable` problems)
- Mobile browsers

### DIFF
--- a/lib/app-layout/style.scss
+++ b/lib/app-layout/style.scss
@@ -112,6 +112,7 @@
 }
 
 .app-layout__note-column {
+  position: relative;
   display: flex;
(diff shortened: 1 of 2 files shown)


roundhill:
> The success criteria for IE11 is "is it usable", rather than "does it look the same as in other browsers".

Love it ;)

PR #1078, opened 2018-12-06
Fix NoteList excerpt caching

mirka:
This is mostly a code maintenance PR, but with three functional fixes:

1. ca2775d — While the result of `noteTitle()` was being cached for row height calculation, the cached results weren't being used inside `render()`. This was aggravating the issue in #1073.
1. cdb4d60 — `noteTitleAndPreviewCache()` wasn't returning updated data when the cache should have been invalidated. This fix should make row height calculation more reliable.
1. 97cd223 — `noteTitleAndPreviewCache()` was comparing the whole note content to decide whether to invalidate the cache. Changing this to compare modification dates should be faster and use less memory.

### Side note

The `render()` in NoteList is being called twice for every click in the editor (!), which seems very strange and is something we should look into...

### DIFF
--- a/lib/note-list/index.jsx
+++ b/lib/note-list/index.jsx
@@ -485,7 +485,7 @@ const mapStateToProps = ({ appState: state, settings: { noteDisplay } }) => {
    *
    * @type {String} preview excerpt for the current note
    */
-  const selectedNoteTitle =
+  const selectedNotePreview =
     selectedNote && getNoteTitleAndPreview(selectedNote).preview;
 
   return {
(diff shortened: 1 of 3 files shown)


roundhill:
> The `render()` in NoteList is being called twice for every click in the editor (!), which seems very strange and is something we should look into...

😱 https://github.com/Automattic/simplenote-electron/issues/1079

lib/note-list/index.jsx
```diff
@@ -515,7 +485,8 @@ const mapStateToProps = ({ appState: state, settings: { noteDisplay } }) => {
    *
    * @type {String} preview excerpt for the current note
    */
-  const selectedNoteTitle = selectedNote && getNoteTitle(selectedNote);
+  const selectedNoteTitle =
+    selectedNote && getNoteTitleAndPreview(selectedNote).preview;
```
roundhill: Should this const be called `selectedNoteTitleAndPreview`? It's a little confusing that it is getting the `.preview` value but is called `selectedNoteTitle`.

mirka: Good catch, thanks!

mirka:
> I wonder how important all of this caching is 🤔. I know we put it in for performance reasons previously, though.

I'll learn how to better profile this stuff!

PR #1081, opened 2018-12-07
Add cross-browser support for TagInput

mirka:
Depends on #1055 

This adds cross-browser support for the TagInput component. It works around inconsistencies in `contenteditable` elements. It also changes the `Add a tag…` placeholder implementation from a pseudo-element to an actual element, for better consistency (especially in IE11).

Tested in Chrome, Firefox, Safari, IE11, and Edge.

### DIFF
--- a/lib/tag-input/style.scss
+++ b/lib/tag-input/style.scss
@@ -23,7 +23,7 @@
 }
 
 .tag-input__entry {
-  min-width: 1px;
+  min-width: 1px; // Needed for the caret to be visible in Safari
 
   & b,
   & strong {


mirka:
@roundhill Rebased onto current master so you can test it with the polyfill 👍

lib/tag-input/style.scss
```diff
@@ -15,12 +15,15 @@
...
+  background: transparent;
+}
+
 .tag-input__entry {
-  &:empty:before {
-    content: attr(placeholder);
-    display: block;
-    color: lighten($gray, 10%);
-  }
+  min-width: 1px;
```
roundhill: What's this for? 🤔

PR #1082, opened 2018-12-07
Remove cut/copy handlers on NoteDetail

mirka:
tl;dr — This removes the cut/copy handlers on NoteContentEditor because a) it doesn't work in IE11, and b) it isn't doing anything necessary (from what I can tell).

### How I came to this conclusion

1. I needed to make the `copy` event handler work with IE. (This handler "replaces `\n` with `\r\n` to keep line breaks on Windows")
1. But it turns out that, due to the way `window.clipboardData.getData()` works in IE, if we get the clipboard data through this function we can't preserve the line endings without adding a bunch of hacky code in the note editor.
1. This is when I noticed that the `\n` to `\r\n` replacement doesn't seem to be necessary at all. Copy & paste is working on Windows without that replacement function. (Tested by copying note text from IE11/Edge/Chrome and pasting into Notepad++ with line endings visible. They are all CRLF.)

### To test

Check that copying NoteDetail text (to Notepad or whatever) on Windows works as expected. 

(Caveat: Doing this in IE11 results in two line breaks for each line break. There is no easy fix for this, because of point number 2 in the section above. I'm inclined to say `wontfix` on this one, but maybe we'll think about it if we get enough complaints...)


roundhill:
Waiting for AppVeyor to rebuild this one 🕥

PR #1084, opened 2018-12-07
Fix styles for many tags

### LINKED ISSUE #1017: Tags: can't use too many tags without breaking/being unable to edit them anymore
On the Linux app (1.2.1), adding a large number of tags will stretch the screen, pushing the view, extra info, share and the all the buttons to the side. Tags stretch off the screen and then it's not possible to add new ones.

Once it happens, according to the user:

> You can delete a tag from the list (by clicking a tag) which will switch the focus to tags menu which allows you to see the tags and add more tags as well as see the menu

> The issue is now there is no way that I have found focus back onto the text and type notes easily. I resolve this by restarting the app. 

An idea would be to increase the tags area height, which is similar to the report on Automattic/simplenote-macos#57 .

#### What I expected

captura de tela 2018-11-14 as 16 38 08

#### What happened instead
captura de tela 2018-11-14 as 16 38 22

User also suggested that:

> If we could hotkey switch focus on the tags and text like on the mac app this could be a workaround?

Which seems similar to #970 

***

#### Simplenote version
1.2.1

#### OS version
Linux

***
user-report 1552536-zen

mirka:
Fixes #1017 

### Minor tweaks (in addition to the #1017 fix)

Adjusts spacing so the tags look nicer when on multiple rows:

screen shot 2018-12-08 at 7 12 02

Sets the max height to about 2.5 rows (so the user can see there is more), and scrolls the rest:

screen shot 2018-12-08 at 7 11 16

### DIFF
--- a/lib/tag-field/style.scss
+++ b/lib/tag-field/style.scss
@@ -10,7 +10,7 @@
   line-height: 1.75em;
   white-space: nowrap;
   overflow: auto;
-  max-height: 78px; // about 2.5 rows
+  max-height: calc(2.5 * 1.75em + 16px); // about 2.5 rows
   padding: 8px 12px;
 
   &:focus {
--- a/lib/tag-field/style.scss
+++ b/lib/tag-field/style.scss
@@ -9,9 +9,9 @@
   flex: 1 1 auto;
   line-height: 1.75em;
   white-space: nowrap;
-  overflow: visible;
-  max-height: 120px;
-	padding: 8px 12px;
+  overflow: auto;
+  max-height: 78px; // about 2.5 rows
+  padding: 8px 12px;
 
   &:focus {
     outline: none;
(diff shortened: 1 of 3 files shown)


lib/tag-field/style.scss
```diff
@@ -9,9 +9,9 @@
   flex: 1 1 auto;
   line-height: 1.75em;
   white-space: nowrap;
-  overflow: visible;
-  max-height: 120px;
-	padding: 8px 12px;
+  overflow: auto;
+  max-height: 78px; // about 2.5 rows
```
roundhill: Not sure if this is better, but you could do something like `calc(2.5 * 1.75em + 16px);` to get the exact 'two and a half lines' value. (`1.75em` is the line height and `16px` of top and bottom padding)

mirka: It'll also be better for when we `em/rem`-ify more `font-size` values for accessibility!

Issue #1095, opened 2018-12-12
Note Sync Nudge Could Cause Constant Websocket Reconnect


roundhill:
I'm still seeing a higher amount of websocket `init` traffic from this app. I think we may have an issue with the `ActivityHooks` code repeatedly reconnecting because it thinks it has an unsynced note.

I can cause many repeated reconnects by adding `v = 0;` [here](https://github.com/Automattic/simplenote-electron/blob/master/lib/utils/sync/nudge-unsynced.js#L21-L26), but I'm not certain this is exactly what is causing this. Still, the way that we connect the client [here](https://github.com/Automattic/simplenote-electron/blob/master/lib/utils/sync/nudge-unsynced.js#L42), and it was added in 1.3.0 when this started occurring, makes me believe that something like this is happening for some users.

Perhaps we should get rid of `ActivityHooks` for now, and just reconnect the websocket during the import process only so we still solve the issue of unsynced notes for large imports?

PR #1098, opened 2018-12-14
Persist Simperium LocalQueue to localStorage

mirka:
This fixes a problem with sync, where offline changes wouldn't get sent to the server at all if you quit the app before going back online.

### Steps to reproduce

1. Start with an empty account.
2. Log in to the account, and go offline.
3. Add a new note.
4. Restart the app.
5. Go online again and wait for the syncing to resume.

The `c` command for the note will not be sent, even when you’re back online.

### Cause

Offline changes were being queued in a `LocalQueue`, but these did not persist across reloads. As a result, these changes would be lost.

### Fix

On the `beforeunload` event, persist the local queue of each bucket to localStorage. Then, restore these queues when the BrowserClient is instantiated.

`localQueueStore` is pretty tightly coupled to the internals of `node-simperium` 😅 I would eventually want `node-simperium` to be enhanced in way that I can just pass it a store provider, like with ghostStore and objectStore.

### In relation to the reconnect loop problem

This was likely the cause of #1095, where an unsynced new note (`v === 0`) stays in limbo forever. While this PR will prevent new occurrences of the reconnect loop, it doesn't address our users' unsynced notes that are already stuck in limbo. So in a separate PR, I'll see what I can do to actually sync over these stuck `v === 0` notes.

### DIFF
--- a/lib/simperium/index.js
+++ b/lib/simperium/index.js
@@ -9,8 +9,23 @@ const Bucket = Client.Bucket;
 export const Auth = simperium.Auth;
 
 export default function(settings) {
-  return new BrowserClient(settings);
+  const browserClient = new BrowserClient(settings);
+  Object.values(browserClient.buckets).forEach(bucket => {
+    bucket.persistLocalQueue();
+  });
+  return browserClient;
 }
+
+Bucket.prototype.persistLocalQueue = function() {
+  const bucketName = this.name;
+  const { queues, sent } = this.channel.localQueue;
+  window.addEventListener('beforeunload', function() {
+    window.localStorage.setItem(
+      `localQueue:${bucketName}`,
+      JSON.stringify({ queues, sent })
+    );
+  });
+};
 
 Bucket.prototype.query = function(fn) {
   this.store.setup.then(fn);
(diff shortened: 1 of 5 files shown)


lib/simperium/index.js
```diff
@@ -56,8 +63,10 @@ function BrowserClient({ appID, token, bucketConfig, database, version }) {
     }.bind(this)
   );
 
-  for (let bucket in bucketConfig) {
-    this.buckets[bucket] = this.client.bucket(bucket);
+  for (let bucketName in bucketConfig) {
+    const bucket = this.client.bucket(bucketName);
+    localQueueStore.restoreTo(bucket);
+    this.buckets[bucketName] = bucket;
```
roundhill: Should we only worry about persisting the `tag` and `note` buckets? Probably doesn't hurt anything to do the preferences one too.

mirka:
> I'm thinking we might want to make sure the localStorage data is deleted when the user signs out or becomes unauthorized?

I was going to add logic for this, but turns out we don't need any because:

1. The local queues are persisted when the user closes/navigates away, so only empty local queues will be persisted if a user is logged out.
2. The local queue data is removed from localStorage immediately after it's been read into memory.

lib/simperium/index.js
mirka: Yes, I have trouble imagining a bucket where I wouldn't want offline changes to be synced, so I made this the default behavior for all buckets.

roundhill:
> I was going to add logic for this, but turns out we don't need any because:
> 
> 1. The local queues are persisted when the user closes/navigates away, so only empty local queues will be persisted if a user is logged out.
> 2. The local queue data is removed from localStorage immediately after it's been read into memory.

I see! :shipit:

PR #1101, opened 2018-12-17
Add rate limiter to imports

mirka:
This adds rate limiting to the CoreImporter so that we don't have to rely on the hacky workaround in `nudgeUnsynced` to get large imports unstuck:

https://github.com/Automattic/simplenote-electron/blob/0b079b92077b8caa0f38f400303c1aec01b2d428/lib/utils/sync/nudge-unsynced.js#L42

With my new familiarity with the LocalQueue 😌, I was able to realize that the stuck import notes were not stuck locally, nor were they being "ignored" by the server. They were sent out correctly, but then each of them were being returned 503 errors from the Simperium server.

@roundhill:
- Should `node-simperium` be patched to handle these errors?
- Should there be an official rate limit for the Simperium API? (In this app, I set the limit to 250 requests/sec. It is a little conservative, but I think it's fine because according to Tracks, imports larger than 100 notes are very rare.)

### To test

1. Set `localStorage.debug="sync:*,analytics"` and reload.
1. Import a file with >500 notes.

The syncing should complete, and `nudgeUnsynced` shouldn't log anything other than `0 unsynced notes`. (Or, you can try commenting out the `client.client.connect()` line quoted above.)

Import file (1000 notes) for your convenience: [few-tags-1000-800kb.json.zip](https://github.com/Automattic/simplenote-electron/files/2686889/few-tags-1000-800kb.json.zip)

### Next steps

Now that this is out of the way, I will modify the `nudgeUnsynced` function so that it specifically addresses the `v === 0` notes that resulted from unpersisted offline changes (#1098). We'll get rid of the `client.connect()` call there, thus fixing the `init` flood once and for all 🔨

### DIFF
--- a/lib/utils/import/index.js
+++ b/lib/utils/import/index.js
@@ -12,7 +12,7 @@ const propertyWhitelist = [
   'tags',
 ];
 
-const MAX_REQUESTS_PER_SEC = 250;
+const MAX_REQUESTS_PER_SEC = 50;
 
 class CoreImporter extends EventEmitter {
   constructor({ noteBucket, tagBucket }) {
--- a/lib/utils/import/index.js
+++ b/lib/utils/import/index.js
@@ -1,3 +1,4 @@
+import Bottleneck from 'bottleneck';
 import { EventEmitter } from 'events';
 import { isEmpty, get, pick } from 'lodash';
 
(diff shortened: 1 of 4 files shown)


roundhill:
As discussed in Slack, let's set `MAX_REQUESTS_PER_SEC` to `50`.

roundhill:
Filed https://github.com/Simperium/node-simperium/issues/70 for our friend `node-simperium`.

PR #1103, opened 2018-12-18
Find and re-push stuck version 0 notes

### LINKED ISSUE #1095: Note Sync Nudge Could Cause Constant Websocket Reconnect

mirka:
Closes #1095 

As mentioned in #1101, this addresses the version 0 notes that resulted from unpersisted offline changes (#1098). This PR branch doesn't include the localQueue persistence commit, so that it is easier to produce a stuck version 0 note for testing purposes.

### To test

1. Start with an empty account.
2. Log in to the account, and go offline.
3. Add a new note.
4. Restart the app.
5. Go online again and wait for the syncing to resume.

After the restart in step 4, the new note is at version 0 with no change in the local queue, but the nudgeUnsynced function will push it over to the server. It should not trigger an infinite loop of the sync activity hooks.

### DIFF
--- a/lib/utils/sync/nudge-unsynced.js
+++ b/lib/utils/sync/nudge-unsynced.js
@@ -1,10 +1,16 @@
 /**
- * Ensure that all imported notes are sent to the server
+ * Ensure that all version 0 notes are sent to the server.
  *
- * Importing over 350-400 notes at once can cause the syncing to get
- * stuck. This function checks if there are any local notes which have not
- * been acknowledged by the server yet, and if so, reconnects the client
- * to resume the syncing.
+ * Version 0 notes (new notes, usually created offline, that have not yet been
+ * acknowledged by the server) could get stuck in that state, probably due to the
+ * fact that the local queues were not persisted. If the app was quit
+ * while offline, those offline changes could be lost and never sent to the
+ * server unless another change is made to that note while online.
+ * See https://github.com/Automattic/simplenote-electron/pull/1098
+ *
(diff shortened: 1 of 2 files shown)


lib/utils/sync/nudge-unsynced.js
```diff
@@ -28,19 +43,10 @@ const nudgeUnsynced = ({ noteBucket, notes, client }) => {
...
 
-    if (hasUnsyncedNotes && debug.enabled) {
-      unsyncedNoteCount = result.filter(note => note.unsynced).length;
-    }
-
-    debug(`${unsyncedNoteCount} unsynced notes`);
-
-    // Re-init the client and prod the syncing to resume
-    if (hasUnsyncedNotes) {
-      client.client.connect();
```
roundhill: 😌

Issue #491, opened 2017-01-18
Tag deletion


drw158:
#389 introduces new behavior for tag deletion.

Before, you could not delete a tag by clicking, you had to use the delete key.

Now, just a single click deletes the tag with no confirmation. The hover state also does not indicate deletion.

drw158:
Here's where the discussion stopped

> in the meantime I will remove the confirmation step so that a single click removes the tags. we can play with it and see how it feels

> @drw158 one issue with not selecting the tags before deleting them is that it currently mirrors the keyboard method for deleting tags

> have the cursor focused in the tag entry with existing tags
backspace to clear out the entry
backspace again and it should select the last tag
backspace again and it confirms and deletes the tag
should we maintain selection for the keyboard but remove it for the mouse?

dmsnell:
@drw158 is there a question or change request here?

drw158:
I don't know the solution, but this is the issue that should be solved:

> just a single click deletes the tag with no confirmation. The hover state also does not indicate deletion.

mirka:
Action item:

The tags should show a small, distinct close button when hovered. Perhaps like the [deletable chips in Material](https://material-ui.com/demos/chips/#chip-array). Only that button area should trigger the deletion.

PR #1106, opened 2018-12-19
Change nav bar icon from Tag to Hamburger Menu

mirka:
This changes the menu icon in the top left from a Tag icon to a Hamburger menu icon.

screen shot 2018-12-20 at 0 22 42

This is in preparation for deploying this as a web app, since we will need to put the Settings button in there.

### Side note

The trade-off for improving discoverability of the Settings button (and Trash!) is that we may be reducing discoverability of the "filter notes by tag" feature. I think we can alleviate this by adding a feature to filter notes by clicking on a note tag in the editor. The current onclick behavior (removes the tag) is problematic anyway #491, so we can maybe add an `x` badge on hover for a separate remove function.

### DIFF
--- a/lib/icons/menu.jsx
+++ b/lib/icons/menu.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+
+export default function MenuIcon() {
+  return (
+    <svg
+      className="icon-menu"
+      xmlns="http://www.w3.org/2000/svg"
+      width="24"
+      height="24"
+      viewBox="0 0 24 24"
+    >
+      <path d="M21 6v2H3V6h18zM3 18h18v-2H3v2zm0-5h18v-2H3v2z" />
+    </svg>
+  );
+}
--- a/lib/search-bar/index.jsx
+++ b/lib/search-bar/index.jsx
@@ -12,7 +12,7 @@ import { tracks } from '../analytics';
 import IconButton from '../icon-button';
 import NewNoteIcon from '../icons/new-note';
 import SearchField from '../search-field';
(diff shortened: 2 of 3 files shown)


roundhill:
It'd be nice if the "top bun" and "bottom bun" aligned with the top and bottom of the other icons in the toolbar. If not possible, maybe just push it down a few more pixels so that it is centered vertically?

screen shot 2018-12-19 at 8 55 05 am

roundhill:
> It'd be nice if the "top bun" and "bottom bun" aligned with the top and bottom of the other icons in the toolbar. If not possible, maybe just push it down a few more pixels so that it is centered vertically?

Noticed that it is perfectly aligned with the search box though 🤔. Maybe we need to adjust the position of the search box as well..

mirka:
Inspecting the alignment graphically, it doesn't look like anything is really off center, except for the New Note icon which seems to be optically adjusted to account for the protruding plus sign. I played around with it a bit, but didn't really find a positioning that looked better overall. Let me know if you have an adjustment that looks better to your eye! 🙂

roundhill:
> except for the New Note icon which seems to be optically adjusted to account for the protruding plus sign.

I think that was the issue. I found that bumping down the search field by one pixel aligned everything nicely:

screen shot 2018-12-19 at 11 36 53 am

I pushed the fix, let me know if you like it.

mirka:
Works for me! Thank you

roundhill:
Mutual :shipit: achieved!

PR #1111, opened 2018-12-20
Add Tools panel to Settings dialog

mirka:
Based on #1109 

This makes the Import/Export tools reachable from a non-Electron environment.

screen shot 2018-12-21 at 1 19 08


lib/dialogs/settings/index.jsx
```diff
@@ -141,6 +142,9 @@ export class SettingsDialog extends Component {
         <div className="dialog-column">
           <DisplayPanel buckets={buckets} />
         </div>
+        <div className="dialog-column">
```
roundhill: Should we move these `dialog-column` divs to within the panel components? 🤔

lib/utils/export/index.js
```diff
@@ -4,7 +4,9 @@ import { get } from 'lodash';
 import exportNotes from './export-notes';
 import exportToZip from './to-zip';
 
-const exportZipArchive = filename => {
+const filename = 'notes.zip';
```
roundhill: Just for fun, what if we called this `Simplenotes.zip`?

lib/dialogs/settings/index.jsx
mirka: True, the CSS is a bit messy! I refactored the TabPanels component so the CSS isn't all over the place.

lib/utils/export/index.js
mirka: Maybe I'm stuffy and overprotective of the Simplenote® brand, but:

> NEVER modify a trademark by changing it to the plural form.

http://www.inta.org/trademarkbasics/factsheets/pages/trademarkusefactsheet.aspx

😂😂😂

PR #1132, opened 2018-12-27
Make text in Markdown preview selectable

### LINKED ISSUE #744: Ctrl+A Does Now Work In Preview Mode
The user reported that they can not select their text with Ctrl+A on Windows app version 1.1.3 in the Preview mode. 

I have tested this both on Windows and Mac apps - the Mac app allows you to select text in both Preview and Edit modes with Cmnd+A; the Windows app allows you to select text with Ctrl+A only in the edit mode. 
#### Steps to reproduce
1. Create a note
2. Enable Markdown on a note
3. Switch to Preview mode
4. Click on Ctrl+A

#### What I expected

To be able to select the text in my note
#### What happened instead

The text was not selected

#### OS version
 Windows 10; Simplenote 1.1.3

The issue was reported by the user in #1061119-zen

mirka:
Partial fix for #744 

_2019-01-07 update:_

While this makes the preview text selectable by mouse, and by "Select All" in a plain web browser, "Select All" in Electron will still not work.

"Select All" not working here is an issue with Electron, and will be fixed when we update to Electron >3.0.0.

### DIFF
--- a/lib/note-detail/style.scss
+++ b/lib/note-detail/style.scss
@@ -139,6 +139,10 @@
     border-top: 1px solid lighten($gray, 20%);
   }
 
+  a {
+    cursor: pointer;
+  }
+
   blockquote {
     font-style: italic;
     border-left: 4px solid lighten($gray, 20%);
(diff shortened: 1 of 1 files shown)


mirka:
@roundhill Thanks for the catch!

I confirmed this is an Electron problem that will fix itself in >3.0.0. I will keep the issue open until then.

PR #1142, opened 2019-01-09
[WIP] Fix strange behavior after a closing modal restores focus on the Editor

### LINKED ISSUE #1134: Opening 'Preferences' leads to strange cursor behaviour, unwanted and strange duplicating of typed text
#### Steps to reproduce
1. Create a new note, activate Markdown for this note
2. Create a # headline further down a ## headline
3. Put in three (or fewer or more) bullet points 
4. Go with the cursor to the end of the line of your last bullet point
5. Open File > Preferences, click on 'Display' and then click back on 'Account', close the preferences dialogue
6. When closing Preferences, the cursor is automatically being placed at the top at the very left in the active note
7. Now simply write something, the text being typed appears in the first line, where the cursor was placed automatically as well as at the beginning in the line where the cursor was before the Preferences tab was opened. Typing is being duplicated but Backspace only affects the line where the cursor actually is (Line 1 in this example)
8. When switching to another programme, e.g. FireFox, and then back to Simplenote, the false behavouir rewinds to normal functionality but first one must place the cursor somewhere else in the note.

#### What I expected
Normal behaviour, alterting the text where the cursor is

#### What happened instead
Copying the text of the active line somewhere else in the note and duplicating it as it is being typed

#### Simplenote version

Simplenote 
simplenote bug
1.3.4 

#### OS version
Windows 10, all updates installed

mirka:
Closes #1134 

**Yet unaddressed problem**: After programmatically restoring focus, the Draft.js editor somehow stops updating its SelectionState. Further clicks in the Editor do not emit change events. The visible caret and the internal SelectionState remains out of sync until a new EditorState is created. This may be a Draft.js bug.

If there is no workaround, we may have to either:

a. Disable the `shouldReturnFocusAfterClose` on `react-modal` Modals 😬
b. Reimplement the `shouldReturnFocusAfterClose` behavior so it works with Draft.js 😬😬

### DIFF
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { ContentState, Editor, EditorState, Modifier } from 'draft-js';
-import { get, includes, invoke, noop } from 'lodash';
+import { get, includes, invoke, noop, once } from 'lodash';
 
 import { filterHasText, searchPattern } from './utils/filter-notes';
 import matchingTextDecorator from './editor/matching-text-decorator';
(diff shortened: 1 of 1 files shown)


belcherj:
Clicking into the editor after modal close appears to be firing events.  Closing for now.

PR #1144, opened 2019-01-09
Improve alphabetical note sorting

### LINKED ISSUE #1050: Re-order Accented Characters Alphabetically
Presently, accented characters are listed after Z when you sort alphabetically, and when you search for keywords with accented characters, the app doesn't find the notes that contain those keywords written with non-accented characters (and vice versa).

Would be great to have better search functionality for accented characters. 

1598900-zen

### LINKED ISSUE #1136: Markdown heading notes are sorted before plaintext notes
#### Steps to reproduce
1. Create a few plaintext and Markdown notes (with a heading on the first line).
2. Sort the notes alphabetically.

#### What I expected

I expect that, since the note titles have Markdown removed, they would be sorted by displayed title.

#### What happened instead

All Markdown notes (with a Markdown title) are sorted *first* followed by all other notes.

#### Simplenote version

1.3.4 (Electron)

#### OS version

macOS 10.14.2

#### Screenshot / Video

screen shot 2019-01-05 at 7 14 25 am

mirka:
Closes #1050 
Closes #1136 

When notes are sorted alphabetically in the Note List, it will now ignore accents/diacritics and the `#` characters used for Markdown headings.

### TODO

- [x] https://github.com/Automattic/simplenote-electron/pull/1144/commits/7737bea24c57f9e7ef98758099fb1fa3ec9a13fa Make it [work in IE11](https://github.com/Automattic/simplenote-electron/pull/1144#discussion_r246902765)

### To test

Try sorting notes with titles like `# asdf` and `àpres`.

### DIFF
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -81,7 +81,7 @@ const client = initClient({
     },
   },
   database: 'simplenote',
-  version: 41,
+  version: 42,
 });
 
 const l = msg => {
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -56,6 +56,7 @@ const client = initClient({
         return {
           ...note,
           contentKey: content
+            .replace(/^\s*#+\s*/, '') // Remove leading whitespace and Markdown heading
             .replace(/\s+/g, ' ')
             .trim()
             .slice(0, 200)
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -60,7 +60,10 @@ const client = initClient({
             .replace(/\s+/g, ' ')
             .trim()
             .slice(0, 200)
-            .toLowerCase(),
+            .toLowerCase()
+            // Remove accents/diacritics (https://stackoverflow.com/questions/990904)
+            .normalize('NFD')
+            .replace(/[\u0300-\u036f]/g, ''),
           pinned: systemTags.indexOf('pinned') !== -1,
         };
       },
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -28,6 +28,7 @@ import { content as welcomeMessage } from './welcome-message';
 
 import appState from './flux/app-state';
 import isDevConfig from './utils/is-dev-config';
+import { normalizeForSorting } from './utils/note-utils';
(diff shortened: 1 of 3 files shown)


lib/boot.js
```diff
@@ -56,10 +56,14 @@ const client = initClient({
         return {
           ...note,
           contentKey: content
+            .replace(/^\s*#+\s*/, '') // Remove leading whitespace and Markdown heading
```
roundhill: Should we include more Markdown characters here like `*`, `-`, `_` etc?

lib/boot.js
```diff
@@ -56,10 +56,14 @@ const client = initClient({
...
             .replace(/\s+/g, ' ')
             .trim()
             .slice(0, 200)
-            .toLowerCase(),
+            .toLowerCase()
+            // Remove accents/diacritics (https://stackoverflow.com/questions/990904)
+            .normalize('NFD')
```
roundhill: **TIL**

mirka: Ugh, I just realized that this won't work in IE11 (because of course it doesn't). I'll add some conditional logic to skip this part, since the shim is huge and I don't want to include it just for this.

lib/boot.js
mirka: I think this is fine for now, since we're only stripping the `#` chars for the [Note List titles](https://github.com/Automattic/simplenote-electron/blob/a5cc00443c0f89719912ab69ff8c9334e8da01be/lib/utils/note-utils.js#L15), and the sorting should reflect the text that's displayed in the Note List. (We can consider doing more aggressive Markdown stripping in both the index and Note List titles if we get a lot of requests!)

lib/boot.js
mirka: @roundhill I [fixed this part](https://github.com/Automattic/simplenote-electron/pull/1144/commits/7737bea24c57f9e7ef98758099fb1fa3ec9a13fa#diff-a597bcf2c27edf0291e524f8897ad4f1R94) so it won't choke in IE11.

Issue #1138, opened 2019-01-07
Add menu item to insert checklist


mirka:
Format ▸ Insert Checklist <kbd>⌘⌥C</kbd>

- Should insert as well as remove checkboxes to the current line.
- If multiple lines are selected, it should add/remove checkboxes to all of them.

PR #1145, opened 2019-01-10
Render clickable checkboxes in Editor

### LINKED ISSUE #1139: New lines after a checklist item should automatically be a checklist item
A new line following a checklist item should be auto-formatted to be a checklist item:

```
- [ ] task 1
- [ ] 
```

mirka:
Closes #1139 

checkbox-decorator

### TODO
- [x] handle backspacing #1154
- [x] handle copy event #1155 

### Next steps
#1138 #1157

### DIFF
--- a/lib/note-detail/toggle-task/constants.js
+++ b/lib/note-detail/toggle-task/constants.js
@@ -3,4 +3,4 @@ export const selectors = {
   markdownRoot: '[data-markdown-root]',
 };
 
-export const taskRegex = /^(\s+)?-[ \t]+\[[xX\s]?\](.+)/gm;
+export const taskRegex = /^(\s+)?(-[ \t]+\[[xX\s]?\])(.+)/gm;
--- a/lib/components/checkbox/index.jsx
+++ b/lib/components/checkbox/index.jsx
@@ -5,18 +5,19 @@ import CheckmarkIcon from '../../icons/checkmark';
 import CircleIcon from '../../icons/circle';
 
 const Checkbox = ({ checked = false, onChange }) => {
+  // A custom checkbox with an ARIA role is used here to work around a bug in
+  // DraftJS, where using a hidden <input type="checkbox"> will trigger a error.
   return (
-    <div className="checkbox">
-      <div className="checkbox__icon" aria-hidden="true">
+    <span
+      className="checkbox"
+      role="checkbox"
+      aria-checked={checked}
+      onClick={onChange}
+    >
+      <span className="checkbox__icon" aria-hidden="true">
         {checked ? <CheckmarkIcon /> : <CircleIcon />}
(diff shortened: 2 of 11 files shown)


roundhill:
The cursor jumps next to the checkbox when clicked, would it be possible to preserve the current cursor position?

cursor-position

roundhill:
At very large font sizes, the checkboxes start to look off-center:

screen shot 2019-01-15 at 9 31 53 pm

lib/note-detail/toggle-task/constants.js
```diff
@@ -3,4 +3,4 @@ export const selectors = {
   markdownRoot: '[data-markdown-root]',
 };
 
-export const taskRegex = /^(\s+)?-[ \t]+\[[xX\s]?\](.+)/gm;
+export const taskRegex = /^(\s+)?(-[ \t]+\[[xX\s]?\])(.+)/gm;
```
roundhill: I don't think we want the whitespace matching in this regex, I'm able to add spaces and then add a checkbox:

spaces-checkbox

mirka: This is because Markdown lists can be nested:

```
- [ ] parent item
  - [ ] sub item
```

- [ ] parent item
  - [ ] sub item

roundhill: Oh, ok. The mobile apps enforce them to be at the start of the line. Should we do that here for consistency?

roundhill:
> This will make the caret go away instead of preserving the position, but HTML-wise this is probably more correct. Do you think that’s ok?

Sounds good.

lib/note-detail/toggle-task/constants.js
mirka: Functionally, I don't think it's unreasonable to enforce that tasklists be flat. Rather, I'm more concerned about straying too much from the GFM spec (or common implementations), as it will hurt our compatibility within the wider Markdown ecosystem. Users do import/export and copy/paste between Markdown editors. Not to mention internal consistency, with the Markdown renderers we're using for previews.

Disregarding code blocks (when rendering checkboxes) was a trade-off we had to make for feasibility reasons, but that’s not an issue here with the leading whitespace. I'm leaning toward consistency with common Markdown implementations, as long as it's feasible. Are there any extra considerations on the mobile side?

roundhill: Let me investigate how hard it'd be to support it on the mobile apps.

roundhill: OK it should be possible 👍I'll create PRs for the other apps to support the nested checklists.

mirka: Awesome. Thank you!

Issue #1116, opened 2018-12-21
Left border should be hidden in focus mode


roundhill:
When you're in focus mode, there's a border to the left of the app:

<img width="103" alt="screen shot 2018-12-21 at 3 16 33 pm" src="https://user-images.githubusercontent.com/789137/50367221-90a95c00-0533-11e9-92d4-74e501da8f2c.png">

We should hide that border while focus mode is active.

PR #1149, opened 2019-01-11
Remove left border in Focus Mode

qualitymanifest:
Fixes https://github.com/Automattic/simplenote-electron/issues/1116

I Included a transition delay so that the border remains while the note list is being "pushed" out of the way, I'm assuming this is the desired behavior.


mirka:
(Side note: Looks like the border management in the CSS could use some refactoring! I posted a separate issue for that — feel free to tackle 🙂 #1150 )

Issue #1150, opened 2019-01-11
CSS: Move the left borders of the note-column elements to an app-layout column


mirka:
The border separating the source-column and note-column are currently set as a `border-left` on the individual child elements:

- `note-toolbar-wrapper`
- `note-detail-wrapper`
- `tag-field`

This should be handled at the app-layout level. It could either be a border-right on the source-column, or a border-left on the note-column. (Not immediately sure which is simpler. Focus Mode and mobile screen widths should be taken into account.)

PR #1152, opened 2019-01-11
Refactor border between Source and Note columns

qualitymanifest:
Closes https://github.com/Automattic/simplenote-electron/issues/1150

In SCSS:
	- Added `border-left` to `.app-layout__note-column` in app-layout
	- Removed `border-left` from `.tag-field` in tag-field, `.note-detail-wrapper` in note-detail, and `.note-toolbar-wrapper` in note-toolbar

In JSX:
	- Added `.theme-color-border` to `.app-layout__note-column` div in app-layout
	- Removed `.theme-color-border` from tag-field and note-detail
	- Left `.theme-color-border` in note-toolbar due to `border-bottom`

Tested on Linux using various screen widths in Focus Mode and regular mode, both in light and dark themes.

### DIFF
--- a/lib/app-layout/index.js
+++ b/lib/app-layout/index.js
@@ -57,7 +57,7 @@ export const AppLayout = ({
             onNoteOpened={onNoteOpened}
           />
         </div>
-        <div className="app-layout__note-column theme-color-bg theme-color-fg">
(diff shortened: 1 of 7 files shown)


qualitymanifest:
Sounds good. I made the change and tested it out, works as expected.

PR #1155, opened 2019-01-14
Copy text from DraftJS EditorState instead of browser selection

### LINKED ISSUE #1129: [Windows] Copying notes from Simplenote 1.3.4 to Blogger will show wide line spacing
#### Steps to reproduce
1. Create a note in Simplenote 1.3.4
2. Paste it in a Blogger post and Preview it.
3. It will show something like this: https://monosnap.com/file/qdYSJMXnpmJ8On5StQONkWqKIpwynd

#### What I expected
To have it like this:
https://monosnap.com/file/AFWIes5psuU5tc28RnwJlYmA8w2kRG

#### What happened instead
The third point above.

**Note: This doesn't happen with Simplenote 1.3.3**

#### Simplenote version
1.3.4

#### OS version
Windows 10

Ref: #1668013-zen

mirka:
Based on #1154 
Closes #1129

This enhances the cut/copy behavior in the Note Editor, by retrieving the selection from the DraftJS state rather than the browser selection. This allows us to:

- Copy the raw Markdown text, even if they are decorated (e.g. `- [ ] task`) in the browser.
- Ensure that plain text (not HTML) is being copied. (Fixes #1129)
- Retain IE11 compatibility. (Related: #1082)

### To test

- Copying text in Dark Mode and pasting it into a rich text editor should not make the text white.
- Tasklist items can be copied as Markdown text. (Caveat: Due to a limitation with the contenteditable, checkboxes cannot be selected unless the selection range starts on the previous line or before.)

### DIFF
--- a/lib/editor/utils.js
+++ b/lib/editor/utils.js
@@ -6,3 +6,36 @@ export function getCurrentBlock(editorState) {
   const key = editorState.getSelection().getFocusKey();
   return editorState.getCurrentContent().getBlockForKey(key);
 }
+
+export function getSelectedText(editorState) {
+  const contentState = editorState.getCurrentContent();
+  const selectionState = editorState.getSelection();
+
+  const startKey = selectionState.getStartKey();
+  const startBlock = editorState.getCurrentContent().getBlockForKey(startKey);
+  const start = selectionState.getStartOffset();
+  const endKey = selectionState.getEndKey();
+  const end = selectionState.getEndOffset();
+
+  if (startKey === endKey) {
+    return startBlock.getText().slice(start, end);
+  }
+
+  let selectedText = '';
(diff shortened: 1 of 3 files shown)


mirka:
Thanks for all the reviews @roundhill !

> I noticed that you can't copy text in the markdown preview. Should we add support for that?

I guess you're referring to [this thing](https://github.com/Automattic/simplenote-electron/issues/744#issuecomment-453437049)? We're going to need to update Electron to fix that, so it probably won't make it into our next release. Hopefully the next one, though.

roundhill:
> We're going to need to update Electron to fix that, so it probably won't make it into our next release. Hopefully the next one, though.

OK 👍

PR #1166, opened 2019-01-18
Add "Insert Checklist" menu item

### LINKED ISSUE #1138: Add menu item to insert checklist

mirka:
Closes #1138 

## Single
insert-checklist-single

## Multiple
insert-checklist-multiple

### DIFF
--- a/lib/editor/insert-or-remove-checkboxes.js
+++ b/lib/editor/insert-or-remove-checkboxes.js
@@ -56,7 +56,7 @@ export function insertOrRemoveForBlock(blockKey, contentState) {
 
   if (matches) {
     // Remove task prefix
-    const leadingWhitespace = get(matches, '1', '');
+    const leadingWhitespace = matches[1];
     const taskPrefix = matches[2];
     const spacesToTrim = matches[3].match(/^\s*/)[0].length;
     const anchorOffset = leadingWhitespace.length;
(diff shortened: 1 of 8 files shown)


lib/editor/insert-or-remove-checkboxes.js
```diff
@@ -0,0 +1,117 @@
...
+  const text = contentState.getBlockForKey(blockKey).getText();
+  taskRegex.lastIndex = 0; // Reset the search
+  const matches = taskRegex.exec(text);
+  let newContentState;
+
+  if (matches) {
+    // Remove task prefix
+    const leadingWhitespace = get(matches, '1', '');
+    const taskPrefix = matches[2];
+    const spacesToTrim = matches[3].match(/^\s*/)[0].length;
```
roundhill: Are we safe from out of index exceptions here?

roundhill:
Not sure if this is the right PR to report this, but I noticed that an extra space is hanging around if you press enter on an empty checklist item and it gets removed:

space-character

lib/editor/insert-or-remove-checkboxes.js
mirka: Should be ok, and unit tests will fail spectacularly if the regex is changed in an unexpected way. I tweaked the regex a bit to ensure that the first capturing group is always a string though (https://github.com/Automattic/simplenote-electron/pull/1166/commits/0a738e6a49d1ae8d22ce0c549003d335e3e7fb63).

mirka:
> I noticed that an extra space is hanging around if you press enter on an empty checklist item and it gets removed

Great catch. I've determined the cause and will fix in a separate PR. Thanks!

PR #1171, opened 2019-01-23
Fix CJK-related crashes after tab character

### LINKED ISSUE #780: Typing CJK after a Tab character can crash the app
#### Steps to reproduce
1. Open your Simplenote app.
2. Change the keyboard layout into Chinese(Simplified Chinese)
3. Now try to type something in a note an increase your speed.
4. The app will crash most of the times.

#### What I expected
To work in a similar way as when using the ABC alphabet.

#### What happened instead
It crashed the app.

#### OS version
MacOS
The user also reports the following:
>Mac and Linux Manjaro.

#### Screenshot / Video

https://monosnap.com/file/D9Wqx5L9idY7WZRYLvFdjZKRSEH4lb#

More details from the user:

>* The version on Windows is 1.16, installed by Windows Store.
>* The version on Mac is 1.16, installed by homebrew.
>* The version on Manjaro is 1.1.6-1, which is maintained by aur-simplenote-electron-bin (https://aur.archlinux.org/packages/simplenote-electron-bin/)

Please note that this won't break the app if you copy-paste the characters so it is important to change the keyboard layout.

Case: #1233037-zen

mirka:
Closes #780 

_Update: This seems to have been a bug in the Chrome version in Electron 2. It doesn't reproduce at all in Electron >3.0.0. So this workaround should be unnecessary once we update._

This is a really unsatisfying workaround that doesn't address any root causes and I don't like it 😐I first suspected a focus management bug in our [`indentCurrentBlock()`](https://github.com/Automattic/simplenote-electron/blob/master/lib/note-content-editor.jsx#L24) code, but as far as I could tell that doesn't seem to be the case here. Hopefully we can dig a little deeper into the issue sometime in the future, but since this is a pretty serious bug for our CJK users this workaround should help in the meantime.

## To test

Refer to this comment for repro steps: https://github.com/Automattic/simplenote-electron/issues/780#issuecomment-449771120

Note that the original bug only applies to Electron, and not Chrome/Firefox/Safari.

### DIFF
--- a/lib/editor/tab-decorator.js
+++ b/lib/editor/tab-decorator.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import SimpleDecorator from 'draft-js-simpledecorator';
+
+/**
+ * Wrap tab characters in a span to prevent CJK-related crashes
+ *
+ * The sole purpose of this decorator is to prevent DraftJS from
+ * crashing when dealing with CJK characters after a tab. The span somehow
+ * prevents DraftJS's internal focus state from getting out of sync with the
+ * visible caret. See below for steps to reproduce the original bug:
+ * https://github.com/Automattic/simplenote-electron/issues/780#issuecomment-449771120
+ *
+ * TODO: Figure out if this is a bug that should be fixed in DraftJS itself.
+ */
+const tabDecorator = () => {
+  const strategy = (contentBlock, callback) => {
+    const tabRegex = /\t/g;
+    const text = contentBlock.getText();
+    let match;
+
+    while ((match = tabRegex.exec(text)) !== null) {
+      const start = match.index;
+      const end = start + match[0].length;
+
+      callback(start, end);
+    }
+  };
+
+  const WrappedTab = ({ children }) => <span>{children}</span>;
+
+  return new SimpleDecorator(strategy, WrappedTab);
+};
+
+export default tabDecorator;
--- a/lib/editor/tab-decorator.js
+++ b/lib/editor/tab-decorator.js
@@ -10,7 +10,7 @@ import SimpleDecorator from 'draft-js-simpledecorator';
  * visible caret. See below for steps to reproduce the original bug:
  * https://github.com/Automattic/simplenote-electron/issues/780#issuecomment-449771120
  *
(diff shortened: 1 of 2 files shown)


roundhill:
Pretty sure there's a bug with this where characters get duplicated, I can't reproduce this in `master`:

1. Using Hiragana input, type `tab` and then `kanji`.
2. Press enter.
3. The phrase is repeated twice, and pressing backspace doesn't work as expected:

tab

mirka:
> Pretty sure there's a bug with this where characters get duplicated, I can't reproduce this in `master`

Gaaah. I can reproduce it in `master`. It just manifests a little differently — no immediate input duplication, but the internal focus state and visible caret are definitely out of sync so it glitches once you hit backspace. I can confirm that this issue is also fixed by updating to Electron >3.0.0.

Despite this, I think we should merge in this PR for v1.4.0 as a band-aid while we work to get the Electron update out ASAP. Two reasons:

- Most users who are using tabs are presumably typing a nested list item, which will trigger the crash rather than the glitch.
- Glitches are at least better than crashes 😬

Thoughts?

roundhill:
> Thoughts?

I agree, duplicated content is better than a WSOD. And we'll have it really fixed soon enough with the electron update. :shipit:

lib/editor/tab-decorator.js
```diff
@@ -0,0 +1,34 @@
...
+ * crashing when dealing with CJK characters after a tab. The span somehow
+ * prevents DraftJS's internal focus state from getting out of sync with the
+ * visible caret. See below for steps to reproduce the original bug:
+ * https://github.com/Automattic/simplenote-electron/issues/780#issuecomment-449771120
+ *
+ * TODO: This workaround can be removed when we update to Electron >3.0.0.
+ */
+const tabDecorator = () => {
+  const strategy = (contentBlock, callback) => {
+    const tabRegex = /\t/g;
```
roundhill: One of the smallest regex I've ever seen ;)

mirka:
The duplication glitch https://github.com/Automattic/simplenote-electron/pull/1171#issuecomment-456982555 is now fixed in v1.5.0.

PR #1090, opened 2018-12-10
[WIP] Add "Check for Updates" menu item

mirka:
Closes #1041 

### Test builds
https://github.com/Automattic/simplenote-electron/commits/feature/check-for-updates-test

Issue #745, opened 2018-04-09
Linux note editor: Korean consonants and vowels represented separately


pauljacobson:
This issue was raised in 1040852-zen

#### Steps to reproduce

1. The user installed Simplenote for Linux using the `simplenote-linux-1.1.3.deb` package
2. Latin characters are rendered correctly.
3. Korean characters are not rendered correctly (user's explanation below).

#### What I expected

For Korean characters to be rendered correctly.

#### What happened instead

From the user's report:

>I downloaded the simplenote package from the website.
>(simplenote-linux-1.1.3.deb)
>I installed the app on my ubuntu os successfully.
>
>Everything is find except typing non-english character.
>When I try to type Korean character, it typed weirdly.
>
>For example, I would like to type "한글", but it typed "한ㄱㅡㄹ" on the editor. The problem is consonants and vowels are inputted separately.

>To input Korean character, I'm using IBusHangul package and my keyboard layout is Hangul keyboard Dubeolsik.

#### OS version

To be confirmed.

#### Related issues

These issues appear to be related:

https://github.com/Automattic/simplenote-electron/issues/561

https://github.com/Automattic/simplenote-electron/issues/560

rumahrumah:
When I try to add a tag in Korean, one stroke produces two characters.
![2018-09-23 07-02-22](https://user-images.githubusercontent.com/14152092/45922133-d471d280-befe-11e8-8936-0a1b093bf95a.png)

I guess it's related to this issue.
I'm using Simplenote 1.1.7 on Ubuntu 18.04.

mirka:
I confirmed this issue on Ubuntu 18, Simplenote v1.2.1. Typing the key sequence `gksrmf` seems to produce the characters `한글` in Text Editor, Firefox, etc., but this fails in Simplenote. Japanese is working fine. I would like to know if Chinese is working.

Also, this does not happen on Windows or Mac.

mirka:
> When I try to add a tag in Korean, one stroke produces two characters.

@rumahrumah Thanks for the report! This is a different issue and will be fixed in #999.

nemonein:
I tested 1.3.4 on KDE Neon(Kubuntu 18.04 based Disto).
Kubuntu's default input system(for Far east Asian languages) is fcitx. I used fcitx and fcitx-hangul(Korean engine).

I can't say it works fine.
[...]

PR #1172, opened 2019-01-24
Update to Electron 4.0.2

### LINKED ISSUE #744: Ctrl+A Does Now Work In Preview Mode

mirka:
Closes #744 
Fixes tangential issues related to #1171 #1090 

The predictable and immediately apparent issues have been addressed. Smoke tested on Windows, Linux, and Mac.

## Known issues

The "Checking for Updates" dialog on Linux is showing a menu bar when [it shouldn't](https://github.com/AndersonMamede/electron-progressbar/blob/master/source/index.js#L248). This is a known Electron bug. (https://github.com/electron/electron/issues/15901 and/or https://github.com/electron/electron/issues/16521)

## TODO

- [x] Check if #745 is fixed → No
- [x] Check if #1142 is fixed → No
- [x] Figure out why the [appx code signing](https://ci.appveyor.com/project/Automattic/simplenote-electron/builds/21860675/job/kvbgo00lbqu2cj91) is failing → Something in `electron-builder` 20.33.1 is breaking it. Doesn't seem like a known issue so it might be something on our end, but I reported it to the electron-builder repo. In the meantime I think we can keep using the last working release.


mirka:
> One weird thing happened which I don't think is related to this PR. I got the app into a state where the Focus Mode button wasn't working 🤷‍♂️

Thanks, I'll keep an eye out for that!

Issue #36, opened 2015-10-20
Restore cursor location when note is updated


beaucollins:
When a change is applied from Simperium, the cursor position should be updated to account for the new content.

For example, if I'm typing at the end of the note and a change comes in that modifies the beginning of the note by adding six characters, the cursor should be advanced six characters so that as I continue typing I'll be in the correct location.

**Reported In**:
 - #2642910-t
 - #864118-zen
 - #958281-zen
 - #757894-zen
 - #808817-zen
 - #1193660-zen

dmsnell:
While working on the Monaco interface I built this so that we wouldn't have to destroy the editor context the way we have been. Many issues will be simultaneously solved: wonky cursors, losing the undo history, accidentally undoing remote changes, etc…

I felt like [having access to the patch itself](https://github.com/Automattic/simplenote-electron/pull/692/files#diff-a4e7d231ffbd43ffee4a46bf0854cb1bR443) would be highly valuable and I'd like to expose more deeply the underlying data changes to the notes. For example, there are good ideas to show activity based on what segments of a note have been recently changed.

Getting the [cursor shift](https://github.com/Automattic/simplenote-electron/pull/692/files#diff-6646bac18aa042ee2a7fb3cc0a5f669cR41) introduced from a patch then became trivial after having the parsed form in app state.

cecilerainon:
Another report in #1193660-zen (they actually thought it was intended and wanted to know how to have it work on other notes...)

jessestu:
I think this is related in 1637228-zen. A Linux user writes,

> I've been using Simplenote for 8 months and I noticed that the cursor don't follow on page scroll.

> It means that if I have a long note, if the cursor is set at the bottom of the page then I hit enter. The page didn't scroll. It hard for me to just continue typing and be not able to see the latest update of my note.

WolfyLPDC:
I have been doing testing and, although the problem persists after reinstall the software, it seems to not appear anymore after deleting the folder `~/.config/Simplenote/`. Of course, you have to login again but the application turns useful, which is nice. 

If another user can report this as a solution, perhaps it can help focusing the debugging.

logTom:
Can confirm.
Deleting the folder ~/.config/Simplenote/ makes the app usable again. Thanks for pointing out this workaround.

BackSeat:
Seeing the same (just started for me with 1.4). Above fix also worked.

rob-01:
[...]

PR #1192, opened 2019-02-05
Fix cursor position jump when setting new EditorState

qualitymanifest:
Closes https://github.com/Automattic/simplenote-electron/issues/1185

As I describe here https://github.com/Automattic/simplenote-electron/issues/1185#issuecomment-460163194 you can reproduce this issue fairly easily if you get your keystroke timing right, and you can verify that it's happening by putting a `console.log` in that `if` statement. After making these changes, the `console.log` would still run, but the cursor maintains the correct position.

No offense taken if you don't want to merge this, it was largely done just to see if I could tackle it. If you do, it seems to be working great but it definitely needs some thorough testing.

### DIFF
--- a/lib/note-content-editor.jsx
+++ b/lib/note-content-editor.jsx
@@ -144,6 +144,34 @@ function continueList(editorState, itemPrefix) {
   );
 }
 
+function getNewSelectionState(oldEditorState, newEditorState) {
+  // Find the block index for the old focus/anchor keys
+  // Use the new focus/anchor keys at that index with the old focus/anchor offsets
+  const oldEditorSelection = oldEditorState.getSelection();
+  const oldAnchorKey = oldEditorSelection.getAnchorKey();
+  const oldFocusKey = oldEditorSelection.getFocusKey();
+  const oldEditorBlocks = oldEditorState.getCurrentContent().getBlocksAsArray();
+  let oldAnchorPosition;
+  let oldFocusPosition;
+  for (let i = 0; i < oldEditorBlocks.length; i++) {
+    if (oldEditorBlocks[i].getKey() === oldAnchorKey) {
+      oldAnchorPosition = i;
+    }
+    if (oldEditorBlocks[i].getKey() === oldFocusKey) {
+      oldFocusPosition = i;
+    }
+  }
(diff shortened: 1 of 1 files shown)


qualitymanifest:
I was curious about the performance of this fix so I filled up a note with 30 large blocks of text and tried it out with my task manager open. 

When you trigger the (previously) problem behavior and continue to "ride" the update cycle as I describe here https://github.com/Automattic/simplenote-electron/issues/1185#issuecomment-460314564 Simplenote's CPU usage goes up to about 10% until you let off and exit the continuous update cycle.

I tried restoring the code to the way it was before (with the problem behavior) and the CPU usage is the same when you ride the update cycle, 10%. So, although my fix is somewhat verbose, I don't think it's having a measurable effect on performance, the main issue there being the continuously updating state.

mirka:
Thank you so much for looking into this! I agree with your assessment that we have a race condition when a user is typing during the short amount of time between the debounced update being fired and the updated note content coming back as a prop. That short window is < 1 ms when I time it, but like you said, there are many opportunities for it to occur.

But to back up a little, this editor state re-creation code is only intended for when the selected note receives a remote change from the server, and not for local changes caused by typing:

https://github.com/Automattic/simplenote-electron/blob/792c900f8494f2ab286b798243b33ec8cb5c2dd4/lib/note-content-editor.jsx#L259-L266

The reason that local changes weren't being propagated back as the `content` prop until now is because _that itself_ was a bug, which I fixed in #1130 😭

Ideally we shouldn't be doing `createNewEditorState()` at all, much less the focus manipulation, for local changes caused by typing. It's unnecessary and also resets the undo stack.

Your cursor location strategy is relevant to #36 though. The dev team back then seemed to be planning on calculating the new location from the update patch coming in from the server, but I'm not immediately sure if that's even possible in our current DraftJS setup. In that case, I think your strategy here could be an alternative.

Let me put this on hold while I see if there's a good way to get rid of the root cause 🙏

qualitymanifest:
Ah, makes sense. I had a hunch I may have been overlooking something, as I've still barely scratched the surface of how everything flows around within simplenote-electron, and it's definitely quite a band-aid fix.

mirka:
Thanks again for your great work on this @qualitymanifest! We'll get the fix out asap.

PR #1201, opened 2019-02-12
Add sync indicator

### LINKED ISSUE #459: Status bar to show sync / autosave status (feature request)
It will be helpful if we have a status bar or a message that shows if a note has been successfully synced or saved in a system.

### LINKED ISSUE #565: If Simplenote can't connect to the server, it should say so in its title bar
#459 isn't done yet, and I presume it may be very difficult to do. Here's a proposed temporary workaround which might be easier to implement.

-  If you start Simplenote, and it can't connect to the server:
    -  It should say "Simplenote — No Internet connection" in the title bar.

-  If server connectivity later becomes available:
    -  The client should retrieve as many updates as possible from the server.
    -  Once the client believes that it's probably up to date, it should change its title to just "Simplenote".
    -  Finally, it should also show an [infobar](https://en.wikipedia.org/wiki/Infobar). This bar should say "Your Internet connection has been restored". It should have an "OK" button.

### LINKED ISSUE #730: Add spinner for network activity
Any time we have local changes which we haven't yet attempted to send to the server and any time we have _sent_ data to the server but not heard back from it then we are basically waiting on network activity in order to synchronize a note's data. **It would be good to present a small spinner to show that activity** - something like the way some mobile apps do it.

We would probably need to provide hooks into `node-simperium` on `change-queued` and `change-confirmed` or something like that in order to get the right triggers.

mirka:
Closes #730 
Closes #565 (Maybe needs a [snackbar](https://material-ui.com/demos/snackbars/) to alert the user if the connection was lost? Would be distracting when on spotty Wi-Fi though.)
Closes #459 

### Has connection:

synced

### Offline with many changes:

many-changes

## TODO
- [x] truncate list when there are lots of notes with unsynced changes
- [x] fix keyboard accessibility

### Last synced

Last synced time is determined by the last time a non-heartbeat message is received from the server. The timestamp is saved to Local Storage, so it will persist if the user quits & relaunches the app while offline.

### "Offline"

"Offline" state is determined by the `connect` and `disconnect` events emitted by the Simperium client. The "sync" icon will change to a "!" icon when offline. When there are no pending changes, the message will say "No connection" instead of "All changes synced".

### Unsynced changes

Unsynced changes are determined from the note bucket's local queue in the Simperium client. The number of unsynced changes is the count of unique note IDs found in the local queue's `queues` and `sent` objects.


lib/icons/sync.jsx
```diff
@@ -0,0 +1,15 @@
...
+      className="icon-sync"
+      xmlns="http://www.w3.org/2000/svg"
+      width="18px"
+      height="12px"
+      viewBox="0 0 18 12"
+    >
+      <path d="M17.625,7.125 L15.3105,9.4395 L14.25,10.5 L10.875,7.125 L11.9355,6.0645 L13.5,7.629 L13.5,6 C13.5,3.519 11.481,1.5 9,1.5 L9,0 C12.3135,0 15,2.6865 15,6 L15,7.629 L16.5645,6.0645 L17.625,7.125 Z M4.5,6 L4.5,4.371 L6.0645,5.9355 L7.125,4.875 L3.75,1.5 L2.6895,2.5605 L0.375,4.875 L1.4355,5.9355 L3,4.371 L3,6 C3,9.3135 5.6865,12 9,12 L9,10.5 C6.519,10.5 4.5,8.481 4.5,6 Z" />
+    </svg>
+  );
+}
```
roundhill: Where is this icon from? Is it a GridIcon?

roundhill: I'm also wondering if we should show this icon. It may look like a button to some that you can click to start a sync?

lib/components/sync-status/popover.jsx
```diff
@@ -0,0 +1,110 @@
...
+      notes,
+      onClose,
+      theme,
+      unsyncedNoteIds,
+    } = this.props;
+    const themeClass = `theme-${theme}`;
+    const open = Boolean(anchorEl);
+    const hasUnsyncedChanges = unsyncedNoteIds.length > 0;
+
+    const QUERY_LIMIT = 10;
```
roundhill: Is this limit so that the popup doesn't become too tall? What if we put the note titles in a scrollable div?

lib/icons/sync.jsx
mirka: Yes, it's from the Gridicons set.
https://github.com/Automattic/gridicons/blob/master/sources/svg/gridicons-sync.svg

> I'm also wondering if we should show this icon. It may look like a button to some users that you can click to start a sync?

I'm a little conflicted on this, too. The button thing is definitely a concern, but I also feel like it's harder to understand what the indicator even is without the sync icon. (It's less of a problem in the native macOS app, since the Navigation Bar is always visible, and we don't have to think about the webapp case where the Settings button and footer links are in close proximity.)

One mitigating factor is that the sync icon is only shown when the websocket is connected. So in most situations, a user would only see that sync icon in conjunction with "All changes synced" — thus no reason to click it.

lib/components/sync-status/popover.jsx
mirka: Good question. A few reasons for the limit, in my mind:

1. The note titles are taken by regex-ing the note contents on each sync indicator hover, and not cached.
1. I feel like having a scrollable area inside a hover-activated popover is a little too much complexity. That would warrant a proper click-activated popover.
1. The intent of this feature is not to provide an exhaustive list of unsynced changes, but rather a way to self-diagnose if something's going wrong. Providing a list of 100 note titles doesn't add much utility over 10 notes. If we really wanted to show every unsynced note in a useful way, we should consider a different UI.

lib/icons/sync.jsx
roundhill: > One mitigating factor is that the sync icon is only shown when the websocket is connected. So in most situations, a user would only see that sync icon in conjunction with "All changes synced" — thus no reason to click it.

That makes sense, let's leave it in then and I'll add it to macOS :)

Donboscow:
Isn't it possible to place the animated moving/rotating (like dropbox sync indicator) spinner or [Saving and Saved] sign before the top [History, Share, Trash, Info] logos? I think it will be simpler and natural than this approach. Also to check the sync status you have to always leave the main view and go to the [Menu] which is quite distracting and unintuitive. In the app.simplenote.com  website when you edit any note it shows live [Saving and Saved] status sign in the top right corner which is more natural and intuitive. @mirka @dmsnell

mirka:
@Donboscow 
Yes, your points are valid and that is something we considered as well. But for our first iteration here, the intent was more to provide a way to self-diagnose when something is going wrong, rather than having an indicator animating on every successful server request. Having a constantly visible, animating sync indicator makes you more aware of the syncing, which itself could be considered a distraction for some people.

There are definitely tradeoffs for each approach (visible/hidden, static/animating, etc). But nothing is set in stone and I appreciate your input 👍

PR #1205, opened 2019-02-15
Auto-detect note language (for spelling and fonts)

### LINKED ISSUE #788: Editor: Simplenote doesn't display correct fonts for CJK languages, for example Japanese
This is a report from a user in 1304762-zen. The user pointed out the following issue:

>On windows app with Japanese language, incorrect and mixed fonts are used.
> Text to reproduce:
>"日本語の漢字とひらがな、カタカナ、Alphabet混在サンプルです。"
>That's well known problem CJK language environment.

This may be an Electron issue, rather than something that's specific to the Simplenote app itself. The user helpfully pointed out these two issues that touch on how the Electron app doesn't adequately cater for CJK languages when it comes to font choices:

* https://github.com/Microsoft/vscode/issues/4433
* https://github.com/Microsoft/vscode/issues/5260

The user included the following screenshot as an illustration of their point:

Unfortunately, I'm not familiar with the correct fonts that _should_ be used for CJK fonts to elaborate on this example.

### OS version

OS: Windows (although this appears to relate to Electron-based Simplenote apps).

### Related issue

This issue is peripherally related to Issue https://github.com/Automattic/simplenote-electron/issues/274

### LINKED ISSUE #949: Ability to change the language used for the Spell Check feature
Reported originally on 1511012-zen . 

This user has the system settings in Spanish, but uses Simplenote in English. What happens is that the new spell check feature defaults to check the notes using Spanish instead of English.

> I was really excited when I saw on the release notes of 1.21 that you have added spellcheck. The thing is, my Windows (7) system is in Spanish and I usually write in Simplenote in English. I guess it's pulling the language settings from my system because it's correcting everything in Spanish.

> Is there an option to change this? Or would you consider adding a language toggle for future versions?

mirka:
Closes #949 
Closes #788

This enhances the spellchecker by adding language auto-detection. It's working pretty well in my testing, and hopefully it will be a better experience than having to manually select a language from a dropdown.

## Bonus: Better fonts on Windows

As a bonus, I'm using the auto-detected language to set a `lang` attribute to the note body. This will help Electron in Windows to select language-appropriate fonts when the primary OS language is not set to that (#788). One limitation to this is that it's only for the Note Editor, and not the NoteList, since that would be too costly. Still, it's an improvement over the current behavior.

## To test

These changes only affect Electron on Windows and Linux. (Don't know why, but the language detection feature in `electron-spellchecker` is [disabled on Mac](https://github.com/electron-userland/electron-spellchecker/blob/5fa6bd78c10dcf8d1c0cfe6b03bf973bd3917132/src/spell-check-handler.js#L365). It's strange because it works if I just comment out that line.)

You can set `localStorage.debug="electron-spellchecker:*"` in the devtools console and reload to see the language switching in action. Language will be determined every time a different note is selected, or whenever the spelling/auto-detection menu items are toggled. If a language cannot be detected (either by being too short or ambiguous), it will use the OS language.

### Test instructions on Linux/Windows

1. Enable `Check Spelling` in the Edit menu.
1. Grab some dummy text in several spell-checkable languages (English, French, German, etc.) and make a new note for each language text. Assuming that your system language is set to English, non-English words should be considered spelling errors and have squiggly lines.
1. Enable `Auto-Detect Language` in the Edit menu.
1. Now see the non-English notes again. Squiggly lines should be gone. (It may take a few seconds because the spelling dictionaries need to be downloaded first.)
1. Delete some characters from a word or two to make spelling errors, and see that they show squiggly lines.

### Test instructions on Mac

1. See that the `Auto-Detect Language` option in the Edit menu is hidden. (Feature is disabled on Mac)

### DIFF
--- a/desktop/menus/edit-menu.js
+++ b/desktop/menus/edit-menu.js
@@ -1,3 +1,4 @@
+const platform = require('../detect/platform');
 const { appCommandSender } = require('./utils');
 
 const buildEditMenu = settings => {
@@ -49,6 +50,13 @@ const buildEditMenu = settings => {
         type: 'checkbox',
         checked: settings.spellCheckEnabled,
         click: appCommandSender({ action: 'toggleSpellCheck' }),
+      },
+      {
+        label: '&Auto-Detect Language',
+        type: 'checkbox',
+        checked: settings.languageDetectionEnabled,
+        click: appCommandSender({ action: 'toggleLanguageDetection' }),
+        visible: !platform.isOSX(), // currently not working on Mac
       },
     ],
   };
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -44,6 +44,7 @@ class NoteContentEditor extends Component<Props> {
       hasRemoteUpdate: PropTypes.bool.isRequired,
       version: PropTypes.number,
     }),
+    detectLanguage: PropTypes.bool.isRequired,
     noteId: PropTypes.string,
     onChangeContent: PropTypes.func.isRequired,
     spellCheckEnabled: PropTypes.bool.isRequired,
@@ -98,6 +99,7 @@ class NoteContentEditor extends Component<Props> {
       this.props.content.text,
       this.props.searchQuery
     ),
+    lang: undefined,
   };
 
   editorKey = 0;
@@ -161,16 +163,51 @@ class NoteContentEditor extends Component<Props> {
   };
 
   componentDidUpdate(prevProps) {
-    const { content, searchQuery, noteId, spellCheckEnabled } = this.props;
+    const {
+      content,
+      detectLanguage,
+      noteId,
+      searchQuery,
+      spellCheckEnabled,
+    } = this.props;
+
     const { editorState } = this.state;
 
-    // To immediately reflect the changes to the spell check setting,
-    // we must remount the Editor and force update. The remount is
-    // done by changing the `key` prop on the Editor.
-    // https://stackoverflow.com/questions/35792275/
-    if (spellCheckEnabled !== prevProps.spellCheckEnabled) {
-      this.editorKey += 1;
-      this.forceUpdate();
+    const updateLanguage = () => {
+      const minimumContentLength = 10;
+      if (!detectLanguage || content.text.length < minimumContentLength) {
+        window.spellCheckHandler.switchLanguage(navigator.language);
+        this.setState({ lang: undefined });
+      } else {
+        // Auto-detect the note content language to switch spellchecker
+        window.spellCheckHandler.provideHintText(content.text).then(() => {
+          // Use the auto-detected language to set a `lang` attribute on the
+          // note, which helps Chromium in Electron pick an appropriate font
+          this.setState({
+            lang: window.spellCheckHandler.currentSpellcheckerLanguage,
+          });
+        });
+      }
+    };
+
+    // Only relevant in Electron
+    if (window.spellCheckHandler) {
+      // To immediately reflect the changes to the spell check setting,
+      // we must remount the Editor and force update. The remount is
+      // done by changing the `key` prop on the Editor.
+      // https://stackoverflow.com/questions/35792275/
+      if (spellCheckEnabled !== prevProps.spellCheckEnabled) {
+        this.editorKey += 1;
+        this.forceUpdate();
+        updateLanguage();
+      }
+
+      if (
+        noteId !== prevProps.noteId ||
+        detectLanguage !== prevProps.detectLanguage
+      ) {
+        updateLanguage();
+      }
     }
 
     // If another note/revision is selected,
@@ -319,6 +356,7 @@ class NoteContentEditor extends Component<Props> {
   render() {
     return (
       <div
+        lang={this.state.lang}
         onCopy={this.copyPlainText}
         onCut={this.copyPlainText}
         style={{ height: '100%' }}
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -32,7 +32,9 @@ import * as S from './state';
 const TEXT_DELIMITER = '\n';
 
 type StateProps = {
+  detectLanguage: boolean;
   searchQuery: string;
+  spellCheckEnabled: boolean;
 };
 
 type Props = StateProps;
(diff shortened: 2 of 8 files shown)


lib/state/settings/actions.ts
```diff
@@ -103,3 +103,16 @@ export const toggleAutoHideMenuBar = () => (dispatch, getState) => {
...
+  languageDetectionEnabled,
+});
+
+export const toggleLanguageDetection = () => (dispatch, getState) => {
+  const {
+    settings: { languageDetectionEnabled },
+  } = getState();
+
+  dispatch(setLanguageDetection(!languageDetectionEnabled));
+};
```
dmsnell: hey! fun fact - we don't ned to create imperative _thunk_-style action creators for toggle-actions like this. reducers are great at keeping things simpler and then our actions remain a basic data object representing the action our users want to take.

```ts
// state/action-types
export type ToggleLanguageDetection = Action<'TOGGLE_LANGUAGE_DETECTION'>;

// …/settings/actions.ts
import * as A from '../action-types';

export const toggleLanguageDetection: A.ActionCreator<A.ToggleLanguageDetection> = () => ( {
	type: 'TOGGLE_LANGUAGE_DETECTION',
} );

// …/settings/reducer.ts
const languageDetection: A.Reducer<boolean> = (state = false, action) => {
	switch (action.type) {
		case 'SET_LANGUAGE_DETECTION':
			return action.languageDetectionEnabled;

		case 'TOGGLE_LANGUAGE_DETECTION':
			return !state;

		default:
			return state;
	}
}
```

Note here that we never need to base our intentions/actions off of the current app state - all we have to do is explain what we're trying to accomplish and then the reducers govern their own data type. It gets murkier when we mash together multiple data values into a single reducer, which is where `combineReducers` can help (you can see how things were prior to #1216 - sometimes the boiler plate is valuable for other reasons, here in keeping separate data types isolated from each other and making it easier to reason about changes from the user's perspective).

It may also be the case that when we're working this way we don't need the `SET_` action at all. This is the css with the edit/preview mode toggle, because the app starts in one mode and all that matters from the user's perspective is the intention to toggle the mode, not necessary to _set_ it.

dmsnell:
@mirka it's so awesome to see you updating this PR - thanks so much!

lib/state/settings/actions.ts
mirka: Yes, that makes sense. And I see the benefits of the `combineReducers` approach better now. Thanks for the tip!

lib/note-content-editor.tsx
```diff
@@ -152,16 +154,51 @@ export default class NoteContentEditor extends Component {
...
+    // Only relevant in Electron
+    if (window.spellCheckHandler) {
+      // To immediately reflect the changes to the spell check setting,
+      // we must remount the Editor and force update. The remount is
+      // done by changing the `key` prop on the Editor.
+      // https://stackoverflow.com/questions/35792275/
+      if (spellCheckEnabled !== prevProps.spellCheckEnabled) {
+        this.editorKey += 1;
+        this.forceUpdate();
+        updateLanguage();
```
dmsnell: could we instead incorporate the language into the `key` itself so that we don't need to track an increasing number and force a `re-render`?

```ts
getEditorKey = () => {
  const { spellCheckEnabled } = this.props;
  const { lang } = this.state;

  return `${ spellCheckEnabled ? 'spellcheck' : 'no_spellcheck' }-${ lang || 'unknown_lang' }`;
}

render() {
  …
    key={ this.editorKey() }
  …
}
```

With this I believe we should be able to get rid of the force-update cycles and allow the prop-changes to trigger the editor update naturally.

lib/note-detail/index.tsx
```diff
@@ -241,6 +245,7 @@ export class NoteDetail extends Component {
 }
 
 const mapStateToProps = ({ appState: state, ui, settings }) => ({
+  detectLanguage: settings.languageDetectionEnabled,
```
dmsnell: we might as well add this inside of `<NoteContentEditor />` so we don't have to pass it _through_ this component.

desktop/menus/edit-menu.js
```diff
@@ -50,6 +51,13 @@ const buildEditMenu = settings => {
         checked: settings.spellCheckEnabled,
         click: appCommandSender({ action: 'toggleSpellCheck' }),
       },
+      {
+        label: '&Auto-Detect Language',
+        type: 'checkbox',
+        checked: settings.languageDetectionEnabled,
+        click: appCommandSender({ action: 'toggleLanguageDetection' }),
+        visible: !platform.isOSX(), // currently not working on Mac
+      },
```
dmsnell: Besides disabling language-detection (and switching back to English?) is there any way to manually choose the language?

I can imagine that if someone wants to write in one language about another language the detection could be consistently wrong but then I'd want to manually fix it.

For example, writing in German about French poetry.

lib/note-detail/index.tsx
dmsnell: updated in ef9fcb7

lib/note-content-editor.tsx
dmsnell: replaced in ef9fcb7

dmsnell:
@mirka I have rebased this PR against the latest `develop` and pushed some changes from my comments. I tested it in macOS with the disable commented out but haven't yet tested it in Windows or Linux

desktop/menus/edit-menu.js
mirka: It is technically feasible, but I think that use case probably isn't common enough to warrant the clutter of an additional menu item. However, I did initially post a UI suggestion for a global manual select (as well as some other, per-note possibilities) [in the original issue](https://github.com/Automattic/simplenote-electron/issues/949#issuecomment-435934703).

This is mostly a UI design decision, so my thinking was to start out with the most simplest design, which is auto detection. If that works well enough for the majority of our users, great. If not, we can revisit.

mirka:
@dmsnell Thanks for the review and improvements! Feel free to move forward however the team wishes 👍

desktop/menus/edit-menu.js
dmsnell: > I did initially post a UI suggestion for a global manual select

Yeah that's what I meant. It's a good idea.

> start out with the most simplest design

Also a great idea. I wouldn't want to hold up this PR for not adding more but I wanted to ask if it was something already available we might have just been overlooking, for example, in the right-click menu of the text view if all we had to do were set some boolean toggle…

mirka: > I wanted to ask if it was something already available we might have just been overlooking

I'm afraid not 😅 Also, I wrote "technically feasible" but I overlooked one possible complication we might run into: Getting a definitive list of available spell checker languages. I haven't looked into it too deeply, but it doesn't seem like a language list is exposed through `electron-spellchecker`. If Chromium doesn't expose a list, it might get complicated. (Just a heads up in case you consider it in the future.)

belcherj:
Closing for now. Too out of date. We can take a look when we are ready to work on this.

PR #1207, opened 2019-02-15
Hide menu bar by default on Windows and Linux

cweagans:
Fixes https://github.com/Automattic/simplenote-electron/issues/293

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -61,6 +61,7 @@ module.exports = function main() {
       icon: iconPath,
       titleBarStyle: 'hidden',
       show: false,
+      autoHideMenuBar: true,
       webPreferences: {
         preload: path.join(__dirname, './preload.js'),
       },


mirka:
Thanks for the PR!

Since we can't just make the menu bar hidden by default (it's completely undiscoverable), we are going to need a few more things:

- A toggle option in the Display Settings ([like in Atom](https://github.com/atom/atom/pull/4215), but it should not be shown when on Mac or when not running in Electron)
- The same option in the View menu (also hidden when on Mac)
- When auto-hide is enabled, the Settings button and footer items in the Navigation Bar should be visible

Would you like to tackle these items, or would you prefer someone to take over?

cweagans:
I probably won't get to it for a while, so if someone else can do it sooner, please definitely jump in!

mirka:
Closing in favor of #1215.

Issue #985, opened 2018-11-06
Keyboard shortcut for preview on / off


rschouwenburg:
I'd love to see a keyboard shortcut to switch between preview and edit modes.

mirka:
Thanks. We actually have this, but it's undocumented at the moment. It's <kbd>Ctrl+Shift+P</kbd>. I'm hoping to make all the shortcuts discoverable in an in-app shortcut cheat sheet, and perhaps also tooltips. (Related: #700)

triangulum:
It's worth adding that after exiting the preview mode, the cursor should return to the original editing position, so the user can keep on writing. This would be a kind of *write-check-write* methodology.

At the moment, the cursor simple disappears and the document becomes inactive, which means you have to move your hands off the keyboard, and use the mouse to reposition the cursor.

mirka:
@triangulum I agree 100%. We're tracking this enhancement at #331

mirka:
A shortcut hint will be added as a tooltip in #1214.

Issue #700, opened 2018-01-27
Shortcut Documentation


AntonioFasano:
Shortcut documentation is outdated (especially for Linux/Windows).
See [simplenote.com/help/#shortcuts](https://simplenote.com/help/#shortcuts). There is also a mismatch in FAQ headers. 

IMHO shortcuts are absolutely necessary for productivity apps and documenting them is a necessary step to _make them consistent_.

mirka:
### TODO
- [ ] comb through the code and figure out what all the shortcuts are
- [ ] sort them into groups (Navigation, Views, Note, etc.)
- [ ] design a keyboard shortcut cheatsheet (modal dialog) to add in the app
- [ ] cheatsheet should be smart about platform-specific shortcuts
- [ ] add it under the Help menu (we’ll only need it in Electron for now)
- [ ] remove the shortcut documentation at [simplenote.com/help/#shortcuts](https://simplenote.com/help/#shortcuts)

rakhi2104:
I'm taking this up :1st_place_medal:

mirka:
@rakhi2104 Awesome! Even partial help with the first one or two tasks would be appreciated. You can post the results in this thread 🙂

rakhi2104:
Definitely :sunglasses: 
Don't know why the third one is also so attractive :drooling_face:

rakhi2104:
Hey @mirka, I made a board of shortcuts to classify between different categories.
Any changes to be made ?
![simplenote-shortcuts](https://user-images.githubusercontent.com/17671710/47256082-81585480-d497-11e8-8352-b5f1697e2bcd.png)

mirka:
Yay, thank you @rakhi2104 ! 😂 

For the documentation, let's remove some of the ones that are not really specific to our app. So remove Minimize, Close, Exit, and Undo, Redo, Cut, Copy, Paste, Select All, Print.

> Previous note, Next note

I'm planning on removing the Ctrl+Shift+Up/Down shortcuts for this, as per #798. So let's keep just the J/K ones.

We also seem to have

Navigation:
- Ctrl+T - Tag field

Note:
- Ctrl+Shift+P - Toggle Markdown preview

rakhi2104:
Oh, thanks ! Will make those changes.

mirka:
Adding `Alt + Ctrl + C` for Insert Checklist

AntonioFasano:
Thanks for the good work!

A couple of suggestions:

[...]

PR #1214, opened 2019-02-19
Add custom tooltips to toolbar buttons

mirka:
These custom tooltips are a small usability enhancement over the current built-in `title` tooltips.

tooltip

Benefits:

1. We can control the delay to be shorter than the built-in tooltips. (I've set it to 200ms)
2. We can add hints for shortcuts to aid discovery. Many users had requested a shortcut for the Markdown Preview button, which already existed but was near impossible to discover.

Related: #700 #985

### DIFF
--- a/lib/search-bar/index.jsx
+++ b/lib/search-bar/index.jsx
@@ -25,7 +25,12 @@ export const SearchBar = ({
   showTrash,
 }) => (
   <div className="search-bar theme-color-border">
-    <IconButton icon={<MenuIcon />} onClick={onToggleNavigation} title="Menu" />
+    <IconButton
+      icon={<MenuIcon />}
+      disableTooltip={true}
+      onClick={onToggleNavigation}
+      title="Menu"
+    />
     <SearchField />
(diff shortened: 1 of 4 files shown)


lib/search-bar/index.jsx
```diff
@@ -25,7 +25,12 @@ export const SearchBar = ({
   showTrash,
 }) => (
   <div className="search-bar theme-color-border">
-    <IconButton icon={<MenuIcon />} onClick={onToggleNavigation} title="Menu" />
+    <IconButton
+      icon={<MenuIcon />}
+      disableTooltip={true}
```
roundhill: Why did you disable it for the menu button?

mirka: I was under the impression that it was a standard Material design pattern to disable tooltips on the hamburger button, but I did a little more research today and it doesn't really seem to be the case 🤔 (There's even variance in Google's own apps!)

I'm going to remove this logic to keep it simple. Thanks!

PR #1218, opened 2019-02-22
Improve search performance on long notes

mirka:
Closes #1107

PR #1220, opened 2019-02-22
Keep editor in sync with selected note in NoteList

mirka:
Based on #1218 

This is a fix for an inconsistency in the UI state, where the selected note in the Note List (highlighted in blue) could get out of sync with the note shown in the editor.

There were at least two cases where this would occur:

- Start searching for a note, and the editor will continue showing the initially selected note even when the query doesn't match it anymore.
- When you switch from All Notes to the Trash view, the first trashed note is selected blue in the Note List, but the editor is blank.

### DIFF
--- a/lib/note-list/index.jsx
+++ b/lib/note-list/index.jsx
@@ -227,6 +227,8 @@ export class NoteList extends Component {
   static displayName = 'NoteList';
 
   static propTypes = {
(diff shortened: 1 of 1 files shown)


roundhill:
> * Start searching for a note, and the editor will continue showing the initially selected note even when the query doesn't match it anymore.

This always bugged me, yay!

Issue #1226, opened 2019-02-28
Lots of tags on a note squashes the notes list


roundhill:
It appears that if you have a lot of tags added to a note, it can cause the notes list to get squashed:

<img width="1023" alt="screen shot 2019-02-27 at 4 22 22 pm" src="https://user-images.githubusercontent.com/789137/53532520-07228700-3aac-11e9-9688-5bba52408665.png">

PR #1227, opened 2019-03-01
Fix large tag-list squashing note-list

qualitymanifest:
Closes https://github.com/Automattic/simplenote-electron/issues/1226

### DIFF
--- a/lib/app-layout/style.scss
+++ b/lib/app-layout/style.scss
@@ -95,14 +95,13 @@
 
 .app-layout__source-column {
   width: 328px;
-  min-width: 328px;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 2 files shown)


qualitymanifest:
Just realized I had missed the `$medium` breakpoint for the `source-column`, so added that in.

qualitymanifest:
Made those changes, everything seems to work as expected. 

Something I noticed while working on this is that if a tag exceeds the width of the tag-field, the entire note-editor grows. If the tag gets long enough the note-list disappears and eventually the note-detail does as well.

Of course that can only really happen with a **really** long tag (takes about 65+ characters in the narrowest screen width), which is so unlikely that it may not be worth addressing.

PR #1231, opened 2019-03-14
An Install section in readme

marioortizmanero:
Added mention to snap install. Much better than manually installing and it should at least be mentioned in the Readme for a faster and simpler install. Maybe not in the `Running` section, but perhaps in a new `Installing` section?


belcherj:
Until we officially manage the snap version I'd prefer that the README not include this section.  I'd like to get towards supporting services like snap officially.

Issue #1238, opened 2019-03-29
Adding @text in a note with Markdown formatting creates a link to GitHub


edequalsawesome:
#### Steps to reproduce
1. Create a new note in Simplenote
2. Add text that has @ in it, such as `@text`
3. Ensure note is formatted for Markdown
4. Preview note

#### What I expected
`@text` would be regular text

#### What happened instead
A link was created to `github.com/text`

#### Simplenote version
1.5.0

#### OS version
macOS 10.14.4

#### Screenshot / Video
![Screen Recording](https://cld.wthms.co/ffS72s+)

#### Other
Initially reported in #1914354-zen as an issue in the web app

gabe31415:
Thank you for accepting my issue and creating the bug report!  You all rock!

mirka:
We should disable the `ghMentions` option in Showdown.
https://github.com/showdownjs/showdown#valid-options

briangzimmerman:
If no one has started this, yet, I'd like to give it a go. Thanks.

belcherj:
This was merged in https://github.com/Automattic/simplenote-electron/pull/1256

PR #1244, opened 2019-04-09
Disable ghMentions option in Showdown converter

lol-russo:
Fixes https://github.com/Automattic/simplenote-electron/issues/1238 — typing @text will no longer create a link to a GitHub user in markdown preview, it will just display standard text.


belcherj:
Duplicate of: https://github.com/Automattic/simplenote-electron/pull/1256 Yours was in first but I was working from the top, my bad.  Thanks for the PR!

Issue #1294, opened 2019-05-31
Replace create-hash with md5.js


belcherj:
create-hash is just including md5.js to create the md5 hash. create-hash includes lots of features that we don't use causing bloat.

### What
***(Required)*** What is the feature you are requesting?  Add a concise description of the feature being requested.
Replace create-hash with md5.js

### Why
***(Required)*** Why are you requesting this feature?  Add a concise description of the problem this feature solves.

Decrease package size

### How
***(Optional)*** If applicable, add screenshots, animations, or videos to help illustrate how the feature could be done.
Use md5.js

PR #1308, opened 2019-06-04
Use md5 node module

belcherj:
### Fix
This uses a node module that only does md5 not a host of other hashing functions. This is only used in the sharing modal to pull down a users Gravatar

### Test
1. Go to https://app.simplenote.com/new
2. Click share icon in top right add a user to share with
3. In Chrome dev tools get the link to the image
4. Build this branch
5. Open app
6. Click share
7. Enter email
8. Get image link
9. Compare hash from production, this branch, and another md5 hash generator
10. They should all match
 
### Review

### Release
`RELEASE-NOTES.txt` was updated in d3adb3ef with:

- md5 hashing module was replaced

Closes: #1294


lib/dialogs/share/index.jsx
```diff
@@ -85,15 +85,11 @@ export class ShareDialog extends Component {
   };
 
   gravatarURL = email => {
-    const hash = createHash('md5');
-
-    hash.update(email.trim().toLowerCase());
-    let digest = hash
-      .digest()
+    let hash = md5(email.trim().toLowerCase())
```
mirka: `const`? (I didn't realize we don't have the prefer const rule enabled in eslint!)

belcherj: I need to add that! I think I'll merge and then add that rule and the fixes in another pr

dmsnell:
What was the purpose of this PR? Were we having trouble with the `create-hash` library? Was it unmaintained?

belcherj:
We were including create-hash that does a lot of things besides creating an md5 hash.  If I remember correctly I did away the middleman and just included the md5 library which just creates an md5 hash.

dmsnell:
Ah. I just noticed now that this merged months ago. I guess @belcherj we're saying that the purpose was to eliminate superfluous dependencies? Did you catch a before-and-after count of dependencies in the `node_modules` directory?

I'm not sure if I did this properly, and with `npm` it's always hard to do testing from code that existed in the past (because the repository is always changing) but I found that this change added three new directories to `node_modules` after applying it. That seems counter-intuitive, but with the way dependencies work it wouldn't surprise me if this is right.

| | Before applying patch | After applying patch |
|---|---|---|
| Directories in `node_modules` | 1037 | 1040 |
| Unique packages in `npm list` | 1071 | 1074 |

For counting unique packages in `npm list` I used the following command to try and get the total number of unique package names. I'm not sure if it's important to count unique versions or if those are even installed. I discovered that counting total number of dependencies is hard.

Either way it looks like this change probably _increased_ our dependency count and did so in a mostly insignificant way.

belcherj:
I did not catch the before and after but: https://www.npmjs.com/package/md5 has three dependencies which in turn have zero dependencies.  

https://www.npmjs.com/package/create-hash has 5 dependencies which in turn have more dependencies.  Not sure how the number of dependencies could have gone up.

dmsnell:
> Not sure how the number of dependencies could have gone up.

`npm` is funny sometimes. one way it could have gone up is that if `create-hash` were already a transitive dependency and already pulled in `md5` then our introduction of `md5` as a new top-level dependency at a specific version may have introduced a _second_ copy of `md5` (as another version) while also not eliminating the cause for the `create-hash` sub-dependency, thus leaving it in there.

I wish I were better at navigating the dependency tree and finding answers to these kinds of questions. 

From some additional basic investigation…
 - many packages depend on `create-hash`: `browserify-`, `public-encrypt`, `pbkdf2`, `parse-asn1`, `create-hmac`. they are pulling in `^1.1.0`
 - `create-hash@1.20` is in the tree and it depends on `md5.js@^1.3.4`
 - `md5.js@1.3.4` is required by `evp_bytestokey`
 - `md5@2.2.1` is a top-level dependency

So I think what happened is we added a new top-level dependency we didn't already have in our dependency tree before but we also didn't remove any of the dependencies for the package we were presumably replacing. In effect, this didn't do anything but introduce a new dependency.

Had we used `md5.js` instead of `md5` we may have eliminated the top-level dependency of `create-hash` without introducing new dependencies, but the net effect would be that we didn't change the total number of dependencies. At best, possibly by using `md5.js` we would leave open the possibility that _if_ several of our dependencies also remove their need for `create-hash` that we could eventually get rid of it too because we wouldn't directly depend on it.

If we continue with the numbers from earlier then by switching `md5` to `md5.js` the new directory count is `1037`, matching the pre-patch count, and the unique packages count is `1857`, so I have no idea how that happened - presumably because it updated the `package-lock.json` based on _today_ when I changed `md5` to `md5.js` vs. what it made _when this merged_.

PR #1313, opened 2019-06-04
Fix linting warning in lib/app

belcherj:
### Fix
- Updates propTypes to include all props.
- Uses UNSAFE_componentWillMount
- Does not redeclare appState variable

### Test
1. Run app
2. Observe no errors in console

### Review


belcherj:
#1315

PR #1314, opened 2019-06-04
Fix linting warning in tag-chip

belcherj:
### Fix
Adds prop types to tag-chip component

### Test
1. Have a note with tags
2. Observe no errors in console

### Review


belcherj:
#1315

PR #1516, opened 2019-08-14
Fixes webpack bug where wrong file is requested

belcherj:
### Fix
Webpack was optimizing dynamic imports renaming the files to 1.js, 2.js etc.  We were getting errors where the file requested was the not the expected file.  I changed the optimizations to use named modules so that it can't get confused. 

### Test
1.  This error is not really reproducible.  @theck13 was the only one to reproduce.  The plan is merge to develop then test using develop.simplenote.com

### Review

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -12,6 +12,15 @@ module.exports = () => {
 
   return {
     context: __dirname + '/lib',
+    optimization: {
+      namedModules: true,
+      namedChunks: true,
+      splitChunks: {
+        cacheGroups: {
+          default: false,
+        },
+      },
+    },
     mode: isDevMode ? 'development' : 'production',
     devtool:
       process.env.SOURCEMAP || (isDevMode && 'cheap-module-eval-source-map'),


sirreal:
The issue sounds like something likely related to caching, and the solution would be for chunks to include something to ensure the wrong cached chunk isn't used.

Renaming all the chunks will temporarily fix the problem —it busts the cache— but I suspect in the long run this may continue to occur.

I'd suggest reverting this chane change here to use `[name].[chunkhash].js`, that should ensure that only valid chunks are used:

https://github.com/Automattic/simplenote-electron/blob/ef224d62045afbcdd3aeb9ef1bb5bd87f1551092/webpack.config.js#L32

One question as someone new to this codebase: Is this for a web UI, an electron application, or both? Some optimizations make sense for certain context but not for others, that's something to keep in mind with changes like these.

sirreal:
> I'd suggest reverting this chane change here to use `[name].[chunkhash].js`, that should ensure that only valid chunks are used:

Opened a PR: https://github.com/Automattic/simplenote-electron/pull/1530

Note that this only makes sense in web context not electron, but based on `develop.simplenote.com` I assume this is used in a web context.

belcherj:
> One question as someone new to this codebase: Is this for a web UI, an electron application, or both? Some optimizations make sense for certain context but not for others, that's something to keep in mind with changes like these.

This repo serves as the code basef for electron app and the web app.

PR #1530, opened 2019-08-16
Fix/webpack chunkname with hash

sirreal:
Add `chunkhash` to `chunkFilename`

Alternative to #1516 (reverted here).

Reports of the wrong chunk being served with a particular build suggest
that cached, stale chunks may be served. This results in a broken
application.

Add chunkhash to the chunkFilename to ensure that the expected chunks
are requested and eliminate issues with cached chunks.

See
https://webpack.js.org/configuration/output/#outputchunkfilename

**Please test and review carefully, I'm new to this codebase**

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -20,7 +20,7 @@ module.exports = () => {
     output: {
       path: __dirname + '/dist',
       filename: 'app.js',
-      chunkFilename: '[name].js',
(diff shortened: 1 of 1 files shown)


belcherj:
This PR would require some changes to the backend.  I will see about making those changes but this may have to wait till I'm back from AFK.  Thank you so much for this. I had been meaning to make this change.

PR #1551, opened 2019-08-20
Support the unicode bullet as a list item bullet in unordered lists

kafoalto:
### Fix
Support the unicode bullet character `•` as a list item bullet because lists copied from HTML or word documents may contain this character.

### Test
1. Open Simplenote
2. Create an unordered list with unicode bullets `•`
```
  • list item a
  • list item b
```
3. Press enter after the first item
4. A new list item should be opened, starting with a unicode bullet `•`

### Review
Everyone can review this if they want :)

### DIFF
--- a/lib/editor/text-manipulation-helpers.js
+++ b/lib/editor/text-manipulation-helpers.js
@@ -4,7 +4,7 @@ import { includes } from 'lodash';
 import { getCurrentBlock } from './utils';
 
 export const isLonelyBullet = line =>
-  includes(['-', '*', '+', '- [ ]', '- [x]'], line.trim());
+  includes(['-', '*', '+', '- [ ]', '- [x]', '•'], line.trim());
 
 export function indentCurrentBlock(editorState) {
   const selection = editorState.getSelection();
--- a/lib/editor/text-manipulation-helpers.js
+++ b/lib/editor/text-manipulation-helpers.js
@@ -4,7 +4,7 @@ import { includes } from 'lodash';
 import { getCurrentBlock } from './utils';
 
 export const isLonelyBullet = line =>
-  includes(['-', '*', '+', '- [ ]', '- [x]', '•'], line.trim());
+  includes(['-', '*', '+', '- [ ]', '- [x]', '\u2022'], line.trim());
 
 export function indentCurrentBlock(editorState) {
   const selection = editorState.getSelection();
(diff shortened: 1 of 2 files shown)


lib/note-content-editor.jsx
```diff
@@ -233,10 +233,10 @@ export default class NoteContentEditor extends Component {
   };
 
   handleReturn = () => {
-    // matches lines that start with `- `, `* `, or `+ `
+    // matches lines that start with `- `, `* `, `+ `, or `•`
     // preceded by 0 or more space characters
     // i.e. a line prefixed by a list bullet
-    const listItemRe = /^[ \t\u2000-\u200a]*[-*+]\s/;
+    const listItemRe = /^[ \t\u2000-\u200a]*[-*+•]\s/;
```
dmsnell: Looks like this is fine when I test in my terminal, so it's probably good to leave, but if we are worried about issues having Unicode symbols in the source we can replace this with `\u2022`, clearing up any confusion about which symbol we wanted.

Related but unlikely and unwanted additions might also be:
 - bullet operator `\u2219` `∙`
 - bullet train `\u1f685` `🚅`
 - bullet bullet `\u204c` `⁌`

lib/editor/text-manipulation-helpers.js
```diff
@@ -4,7 +4,7 @@ import { includes } from 'lodash';
 import { getCurrentBlock } from './utils';
 
 export const isLonelyBullet = line =>
-  includes(['-', '*', '+', '- [ ]', '- [x]'], line.trim());
+  includes(['-', '*', '+', '- [ ]', '- [x]', '•'], line.trim());
```
belcherj: Please replace the unicode character with: `\u2022`

lib/note-content-editor.jsx
```diff
@@ -233,10 +233,10 @@ export default class NoteContentEditor extends Component {
   };
 
   handleReturn = () => {
-    // matches lines that start with `- `, `* `, or `+ `
+    // matches lines that start with `- `, `* `, `+ `, or `•`
```
belcherj: Please replace the unicode character with: `\u2022`

lib/note-content-editor.jsx
belcherj: Please replace the unicode character with: `\u2022`

PR #1563, opened 2019-09-03
Another attempt at fixing white screen error

belcherj:
### Fix
Attempts to fix issue where this function cannot be found causing a js failure. In the image below Ks is the function this PR moves.  

image

### Test
This failure is intermittent and cannot be replicated

### Review
Only one developer ise required to review these changes, but anyone can perform the review.

### Release
These changes do not require release notes as we don't know if this will resolve the issue

### DIFF
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -42,6 +42,19 @@ const appProvider = 'simplenote.com';
 const appID = config.app_id;
 let token = cookie.token || localStorage.access_token;
 
+// Signs out the user from app engine and redirects to signin page
+const redirectToWebSigninIfNecessary = () => {
+  if (!config.is_app_engine) {
+    return;
+  }
+
+  if (window.webConfig && window.webConfig.signout) {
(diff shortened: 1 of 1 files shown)


dmsnell:
@belcherj did we get a report of this breaking? how were we made aware.

it appears to be an issue with the Babel process, whereby we formerly were allowed to move `const` declarations below their first use _because_ Babel rewrote them as `var` statements, and the bindings were _hoisted_ up above in scope.

@beaucollins has been doing some work to remove other related hoisting issues in #1596. These defects have always been in the code, but `babel` hid them during the code transformation.

Was it possible that this appeared when trying to run the app without the Babel transform?

belcherj:
That could have been the case.  It was very intermittent and I never experienced the issue. @dmsnell

dmsnell:
> We were made aware via support requests.

This is interesting. It's worth keeping an eye on this and related issues, though I think that the work here and in #1596 probably took care of most of it.

PR #1565, opened 2019-09-11
Update colours to color-studio.blog

belcherj:
### Fix
Update CSS colors to use [color-studio.blog](https://color-studio.blog/) in line with the [new style guidelines](P2XJRt-11W-p2).

### Test
* Verify that the app works and the colors look pretty much the same :)
* Test in both light and dark modes

### Screenshots
Before:
Screen Shot 2019-09-25 at 10 44 46 PM
        
After:      
Screen Shot 2019-09-25 at 10 30 52 PM

Before:
Screen Shot 2019-09-25 at 10 48 10 PM

After:
Screen Shot 2019-09-25 at 10 30 01 PM

Before:
Screen Shot 2019-09-25 at 10 50 46 PM

After:
Screen Shot 2019-09-25 at 10 30 10 PM

Before:
Screen Shot 2019-09-25 at 10 47 59 PM

After:
Screen Shot 2019-09-25 at 10 30 30 PM

### Release
> Updated colors to use Color Studio, the color palette for Automattic products

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -8,7 +8,7 @@
 - Renamed CHANGELOG.md to RELEASE-NOTES.txt
 - Open new note automatically upon creation
 - Open new note automatically upon creation [1566](https://github.com/Automattic/simplenote-electron/issues/1566)
-- Updated colors to use Color Studio in alignment with new design guidelines
+- Updated colors to use Color Studio, the color palette for Automattic products
 
 ## [v1.7.0](https://github.com/Automattic/simplenote-electron/releases/tag/v1.7.0) (2019-08-12)
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -8,6 +8,7 @@
 - Renamed CHANGELOG.md to RELEASE-NOTES.txt
 - Open new note automatically upon creation
 - Open new note automatically upon creation [1566](https://github.com/Automattic/simplenote-electron/issues/1566)
+- Updated colors to use Color Studio in alignment with new design guidelines
 
 ## [v1.7.0](https://github.com/Automattic/simplenote-electron/releases/tag/v1.7.0) (2019-08-12)
 
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1,7 +1,6 @@
 // Grays
 $gray: #808080;
 
-$gray-lightest: lighten($gray, 38%);
 $gray-darkest: darken($gray, 38.5%);
 
 // Secondary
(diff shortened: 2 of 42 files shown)


scss/_variables.scss
```diff
@@ -1,32 +1,8 @@
...
-// lighten($gray, 30%)
-// lighten($gray, 20%)
-// lighten($gray, 10%)
-// darken($gray, 10%)
-// darken($gray, 20%)
-// darken($gray, 30%)
-
-// Secondary
+// Secondary colors
 $yellow: #f0b849;
```
belcherj: Can we transition the rest of these colors to color studio? A separate PR may be best given the size of this PR as it stands.

SylvesterWilmott:
Looks good!

I'm unsure if you want to tackle this within this PR but the way we use colours in dark mode should ideally be adjusted for better accessibility/contrast. The rules of colour use in dark mode can be found in the Design Guidelines. Otherwise, it looks great, thanks @belcherj

belcherj:
I cannot approve since I opened the PR but the code looks and the UI seems consistent with production.  I do think we have more serious contrast concerns but this is outside the scope of this PR. I think that @SylvesterWilmott is addressing these as part of a future effort.

belcherj:
Look like Sly and I was thinking the same thing.  For proper attribution @codebykat did most of the heavy lifting on this PR.

Let's address the color contrast issues in another PR.

belcherj:
If there is no other comment by end of day, please merge.

RELEASE-NOTES.txt
```diff
@@ -8,6 +8,7 @@
 - Renamed CHANGELOG.md to RELEASE-NOTES.txt
 - Open new note automatically upon creation
 - Open new note automatically upon creation [1566](https://github.com/Automattic/simplenote-electron/issues/1566)
+- Updated colors to use Color Studio in alignment with new design guidelines
```
dmsnell: When I do a web search for `color studio` I don't find anything that would explain this. Could we add something that indicates what it is or more importantly, what impact it has on our customers?

Something like this maybe…
```
 - Updated colors to use Color Studio, the color palette for Automattic products
```

scss/_variables.scss
codebykat: Yeah, these aren't used in many places so I can update them. I wasn't sure if we wanted to use color studio for the secondary stuff like greens.

codebykat:
> I cannot approve since I opened the PR but the code looks and the UI seems consistent with production. I do think we have more serious contrast concerns but this is outside the scope of this PR. I think that @SylvesterWilmott is addressing these as part of a future effort.

Yeah, I don't think dark mode is great. I fixed a few of the worst offenders in this PR (the search bar and the tag listing) but I agree it needs a more serious overhaul at a future date.

codebykat:
Did a last round of updates -- could use a smoke test but I think it's good to go and we can address the contrast in a follow-up PR.

belcherj:
Ship it!!

PR #1573, opened 2019-09-15
Fix: Cursors over UI elements should indicate clickable or text entry

belcherj:
### Fix
Cursors in the UI do not use the proper type to indicate when elements are clickable or editable. This PR fixes those issue across the app.

### Test
1. Run the branch
2. Step through app hovering over anything that is clickable and verify that the cursor changes to "pointer"
3. Hover over anything that allows text entry and verify that the cursor changes to "text"

### Review

### Release

- Hovering over a clickable or editable UI element now show the correct cursor for its type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,12 @@
 ## Future Release
 
 - Updated Log in and Sign up form to match current styling
+- Fixed cursors to use the proper type
 
 ## [v1.7.0](https://github.com/Automattic/simplenote-electron/releases/tag/v1.7.0) (2019-08-12)
 
 ### Fixes
+
 - Updates to dark mode styling [#1452](https://github.com/Automattic/simplenote-electron/pull/1452)
 - Updated several dependencies
 
--- a/lib/components/checkbox/style.scss
+++ b/lib/components/checkbox/style.scss
@@ -1,4 +1,9 @@
+input[type="checkbox"] {
+  cursor: pointer;
+}
+
 .checkbox {
+  cursor: pointer;
   line-height: inherit;
 
   &[aria-checked="true"] {
--- a/lib/components/slider/style.scss
+++ b/lib/components/slider/style.scss
@@ -1,6 +1,7 @@
 .slider {
   @extend %smart-focus-outline;
 
+  cursor: pointer;
   width: 100%;
   max-width: $max-content-width;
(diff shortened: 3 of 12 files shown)


CHANGELOG.md
```diff
@@ -3,10 +3,12 @@
 ## Future Release
 
 - Updated Log in and Sign up form to match current styling
+- Fixed cursors to use the proper type
```
dmsnell: Can we provide a stronger explanation of this? Where will people expect to find changes? Why were they wrong?

Something simple would probably go a long way…

```
Changed cursor over non-editable items to pointer
```

What were the cursors even before-hand?

dmsnell:
I'm not sure what to expect. Are there any before/after screenshots? I don't particularly know what "the proper type" means or how I would know that a cursor is the wrong type. Is there a way we could provide any more description in here?

codebykat:
@dmsnell There were essentially no pointer cursors over any clickable elements. I'll 👀

CHANGELOG.md
codebykat: Something like this, maybe?

> Hovering over a clickable or editable UI element now show the correct cursor for its type

belcherj:
> The note list on the sidebar (to switch notes, or to pin/unpin a note -- should be pointer, is default)

This should have resolved that https://github.com/Automattic/simplenote-electron/pull/1573/files#diff-b4aef3f86a137f3edaf5a4e2fbb2a23fR104 Do you not see the pointer

> Tags at the bottom of a note, when it has tags (should be pointer, is text cursor)

> "Add a tag..." at the bottom of a note (should be text cursor, is default -- only in Firefox, it's correct in Safari)

Made Tags that are on a note pointer since clicking on them removes from note. In the UI spring cleaning, this interaction will be cleaned up. Where you enter new tags I changed the cursor to text.

codebykat:
I do not see the pointer over the notes sidebar, over the pin/unpin or over the note itself:

Screen Shot 2019-09-22 at 8 56 33 PM

I tested this in both Firefox and Safari.

Verified that the other two are working now 👍

codebykat:
Okay so for whatever reason it's not working to add it to `.note-list-item` in `theme.scss`, but it works if added to `.note-list-item` in `note-list/style.scss`.

PR #1577, opened 2019-09-20
Note Updates: Resequence async calls to match execution flow

dmsnell:
The `noteUpdateRemotely` action will attempt to update a note's contents
in the `noteBucket` if there are changes in the local copy that need to
rebase against updates from the remote server.

The code as written has made it appear as though this update occurs
before calling `loadAndSelectNote` but because of the asynchronous
behavior of the `noteBucket` the update occurs after that action is
dispatched. The `loadAndSelectNote` action further complicates things by
calling `noteBucket.get()` which is itself asynchronous and so ordering
of events is even more dubious.

In this change I'm rearranging some of the control flow and performing
an early-abort in the case where we have no rebasing to apply. This
should hopefully make it slightly clearer in what sequence the events
are occurring. This exists to clear up the code so we can more easily find
a source of data loss and/or synchronization issues and resolve them.

**Likely this will have no impact on the application**: the change is mostly
duplicating some code in a way which shouldn't actually change the
runtime operation.

**Testing**

Smoke testing should confirm that nothing major is broken.
Testing the actual changes are harder because the purpose here is
to track down data races and untangle them. Nonetheless, the code
in question runs when we receive external updates from Simperium.

Thus, to test we can try a few primary scenarios:
 - open a note and disconnect from the network
 - make changes to that note in another network-connected device
 - reconnect this version of the app and ensure that the update is received and applied
 - repeat, but make changes locally while offline that represent the same changes applied on the connected device
 - repeat, but make changes locally while offline that don't conflict with the remote updates
 - repeat, but make changes locally while offline that do conflict with the remote updates

In the following tests I'm using `AC` as a start version and adding `B` remotely and `D` locally where necessary.

| Change | Behavior in `develop` | Behavior in Branch |
|---|---|---|
| Remote change, no local changes | ✅ `ABC` | |
| Remote change, matching local change | ⭕️ crash - `ABBC` | |
| Remote change, non-conflicting local change | ⭕️ crash - `ACDD` | |
| Remote change, conflicting local change  | ⭕️ crash - `ADC` | |

In summary I guess this means that the changes in this branch do not introduce the errors; they are present in `develop` already.

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -413,40 +413,46 @@ export const actionMap = new ActionMap({
             return;
           }
 
-          let working = note.data;
-
           // diff of working and original will produce the modifications the client has currently made
-          let working_diff = simperiumUtil.change.diff(original, working);
+          const working_diff = simperiumUtil.change.diff(original, note.data);
 
           // Check for equal diffs so we don't duplicate changes
           const diffsAreEqual =
             get(working_diff, 'content.v', '') === get(patch, 'content.v', '');
 
-          if (!diffsAreEqual) {
-            // generate a patch that composes both the working changes and upstream changes
-            const newPatch = simperiumUtil.change.transform(
-              working_diff,
-              patch,
-              original
+          if (diffsAreEqual) {
+            return dispatch(
+              this.action('loadAndSelectNote', {
+                noteBucket,
+                noteId,
+                hasRemoteUpdate: true,
+              })
             );
-            // apply the new patch to the upstream data
-            let rebased = simperiumUtil.change.apply(newPatch, data);
-
-            // TODO: determine where the cursor is and put it in the correct place
-            // when applying the rebased content
-
-            state.note.data = rebased;
-
-            // update the bucket and sync
-            noteBucket.update(noteId, rebased);
           }
 
-          dispatch(
-            this.action('loadAndSelectNote', {
-              noteBucket,
-              noteId,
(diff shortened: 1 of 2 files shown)


dmsnell:
Thanks @sirbrillig - I took your advice but in a different way. This update matches my original intent, which was to run both the callback and the promise, but I have removed the callback-invocations from this codebase so that we only use the Promise-based approach.

Sadly the `node-simperium` library we depend on calls up through the stack in an odd Promise wrapper, ironically preventing us from removing the callback argument entirely as it attempts to expose the call as a Promise. 🤦‍♂ 

In the process of removing the callbacks several latent errors were exposed in the app: namely that we have the (common) pattern of `callback = ( error, data ) => assumeNoError( data )`. Those flaws are now exposed even though the app flows haven't been fixed. For example, if the preferences don't load then we still end up in the same predicament, but now we will at least prevent continuing the process with undefined data representing the promises. This could clear up some existing and hard-to-replicate bugs but we won't likely be able to tie them back to this change.

@belcherj @sirbrillig I think is out of office at the moment; sorry to change the PR after you approved it. the newest changes should only affect behavior when the app is already in a failure mode

dmsnell:
never mind @sirbrillig and @belcherj - I reset the branch because the additional changes were too big and also causing spurious errors in testing

dmsnell:
Not relevant in light of #1598

PR #1580, opened 2019-09-21
Add Tests to Checkbox component

belcherj:
### Fix
This PR does the following to the Checkbox component:
- Small css cleanup
- Adds tests

### Test
1. Add checkbox to a note
2. Check it off
3. Uncheck it

### Review

### Release

- Added tests to Checkbox component

### DIFF
--- a/lib/components/checkbox/test.js
+++ b/lib/components/checkbox/test.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import Checkbox from './';
+import renderer from 'react-test-renderer';
+
+it('renders Checkbox correctly when checked', () => {
+  const tree = renderer.create(<Checkbox checked={false}></Checkbox>).toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it('renders Checkbox correctly when not checked', () => {
+  const tree = renderer.create(<Checkbox checked={true}></Checkbox>).toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it('should call onChange prop when span is clicked', () => {
+  const noop = jest.fn();
+  const output = renderer.create(<Checkbox onChange={noop}></Checkbox>);
+  output.root.findByType('span').props.onClick();
+  expect(noop).toHaveBeenCalled();
+});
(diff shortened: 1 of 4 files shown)


lib/components/checkbox/test.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Checkbox from './';
+import renderer from 'react-test-renderer';
+
+it('renders Checkbox correctly when checked', () => {
```
dmsnell: just noting that this test doesn't validate that the Checkbox correctly renders but rather that it renders the same as it previously did. for snapshot testing I personally like naming the tests in a way that corresponds to that…

```js
it( `hasn't had its output unintentionally altered`, … );
```

on the other hand I always find intentional unit tests to be valuable as they can identify the behaviors we intend and consider correct

```js
it( 'renders the circle icon when unchecked', () => {
  const checkbox = shallow( <Checkbox checked={false}></Checkbox> );
  expect( checkbox.someWhere( n => n.hasClass( 'icon-circle' ) ) ).toBeTrue();
  expect( checkbox.someWhere( n => n.hasClass( 'icon-checkmark' ) ) ).toBeFalse();
} );

it( 'renders the checkmark icon when checked', () => {
  const checkbox = shallow( <Checkbox checked={false}></Checkbox> );
  expect( checkbox.someWhere( n => n.hasClass( 'icon-circle' ) ) ).toBeFalse();
  expect( checkbox.someWhere( n => n.hasClass( 'icon-checkmark' ) ) ).toBeTrue();
} );
```

Obviously these tests too are deficient but they should stand more resilient than the snapshot tests which will fail with a high false positive rate (for instance, this test will fail when the checkmark icon gets updated even though the behavior of the app is still good).

Personally I'd be happy to see one of those two options taken here: rename the snapshot tests to indicate what they are doing (they are definitely _not_ telling us if the render is "correct"); or test specific behaviors we want to enforce and allow unrelated changes to not fail the tests.

belcherj: I agree. Tests updated.

dmsnell: cool - thanks!

Issue #1568, opened 2019-09-12
Web: Autodetect Dark Mode in Safari


codebykat:
### What
There's apparently now a way for websites to support light and dark themes in Safari via a meta tag. This came up in conversation so I'm not sure I thoroughly grok it but it appears to be this: https://webkit.org/blog/8840/dark-mode-support-in-webkit/

### Why
Feature request from a user :) If you use dark mode I guess this means the webapp would always match your system settings.

belcherj:
Done and done

PR #1581, opened 2019-09-23
Make system setting the default theme preference

belcherj:
### Fix
https://github.com/Automattic/simplenote-electron/issues/1568

This PR makes the following changes:
- Add a system option for the theme setting
- Makes system the default for the theme setting

Note: This only applies to the web since our current version of Electron and Electron Builder do not support getting theme settings from the OS.

### Test
1. Have os dark mode enabled, have app theme setting set to light
2. The app should be in light mode
3. Have os dark mode disabled, have app theme setting set to dark
4. The app should be in dark mode
5. Have os dark mode enabled, have app theme setting set to system
6. The app should be in dark mode
7. Have os dark mode disabled, have app theme setting set to system
8. The app should be in light mode

Screen Shot 2019-09-23 at 3 01 12 PM

### Review

### Release
- Add ability to select system as a theme option and make it the default


desktop/menus/view-menu.js
```diff
@@ -106,6 +106,10 @@ const buildViewMenu = settings => {
             label: '&Dark',
             id: 'dark',
           },
+          {
+            label: '&System',
+            id: 'system',
+          },
```
dmsnell: Just a thought, but I've seen the "system" setting at the top of the list in other apps - especially if it's the default

lib/dialogs/settings/panels/display.jsx
```diff
@@ -104,6 +104,7 @@ const DisplayPanel = props => {
       >
         <Item title="Light" slug="light" />
         <Item title="Dark" slug="dark" />
+        <Item title="System" slug="system" />
```
dmsnell: This will appear in the web app too. Do we care that it may or may not have any impact there?

lib/utils/get-theme.js
```diff
@@ -0,0 +1,12 @@
...
+    if (isElectron) {
+      const { remote } = __non_webpack_require__('electron'); // eslint-disable-line no-undef
+      return remote.systemPreferences.isDarkMode ? 'dark' : 'light';
+    }
+    return window.matchMedia('(prefers-color-scheme: dark)').matches
+      ? 'dark'
+      : 'light';
+  }
+  return themeSetting;
+};
```
dmsnell: `isElectron` is something that won't change/can't change during runtime. because of that we won't need to pass it in to the function call on every invocation.

maybe it's easier to pass it in from `app.jsx` where we make the check, though it seems awkward.

it's probably also awkward to think about defining this function in `app.jsx`, but it would remove the arg-passing and place this directly next to where it's called (or in the same file at least vs. through a redirect)

okay so this is bad but here's an idea. it hooks into the point where we are supposedly initializing Electron specifics and replaces the `getTheme` function; by keeping it contained in `app.jsx` it removes some abstraction that maybe doesn't need to be there. it's less testable, but we're already depending on `remote.systemPreferences` and `window.matchMedia` which are themselves difficult to test

```js
class App extends Component {

  initializeElectron = () => {
    const { remote } = __non_webpack_require('electron');

    // replace theme selection to get system setting
    this.getTheme = () => {
      const { settings: { theme } } = this.props;

      return 'system' === theme
        ? ( remote.systemPreferences.isDarkMode ? 'dark' : 'light' ) 
        : theme;
    }

    this.setState( … );
  }

  …

  getTheme = () => {
    const { settings: { theme } } = this.props;

    return 'system' === theme
      ? ( window.matchMedia( '(prefers-color-scheme: dark)' ).matches ? 'dark' : 'light' )
      : theme;
  }

  …
}
```

broken apart…

```js
getSystemColorMode = () => window.matchMedia('(prefers-color-scheme: dark)').match ? 'dark' : 'light';

getTheme = () => {
  const { settings: { theme } } = this.props;

  return 'system' === theme
    ? this.getSystemColorMode()
    : theme;
}

initializeElectron = () => {
  …

  this.getSystemColorMode = () => remote.systemPreferences.isDarkMode ? 'dark' : 'light';

  …
}
```

codebykat:
🤔 I just tested this and it seems to think my OS is in dark mode, though it isn't:
Screen Shot 2019-10-04 at 11 42 48 AM

Switching to light or dark mode works. If I switch to Light Mode and then System it changes it to dark.

Changing to dark mode and back doesn't change this behavior.

I'm on OSX Mojave 10.14.6.

desktop/menus/view-menu.js
belcherj: Done

desktop/menus/view-menu.js
```diff
@@ -98,6 +98,10 @@ const buildViewMenu = settings => {
       {
         label: 'T&heme',
         submenu: [
+          {
+            label: '&System',
```
dmsnell: Is this the label we use on the other apps? Is `System` enough to explain `System setting`?

belcherj: I struggled with what to call this option for the setting. Is it `Default`? Is it `OS Setting`? System seemed to be the best answer.

Issue #1566, opened 2019-09-12
New notes should open immediately 


forteller:
Thank you for a great product!

### What
When the Linux application is resized so that there is no sidebar but only a list of notes (or a note itself), clicking on the New Note button only creates a new note in the list, and you have to manually click on it to open it and start typing.

### Why
It is very counterintuitive, and not at all usefull, that you have to open the new note after creating it. The thing you expect after clicking the butto is to be able to start typing immediately, which is the case when the application window is sized large enough that there is a sidebar. This throws me a tiny bit off every time it happens.

Thank you!

SylvesterWilmott:
@belcherj this is a great idea.

dmsnell:
Resolved in #1582

forteller:
Thanks for fixing this! But it only works when clicking the New note icon. When using Ctrl+N it still exhibits the old behavior. Hope this can be fixed. Thanks!

PR #1582, opened 2019-09-23
Open new note upon creation

belcherj:
### Fix
https://github.com/Automattic/simplenote-electron/issues/1566

This PR updates the logi so that clicking new note not only creates that note but opens it. This was only a problem on smaller screens where only the note or the note list show.

### Test
1. Decrease the viewport of your browser or electron app so that it is single-paned
2. Click the new note icon
3. Did the new open so that you can edit?
4. Repeat 1 to 3 using production
5. Observe note does not open

### Review

### Release

- Open new note automatically upon creation [1566](https://github.com/Automattic/simplenote-electron/issues/1566)

### DIFF
--- a/lib/search-bar/index.jsx
+++ b/lib/search-bar/index.jsx
@@ -41,10 +41,11 @@ const mapStateToProps = ({ appState: state }) => ({
   showTrash: state.showTrash,
 });
 
-const mapDispatchToProps = (dispatch, { noteBucket }) => ({
+const mapDispatchToProps = (dispatch, { noteBucket, onNoteOpened }) => ({
   onNewNote: content => {
     dispatch(search({ filter: '' }));
     dispatch(newNote({ noteBucket, content }));
+    onNoteOpened();
     recordEvent('list_note_created');
   },
   onToggleNavigation: () => dispatch(toggleNavigation()),
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,7 @@
 - Hovering over a clickable or editable UI element now show the correct cursor for its type
 - Added tests to Checkbox component
 - Renamed CHANGELOG.md to RELEASE-NOTES.txt
+- Open new note automatically upon creation
 
 ## [v1.7.0](https://github.com/Automattic/simplenote-electron/releases/tag/v1.7.0) (2019-08-12)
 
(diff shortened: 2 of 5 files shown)


lib/search-bar/index.jsx
```diff
@@ -41,10 +41,11 @@ const mapStateToProps = ({ appState: state }) => ({
   showTrash: state.showTrash,
 });
 
-const mapDispatchToProps = (dispatch, { noteBucket }) => ({
+const mapDispatchToProps = (dispatch, { noteBucket, onNoteOpened }) => ({
   onNewNote: content => {
     dispatch(search({ filter: '' }));
     dispatch(newNote({ noteBucket, content }));
+    onNoteOpened();
```
dmsnell: huh. would it be worth creating a new semantic action for this? `onNoteOpened()` leaves me flabbergasted when trying to understand what it's supposed to be doing. on the other hand, some action like `openNote()`, better `openNoteView()`, or something else that makes sense could clear it up some. is it worth the extra steps intros PR? not sure.

of note, `onNoteOpened` updates state to `isNoteOpen: true` which ends up being passed around and then sets a CSS class name - nothing else.

seems like we can go back to some of the naming troubles at the time and realize we lack a few simple terms to describe the views.

further, is there any time we would want to create a new note without opening up the note editor view? we've passed this prop down through the UI hierarchy but we're already calling `newNote()` in `app-state` - we could also listen for that action in `app-state` and _also_ update the current view.

for previous work see https://github.com/Automattic/simplenote-electron/pull/443/files#diff-ad1834ceae17437ce4883908cceeb573

belcherj: My concern is that doing so will take us down a path that will require an large effort and significant refactor that is not within the scope of this sprint.  

I used the existing method for opening the note:  https://github.com/Automattic/simplenote-electron/blob/develop/lib/note-list/index.jsx#L172

I would like to clean up the redux and flux state but should we stall this fix till that can be completed?

dmsnell: sure. that's fine as long as we acknowledge it. it caught my eye because of the duplicated logic and the hard time I have knowing what "onNoteOpened" _should_ mean.

👍

belcherj: 100% agree

PR #1594, opened 2019-09-24
Build: stop failing the build on eslint errors

dmsnell:
`eslint` provides helpful warnings that prevent bugs in our software.
However by running it as a precondition for the build we introduce
artificial obstacles to quick testing and iteration. For instance, we
will fail to build the app if we have an unused variable even though
there's no reason it would break the build.

In this change we're removing the precondition for passing `eslint`
before building the app. This should allow us to more rapidly made small
debugging/testing adjustments while working on fixing or altering the
app.

`eslint` still belongs in the process and we might follow-up on this PR
by introducing it as a `git` hook on `precommit`, a more appropriate
place for the safety check. For now though I am foregoing that work in
order to remove what has been a persistent frustration while working on
the app.

**This should have no visual or functional change on the app. This should
only impact the build of the application while developing and debugging.**

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -28,21 +28,6 @@ module.exports = () => {
     module: {
       rules: [
         {
-          test: /lib\/.+\.jsx?$/,
-          exclude: /node_modules|lib\/simperium/,
-          enforce: 'pre',
-          use: [
-            {
-              loader: 'eslint-loader',
-              options: {
-                cache: true,
-                configFile: '.eslintrc.json',
-                quiet: true,
-              },
-            },
-          ],
-        },
-        {
           test: /\.jsx?$/,
           exclude: /node_modules\/core-js/,
           use: [
--- a/package-lock.json
+++ b/package-lock.json
@@ -5990,19 +5990,6 @@
         }
       }
     },
-    "eslint-loader": {
-      "version": "2.2.1",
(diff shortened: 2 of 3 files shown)


beaucollins:
Related #1596

belcherj:
> eslint still belongs in the process and we might follow-up on this PR
by introducing it as a git hook on precommit, a more appropriate
place for the safety check. For now though I am foregoing that work in
order to remove what has been a persistent frustration while working on
the app.

This is on my list of wants.  Seems like an easy GitHub Action. I will follow up with a precommit hook as I've written that code a number of times. Probably won't get to the server check for a while.

webpack.config.js
```diff
@@ -27,21 +27,6 @@ module.exports = () => {
     },
     module: {
       rules: [
-        {
-          test: /lib\/.+\.jsx?$/,
-          exclude: /node_modules|lib\/simperium/,
-          enforce: 'pre',
-          use: [
-            {
-              loader: 'eslint-loader',
```
sirreal: This is likely a dependency that can be removed as well.

dmsnell: Thanks. I'll update.

dmsnell: removed dependency in 065321e

beaucollins:
`lint-staged` does the trick for us https://github.com/Automattic/happy-tools/blob/d3df67756ceab0350bd0588a504ac2cf4c1f3e09/bin/pre-commit

PR #1596, opened 2019-09-26
Fix ESLint Errors

beaucollins:
### Fix

- Fixes all ES Lint errors that exist in master, they were all `no-use-before-define`

Generally using the arrow function syntax is more concise, however it means the functions are defined as constants which means they aren't hoisted in the runtime, therefore we get a `no-use-before-define`.

In each case the arrow functions that caused this problem could be changed to boring old `functions`.

The one other case was use of the `BucketStore` constructor before the class definition. The `StoreProvider` and `BucketStore` have been split into their own modules and the import order has been fixed.

### Test

1. `make run`
2. If you can edit and sync a note all codepaths in this change have been executed.

To test the dropzone change

1. Open simplenote
2. Go to settings -> tools -> import -> plain text files
3. Drop a `.txt` file in the dropzone
4. You should see a new note

### Review

There should be no noticeable changes to the application.

### Release

### DIFF
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -219,12 +219,12 @@ let props = {
 };
 
 // If we sign in with a different username, ensure storage is reset
-const resetStorageIfAccountChanged = newAccountName => {
+function resetStorageIfAccountChanged(newAccountName) {
   const accountName = get(store.getState(), 'settings.accountName', '');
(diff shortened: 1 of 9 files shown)


beaucollins:
From what I can tell `develop` no longer reports these as errors. So we can abandon these changes.

PR #1598, opened 2019-09-27
Fix: Simperium sync issues

### LINKED ISSUE #1579: Sync: diff_match_patch crashes when rebasing
`diff_match_patch` crashes when rebasing an update while there are local changes. I am not certain yet what is causing the crash but I can reproduce it with a minimal test.

 1. Open note with contents `AC` in app Left and app Right
 2. Disconnect Left from the network
 3. In Left, update note to `ACD` by adding `D` at the end
 4. In Right, update note to `ABC` by inserting `B` in the middle
 5. Reconnect Left and wait for update from server

This is a _non-conflicting_ merge with changes in the local client that should be rebased with the update from the server but during the process `diff_match_patch` is crashing…

Screen Shot 2019-09-20 at 5 11 39 PM

The resulting note revision becomes `ACDD` which _never_ existed and is _whole incorrect_. Likely this is related to murky data flows in the app updating based on shared mutable state.

---

I have confirmed that **the iOS clients are not suffering from this error**. When recreating the steps to produce the invalid sequence I found the expected `ABCD` as a result.

beaucollins:
**Status**

Ready to merge 🎉 

---

Closes #1579 

This PR changes the way that the editor's contents and Simperium interact; it's dependent on the changes made in `simperium@1.0.0`.

Previously we held content in the editor component and would delay sending those changes to the underlying Simperium client library (`node-simperium`) until the timeout flushed. This made for awkward interactions when we would received remote updates from the server that happened before that timeout flushed; this led to note corruption.

After this PR we are announcing immediately to the underlying client library when we make changes and telling it (via `{ sync: false }` in `noteBucket.update()`) that it's okay to wait to synchronize those changes with the server. We still maintain our debounce, but instead of sending the editor component's text on the timeout we more simply tell the underlying library to flush its changes (via `noteBucket.touch()`). The effect is that we've removed a source of concurrency bugs with those remote updates.

After this PR we should be removing a number of related data loss and note corruption issues all relating to the interplay of queued and received changes.

**Testing**

See detailed testing instructions in p2XJRt-1jm-p2.
Most testing needs are to open up builds of the app with these changes and to exercise as many update interactions as possible. In #1579, for example, we triggered these updates via the following test:

Open the app on two devices: call one `Local` and the other `Remote`

1. Create a note with contents `AC` and let it synchronize to both apps/devices
2. Disconnect `Local` from the network
3. On `Local` add a `D` to the end of the note to make `ACD`
4. With `Remote` still connected to the network, add a `B` in the middle to make `ABC`
5. After a few seconds (to ensure that `Remote` synchronizes with the server) re-connect `Local`
6. Verify the state of the note on both apps/devices

Before this change we were finding variations of `ACDD`, `ACD`, `AC`, and more incorrect end-states. After this change we should find what we expect: `ABCD`

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -151,7 +151,12 @@ export const App = connect(
         .on('index', this.onNotesIndex)
         .on('update', this.onNoteUpdate)
         .on('update', debounce(this.onNotesIndex, 200, { maxWait: 1000 })) // refresh notes list
-        .on('remove', this.onNoteRemoved);
+        .on('remove', this.onNoteRemoved)
+        .beforeNetworkChange(noteId =>
+          this.props.actions.onNoteBeforeRemoteUpdate({
+            noteId,
+          })
+        );
 
       this.props.preferencesBucket.on('update', this.onLoadPreferences);
 
@@ -322,6 +327,10 @@ export const App = connect(
         content,
       });
 
+    syncNote = noteId => {
+      this.props.noteBucket.touch(noteId);
+    };
+
     // gets the index of the note located before the currently selected one
     getPreviousNoteIndex = note => {
       const filteredNotes = filterNotes(this.props.appState);
@@ -411,6 +420,7 @@ export const App = connect(
                 onNoteClosed={() => this.setState({ isNoteOpen: false })}
                 onNoteOpened={() => this.setState({ isNoteOpen: true })}
                 onUpdateContent={this.onUpdateContent}
+                syncNote={this.syncNote}
               />
               {state.showNoteInfo && <NoteInfo noteBucket={noteBucket} />}
             </div>
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -417,6 +417,28 @@ export const actionMap = new ActionMap({
       },
     },
 
+    /**
+     * A note is being changed from somewhere else! If the same
+     * note is also open and being edited, we need to make sure
+     * any in-memory changes don't get blown away. This is our
+     * chance to tell node-simperium what we want the note to
+     * be.
+     *
+     * node-simperium will compare these changes with the changes
+     * from the server and merge them together.
+     */
+    onNoteBeforeRemoteUpdate: {
+      creator({ noteId }) {
+        return (dispatch, getState) => {
+          const state = getState();
+          if (state.selectedNoteId !== noteId) {
+            return null;
+          }
+          return getState().note.data;
+        };
+      },
+    },
+
     updateNoteContent: {
       creator({ noteBucket, note, content }) {
         return (dispatch, getState) => {
@@ -424,8 +446,11 @@ export const actionMap = new ActionMap({
             note.data.content = content;
             note.data.modificationDate = Math.floor(Date.now() / 1000);
 
-            // update the bucket and sync
-            noteBucket.update(note.id, note.data);
+            // update the bucket don't sync right away
+            // as this happens per keystroke when the user is editing
+            // a note. The NoteEditor notify via props when its time
+            // to sync via Simperium
+            noteBucket.update(note.id, note.data, {}, { sync: false });
 
             // Check if note is still selected (to avoid race conditions)
             if (get(getState().appState, 'note.id') === note.id) {
--- a/lib/note-detail/index.jsx
+++ b/lib/note-detail/index.jsx
@@ -12,7 +12,7 @@ import SimplenoteCompactLogo from '../icons/simplenote-compact';
 import renderToNode from './render-to-node';
 import toggleTask from './toggle-task';
 
-const saveDelay = 2000;
+const syncDelay = 2000;
 
 export class NoteDetail extends Component {
   static displayName = 'NoteDetail';
@@ -23,6 +23,7 @@ export class NoteDetail extends Component {
     fontSize: PropTypes.number,
     isViewingRevisions: PropTypes.bool.isRequired,
     onChangeContent: PropTypes.func.isRequired,
+    syncNote: PropTypes.func.isRequired,
     onNotePrinted: PropTypes.func.isRequired,
     note: PropTypes.object,
     noteBucket: PropTypes.object.isRequired,
@@ -41,7 +42,7 @@ export class NoteDetail extends Component {
   };
 
   componentWillMount() {
-    this.queueNoteSave = debounce(this.saveNote, saveDelay);
+    this.queueNoteSync = debounce(this.syncNote, syncDelay);
     document.addEventListener('copy', this.copyRenderedNote, false);
   }
 
@@ -51,7 +52,7 @@ export class NoteDetail extends Component {
     this.props.storeHasFocus(this.hasFocus);
 
     // Ensures note gets saved if user abruptly quits the app
-    window.addEventListener('beforeunload', this.queueNoteSave.flush);
+    window.addEventListener('beforeunload', this.queueNoteSync.flush);
 
     if (previewingMarkdown) {
       this.updateMarkdown();
@@ -64,8 +65,14 @@ export class NoteDetail extends Component {
 
   isValidNote = note => note && note.id;
 
-  componentWillReceiveProps() {
-    this.queueNoteSave.flush();
+  componentWillReceiveProps(nextProps) {
+    const isEditingNote = get(this.props, ['note', 'id'], false);
+    if (isEditingNote === false) {
+      return;
+    }
+    if (get(nextProps, ['note', 'id']) !== isEditingNote) {
+      this.queueNoteSync.flush();
+    }
   }
 
   componentDidUpdate(prevProps) {
@@ -90,7 +97,7 @@ export class NoteDetail extends Component {
   }
 
   componentWillUnmount() {
-    window.removeEventListener('beforeunload', this.queueNoteSave.flush);
+    window.removeEventListener('beforeunload', this.queueNoteSync.flush);
     document.removeEventListener('copy', this.copyRenderedNote, false);
   }
 
@@ -153,7 +160,16 @@ export class NoteDetail extends Component {
     if (!this.isValidNote(note)) return;
 
     this.props.onChangeContent(note, content);
+    this.queueNoteSync();
     analytics.tracks.recordEvent('editor_note_edited');
+  };
+
+  syncNote = () => {
+    const { note } = this.props;
+
+    if (!this.isValidNote(note)) return;
+
+    this.props.syncNote(note.id);
   };
 
   storeEditorHasFocus = f => (this.editorHasFocus = f);
@@ -222,7 +238,7 @@ export class NoteDetail extends Component {
                   noteId={get(note, 'id', null)}
                   content={content}
                   filter={filter}
-                  onChangeContent={this.queueNoteSave}
+                  onChangeContent={this.saveNote}
                 />
               </div>
             )}
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,16 +2,30 @@
 
 ## Future Release
 
-- Updated Log in and Sign up form to match current styling
-- Hovering over a clickable or editable UI element now show the correct cursor for its type
-- Added tests to Checkbox component
-- Renamed CHANGELOG.md to RELEASE-NOTES.txt
-- Open new note automatically upon creation
-- Open new note automatically upon creation [1566](https://github.com/Automattic/simplenote-electron/issues/1566)
-- Updated colors to use Color Studio, the color palette for Automattic products
+### Enhancements
+
+- Open new note automatically upon creation [1582](https://github.com/Automattic/simplenote-electron/pull/1582)
+- Updated colors to use Color Studio, the color palette for Automattic products [#1565](https://github.com/Automattic/simplenote-electron/pull/1565)
+
+### Fixes
+
+- Hovering over a clickable or editable UI element now show the correct cursor for its type [#1573](https://github.com/Automattic/simplenote-electron/pull/1573)
 - Fixes vertical spacing with nested markdown lists
-- Added a GitHub Action to deploy develop and master branches
 - Fixes sort order on revision slider when the timestamps don't match the change sequence [#1605](https://github.com/Automattic/simplenote-electron/pull/1605)
+- Prevents note corruption when receiving remote updates when local updates are pending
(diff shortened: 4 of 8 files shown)


lib/app.jsx
```diff
@@ -151,7 +151,12 @@ export const App = connect(
         .on('index', this.onNotesIndex)
         .on('update', this.onNoteUpdate)
         .on('update', debounce(this.onNotesIndex, 200, { maxWait: 1000 })) // refresh notes list
-        .on('remove', this.onNoteRemoved);
+        .on('remove', this.onNoteRemoved)
+        .subscribe(noteId =>
```
beaucollins: I'm happy to change this to be something analogous to `.on( 'beforeNetworkApplyChange', ...)`.

The result will probably mean passing a function in that `onNoteBeforeRemoteUpdate` will need to call with the latest data.

lib/flux/app-state.js
```diff
@@ -397,49 +396,15 @@ export const actionMap = new ActionMap({
...
-          const { original, patch } = remoteUpdateInfo;
+          const { patch } = remoteUpdateInfo;
 
           debug('noteUpdatedRemotely: %O', data);
 
           if (state.selectedNoteId !== noteId || !patch) {
             return;
           }
 
-          // working is the state of the note in the editor
```
beaucollins: This is all handled by `node-simperium`.

lib/flux/app-state.js
```diff
@@ -397,49 +396,15 @@ export const actionMap = new ActionMap({
       creator({ noteBucket, noteId, data, remoteUpdateInfo = {} }) {
         return (dispatch, getState) => {
           var state = getState().appState;
-          const { original, patch } = remoteUpdateInfo;
+          const { patch } = remoteUpdateInfo;
```
beaucollins: The `patch` should still be used to maintain the cursor location between updates.

lib/flux/app-state.js
```diff
@@ -452,6 +417,28 @@ export const actionMap = new ActionMap({
...
 
+    /**
+     * A note is being changed from somewhere else! If the same
+     * note is also open and being edited, we need to make sure
+     * any in-memory changes don't get blown away. This is our
+     * chance to tell node-simperium what we want the note to
+     * be.
+     *
+     * node-simperium will compare these changes with the changes
+     * from the server and merge them together.
```
dmsnell: this is a wonderful beautiful comment

lib/flux/app-state.js
```diff
@@ -452,6 +417,28 @@ export const actionMap = new ActionMap({
...
+     */
+    onNoteBeforeRemoteUpdate: {
+      creator({ noteId }) {
+        return (dispatch, getState) => {
+          const state = getState();
+          if (state.selectedNoteId !== noteId) {
+            return null;
+          }
+          return getState().note.data;
```
dmsnell: sadly we need more changes to the app because we're trapping content inside the `debounce` of `<NoteDetail />`

what's _encouraging_ is that I think we can completely get rid of the debounce inside the editor and move it here or somewhere else. what can happen while these async things are floating in lala land is that we can receive a remove update from the network, send a copy of our note from the app memory to simperium via `onNoteBeforeRemoteUpdate`, get the `update` from `node-simperium` and update the editor contents…_and then_ as we update the editor and _after_ simperium has finished doing its thing the editor says, "oh hey, I have changes waiting to send, I'll push them out now." this cycles back up the app and flushes to Simperium and ends up wiping out the new changes

```
= START =
simperium: AC
app: AC
editorDebounce: [empty]

= TYPE =
simperium: AC
app: AC
editorDebounce: ACD

= NETWORK UPDATE =
simperium: ABC
app: AC
editorDebounce: ACD

= BEFORE NETWORK CHANGE =
app: "I have 'AC'"
simperium: "Okay, no changes, apply update and we now have ABC"
editorDebounce: ACD

= UPDATE =
simperium: ABC "hey app, this is the new version"
app: ABC "okay, thanks, I'll replace the text in my editor"
editorDebounce: ACD

= UPDATE EDITOR COMPONENT PROPS =
app: ABC "hey I have changes, I should flush them"

= DEBOUNCE FLUSH =
simperium: ACD [-B]
app: ABC
editorDebounce: [empty]

= SIMPERIUM RESPONDS =
simperium: ACD
app: ACD
editorDebounce: [empty]
```

I predict this cycle. This patch is probably _better_ than what we have, but I think we need to make sure we can flush the editor debounce during this `subscribe` otherwise we'll immediately undo the update from the network

beaucollins: Ideally the debounce would be related to UI flushes while the state can be manipulated freely (isn't React suspense supposed to handle this in the future anyway).

dmsnell: _If_ we remove the debounce from the editor itself then it should mirror the app's note state and we can think of it as always being immediately in sync.

_Then_ if we go down ten lines in `updateNoteContent` (which will now be called on every keypress or note content change) we can wrap `noteBucket.update` in the debounce. We'd have to expose that debounce somehow to `onNoteBeforeRemoteUpdate` so it can cancel it though.

this would have to happen _per note_ and _might_ be reasonable to share via `app.jsx`.

```js
// app.jsx
class NoteChangeQueue {
	constructor( noteBucket ) {
		this.queue = new Map();
		this.noteBucket = noteBucket;
	}

	add( noteId, data ) {
    this.remove( noteId );

		const runner = setTimeout( () => {
			const [ hasQueue, noteData ] = this.remove( noteId );
			if ( hasQueue ) {
				this.noteBucket.update( noteId, noteData );
			}
		}, 2000 );

		this.queue.add( noteId, [ runner, data ] );
  }

	remove( noteId ) {
		const hasQueue = this.queue.has( noteId );
		const [ runner, noteData ] = hasQueue ? this.queue.get( noteId ) : [ null, null ];

		if ( runner ) {
			clearTimeout( runner );
		}

		this.queue.delete( noteId );
		return [ hasQueue, noteData ];
	}
}

…

this.queuedNotes = new NoteChangeQueue( this.props.noteBucket );

this.props.noteBucket.on(…).….subscribe( noteId => {
	this.props.actions.onNoteBeforeRemoteUpdate({
		noteId,
		queuedNotes: this.queuedNotes,
	})
} )

…

onUpdateContent = ( note, content ) =>
	this.props.actions.updateNoteContent( {
		noteBucket: this.props.noteBucket,
		note,
		content,
		queuedNotes: this.queuedNotes
	} );


// app-state.js
onNoteBeforeRemoteUpdate: {
	creator({ noteId, queuedNotes }) {
		return (dispatch, getState) => {
			…

			const [ wasQueued, noteData ] = queuedNotes.remove( noteId );

			// if we have changes we haven't yet sent to Simperium then return
			// them as the new `local` value, but otherwise use the existing app state
			return wasQueued ? noteData : getState().note.data;
		}
	} 
}

updateNoteContent: {
	creator({ noteBucket, note, content, queuedNotes }) {
		return (dispatch, getState) => {
			…

			// update the bucket and sync
			// noteBucket.update(note.id, note.data); <<< no longer this
			// instead we queue up the changes and update app state
			queuedNotes.add( note.id, note.data );
			dispatch(update(getState(), {
				note: { $set: note.data }
			})
			…
		};
	}
}
```

dmsnell: > Ideally the debounce would be related to UI flushes while the state can be manipulated freely (isn't React suspense supposed to handle this in the future anyway).

Maybe, but we still need to rearrange where the debounce happens and what it does else we'll repeat the bug I have above.

```js
// app-state.jsx
updateNoteContent: {
	creator({ noteBucket, note, content }) {
		return (dispatch, getState) => {
			// immediately update app state
			dispatch(update(getState(), {note: {$set: {...note.data,content}}}))
		}
	}
}

// because this is just notifying the `node-simperium` library we don't have to
// do anything. our app state is already updated and won't change just because
// we submit the changes. we only care about updates through `.on( 'update' )`
const sendNoteToSimperium = note => noteBucket.update( note.id, note.data );

// note-detail/index.js
saveNote = content => {
	const { note } = this.props;

	if ( ! this.isValidNote( note ) ) {
		return;
	}

	this.props.onNoteChange( note, content ); // immediately update app state
	this.queueUpdate( note.id );
}

queueUpdate = debounce( () => {
	const { note } = this.props;

	// tell the app to run noteBucket.update
	this.props.sendNoteToSimperium( note );
}, 2000 )

componentWillUnmount() {
	// when closing this note make sure we get our updates to Simperium
	this.queueUpdate.flush();
}
```

this way I think would require fewer changes and fewer new concepts - a new updater function in `app-state` that gets passed down to the editor component - and may be closer to the way the mobile libraries work.

beaucollins: > we can wrap noteBucket.update in the debounce. 

This sounds like the correct thing. If it's sending the data to simperium that we want to debounce then debouncing the state was an indirect way of doing that.

> We'd have to expose that debounce somehow to onNoteBeforeRemoteUpdate so it can cancel it though.

Why de we need to be able to cancel it? Calling extra `noteBucket.updates` isn't a problem.


```js
			// if we have changes we haven't yet sent to Simperium then return
			// them as the new `local` value, but otherwise use the existing app state
			return wasQueued ? noteData : getState().note.data;
```

I don't think we need to do this. If we return `false`y from here, the default implementation will use what is in the `noteBucket` as the "current state". If a note isn't open it doesn't have a in-memory representation we need to worry about.

beaucollins:
This is the debounce we're concerned with correct?

https://github.com/Automattic/simplenote-electron/blob/develop/lib/note-detail/index.jsx

lib/flux/app-state.js
beaucollins: This is the debounce we're concerned with correct?

https://github.com/Automattic/simplenote-electron/blob/31e6bc7b2dfaae2059bbee8499308eaaf1898547/lib/note-detail/index.jsx#L44

beaucollins: **This** is where we should "debounce". And _not_ debounce the `update` function itself but  use the `sync: true|false` to reduce sync operations, because that is the point of the debounce right?

https://github.com/Automattic/simplenote-electron/blob/31e6bc7b2dfaae2059bbee8499308eaaf1898547/lib/flux/app-state.js#L463

lib/app.jsx
```diff
@@ -327,6 +327,13 @@ export const App = connect(
         content,
       });
 
+    onSyncNote = noteId => {
```
dmsnell: I know we have precedent with all of our other actions, but `onSyncNote` seems misleading here. in these cases we're not responding to things but rather manually triggering them. did you consider naming the prop `syncNote`?

lib/flux/app-state.js
```diff
@@ -439,15 +439,27 @@ export const actionMap = new ActionMap({
       },
     },
 
+    syncNote: {
+      creator({ noteBucket, noteId }) {
+        console.log('sync the note');
```
dmsnell: if we do this we should add `debug()` - taking out the log statement altogether is also fine

lib/flux/app-state.js
```diff
@@ -439,15 +439,27 @@ export const actionMap = new ActionMap({
...
+        return () => {
+          console.log('syncing');
+          noteBucket.touch(noteId);
+        };
```
dmsnell: I'm pretty sure we don't need this thunk here and can just run the code and return

on that note when I was visualizing the change I had `syncNote()` inside of `app.jsx` - it doesn't change the app state in any way and it's independent from all the data flow in the app, therefore I think it may just complicate things by adding the additional step.

do you have strong feelings either way? keeping it here as a new action creator vs. leaving it in `app.jsx`

```js
syncNote = noteId => {
  const { noteBucket } = this.props;

  noteBucket.touch( noteId );
};

<AppLayout
	…
	syncNote={ this.syncNote }
	…
/>
```

lib/flux/app-state.js
dmsnell: all this sounds good. _if_ we inform `node-simperium` of our changes immediately upon making them then there will never be a time when we have a different copy of the local data and therefore no need to send an updated copy of the local changes `beforeNetworkChange`. well, that's true _mostly_ but while we still have indexedDB in the picture we have very small slivers of time when we could have multiple copies - that's nothing to worry about here.

lib/flux/app-state.js
beaucollins: > do you have strong feelings either way?

I was naively copying the action for updating the note content so no strong opinions.

dmsnell: if you have no strong opinion let's keep it in `app.jsx` since we don't have any real reason to add the indirection

if it impacted app state at all or didn't represent something that might be happening in the background anyway I'd feel more compelled to put it in an action creator, but `app-state` is already confusing enough

lib/note-detail/index.jsx
```diff
@@ -64,8 +65,11 @@ export class NoteDetail extends Component {
 
   isValidNote = note => note && note.id;
 
-  componentWillReceiveProps() {
-    this.queueNoteSave.flush();
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.note.id !== this.props.note.id) {
```
dmsnell: can this be called when `note` is undefined.

beaucollins: Most likely, it's JavaScript.

lib/flux/app-state.js
beaucollins: Do you want all of the props to be called `syncNote` as well?

dmsnell: to me that makes sense. it's not even arguable in the way that `onNoteChanged` is, where the author changed the note and the app is responding. in this case it's purely the app acting and doing so actively, so `onNoteSync` leads me to believe there is some other actor involved when actually there isn't

beaucollins: For me when it comes to React components I tend to treat callback functions as a signal. The `NoteDetail` isn't trying to sync the note, but it is saying "whoever is giving me my props, it's time to sync the note".

In that regard, it is not doing the syncing itself but controlling the timing of it. The convention I end up with is: A React components props that are callbacks for future things that happen start with `on`, so `onMySignal`.

Wether the note is synced or not `NoteDetail` no longer worries about that.

Given that `AppLayout` is now doing the call to `noteBucket.touch()` I think having its function named `syncNote` makes the most sense. I'm still partial to component props having `onMySignal`. But no to the point that I'll fight very hard for it.

TL;DR I'll change it all to `syncNote={}`.

PR #1602, opened 2019-09-30
Fix spacing on markdown nested lists

belcherj:
### Fix
When you have a nested Markdown list the spacing is incorrect. This fixes that issue. The issue occurs when an item is rendered inside a paragraph tag.

Prettier also caused a bunch of other changes.

Here is a test markdown snippet:
```
* Fruit
  * Apple
  * Orange
  * Banana
* Dairy
  * Milk
  * Cheese
```

### Test
1. Add snippet from above to a note with markdown
2. Observe note on prod
3. Observe note on this branch
4. There should not be extra space in the lists.

After:
Screen Shot 2019-09-30 at 11 56 57 AM

Before:
Screen Shot 2019-09-30 at 11 57 44 AM

### Review

### Release

- Fixes vertical spacing with nested markdown lists

### DIFF
--- a/lib/note-detail/style.scss
+++ b/lib/note-detail/style.scss
@@ -17,7 +17,7 @@
   display: flex;
   flex-direction: row;
   justify-content: center;
-  transition: all .3s ease-in-out;
+  transition: all 0.3s ease-in-out;
 
   &.is-viewing-revisions {
     padding-top: 51px;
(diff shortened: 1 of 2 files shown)


lib/note-detail/style.scss
```diff
@@ -99,6 +98,10 @@
     margin: 0 0 1.5em;
   }
 
+  ul p {
```
belcherj: This was the only change made by this PR the rest are Prettier

PR #1613, opened 2019-10-02
Refactor `trashNote` to stop directly mutating note object

dmsnell:
See #1614 

should be occurring.

---

Previously we have been directly mutating the note object in the
action creator when trashing a note.

In this patch we're building a new note object when trashing a note in
order to maintain our consistency.

At the same time I have removed the thunk from the action creator since
it wasn't doing anything and didn't need to exist. This shouldn't have
a noticeable impact on the application.

**Testing**

The goal of this PR is to refactor code without changing behaviors, so if we
notice any behavioral changes that will be problematic.

We're affecting note trashing, so build the app at this branch and attempt
to trash and untruth a note and make sure the app updates properly in
response to that.

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -463,14 +463,11 @@ export const actionMap = new ActionMap({
 
     trashNote: {
       creator({ noteBucket, note, previousIndex }) {
-        return dispatch => {
-          if (note) {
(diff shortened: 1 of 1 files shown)


dmsnell:
Closing in favor of combined work in #1615

Issue #1614, opened 2019-10-02
Tracking Issue: Stop directly mutating app state


dmsnell:
The app is ostensibly based on the Redux UI model which assumes atomic state transitions supplying immutable app states. Unfortunately it's the case that in several places throughout the app we are directly mutating that app state. This creates concurrency errors by allowing the app to exist where _partial_ state updates have been applied.

While it's not entirely clear how many, if any, bugs are due to this unidiomatic mutation, it's clear that the mutation further complicates the data flow through the app and makes it hard to determine _if_ these defects are the cause for inconsistencies we observe.

As part of untangling our data flows I'd like to remove all of these places of direct mutation and replace them with idiomatic _immutable swaps_. When this work is finished it should be easier to trace updates throughout the app and we will know that concurrency errors are not due to this concern.

 - [x] `app-state::pinNote` - `note.data.systemTags` ~#1611~ #1615 
 - [x] `app-state::markdownNote` - `note.data.systemTags` #1615
 - [x] `app-state::publishNote` - `note.data.systemTags` #1615
 - [x] `app-state::noteUpdatedRemotely` - `note.data` #1634 
 - [x] `app-state::updateNoteContent` - `note.data.content`, `note.data.modificationDate` #1634 
 - [x] `app-state::trashNote` - `note.data.systemTags` ~#1613~ #1615
 - [x] `domain/notes::updateNoteTags` - `note.data.tags`, `note.data.modificationDate` #1639
 - [x] `domain/tags::renameTag` - `note.data.tags` #1638
 - [x] `domain/tags::trashTag` - `note.data.tags` #1638

belcherj:
Closing as @dmsnell fixed what he found!

PR #1615, opened 2019-10-02
Refactor `systemTag` changes to stop directly mutating note object

dmsnell:
See #1614 

should be occurring.

---

Replaces #1611
Replaces #1613 

Previously we have been directly mutating the note object in the action
creator when toggling a note's pinned status, markdown status, publish
status, and trash status.

In this patch we're building a new note object when toggling these
system tags in order to maintain our consistency.

At the same time I have removed the thunks from the action creators
since they weren't doing anything and didn't need to exist. This
shouldn't have a noticeable impact on the application.

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -7,6 +7,25 @@ import analytics from '../analytics';
 
 const debug = Debug('appState');
 
+const toggleSystemTag = (note, systemTag, shouldHaveTag) => {
+  const {
+    data: { systemTags = [] },
+  } = note;
+  const hasTagAlready = systemTags.includes(systemTag);
+
+  return hasTagAlready !== shouldHaveTag
+    ? {
+        ...note,
+        data: {
+          ...note.data,
+          systemTags: shouldHaveTag
+            ? [...systemTags, systemTag]
+            : systemTags.filter(tag => tag !== systemTag),
+        },
+      }
+    : note;
+};
+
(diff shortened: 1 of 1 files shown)


lib/flux/app-state.js
```diff
@@ -7,6 +7,25 @@ import analytics from '../analytics';
 
 const debug = Debug('appState');
 
+const toggleSystemTag = (note, systemTag, shouldHaveTag) => {
+  const {
+    data: { systemTags = [] },
+  } = note;
+  const hasTagAlready = systemTags.includes(systemTag);
+
+  return hasTagAlready !== shouldHaveTag
```
belcherj: This took me a while to parse.  No idea how it could be any simpler but it took me some time to figure out what was going on.

dmsnell: yeah I don't know the best phrasing. This was the simplest I got.

```js
if ( noteIsAlreadyInAStateWhereThePresenceOrAbsenceOfTheTagMatchesTheExpectationOnceThisFunctionEnds )
```

```js
const tagIsAlreadyInNote = systemTags.includes(systemTag);
const shouldAddTagToNote = ?
```

In my previous work at least there was convention that `should_` indicates a desire to enact the thing and that it's boolean. Here in JS land we only have a single type `any` so the language doesn't help us out like it does in statically-typed languages.

Any recommendation or just noting that it's something that could be better some how?

PR #1627, opened 2019-10-04
Fix: Broken oAuth flow

### LINKED ISSUE #1229: Simplenote users not able to log in with WordPress.com account on Mac, Windows, Linux
Two reports so far for Linux (reproducible on Mac and Windows as well)
Ubuntu 18.10 with 4.18 kernel. Simplenote 1.5.0
Linux Mint and Manjaro. Both are the latest LTS version

I can reproduce this on the Electron Mac version as well, and there is a report for Windows, so it's all 3 Electron apps.

#### Steps to reproduce
1. Try to sign in via WordPress.com on Linux Simplenote app
2. Sign in, click Auth

#### What happened
Linux: "the authorization gets stuck when after clicking "Authorize" button"
Linux: "It goes berserk and issues thousands of requests to `auth`"
Mac Electron: for me, nothing happens at all

#### Simplenote version
1.5.0 (is not an issue in 1.4.1)

### **WORKAROUND**:
You can still log into the Simplenote app with your account details, using your email address and password used to originally sign up.

dmsnell:
Resolves #1229 

We have been experiencing problems when trying to login with the
WordPress.com signin. Something appears to have changed in Electron such
that the older versions of the app still work but newer versions are
failing.

In this patch we're rewriting the authentication flow to simplify it and
prepare ourselves for better handling of the failure cases.

In production we are seeing strange behaviors on failure and some on
success: unending re-requests to `simplenote://auth` which trigger full
CPU load; and no response after authentication.

After this patch we should be able to wrangle in errors and add a
timeout to better communicate when things are failing.

Additionally, the unending loop should be closed due to a replacement of
the old network intercept code with a single simplified model.

---

## Testing

This feature isn't currently working on the `history-analyst-dad` application
configured by default in this repository. In order to test the app you'll have to
build it with a custom `config-local.json` including the details to connect to
the actual production Simplenote system.

 - Logged in with my normal WordPress.com account ✅
 - Created a new WordPress.com account but didn't verify my email address, tried to login here and it failed, responding with a message indicating as much. ✅

 - Logging in to WordPress.com with a non-production app (`history-analyst-dad`) successfully authenticates via oAuth and returns my production-app Simperium token, but that fails when logging in but there's no message. ⭕️ (This is probably best scoped for a separate update as this has been the case before this PR existed and it's not a customer-facing issue, mostly for devs)

### DIFF
--- a/lib/auth/index.jsx
+++ b/lib/auth/index.jsx
@@ -216,84 +216,73 @@ export class Auth extends Component {
   };
 
   setupAuthWindow = () => {
-    const remote = __non_webpack_require__('electron').remote; // eslint-disable-line no-undef
-    const BrowserWindow = remote.BrowserWindow;
-    const protocol = remote.protocol;
+    // eslint-disable-next-line no-undef
+    const { BrowserWindow, session } = __non_webpack_require__(
+      'electron'
+    ).remote;
+
     this.authWindow = new BrowserWindow({
       width: 640,
       height: 640,
       show: false,
       webPreferences: {
         nodeIntegration: false,
+        session: session.fromPartition(`fresh-session-${Math.random()}`),
       },
     });
 
-    // Register simplenote:// protocol
-    protocol.registerHttpProtocol('simplenote', req => {
-      this.authWindow.loadURL(req.url);
+    this.authWindow.on('closed', () => {
+      // make sure to release this from memory
+      this.authWindow = null;
     });
 
-    this.authWindow.webContents.on('will-navigate', (event, url) =>
-      this.onBrowserNavigate(url)
-    );
+    this.authWindow.webContents.session.protocol.registerHttpProtocol(
+      'simplenote',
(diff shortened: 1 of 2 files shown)


lib/auth/index.jsx
```diff
@@ -216,84 +216,73 @@ export class Auth extends Component {
...
-          return this.authError('An error was encountered while signing in.');
+    this.authWindow.webContents.session.protocol.registerHttpProtocol(
+      'simplenote',
+      (req, callback) => {
+        const { searchParams } = new URL(req.url);
+
+        // cancel the request by running callback() with no parameters
+        // we're going to close the window and continue processing the
+        // information we received in args of the simplenote://auth URL
+        callback();
```
roundhill: So this is what fixed the looping issue?

dmsnell: I believe so, that and our manual redirect.

PR #1634, opened 2019-10-08
Refactor updating note content to stop directly mutating note object

dmsnell:
See #1614

should be occurring.

---

Previously we have been directly mutating the note object when updating
its content. This may have been an attempt to work around confusing
data-flow issues that thankfully don't exist anymore. We have also been
performing inline checks to make sure that we update the editor's
contents if we receive these updates.

In this patch we're building a new note object when we update a note
in order to maintain our consistency. In light of #1598 we're also
removing some work-around code that attempted to force consistency when
it didn't exist; that consistency now exists since we're tracking the
underlying Simperium data closely now vs. storing it in separate
places.

When updating checklist items we're forcing a sync so that those changes
will propagate immediately. We don't have a need to debounce those
clicks.

---

## Testing

There should be no significant visual or behavioral changes with this PR. We
adjust updating the note in the editor when receiving remote changes and when
making local changes. We touch the code which toggles checkboxes.

In testing verify that with separate sessions the updates appear as expected.
Toggle checkboxes on and off, from the editing screen and from the Markdown
preview screen as well.
Smoke-test to verify no breakage.

Make some changes, wait a moment, make other changes, wait a few moments more,
and verify that your edits make it through. Repeat but with conflicting edits. This is
to test the debouncing.

## Notes

Since the only "action" that `onUpdateContent` was doing was `loadAndSelectNote`
when the edits occurred on the opened note I have remove that entirely. As of #1598 
those edits should be mirrored exactly in the editor and so we don't need to re-select
it in order to force an update.

### DIFF
--- a/lib/app.jsx
+++ b/lib/app.jsx
@@ -320,12 +320,28 @@ export const App = connect(
       });
     };
 
-    onUpdateContent = (note, content) =>
-      this.props.actions.updateNoteContent({
-        noteBucket: this.props.noteBucket,
-        note,
-        content,
-      });
+    onUpdateContent = (note, content) => {
+      if (!note) {
+        return;
+      }
+
+      const updatedNote = {
+        ...note,
+        data: {
+          ...note.data,
+          content,
+          modificationDate: Math.floor(Date.now() / 1000),
+        },
+      };
+
+      // update the bucket but don't force sync right away
+      // as this happens per keystroke when the user is editing
+      // a note. The NoteEditor will notify via props when
+      // it's time to sync via Simperium
+      const { noteBucket } = this.props;
+
+      noteBucket.update(note.id, updatedNote.data, {}, { sync: false });
+    };
 
     syncNote = noteId => {
       this.props.noteBucket.touch(noteId);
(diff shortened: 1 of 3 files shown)


dmsnell:
Thanks @sirbrillig - that flow actually bothered me and maybe what's causing that distress is the realization that we're not updating local app state with the new note contents. The old one wasn't doing this either, at least not through the recommended means but rather through direct mutation which could be wiped out immediately.

I'll consider this and see if we can't make up for it. All these data flows are still pretty hairy so I suspect that in the process of cleaning them up we may remove some flows, add some flows, add some flows that we removed, etc…

dmsnell:
After more consideration @sirbrillig I'd like to leave this as-is because I think adding the fix you hint at would require substantially more out of this PR than would be manageable. I think it's good work but right now we're mostly updating the code and not changing behavior and I think doing what know we would prefer involve changing behavior.

That is, given the state-of-the-state the app isn't setup to handle `dispatch( updateNoteContents( ) )` and the editor is already linked in to the underlying Simperium store. I'm afraid that if we _added_ what I think we should add, a call to `loadAndSelectNote` then we would risk introducing a cycle into the app whereas right now things are stable enough as is.

I know you didn't require these changes but I thought you had a good point and I wanted to explore the idea.

PR #1638, opened 2019-10-08
Refactor tag operations to stop directly mutating tag objects

dmsnell:
See #1614

should be occurring.

---

Previously we have been directly mutating note and tag objects when
editing those tags. This mutation can lead to concurrency defects which
expose themselves as inconsistent UI state. This breaks our Redux model
which assumes that all UI updates happen atomically.

In this patch we're building new note objects and tag objects when we
make these updates in order to maintain our consistency.

---

There should be no significant visual or behavioral changes with this PR. We
are changing code related to removing tags, renaming tags, and
reordering tags.

In testing verify that with separate sessions the updates appear as expected.
Add, reorder, and remove tags to make sure the changes synchronize.

### DIFF
--- a/lib/state/domain/tags.js
+++ b/lib/state/domain/tags.js
@@ -28,54 +28,47 @@ export const createTag = ({ name }) => {
   tagBucket().update(tagId, { name });
 };
 
-export const reorderTags = ({ tags }) => () => {
-  for (let i = 0; i < tags.length; i++) {
-    let tag = tags[i];
-    tag.data.index = i;
-    tagBucket().update(tag.id, tag.data);
-  }
-};
+export const reorderTags = ({ tags }) => () =>
+  tags.forEach((tag, index) =>
+    tagBucket().update(tag.id, { ...tag.data, index })
+  );
 
-export const renameTag = ({ tag, name }) => (dispatch, getState) => {
-  let oldTagName = tag.data.name;
-  if (oldTagName === name) {
+export const renameTag = ({ tag, name: newName }) => (dispatch, getState) => {
+  const { notes } = getState().appState;
+  const tagName = tag.data.name;
+
+  if (tagName === newName) {
     return;
   }
 
-  let { notes } = getState().appState;
+  tagBucket().update(tag.id, { ...tag.data, name: newName });
 
-  tag.data.name = name;
-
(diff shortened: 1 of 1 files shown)


lib/state/domain/tags.js
```diff
@@ -28,54 +28,47 @@ export const createTag = ({ name }) => {
...
-    let noteTags = note.data.tags || [];
-    let tagIndex = noteTags.indexOf(oldTagName);
-
-    if (tagIndex !== -1) {
-      noteTags.splice(tagIndex, 1, name);
-      note.data.tags = noteTags.filter(noteTag => noteTag !== oldTagName);
-      noteBucket().update(note.id, note.data);
-    }
-  }
+  notes
```
belcherj: Why doesn't this happen in Simperium and then update the client?

belcherj:
Issue:

1. Open two instances of the app in two tabs
2. Edit tag by adding a `b` to the end of a tag
3. Observe that the tag is not updated in the other tab
4. Refresh, observe that the tag is now `xxxxbbbbbbbbb`. It adds a bunch of extra `b`'s

So, two issues:

1. An edited tag does not update in other tabs.
2. An edited tag adds a bunch of duplicate changes upon reload.

Screen Shot 2019-10-09 at 1 13 12 PM
Screen Shot 2019-10-09 at 1 13 19 PM

belcherj:
https://github.com/Automattic/simplenote-electron/issues/1641
https://github.com/Automattic/simplenote-electron/issues/1640

lib/state/domain/tags.js
dmsnell: this is a point where Simperium and Simplenote diverge. Simperium isn't aware of `notes` or `tags` - it only knows about `buckets` and `entities`, so as an application-specific semantic we need to enforce it here.

it's also possible to add supervisory _watchers_ which are app-level admins that can watch the changes as they come in from clients. these watchers could look for tag changes, make additional note updates, and then have those changes distributed, but doing so from a practical standpoint would probably be significantly more complicated than simply repeating the logic here in the clients when it happens.

this is a subjective call and most reasoning probably goes along the lines of "because the data model is so small and contained, duplicating the logic in each client is the most straightforward way to enforce the behaviors"

good question!

PR #1643, opened 2019-10-15
[WIP] Search: Add dropdown with search suggestions

### LINKED ISSUE #1624: Search: Add tag suggestions to search results
See designs at p2XJRt-1k0-p2

Screen Shot 2019-10-04 at 1 07 52 PM

Clicking on the search string magnifying glass will go to the existing results screen.

Clicking on a tag suggestion will go to the results screen with `tag:someTag` filled in.

Related: #54

codebykat:
### Fix
This PR removes the "filter as you type" behavior from the search box in favor of a dropdown suggestion list that includes matching tags.

The first suggestion in the dropdown preserves the existing search behavior and can also be accessed by pressing "enter" on the search field.

Clicking a suggested tag fills in the search field with the `tag:` syntax.

I removed the debounce from searching since it no longer searches as you type, but we'll probably need to add it back to the tag suggestions.

Todo:
* ~Filter tag suggestions for those that match search string~
* Make tag suggestion search a littttttle less broad probably (e.g. don't search in the middle of words)
* Debounce tag searching / performance test overall
* A11y (screen reader labels and keyboard navigation)
* Clean up styles for the suggestion popup

Fixes #1624

### Test
TBD

### Review
TBD

### DIFF
--- a/lib/icons/tag.jsx
+++ b/lib/icons/tag.jsx
@@ -1,16 +1,16 @@
 import React from 'react';
 
-// TODO we need a single tag icon
 export default function TagIcon() {
   return (
     <svg
       className="icon-tag"
       xmlns="http://www.w3.org/2000/svg"
-      width="22"
-      height="22"
-      viewBox="0 0 22 22"
+      width="24"
+      height="24"
+      viewBox="0 0 24 24"
     >
-      <path d="M20 5v5.172l-7 7L7.828 12l7-7H20m2-2h-8l-9 9 8 8 9-9V3zM3.763 13.237L2.525 12l1.237-1.237L11.525 3H9l-9 9 8 8 1.263-1.263-5.5-5.5zM17.5 9a1.5 1.5 0 1 0 0-3 1.5 1.5 0 0 0 0 3z" />
+      <rect x="0" fill="none" width="24" height="24" />
+      <path d="M18.29,5.71v5.17l-7,7L6.12,12.71l7-7h5.17m2-2h-8L4,12a1,1,0,0,0,0,1.41L10.59,20A1,1,0,0,0,12,20l8.29-8.29v-8Zm-4.5,5.5a1,1,0,1,0-1-1A1,1,0,0,0,15.79,9.21Z" />
     </svg>
   );
 }
--- a/lib/icons/tag.jsx
+++ b/lib/icons/tag.jsx
@@ -1,0 +1,16 @@
+import React from 'react';
+
+// TODO we need a single tag icon
+export default function TagIcon() {
+  return (
+    <svg
+      className="icon-tag"
+      xmlns="http://www.w3.org/2000/svg"
+      width="22"
+      height="22"
+      viewBox="0 0 22 22"
+    >
+      <path d="M20 5v5.172l-7 7L7.828 12l7-7H20m2-2h-8l-9 9 8 8 9-9V3zM3.763 13.237L2.525 12l1.237-1.237L11.525 3H9l-9 9 8 8 1.263-1.263-5.5-5.5zM17.5 9a1.5 1.5 0 1 0 0-3 1.5 1.5 0 0 0 0 3z" />
+    </svg>
+  );
+}
--- a/lib/search-field/index.jsx
+++ b/lib/search-field/index.jsx
@@ -1,15 +1,16 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { debounce, isEmpty } from 'lodash';
+import { isEmpty } from 'lodash';
 import SmallCrossIcon from '../icons/cross-small';
 import appState from '../flux/app-state';
 import { tracks } from '../analytics';
+import SearchSuggestions from '../search-suggestions';
 
 const { search, setSearchFocus } = appState.actionCreators;
 const { recordEvent } = tracks;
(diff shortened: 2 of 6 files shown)


dmsnell:
Reviewing the changes here the code looks fine, and probably what I would have done had I set out to do it.

I'm wondering though about the bit of changing the search-as-you-type behavior. People may be relying on that. Do you think it's possible to introduce the basic part of this change without removing that behavior? Could we do both, at least for a while so that the changes would happen more gradually? Or is this something where it's so hard to fit in the two paradigms that it's unreasonable?

lib/icons/tag.jsx
```diff
@@ -0,0 +1,16 @@
...
+      xmlns="http://www.w3.org/2000/svg"
+      width="24"
+      height="24"
+      viewBox="0 0 24 24"
+    >
+      <rect x="0" fill="none" width="24" height="24" />
+      <path d="M18.29,5.71v5.17l-7,7L6.12,12.71l7-7h5.17m2-2h-8L4,12a1,1,0,0,0,0,1.41L10.59,20A1,1,0,0,0,12,20l8.29-8.29v-8Zm-4.5,5.5a1,1,0,1,0-1-1A1,1,0,0,0,15.79,9.21Z" />
+    </svg>
+  );
+}
```
dmsnell: Fun fact! Even though this code is 100% fine it's also possible to rid ourselves of some of the extra syntax. This is a style thing.

```js
export const TagIcon = () => (
	<svg
		…
	</svg>
);
```

lib/search-field/index.jsx
```diff
@@ -36,58 +38,68 @@ export class SearchField extends Component {
     }
   }
 
+  doSearch = query => {
+    this.setState({ query, searchSelected: true });
+    this.props.onSearch(query);
```
dmsnell: It appears as though we may have been using `query` because we were splitting the data flow of the search bar text with the debounced copy. Now that we're taking the debounce out I question whether we need to continue to split the data stream. We might be able to eliminate all of the `this.state.query` and rely on the `filter` provided by Redux.

The `this.props.onSearch` function doesn't appear to do anything except store this `filter` or `query` so I think if we add `state.filter` in `mapStateToProps` we can rely on it instead of `this.state.query`

lib/search-field/index.jsx
```diff
@@ -36,58 +38,68 @@ export class SearchField extends Component {
...
+          />
+          <button
+            aria-label="Clear search"
+            hidden={!hasQuery}
+            onClick={this.clearQuery}
+          >
+            <SmallCrossIcon />
+          </button>
+          {shouldShowSuggestions && (
+            <SearchSuggestions query={query} onSearch={this.doSearch} />
```
dmsnell: Likewise, if we really want to rely on Redux state for providing this information we won't have to pass it along here. Thankfully this doesn't imply any changes to `SearchSuggestions` other than we add the appropriate props to `connect()` instead of passing them _through_ components like this.

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+        {tags
+          .filter(function(tag) {
+            return tag.id.includes(query);
+          })
+          .map(tag => (
+            <div
+              key={tag.id}
+              className="search-suggestion-row"
+              onClick={() => onSearch(`tag:${tag.id}`)}
```
dmsnell: Does this imply we can only search for _either_ one tag at a time _or_ a query string?

We could consider appending it to the current search string instead…

```js
onClick={ () => onSearch( `${ query } tag:${ tag.id }` );
```

It would leave some ugliness in there, like maybe we add the same tag twice, in which case we could handle this in other ways (remove the tag from the suggestion list if it's already in the query; toggle its inclusion when it's clicked so that a second click would remove it; update the `search` reducer to eliminate duplicate clauses; etc…)

Just noting the behavior that I think this implies which I didn't expect.

lib/icons/tag.jsx
codebykat: Gonna leave this as it is since it's the way all the other icons are doing it.

lib/search-suggestions/index.jsx
codebykat: This... is a really good point and definitely a case we need to handle.

codebykat:
I've spun up a alternative version of this PR, https://github.com/Automattic/simplenote-electron/pull/1648/, which preserves the filter-as-you-type behavior for now.

lib/search-field/index.jsx
codebykat: Tabling this for now (see [below](https://github.com/Automattic/simplenote-electron/pull/1643#issuecomment-543523665)) but will revisit if we do decide to remove the filtering.

lib/search-field/index.jsx
codebykat: [Ditto](https://github.com/Automattic/simplenote-electron/pull/1643#discussion_r336329225).

lib/search-suggestions/index.jsx
codebykat: Wait, no, this doesn't make sense because the suggestion will be a matching string of the existing query. So we're sort of saying "oh, were you trying to find notes with this tag?"

I need to do more testing with multiple query arguments, it's not super clear how (or whether) that is supposed to work...

codebykat:
Closing for now to avoid confusion. See https://github.com/Automattic/simplenote-electron/pull/1648

Issue #235, opened 2016-03-18
Note display: number of lines vary for Expanded display 


davewhitley:
Is Expanded 4 lines? I forgot what we decided.

It seem strange that some notes don't have 4 lines even though there is more text.

![screen shot 2016-03-18 at 10 15 30 am](https://cloud.githubusercontent.com/assets/618551/13882219/93be01ea-ecf2-11e5-826b-78a8d58c324c.png)

drw158:
Another example 

![screen shot 2016-03-18 at 10 18 58 am](https://cloud.githubusercontent.com/assets/618551/13882297/e34fe9e4-ecf2-11e5-9be6-5b7a2090f869.png)

PR #1647, opened 2019-10-16
Show more than first line in note list preview

belcherj:
### Fix
fixes: https://github.com/Automattic/simplenote-electron/issues/235

The regex that captures the note list preview was only selecting the first line. This selects the remainder of the note.  That string then gets sliced down to the `previewCharacterLimit`.

This PR also decreases the performance cost of creating the title and preview from 1ms to 0.2ms.

### Test
1. Load production
2. Have note that has a few characters per line for a couple of lines
3. Have Settings -> Display -> Note Display set to Expanded
4. Observe only the first line of text is shown
5. Do the same on the this branch
6. Observe that multiple lines are shown up to 300 characters
7. This is the same as the Android app

**Before:**
Screen Shot 2019-10-18 at 10 53 21 AM
**After:**
Screen Shot 2019-10-18 at 10 53 29 AM

### Review

### Release

- Fixed bug that only shows the first line of text in note list preview

### DIFF
--- a/lib/utils/note-utils.js
+++ b/lib/utils/note-utils.js
@@ -1,37 +1,54 @@
 import removeMarkdown from 'remove-markdown';
 import { isFunction } from 'lodash';
 
-/**
- * Matches the title and excerpt in a note's content
- *
- * Both the title and the excerpt are determined as
- * content starting with something that isn't
- * whitespace and leads up to a newline or line end
- *
- * @type {RegExp} matches a title and excerpt in note content
- */
-const noteTitleAndPreviewRegExp = /\s*(\S[^\n]*)\s*(\S[^\n]*)?/;
+const maxTitleChars = 64;
+const maxPreviewChars = 200;
+const maxPreviewLines = 4; // probably need to adjust if we're in the comfy view
+
+const matchUntilLimit = (pattern, source, preview = '', lines = 0) => {
+  const match = pattern.exec(source);
+  // must match, must have no more than four lines, must not be longer than N=10 characters
+  if (!match || lines > maxPreviewLines || preview.length > maxPreviewChars) {
+    return preview;
+  }
+
+  const [, chunk] = match;
+  return matchUntilLimit(pattern, source, preview + chunk, lines + 1);
+};
 
 /**
- * Matches the title and excerpt in a note's content skipping opening # from title
+ * Returns a string with markdown stripped
  *
- * Both the title and the excerpt are determined as
- * content starting with something that isn't
- * whitespace and leads up to a newline or line end
- *
- * @type {RegExp} matches a title and excerpt in note content skipping opening #'s from title and preview
+ * @param {String} inputString string for which to remove markdown
+ * @returns {String} string with markdown removed
  */
-const noteTitleAndPreviewMdRegExp = /(?:#{0,6}\s+)?(\S[^\n]*)\s*(?:#{0,6}\s+)?(\S[^\n]*)?/;
+const removeMarkdownWithFix = inputString => {
+  // Workaround for a bug in `remove-markdown`
+  // See https://github.com/stiang/remove-markdown/issues/35
+  return removeMarkdown(inputString.replace(/(\s)\s+/g, '$1'), {
+    stripListLeaders: false,
+  });
+};
 
-export const titleCharacterLimit = 200;
+const getTitle = (content, isNoteMarkdown) => {
+  const titlePattern = new RegExp(`\s*([^\n]{1,${maxTitleChars}})`, 'g');
+  const titleMatch = titlePattern.exec(content);
+  if (!titleMatch) {
+    return 'New Note…';
+  }
+  const [, rawTitle] = titleMatch;
+  return isNoteMarkdown
+      ? removeMarkdownWithFix(rawTitle) || rawTitle
+      : rawTitle;
+};
 
-// Ample amount of characters for the 'Expanded' list view
-export const previewCharacterLimit = 300;
-
-// Defaults for notes with empty content
-export const defaults = {
-  title: 'New Note...',
-  preview: '',
+const getPreview = (content, isNoteMarkdown) => {
+  const previewPattern = new RegExp(
+      `[^\n]*\n+[ \t]*([^]{0,${maxPreviewChars}})`,
+      'g'
+  );
+  let preview = matchUntilLimit(previewPattern, content);
+  return isNoteMarkdown ? removeMarkdownWithFix(preview) : preview;
 };
 
 /**
@@ -42,34 +59,13 @@ export const defaults = {
  */
 export const noteTitleAndPreview = note => {
   const content = (note && note.data && note.data.content) || '';
-
-  const match = isMarkdown(note)
-    ? noteTitleAndPreviewMdRegExp.exec(content || '')
-    : noteTitleAndPreviewRegExp.exec(content || '');
-
-  if (!match) {
-    return defaults;
-  }
-
-  let [, title = defaults.title, preview = defaults.preview] = match;
-
-  title = title.slice(0, titleCharacterLimit);
-  preview = preview.slice(0, previewCharacterLimit);
-
-  if (preview && isMarkdown(note)) {
-    // Workaround for a bug in `remove-markdown`
-    // See https://github.com/stiang/remove-markdown/issues/35
-    const previewWithSpacesTrimmed = preview.replace(/(\s)\s+/g, '$1');
-
-    preview = removeMarkdown(previewWithSpacesTrimmed, {
-      stripListLeaders: false,
-    });
-  }
-
+  const isNoteMarkdown = isMarkdown(note);
+  const title = getTitle(content, isNoteMarkdown);
+  const preview = getPreview(content, isNoteMarkdown);
   return { title, preview };
 };
 
-export function isMarkdown(note) {
+function isMarkdown(note) {
   return note && note.data && note.data.systemTags.includes('markdown');
 }
 
--- a/lib/utils/note-utils.js
+++ b/lib/utils/note-utils.js
@@ -37,19 +37,20 @@ const getTitle = (content, isNoteMarkdown) => {
     return 'New Note…';
   }
   const [, rawTitle] = titleMatch;
-  return isNoteMarkdown
-      ? removeMarkdownWithFix(rawTitle) || rawTitle
-      : rawTitle;
+  return rawTitle;
 };
 
 const getPreview = (content, isNoteMarkdown) => {
   const previewPattern = new RegExp(
-      `[^\n]*\n+[ \t]*([^]{0,${maxPreviewChars}})`,
-      'g'
+    `[^\n]*\n+[ \t]*([^]{0,${maxPreviewChars}})`,
+    'g'
   );
-  let preview = matchUntilLimit(previewPattern, content);
-  return isNoteMarkdown ? removeMarkdownWithFix(preview) : preview;
+  return matchUntilLimit(previewPattern, content);
 };
+
+const removeMarkdownFromOutput = isMarkdown
+  ? s => removeMarkdownWithFix(s) || s
+  : s => s;
 
 /**
  * Returns the title and excerpt for a given note
@@ -60,8 +61,8 @@ const getPreview = (content, isNoteMarkdown) => {
 export const noteTitleAndPreview = note => {
   const content = (note && note.data && note.data.content) || '';
   const isNoteMarkdown = isMarkdown(note);
-  const title = getTitle(content, isNoteMarkdown);
-  const preview = getPreview(content, isNoteMarkdown);
+  const title = removeMarkdownFromOutput(getTitle(content), isNoteMarkdown);
+  const preview = removeMarkdownFromOutput(getPreview(content), isNoteMarkdown);
   return { title, preview };
 };
 
--- a/lib/utils/note-utils.js
+++ b/lib/utils/note-utils.js
@@ -1,8 +1,8 @@
 import removeMarkdown from 'remove-markdown';
 import { isFunction } from 'lodash';
 
-const maxTitleChars = 64;
-const maxPreviewChars = 200;
+export const maxTitleChars = 64;
+export const maxPreviewChars = 200;
 const maxPreviewLines = 4; // probably need to adjust if we're in the comfy view
 
 const matchUntilLimit = (pattern, source, preview = '', lines = 0) => {
@@ -30,7 +30,7 @@ const removeMarkdownWithFix = inputString => {
   });
 };
 
-const getTitle = (content, isNoteMarkdown) => {
+const getTitle = content => {
   const titlePattern = new RegExp(`\s*([^\n]{1,${maxTitleChars}})`, 'g');
   const titleMatch = titlePattern.exec(content);
   if (!titleMatch) {
@@ -40,7 +40,7 @@ const getTitle = (content, isNoteMarkdown) => {
   return title;
 };
 
-const getPreview = (content, isNoteMarkdown) => {
+const getPreview = content => {
   const previewPattern = new RegExp(
     `[^\n]*\n+[ \t]*([^]{0,${maxPreviewChars}})`,
     'g'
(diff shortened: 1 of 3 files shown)


lib/utils/note-utils.js
```diff
@@ -10,7 +10,7 @@ import { isFunction } from 'lodash';
  *
  * @type {RegExp} matches a title and excerpt in note content
  */
-const noteTitleAndPreviewRegExp = /\s*(\S[^\n]*)\s*(\S[^\n]*)?/;
+const noteTitleAndPreviewRegExp = /\s*(\S[^\n]*)\s*([\s\S]*)?/;
```
dmsnell: This is a curious change here. What are you intending to communicate with the `([\s\S]*)` group? To me that reads the same as `(.*)`

belcherj: `.` does not include line terminators such as `/n`.  I thought the exact same thing!

belcherj: https://regex101.com/r/Gtac3V/1

dmsnell: As an historical note the single first line I think was chosen for pragmatic reasons when considering performance. We don't want a ten-thousand line note to slow down the note list.

If we want to chop these at a specific limit we can build that into the RegExp as well or feed a chopped version of the input string into the RegExp.

This may be a good time to improve the overall quality of this algorithm if indeed we're going to touch it. In reviewing what we want we could consider something like this…

```js
const noteTitleAndPreviewRegExp = /\s*([^\n]{1,32})[^\n]*\n+[ \t]*([\s\S]{1,300})/;
```

Put another way…

```js
const noteTitleAndPreviewRegExp =new RexExp(
  '\s*' + // ignore leading whitespace
  '([^\n]{1,32})' + // then grab up to so many title characters (on a single line)
  '[^\n]*\n+' + // then skip the remainder of the title characters and any following newlines
  '[ \t]*' + // skip the leading whitespace
  '([\s\S]{1,300})' // and grab the first 300 characters of the note
);

…

const [ , title, rawPreview ] = match;
const preview = rawPreview.split('\n').slice(0,4).join('\n'); // the RegExp allows for 300 one-character lines
```

I'm trying to do this all in one RegExp: allow up to four lines maxing out at a total number of characters. I'll update if I figure it out

dmsnell: Here are some thoughts on behaviors…
 - we want a max length of the title
 - we want a max length of the preview characters
 - we also want a max number of lines in the preview characters
 - we want to skip over blank lines when making the preview

```js
const getTitleAndPreview = content => {
  const titlePattern = /\s*([^\n]{1,10})/g; // grab the first N=10 characters of the first line
  const titleMatch = titlePattern.exec( content );
  if ( ! titleMatch ) {
    return [ 'Untitled', '' ]; // do something appropriate here
  }
  const [ , title ] = titleMatch;

  const previewPattern = /[^\n]*\n+[ \t]*([^\n]{0,10})/g; // skip leading whitespace, then grab first N=10 characters of body
  previewPattern.lastIndex = titlePattern.lastIndex; // start where the title ended

  const go = ( pattern, source, preview = '', lines = 0 ) => {
    const match = pattern.exec( source );
    // must match, must have no more than four lines, must not be longer than N=10 characters
    if ( ! match || lines > 3 || preview.length > 10 ) {
      return preview;
    }

    const [ , chunk ] = match;
    return go( pattern, source, preview + chunk, lines + 1 );
  }

  return [ title, go( previewPattern, content ) ];
}
```

there's cleanup work here but it addresses one thing I don't like about the existing solution or my previous comment: the logic is self-contained and doesn't rely on tricks

this looks complicated and can probably be simplified quite a bit but I think it produces our exactly-desired behaviors and does so without running into performance issues for massive notes. the `N=10` thing is controllable by us

dmsnell:
```js
const maxTitleChars = 64;
const maxPreviewChars = 200;
const maxPreviewLines = 4; // probably need to adjust if we're in the comfy view

const matchUntilLimit = ( pattern, source, preview = '', lines = 0 ) => {
    const match = pattern.exec( source );
    // must match, must have no more than four lines, must not be longer than N=10 characters
    if ( ! match || lines > maxPreviewLines || preview.length > maxPreviewChars ) {
      return preview;
    }

    const [ , chunk ] = match;
    return matchUntilLimit( pattern, source, preview + chunk, lines + 1 );
  }

const getTitleAndPreview = content => {
  const titlePattern = new RegExp( `\s*([^\n]{1,${ maxTitleChars }})`, 'g' ); // grab the first N=10 characters of the first line
  const titleMatch = titlePattern.exec( content );
  if ( ! titleMatch ) {
    return [ 'Untitled', '' ]; // do something appropriate here
  }
  const [ , title ] = titleMatch;

  const previewPattern = new RegExp( `[^\n]*\n+[ \t]*([^\n]{0,${ maxPreviewChars }})`, 'g' ); // skip leading whitespace, then grab first N=10 characters of body
  previewPattern.lastIndex = titlePattern.lastIndex; // start where the title ended

  return [ title, matchUntilLimit( previewPattern, content ) ];
}
```

belcherj: Your version:
```
The/long Note!.
Call to getTitleAndPreview took 0.2849999873433262 milliseconds.
```
My version:
```
The/long Note!.
Call to getTitleAndPreview took 0.9950000094249845 milliseconds.
```

dmsnell: That being said, if we identified a performance issue then I think a big one is under our noses: I think we calculate this on ever render of the sidebar, when indeed we only need run it on our initial index and then whenever note content changes.

We could store this in a separate state value and eliminate any performance issues altogether. Work for a future iteration.

belcherj:
Used the Dennis solution.

belcherj:
Hold off on reviews. I need to fix the remove markdown functionality.

belcherj:
I looked at how many times this actually gets called.  It only gets called a dozen or so times for the items in the list that are visible.  Let's get this fixed in.

lib/utils/note-utils.js
```diff
@@ -1,38 +1,58 @@
...
-export const previewCharacterLimit = 300;
+const getTitle = ( content, isNoteMarkdown ) => {
+  const titlePattern = new RegExp( `\s*([^\n]{1,${ maxTitleChars }})`, 'g' );
+  const titleMatch = titlePattern.exec( content );
+  if ( ! titleMatch ) {
+    return 'New Note...';
+  }
+  const [ , rawTitle ] = titleMatch;
+  if ( isNoteMarkdown ) {
+    let parsed = removeMarkdownWithFix( rawTitle )
```
dmsnell: no need for a `let` here

it's not wrong, though it's nice to leave things `const` to indicate to the reader that the binding won't change

lib/utils/note-utils.js
```diff
@@ -1,38 +1,58 @@
...
+    return parsed.length && parsed || rawTitle;
+  }
+  return rawTitle;
```
dmsnell: just an observation, we can remove some of the nesting by rearranging the conditional…

```js
if ( ! isNoteMarkdown ) {
  return rawTitle;
}

const title = removeMarkdownWithFix( rawTitle );
return title.length ? title : rawTitle;
```

dmsnell: iterating…

```js
return isNoteMarkdown
  ? ( removeMarkdownWithFix( rawTitle ) || rawTitle )
  : rawTitle;
```

lib/utils/note-utils.js
```diff
@@ -1,38 +1,58 @@
...
-  title: 'New Note...',
-  preview: '',
-};
+const getPreview = ( content, isNoteMarkdown ) => {
+  const previewPattern = new RegExp( `[^\n]*\n+[ \t]*([^]{0,${ maxPreviewChars }})`, 'g' );
+  let preview = matchUntilLimit( previewPattern, content );
+  if ( isNoteMarkdown ) {
+    preview = removeMarkdownWithFix( preview );
+  }
+  return preview;
```
dmsnell: by combining we have the chance to eliminate the mutation and clarify the control-flow

```js
const preview = matchUntilLImit( previewPattern, content );

return isNoteMarkdown
  ? removeMarkdownWithFix( preview )
  : preview;
```

lib/utils/note-utils.js
```diff
@@ -1,38 +1,58 @@
...
+  return rawTitle;
+}
 
-// Defaults for notes with empty content
-export const defaults = {
-  title: 'New Note...',
-  preview: '',
-};
+const getPreview = ( content, isNoteMarkdown ) => {
+  const previewPattern = new RegExp( `[^\n]*\n+[ \t]*([^]{0,${ maxPreviewChars }})`, 'g' );
```
dmsnell: It's not essential that we do this, but if we return the `.lastIndex` of the RegExp object we can start where it left off and not scan past the title again. `matchUntilLimit` depends on this statefulness during its own execution which is why we have to create the pattern outside of that function and pass it in

In this case I can see value in separating them as you have done because it removes the coupling between separate parts of the file 👍

lib/utils/note-utils.js
```diff
@@ -42,34 +62,13 @@ export const defaults = {
...
-    const previewWithSpacesTrimmed = preview.replace(/(\s)\s+/g, '$1');
-
-    preview = removeMarkdown(previewWithSpacesTrimmed, {
-      stripListLeaders: false,
-    });
-  }
-
+  const isNoteMarkdown = isMarkdown(note);
+  const title = getTitle( content, isNoteMarkdown );
+  const preview = getPreview( content, isNoteMarkdown );
```
dmsnell: in the previous code we had somewhat more interleaved the computation of the title and preview but now they are separated. I wonder if instead of passing in this boolean flag and crossing layers of abstraction it would make more sense to let `getTitle` and `getPreview` focus specifically on the text, since it looks like all we're doing is transforming the output, not changing how we find it…

```js
const transformOutput = isMarkdown
  ? ( s => removeMarkdownWithFix( s ) || s )
  : s => s;

const title = transformOutput( getTitle( content ) );
const preview = transformOutput( getPreview( content ) );
```

the `getTitle` and `getPreview` functions would then collapse into a few lines each and we wouldn't have `rawTitle` and `title` and that branching code.

not sure what a good name for `transformOutput` would be, though with `lodash` we could approach it differently yet again

```js
const title = getTitle( content );
const preview = getPreview( content );

return isMarkdown
  ? mapValues( { title, preview }, s => removeMarkdownWithFix( s ) || s )
  : { title, preview };
```

lib/utils/note-utils.test.js
```diff
@@ -19,7 +16,7 @@ describe('noteTitleAndPreview', () => {
 
   it('should return default values if note content is empty', () => {
     const result = noteTitleAndPreview(note);
-    expect(result).toEqual(defaults);
+    expect(result).toEqual({"preview": "", "title": "New Note..."});
```
dmsnell: at some point we need to get our auto-formatter working and then this will be `{preview: '', title: 'New Note...' }` automatically

lib/utils/note-utils.js
```diff
@@ -1,38 +1,58 @@
...
+  ;
+}
 
-// Ample amount of characters for the 'Expanded' list view
-export const previewCharacterLimit = 300;
+const getTitle = ( content, isNoteMarkdown ) => {
+  const titlePattern = new RegExp( `\s*([^\n]{1,${ maxTitleChars }})`, 'g' );
+  const titleMatch = titlePattern.exec( content );
+  if ( ! titleMatch ) {
+    return 'New Note...';
```
dmsnell: Do we prefer the ellipsis?

```js
return 'New Note…';
```

I don't know

lib/utils/note-utils.js
belcherj: Oops, fixed

lib/utils/note-utils.js
belcherj: Yaasss, TY!

lib/utils/note-utils.js
belcherj: Wow, I always prefered my version. Never thought about the mutation.

lib/utils/note-utils.test.js
belcherj: I never turned prettier on after my system upgrade.  Agreed that we need to do this on commit.

lib/utils/note-utils.js
belcherj: Don't know either. An ellipsis sounds good, changed.

lib/utils/note-utils.js
belcherj: I removed this because it was introducing some bugs.  Will play a bit more with it.

lib/utils/note-utils.js
dmsnell: mutation scares me and gives me a headache. I like knowing that once I find an assignment in the code I don't have to worry about interpreting the rest of the code in my head to learn if the value might be reassigned outside my purview

lib/utils/note-utils.js
dmsnell: I actually do like what you did by separating them…

lib/utils/note-utils.js
```diff
@@ -37,20 +37,21 @@ const getTitle = (content, isNoteMarkdown) => {
     return 'New Note…';
   }
   const [, rawTitle] = titleMatch;
-  return isNoteMarkdown
-      ? removeMarkdownWithFix(rawTitle) || rawTitle
-      : rawTitle;
+  return rawTitle;
```
dmsnell: nice change - might as well rename this to `title` no?

lib/utils/note-utils.js
```diff
@@ -1,39 +1,57 @@
...
+  // Workaround for a bug in `remove-markdown`
+  // See https://github.com/stiang/remove-markdown/issues/35
+  return removeMarkdown(inputString.replace(/(\s)\s+/g, '$1'), {
+    stripListLeaders: false,
+  });
+};
 
-// Ample amount of characters for the 'Expanded' list view
-export const previewCharacterLimit = 300;
+const getTitle = (content, isNoteMarkdown) => {
```
dmsnell: still have this unnecessary `isNoteMarkdown` argument

lib/utils/note-utils.js
```diff
@@ -42,34 +60,13 @@ export const defaults = {
...
-    const previewWithSpacesTrimmed = preview.replace(/(\s)\s+/g, '$1');
-
-    preview = removeMarkdown(previewWithSpacesTrimmed, {
-      stripListLeaders: false,
-    });
-  }
-
+  const isNoteMarkdown = isMarkdown(note);
+  const title = removeMarkdownFromOutput(getTitle(content), isNoteMarkdown);
+  const preview = removeMarkdownFromOutput(getPreview(content), isNoteMarkdown);
```
dmsnell: still have these unnecessary `isNoteMarkdown` arguments

lib/utils/note-utils.test.js
```diff
@@ -55,13 +50,13 @@ describe('noteTitleAndPreview', () => {
 
   it('should have enough text for an Expanded preview, even if the title is very long', () => {
     // Longer than the char limits
-    const title = 'foo'.repeat(titleCharacterLimit);
-    const paragraph = 'foo'.repeat(previewCharacterLimit);
+    const title = 'foo'.repeat(64);
+    const paragraph = 'foo'.repeat(200);
```
dmsnell: the fact that these numbers match our lengths makes me worry that someone will assume there is more connection here than there is.

how about something like `.repeat( 500 )` or an otherwise big number

```js
const title = 'A really long title'.repeat( 100 );
const paragraph = 'A really long paragraph'.repeat( 100 );
```

lib/utils/note-utils.test.js
```diff
@@ -55,13 +50,13 @@ describe('noteTitleAndPreview', () => {
...
 
     note.data.content = title + '\n' + paragraph;
     const result = noteTitleAndPreview(note);
-    expect(result.title).toHaveLength(titleCharacterLimit);
-    expect(result.preview).toHaveLength(previewCharacterLimit);
+    expect(result.title).toHaveLength(64);
+    expect(result.preview).toHaveLength(200);
```
dmsnell: if we export these values in `note-utils` then we can import them here and then our test won't break when we make an intentional change to the note length and also we will couple the test more tightly with the code it's testing.

lib/utils/note-utils.test.js
belcherj: For sure

lib/utils/note-utils.test.js
belcherj: I don't like that we export something just to use it in a test.

belcherj: I made the change but I don't like it.

lib/utils/note-utils.js
belcherj: 🤦‍♂

lib/utils/note-utils.test.js
dmsnell: > I don't like that we export something just to use it in a test.

why's that?

Issue #895, opened 2018-10-09
Make Tags Searchable


ChrissiePollock:
If I add a tag to a note which does not appear in the main text of the same note, I cannot locate that note with a search.

#### What I expected
When I do a search, my tagged note will show.

#### What happened instead
The tag does not show up in results.

#### OS version
1.3.7 on MacBook Pro and also on Windows app

iamgabrielma:
Adding to this request, would be great to be able to found untagged notes as well, as per 1816199-zen:

> I use "tag" feature. But for one of the notes I forgot to add tag and it was difficult to find it.
> Is there any way to select/separate untagged notes?

iamgabrielma:
Requested in 2128442-zen

codebykat:
Tags are now part of search results as of #1648. But I'm leaving this open because it's still true that the notes filter doesn't match notes that contain a tag, which seems incorrect.

SylvesterWilmott:
note: it appears that searching tags is case sensitive

codebykat:
note to the above note: not anymore (#1905)

belcherj:
@SylvesterWilmott You think we can close this?

codebykat:
IMO it's still weird that the search doesn't search both notes and content, even though we do have tag search now. It feels somewhat different.

SylvesterWilmott:
I see where you are coming from @codebykat and I agree. I think we could **probably** safely allow tags to be included in keyword searches but maybe it should be a point of discussion for other platforms too.

PR #1648, opened 2019-10-18
Search: Add tag suggestions to notes list

### LINKED ISSUE #1624: Search: Add tag suggestions to search results

codebykat:
### Fix
This PR adds matching tag results to the notes list.

Screen Shot 2019-11-13 at 1 57 00 PM

It shows a maximum of five and hides the section if there are no matches.

Clicking a suggested tag fills in the search field with the `tag:` syntax.

Fixes #1624, #895

### Test
1. Perform some searches and make sure everything works in a way that makes sense! Test with weird tag names, long tag names, tags that aren't added to any notes, etc.
2. Change note layout (comfy, condensed, expanded) and make sure the left panel is laid out properly (all elements the correct heights, no notes missing, no text overflow).
3. Test in dark mode as well

### Screenshots

More [here](https://github.com/Automattic/simplenote-electron/pull/1648#issuecomment-555778198).
    
Screen Shot 2019-11-13 at 1 56 46 PM

Screen Shot 2019-11-13 at 1 55 38 PM

Screen Shot 2019-11-13 at 1 55 24 PM


codebykat:
This has a couple bugs still but I think it's ready to not be a draft anymore! Would love opinions on the open questions above, thoughts on how the code looks, and testing. I'll try to mop up the known issues tomorrow (assuming we even want to keep the keyboard stuff, it's kinda janky).

belcherj:
Can you rebase and move your Release Note to 1.11.0?

lib/icons/tag.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
belcherj: Is this from the new Icon pack? Can this be added as a separate PR?

codebykat: Which new Icon pack? It's the tag icon shown next to a tag name (differentiates tag suggestions from search suggestions). Tyler shared the SVG with me in Slack.

codebykat:
Noting style changes requested by @SylvesterWilmott in Slack:

- Each row should be 40px height
- The entire list inside the dropdown should have top and bottom padding of 8px
- I'm not sure if the alignment of the row elements in the design will work here untill we do the UI refresh because the space is much tighter here but we can try, it's like --16px padding-- [24px icon]--16px--[label]--16px padding

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+  }
+
+  nextItem = () => {
+    const { filteredTags } = this.props;
+    const { selectedItem } = this.state;
+    const newItem =
+      selectedItem === filteredTags.length
+        ? filteredTags.length
+        : selectedItem + 1;
```
dmsnell: I'm not sure what we're trying to do here. It looks like this will lead us to pick an index just past the end of the array. If we're trying to _clamp_ the index to the length of the array we can do so in one step with a little math

```js
const newItem = Math.min( selectedItem + 1, filteredTags.length - 1 );
```

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    this.setState({ selectedItem: newItem });
+  };
+
+  prevItem = () => {
+    const { selectedItem } = this.state;
+    const newItem = selectedItem === 0 ? 0 : selectedItem - 1;
```
dmsnell:
```js
const newItem = Math.max( 0, selectedItem - 1 );
```

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+          filteredTags.map((tag, index) => (
+            <div
+              key={tag.id}
+              id={tag.id}
+              className={
+                selectedItem === index + 1
+                  ? 'search-suggestion-row search-suggestion-row-selected'
+                  : 'search-suggestion-row'
+              }
+              onClick={() => onSearch(`tag:${tag.id}`)}
```
dmsnell: fun fact, not too important here, but can be useful especially since this event originates on a `div` - we can pass data through custom attributes and access them in the callback…

```js
<div
	key={tag.id}
	id={tag.id}
	data-tag-id={tag.id}
	onClick={this.doSearch}
>
```

```js
doSearch = ( { target: { dataset: { tagId } } } ) => this.props.onSearch( `tag:${ tagId }` ));
```

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+            </div>
+          ))}
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = ({ appState: state }, ownProps) => ({
+  filteredTags: state.tags.filter(function(tag) {
+    return tag.id.startsWith(encodeURIComponent(ownProps.query));
```
dmsnell: if we have `encodeURIComponent` here in this component it probably means we're neglecting to convert it on entry into the application. did you come across this incidentally or did you anticipate the need for this?

SylvesterWilmott:
Some things that were discussed away from GitHub:

- The designs show a `tag:` prefix on the tag items in the list, the designs are based on an updated UI which has a wider search bar. This doesn't work well with the current width of the search bar so this detail will be omitted until the UI changes have been made.
- Desktop has the advantage of being able to show search results and suggestions simultaneously. The filter as you type behaviour will be retained until we understand if the point scoring system of the default "relevance" search will cause problems with performance.

Feedback:

- The tag suggestions considers character case where it should not
- Let's allow matches anywhere in the tag name and matching with 2 characters
-  Regarding the style changes request, [here is a pen](https://codepen.io/Sylvstr/pen/eYYrjJQ) I whipped up

lib/search-suggestions/index.jsx
codebykat: Oh it's because the search is the first select-able element... then the tag array... so it's the index plus one. This could be clearer I admit >_>

lib/search-suggestions/index.jsx
codebykat: Ooh this is something I have needed in attempting to translate my jQuery brain into React.

lib/search-suggestions/index.jsx
codebykat: I found the need for it in testing... you're saying the query should get encoded upstream somewhere?

codebykat: I think I got it here, is that what you had in mind? Any thoughts on the fact that this input was not formerly encoded? https://github.com/Automattic/simplenote-electron/pull/1648/commits/8a7493ccec8472600b3aa22f7287d19f5557ce70

codebykat:
Done:
* Style fixes
* Case insensitive matching anywhere in the string, starting from 2 characters typed
* Made some code less terrible

codebykat:
@dmsnell your advice on the correct way to do this would be very welcome ❤  --> https://github.com/Automattic/simplenote-electron/pull/1648/commits/30e2bf8ac8fefcfa299d9aedc616b8d617721153

lib/search-field/index.jsx
```diff
@@ -36,21 +42,37 @@ export class SearchField extends Component {
...
 
-  interceptEsc = event => {
-    if (KEY_ESC === event.keyCode) {
-      if (this.state.query === '') {
-        this.inputField.blur();
-      }
-      this.clearQuery();
+  doSearch = query => {
+    this.setState({ query, searchSelected: true });
+    this.debouncedSearch(query);
```
dmsnell: let's remove the `query` from the debounce to prevent the parallel-universes effect when we have multiple calls to `doSearch` in a row. granted, we shouldn't have the same problems as with the note content (which could be updated by remote calls) but this keeps a single source of data instead of two

`debouncedSearch()` also has to change but not in a big way

```js
debouncedSearch = debounce(() => this.props.onSearch(this.state.query), SEARCH_DELAY);
```

dmsnell: is `debounce` the appropriate idiom for search? if we keep on typing then it won't search until we stop. a `throttle` might be more appropriate so that we can see search results as we continue to type but it will still make sure we don't send a new search on every keystroke

`debounce` will prevent the search from interfering with our typing but won't update until we have finished typing
`throttle` will allow a slow search to interfere with our typing but provide us partial results as we type

lib/search-field/index.jsx
```diff
@@ -36,21 +42,37 @@ export class SearchField extends Component {
...
+  };
+
+  interceptKeys = event => {
+    switch (event.keyCode) {
+      case KEY_ESC:
+        if (this.state.query === '') {
+          this.inputField.blur();
+        }
```
dmsnell: some people don't like my style here but I don't like the way nesting slants the code

```js
case KEY_ESC:
	this.state.query === '' && this.inputField.blur();
	return this.clearQuery();
```

take it or leave it - just thought I'd share this fun-fact of how JavaScript short-circuiting logic works and how it can let us build conditional evaluation into a single line

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+    this.props.storeKeyHandler({
+      next: this.nextItem,
+      prev: this.prevItem,
+      select: this.selectItem,
+    });
+  }
+
+  componentDidUpdate() {
+    // scroll selected item into view
+    var selectedItem = this.refs.selectedItem;
```
dmsnell: check out the [docs for `createRef()`](https://reactjs.org/docs/react-api.html#reactcreateref) to see how we can eliminate some of the string processing here. I updated my understanding of `forwardRef` as well and sadly I don't see doing what we want here. We could get access to an underlying DOM node but it doesn't resolve passing functions up to the parent.

```js
componentDidMount() {
	this.selectedItem= createRef();
}

componentDidUpdate() {
	if ( this.selectedItem.current ) {
		this.selectedItem.current.scrollIntoView(false);
	}
}

render() {
	…

	ref={ selectedItem === index + 1 ? this.selectedItem : null }

	…
}
```

I _think_ this will work and you can see that it removes some manual accounting that we're doing.

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      domNode = ReactDOM.findDOMNode(selectedItem);
+      domNode.scrollIntoView(false);
+    }
+
+
+    this.props.storeKeyHandler({
+      next: this.nextItem,
+      prev: this.prevItem,
+      select: this.selectItem,
+    });
```
dmsnell: if we extract the selected item from this component then I think we can probably get rid of all of the `prev/next/select` passing. 🤔 

to do that I think we'd have to just pass `onSelectionChanged` to circle back up update when we've passed invalid ranges (the index clamping) though we can eliminate the possibility of an invalid index by obliterating the index itself…

```js
// search-field
this.state = {
	activeSelection: null,
}

…

const { activeSelection } = this.state;
const { filteredTags } = this.props;
case KEY_UP:
	const prevSelection = Math.max( 0, filteredTags.indexOf( activeSelection ) - 1 );
	return this.setState( { activeSelection: filteredTags[ prevSelection ] } );

case KEY_DOWN:
	const nextSelection = Math.min( filteredTags.length - 1, filteredTags.indexOf( activeSelection ) + 1 );
	return this.setState( { activeSelection: filteredTags[ nextSelection ] } );

…

const query = …
const filteredTags = …

<SearchSuggestions query={ query } tags={ filteredTags } activeSelection={ this.state.activeSelection } />

…

// search-suggestions
const { query, tags, activeSelection } = this.props;

…

className={ activeSelection === query ? '…' : '…' }

…

ref={ activeSelection === tag ? this.selectedItem : null }
```

This involves some more head-work but also unravels other headwork. It's inherently safe because in the event that send an active selection which has somehow disappeared then we're not messing with indices and mismatches.

A change in behavior is that if our search results change the either the active selection will follow the previously-selected tag around with the results _or_ if that previously-selected tag disappears then so will our active selection. I don't know if this is better or worse than keeping the _position_ of the selection constant. I think there are advantages both ways.

What it would do though is remove some back-and-forth with prop passing. It looks like you have boiled this down to a pretty minimal set of exchanges so I wouldn't worry too much about it, though I think if we could do the refactor (maybe after this PR lands unless you are willing to try now) then the entire combination of the two components and their logic might become clearer and easier to reason about - we won't have to couple the parent and children together so tightly.

lib/search-suggestions/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+    const { selectedItem } = this.state;
+    const newItem = Math.min(filteredTags.length, selectedItem + 1);
+    this.setState({ selectedItem: newItem });
+  };
+
+  prevItem = () => {
+    const { selectedItem } = this.state;
+    const newItem = Math.max(0, selectedItem - 1);
+    this.setState({ selectedItem: newItem });
+  };
```
dmsnell: since these are small and next to each other it's not a big problem but they do share a fair amount of logic and I find that it's a wee bit unclear due to the syntax noise to tell what we're intending on doing. extracting out the common operation (adjust the index) could remove some of that noise and bring more clarity and enforce common behavior

```js
moveSelection = ( direction = 'up' ) => {
	const { filteredTags } = this.props;
	const { selectedItem } = this.state;
	const step =
		'up' === direction ? -1 :
		'down' === direction ? 1 :
		0; // should only be up or down
	const newItem = Math.max( 0, Math.min(filteredTags.length, selectedItem + step) );
	this.setState({ selectedItem: newItem });
}

nextItem = () => this.moveSelection('down');

prevItem = () => this.moveSelection('up');
```

codebykat:
Totally updated with the new design pivot (see Slack / Zeplin)!

I think this is feeling pretty slick. Code-wise I didn't remove the dropdown since we'll want it for recent searches later...

It doesn't properly handle multiple tag searches, and the sidebar selected tag still behaves differently, which I need to fix. Those might be the only issues that I'm aware of. Please test and try to find more!

lib/app-layout/index.js
```diff
@@ -19,6 +22,7 @@ const NoteEditor = React.lazy(() =>
 );
 
 export const AppLayout = ({
+  filter = appState.filter,
```
dmsnell: I'm kind of surprised that this works - it may be a by-product of our funky Redux setup. This is already being passed-in from `connect` and if we need a default it would probably be best to set it to something like `null` or an empty string rather than trying to grab `appState.filter`

lib/search-field/index.jsx
```diff
@@ -27,13 +27,19 @@ export class SearchField extends Component {
...
+    const { searchFocus, onSearchFocused, filter } = this.props;
 
     if (searchFocus && this.inputField) {
       this.inputField.select();
       this.inputField.focus();
       onSearchFocused();
     }
+
+    // check to see if the filter has been updated (by a tag being clicked from suggestions)
+    // this is a hack to work around query not being in app state (yet)
```
dmsnell: good comment!

lib/search-field/index.jsx
```diff
@@ -27,13 +27,19 @@ export class SearchField extends Component {
...
+    if (filter !== this.state.query) {
+      this.inputField.value = filter;
```
dmsnell: we can probably also accomplish this here by manually updating local state…

```js
if ( filter !== this.state.query ) {
	this.debouncedSearch.cancel();
	this.setState( { query: filter } );
}
```

lib/tag-suggestions/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+const mapStateToProps = ({ appState: state }, ownProps) => ({
+  filteredTags: state.tags
+    .filter(function(tag) {
+      // todo split on spaces to support additive query args
+
+      // prefix tag ID with "tag:" (encoded) this allows us to match if the user typed the prefix
+      // n.b. doing it in this direction instead of stripping off any "tag:" prefix allows support
+      // of tags that contain the string "tag:" ¯\_(ツ)_/¯
+      var testID = 'tag%3A' + tag.id;
```
dmsnell: note that there is `tag.id` and there should also be `tag.name`. `tag.name` is probably what we want - probably what renders when viewing tags.

lib/search-field/index.jsx
codebykat: Hahaha I tried exactly this first and it destroyed the universe.

codebykat: No wait, maybe I tried .flush (which is what clear search does). Lemme see...

dmsnell: `flush()` runs the currently-queued function while `cancel()` prevents it from happening

if we did `flush()` it would try and update the search filter and immediately clear out the results then set state and reset the search field, I think

lib/app-layout/index.js
codebykat: Apparently I don't need to set it, but I do need to pass it in ¯\_(ツ)_/¯

lib/search-field/index.jsx
codebykat: Fascinatingly, if I do `.cancel` and `setState` instead I am unable to type anything at all in the search field.

lib/tag-suggestions/index.jsx
codebykat: `tag.data.name` good catch

codebykat: (though they appear to be identical in all cases)

codebykat: actually, that's a lie, the ID is a URI encoded version of the name?
Screen Shot 2019-11-14 at 9 25 03 PM

dmsnell: there is some old confusion on tag `id` vs. `name` and the gist of it is that the `id` is the account-unique id value and is meaningless. the `name` is what displays and can actually change.

there were times where one app used `id` and another `name` and then we had duplicate tags and we lost tags during tag-rename operations.

lib/app-layout/index.js
```diff
@@ -52,6 +56,7 @@ export const AppLayout = ({
       <Suspense fallback={placeholder}>
         <div className="app-layout__source-column theme-color-bg theme-color-fg">
           <SearchBar noteBucket={noteBucket} onNoteOpened={onNoteOpened} />
+          {filter.length > 0 && <TagSuggestions query={filter} />}
```
dmsnell: for the sake of data flow and consistency we can keep the `filter` check here but pass in `filter` to `TagSuggestions` in its own `connect()`

lib/search-field/index.jsx
```diff
@@ -93,6 +99,7 @@ export class SearchField extends Component {
 }
 
 const mapStateToProps = ({ appState: state }) => ({
+  filter: state.filter,
```
dmsnell: ah very good - so this copy replaces the copy passed in as ` query={ filter }` in the parent component - we can remove that property in `app-layout.js`

lib/search-field/index.jsx
dmsnell: oh wow. that makes sense now because I see that `.cancel` stops the entire debounced function when we only want to stop a specific invocation.

this probably all sounds more complicated than it needs to be

```js
if ( filter !== this.state.query ) {
	this.debouncedSearch( filter );
  this.debouncedSearch.flush();
}
```

^^^ how about that? update the search through the normal means but also immediately flush it to commit those changes

lib/tag-suggestions/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+import { tracks } from '../analytics';
+
+const { search } = appState.actionCreators;
+const { recordEvent } = tracks;
+
+export class TagSuggestions extends Component {
+  static displayName = 'TagSuggestions';
+
+  static propTypes = {
+    query: PropTypes.string.isRequired,
```
dmsnell: this shouldn't be needed here since we're not using it. it can be "swallowed up" in `filteredTags` which gets set in `connect`

lib/tag-suggestions/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+              {filteredTags.map(tag => (
+                <li
+                  key={tag.id}
+                  id={tag.id}
+                  className="tag-suggestion-row"
+                  onClick={() => this.props.onSearch(`tag:${tag.id}`)}
+                >
+                  <div
+                    className="tag-suggestion"
+                    title={decodeURIComponent(tag.id)}
```
dmsnell: `id` vs. `name`

lib/tag-suggestions/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+      return (
+        testID.search(new RegExp('(tag:)?' + ownProps.query, 'i')) !== -1 &&
+        // discard exact matches -- if the user has already typed or clicked
+        // the full tag name, don't suggest it
+        testID !== ownProps.query
+      );
+    })
+    .slice(0, 5),
+});
```
dmsnell: I'd like to try and make some updates to help you here, but there's one thing we should be careful of here and that's how this will cause the tag suggestions list to re-render on every state change even if it doesn't need to. the reason is that every time we `.filter` or `.slice` we create a new array in JavaScript and Redux is going to use _shallow equality_ to determine if things change.

that means that even though we return the same values, it will be treated as a brand new list of tags here.
the solution to prevent this is to return the same _reference-equal_ value when nothing changes. this is typically done in a number of ways: through memoization or through moving the filtered tags into Redux state.

the easier change here is memoization and that would be a cache based on the list of tags and the filter.

```js
const filterTags = ( tags, query ) => tags.filter( tag => {
	…
} ).slice( 0, 5 );

let lastTags = null;
let lastQuery = null;
let lastMatches = [];
const getMatchingTags = ( tags, query ) => {
	if ( lastTags === tags && lastQuery === query ) {
		return lastMatches;
	}

	lastTags = tags;
	lastQuery = query;
	lastMatches = filterTags( tags, query );
	return lastMatches;
}

const mapStateToProps = ( { appState: { filter, tags } } ) => ( {
	filteredTags: getMatchingTags( tags, filter ),
} );
```

lib/search-field/index.jsx
codebykat: Doesn't work either :/

belcherj (approved):
Ship it. Let us revisit the AutoSizer workaround in the future.

codebykat:
Screenshots and gif for @SylvesterWilmott 

ScreenRecording2019-11-19at41724

Screen Shot 2019-11-19 at 4 17 00 PM
Screen Shot 2019-11-19 at 4 16 46 PM
Screen Shot 2019-11-19 at 4 23 54 PM

SylvesterWilmott:
Looks great @codebykat ! Thanks

PR #1649, opened 2019-10-18
Show Notes Loading not No Notes when notes are loading

belcherj:
### Fix
When notes are loading we should not display the text "No Notes" as there may be notes but they have yet to load.  This PR makes the text "Notes Loading" until notes are loaded and then changes to "No Notes" if there are indeed no notes.

### Test
1. Have lots of notes and throttle your connection, load the app
2. Observe the text in the notes list no says "Notes Loading"
3. Have an account with no notes
4. Observe the text says "No Notes" once notes have loaded

### Review
Only one developer id required to review these changes, but anyone can perform the review.

### Release

- Display a notice that notes are loading when notes are loading

### DIFF
--- a/lib/flux/app-state.js
+++ b/lib/flux/app-state.js
@@ -39,6 +39,7 @@ export const actionMap = new ActionMap({
     revision: null,
     showTrash: false,
     listTitle: 'All Notes',
+    notesLoaded: false,
     showNavigation: false,
     showNoteInfo: false,
     isViewingRevisions: false,
@@ -296,6 +297,7 @@ export const actionMap = new ActionMap({
         notes: { $set: pinSortedNotes },
         note: { $set: selectedNote },
         selectedNoteId: { $set: get(selectedNote, 'id', null) },
+        notesLoaded: { $set: true },
       });
     },
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -8,6 +8,7 @@
 - Updated colors to use Color Studio, the color palette for Automattic products
    - [#1565](https://github.com/Automattic/simplenote-electron/pull/1565)
    - [#1612](https://github.com/Automattic/simplenote-electron/pull/1612)
+- Display a notice that notes are loading when notes are loading
 
 ### Fixes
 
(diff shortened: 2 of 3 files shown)


lib/flux/app-state.js
```diff
@@ -39,6 +39,7 @@ export const actionMap = new ActionMap({
     revision: null,
     showTrash: false,
     listTitle: 'All Notes',
+    notesLoaded: false,
```
dmsnell: Did we consider starting `notes: null`?

There's a chance to indicate whether or not we've loaded using this _tri-state_ value and that could preclude the need for creating a second value intrinsically bound to the same data.

```js
const noteListStatus = notes =>
  null === notes ? 'Loading…' :
  0 === notes.length ? 'No Notes' :
  'Has notes'
```

dmsnell: Just an explanation: I usually try to find ways of gleaning more information from a value without adding additional flags or meta as they tend to get out of sync and because people tend to zoom-in on the value they think they need as soon as they find a value that seems to work, so for dev communication purposes I have found the tai-state value quite useful when there is a separate unknown or pending state from the empty state.

belcherj: Making the default value for notes null causes some plumbing issues. Setting the default to null doesn't do much. It immediately gets overwritten to an array. Changing the way this works I think is outside the scope of this PR.

belcherj: A month or so ago I spent several hours trying to get app state to function as you describe and I created more bugs than I was solving.

dmsnell: ah interesting. I wondered what would break if we made the change.

where does it get immediately overwritten? on `authChanged` I can see we'd have to set it `null` there too.

I wouldn't question this further other than I recognize that we're adding new stuff into the app and doing so with two variables that are storing some of the same data, hypothetically.

I'll do some quick checks in a running copy locally and then report back.

Nice work, and I did at one point see your old state PR. state here in general is in a place it shouldn't be. long-term we can trim it way down

dmsnell:
@belcherj I created #1650 to tease out the ideas in my comment. I had to update `filterNotes()` but beyond that I don't think there was anything else that was broken.

in this PR one thing we need to do is update `authChanged` in `app-state` so that we reset the flag otherwise we'll get `No Notes` again while waiting to receive the first index from Simperium.

Issue #1654, opened 2019-10-21
Can't move app window while logged out


rachelmcr:
If I'm on the login/signup screen in the app, I can't drag/move the app window.

Related past issue: https://github.com/Automattic/simplenote-electron/issues/453

### Expected

I expect to be able to click at the top of the window and drag to move it.

### Observed

Clicking/dragging does nothing if I'm logged out. After logging in, I can move the app window as expected.

OS|OS Version|App Version
-|-|-
macOS|10.15|Electron 1.9.0-beta1

PR #1655, opened 2019-10-21
Fix to make macOS app movable on log in screen

belcherj:
### Fix
Fixes: https://github.com/Automattic/simplenote-electron/issues/1654

This PR makes the macOS app movable when on the log in screen. With not title bar you must have an html element with: `-webkit-app-region: drag;` to make the window draggable. We have this once logged in but no on the login screen.

It appears that the element: `.login__draggable-area` was added for this purpose but the css was never implemented.

### Test
1. `npm run dev`
2. Close chrome dev tools
3. Is app draggable?
4. Run the same test on develop
5. App should not be draggable. 

### Review

### Release
`RELEASE-NOTES.txt` To be added.

### DIFF
--- a/scss/_general.scss
+++ b/scss/_general.scss
@@ -104,6 +104,14 @@ optgroup {
   }
 }
 
+.login__draggable-area {
+  height: 100px;
+  position: absolute;
+  top: 0;
+  width: 100%;
+  -webkit-app-region: drag;
+}
+
 .gridicon {
   fill: currentColor;
 }


dmsnell:
I tried to test this but I couldn't tell if the change wasn't working for me or if I was testing wrong. It appears like the existing behavior remains: if I try and drag the app around by clicking-and-dragging on the app's main panel it does nothing; but if I move the mouse until the cursor turns into a resize pointer and click-and-drag in a motion perpendicular to the resize then I can move the window.

https://cldup.com/JOwtije7P1.mov
(couldn't fit my test video in a small-enough GIF to embed here)

belcherj:
I do not think you were close enough to the top. You have to be in the first 100 pixels.

belcherj:
I experience the same result just now.  I opened Chrome dev tools again and closed it and it works. I also tested with Chrome dev tools in detached mode and it works. I think dev tools is doing something funky.

belcherj:
This should resolve the issue:  https://github.com/Automattic/simplenote-electron/pull/1660

dmsnell:
I tried merging this branch into `update/open-dev-tools-inseperate-window` and still had the same trouble moving the window. Tried with the dev tools closed too. They _do_ open in detached mode as expected, but I still can't drag the window around and move it.

belcherj:
Can you give it another test? Detached mode has been merged and this branch is now up to date with develop.

belcherj:
Oct-28-2019 15-40-12

belcherj:
It is only the top that should move like native apps. When selecting text in the input fields we wouldn't want the app to move.

This gets us exactly where we want to be. 😉

PR #1682, opened 2019-10-30
Update signing certificate

dmsnell:
Replaces the expired signing certificate with a new one and removes one superfluous logging statement.

Props to @loremattei for creating these changes


dmsnell:
@belcherj when I went to "View File" it showed the proper size of 3.2 KB so maybe that's just saying there are zero additions? It looks correct to me.

PR #1700, opened 2019-11-07
Ensure that only one tab or window is open per browser

belcherj:
### Fix
When two tabs are open a bug occurs that changes will repeat infinitely. This
is caused by each tab or window having it's own web socket while sharing the
same instace of indexedDB. When a change happens in one tab it gets made in the
second tab via the web socket and indexedDB.

Screen Shot 2019-11-11 at 1 01 42 PM

### Test
1. Open the app in a tab
2. Open app in a second tab
3. You should see a message stating that the app can only be opened in one tab.

### Review
Only one developer and one designer are required to review these changes, but anyone can perform the review.

### Release
***(Required)*** Add a concise statement to `RELEASE-NOTES.txt` if the changes should be included in release notes. Include details about updating the notes in this section. For example:
> `RELEASE-NOTES.txt` was updated in d3adb3ef with:
> 
> > Added markdown support

If the changes should not be included in release notes, add a statement to this section. For example:

> These changes do not require release notes.

### DIFF
--- a/lib/boot.js
+++ b/lib/boot.js
@@ -1,3 +1,4 @@
+import './utils/ensure-single-browser-tab-only';
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 import 'unorm';
--- a/lib/components/boot-warning/index.jsx
+++ b/lib/components/boot-warning/index.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './style';
+
+const BootWarning = () => {
+  return (
+    <h3 className="boot-warning__message">
+      Simplenote cannot be opened simultaneously in more than one tab or window
+      per browser.
+    </h3>
+  );
+};
+
+export default BootWarning;
--- a/lib/utils/ensure-single-browser-tab-only.js
+++ b/lib/utils/ensure-single-browser-tab-only.js
@@ -1,0 +1,77 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import BootWarning from '../components/boot-warning';
+
+const HEARTBEAT_DELAY = 1000;
+const clientId = uuidv4();
+const emptyLock = [null, -Infinity];
+const foundElectron = window.process && window.process.type;
+
+if (!foundElectron) {
+  if ('lock-acquired' !== grabSessionLock()) {
+    ReactDOM.render(<BootWarning />, document.getElementById('root'));
+    throw new Error('Simplenote can only be opened in one tab');
+  }
+  let keepGoing = true;
+  loop(() => {
+    if (!keepGoing) {
+      return false;
+    }
+    switch (grabSessionLock()) {
+      case 'lock-acquired':
+        return true; // keep updating the lock and look for other sessions which may have taken it
+
+      default:
+        window.alert(
+          "We've detected another session running Simplenote, this may cause problems while editing notes. Please refresh the page."
+        );
+        return false; // stop watching - the user can proceed at their own risk
+    }
+  });
+
+  window.addEventListener('beforeunload', function() {
+    keepGoing = false;
+    const [lastClient] =
+      JSON.parse(localStorage.getItem('session-lock')) || emptyLock;
+
+    lastClient === clientId && localStorage.removeItem('session-lock');
+  });
+}
+
+function uuidv4() {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
+    var r = (Math.random() * 16) | 0,
+      v = c == 'x' ? r : (r & 0x3) | 0x8;
+    return v.toString(16);
+  });
+}
+
+function loop(f, delay = HEARTBEAT_DELAY) {
+  f() && setTimeout(() => loop(f, delay), delay);
+}
+
+function grabSessionLock() {
+  const [lastClient, lastBeat] =
+    JSON.parse(localStorage.getItem('session-lock')) || emptyLock;
+  const now = Date.now();
+  // easy case - someone else clearly has the lock
+  // add some hysteresis to prevent fighting between sessions
+  if (lastClient !== clientId && now - lastBeat < HEARTBEAT_DELAY * 5) {
+    return 'lock-unavailable';
+  }
+
+  // maybe nobody clearly has the lock, let's try and set it
+  localStorage.setItem('session-lock', JSON.stringify([clientId, now]));
+
+  // hard case - localStorage is shared mutable state across sessions
+  const [thisClient, thisBeat] =
+    JSON.parse(localStorage.getItem('session-lock')) || emptyLock;
+
+  // someone else set localStorage between the previous two lines of code
+  if (!(thisClient === clientId && thisBeat === now)) {
+    return 'lock-unavailable';
+  }
+
+  return 'lock-acquired';
+}
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,6 +7,7 @@
 ### Fixes
 
 - Fixed markdown code styles [#1702](https://github.com/Automattic/simplenote-electron/pull/1702)
+- Only allow app to load in one instance per browser [#1700](https://github.com/Automattic/simplenote-electron/pull/1700)
 
 ### Other Changes
 
--- a/lib/components/boot-warning/style.scss
+++ b/lib/components/boot-warning/style.scss
@@ -1,0 +1,4 @@
+h3.boot-warning__message {
+  margin: 50px auto;
+  width: 50%;
+}


lib/boot.js
```diff
@@ -1,3 +1,19 @@
...
+    message.innerText =
+      'Simplenote cannot currently be opened simultaneously in more than one tab or window per browser';
+    document.body.appendChild(message);
+    throw new Error('Simplenote can only be opened in one tab');
+  }
+  localStorage.setItem('appOpened', true);
+  window.addEventListener('unload', function(event) {
+    localStorage.removeItem('appOpened');
+  });
+}
```
dmsnell: it might be helpful to split this into a separate module to keep the logic separate and avoid the noise…

```js
import './ensure-single-browser-tab-only';
import 'core-js/stable';
import 'regenerator-runtime/runtime';
…
```

dmsnell: I'm not a fan of relying on `unload` behavior. If the browser dies or someone does something funny then we're going to leave stale state in `localStorage`. In order to keep consistency I think we'll probably need to create a time-based heartbeat. This is going to get really wicked really quickly and we're going to suffer when the clock lies.

```js
const HEARTBEAT_DELAY = 1000;
const clientId = uuid4(); // something (pseudo) random
const emptyLock = [ null, -Infinity ];

if ( 'lock-acquired' !== grabSessionLock() ) {
	ReactDOM.render(
		<ModalBootWarning />
		document.getElementById( 'root' )
	);
	throw new Error( 'Simplenote can only be opened in one tab' );
}

loop( () => {
	switch ( grabSessionLock() ) {
		case 'lock-acquired':
			return true; // keep updating the lock and look for other sessions which may have taken it

		default:
			window.alert( "We've detected another session running Simplenote. This shouldn't be. Expect problems." );
			return false; // stop watching - the user can proceed at their own risk
	}
} );

function loop( f, delay ) {
	f() && setTimeout( () => loop( f, delay ), delay );
}

function grabSessionLock( clientId ) {
	const [ lastClient, lastBeat ] = localStorage.getItem( 'session-lock' ) || emptyLock;
	const now = Date.now();

	// easy case - someone else clearly has the lock
  // add some hysteresis to prevent fighting between sessions
	if ( lastClient !== clientId && now - lastBeat < HEARTBEAT_DELAY * 1.5 ) {
		return 'lock-unavailable';
	}

	// maybe nobody clearly has the lock, let's try and set it
	localStorage.setItem( 'session-lock', [ clientId, now ] );

	// hard case - localStorage is shared mutable state across sessions
	const [ thisClient, thisBeat ] = localStorage.getItem( 'session-lock' ) || emptyLock;

	// someone else set localStorage between the previous two lines of code
	if ( ! ( thisClient === clientId && thisBeat === now ) ) {
		return 'lock-unavailable';
	}

	return 'lock-acquired';
}
```

lib/boot.js
```diff
@@ -1,3 +1,19 @@
+const foundElectron = window.process && window.process.type;
+
+if (!foundElectron) {
+  if (localStorage.getItem('appOpened')) {
+    const message = document.createElement('h1');
+    message.innerText =
+      'Simplenote cannot currently be opened simultaneously in more than one tab or window per browser';
+    document.body.appendChild(message);
```
dmsnell: this is fine though we are free to use React since it will have been delivered in this bundle already

dmsnell:
Some thoughts for you @belcherj in the comments.

Once we buy some time with this I had an idea: we might be able to push `node-simperium` entirely into a `SharedWorker` or `ServiceWorker` (all major browsers support one or the other or both).

Since our API is already asynchronous it wouldn't change the interface to Simperium. We may or may not have problems with sending changes to it from multiple clients but it's one idea.

Otherwise we'll have to do something with the `Bucket` / `objectStore` and eliminate the shared state. I think the `ghost` is safe being shared.

lib/boot.js
belcherj: Done.

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,69 @@
...
+        return true; // keep updating the lock and look for other sessions which may have taken it
+
+      default:
+        return false; // stop watching - the user can proceed at their own risk
+    }
+  });
+
+  window.addEventListener('unload', function(event) {
+    localStorage.removeItem('session-lock');
+  });
```
dmsnell: we need to be careful here. it's a good point you bring up.

while this isn't needed to free the lock generally, you've pointed out that we have the opportunity to free it for ourselves in some or most cases

I'm worried that in unloading we could run after another session has grabbed the lock and then destroy that session. we may not be able to eliminate this but we can probably add one tiny guard to mitigate most races

```js
let keepGoing = true;

…
loop(() => {
	if ( ! keepGoing ) {
		return false;
	}

	…
} );
…

window.addEventListener('unload', function(event) {
	keepGoing = false;
	const [ lastClient ] = localStorage.getItem( 'session-lock' );
	lastClient === clientId && localStorage.removeItem( 'session-lock' );
} );
```

this is based on the untested assumption that we might incur a delay between killing the page, shutting down our timers, and running the `unload` code

probably also responsible to make sure we don't finish running our event queue and inadvertently grab the lock after we've killed the page.

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+function uuidv4() {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
+    var r = (Math.random() * 16) | 0,
+      v = c == 'x' ? r : (r & 0x3) | 0x8;
+    return v.toString(16);
+  });
+}
```
dmsnell: nice. this should also be available in one of our transient dependencies `uuid`

that is, it's great that there are no dependencies, though we already pull this in, so maybe we should just use it.

you're call. don't think either one is clearly better but wanted to point out that it's in there already

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,69 @@
...
+  const now = Date.now();
+
+  // easy case - someone else clearly has the lock
+  // add some hysteresis to prevent fighting between sessions
+  if (lastClient !== clientId && now - lastBeat < HEARTBEAT_DELAY * 1.5) {
+    return 'lock-unavailable';
+  }
+
+  // maybe nobody clearly has the lock, let's try and set it
+  localStorage.setItem('session-lock', JSON.stringify([clientId, now]));
```
dmsnell: > JSON.stringify

good catch

lib/components/boot-warning/index.jsx
```diff
@@ -0,0 +1,15 @@
...
+import './style';
+
+const BootWarning = () => {
+  return (
+    <h1>
+      Simplenote cannot currently be opened simultaneously in more than one tab
+      or window per browser.
+    </h1>
+  );
+};
```
dmsnell: @SylvesterWilmott would you be willing to review this? We have this ongoing problem caused by the fact that if the app is open in more than two tabs/sessions/windows _of the same browser_ (multiple copies in Firefox /or/ multiple copies in Chrome /or/ multiple copies in Safari etc…) then the workers that store updates fight and cause the _infinite duplication_ bug.

Just need to communicate that.

lib/utils/ensure-single-browser-tab-only.js
belcherj: If this code was intended to live forever I'd pull it in.

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import BootWarning from '../components/boot-warning';
+
+const HEARTBEAT_DELAY = 1000;
+const clientId = uuidv4();
+console.log(clientId);
```
dmsnell: if we want to leave this in we should turn it into `debug`

```js
import debugFactory from 'debug';
const debug = debugFactory('session-lock');

…

debug(`clientId: ${ clientId }`);
```

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,78 @@
...
+const emptyLock = [null, -Infinity];
+const foundElectron = window.process && window.process.type;
+
+if (!foundElectron) {
+  const sesh = grabSessionLock();
+  console.log(sesh);
```
dmsnell:
```js
debug(`tried to acquire lock: ${ lockAttempt }`);
```

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,78 @@
...
+import BootWarning from '../components/boot-warning';
+
+const HEARTBEAT_DELAY = 1000;
+const clientId = uuidv4();
+console.log(clientId);
+const emptyLock = [null, -Infinity];
+const foundElectron = window.process && window.process.type;
+
+if (!foundElectron) {
+  const sesh = grabSessionLock();
```
dmsnell: can we use something less abbreviated than `sesh`? I'm afraid this might be confusing in the code

 - `lockAttempt`
 - `lockStatus`
 - `sessionLock`
 - something better

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,78 @@
...
+  loop(() => {
+    if (!keepGoing) {
+      return false;
+    }
+    switch (grabSessionLock()) {
+      case 'lock-acquired':
+        return true; // keep updating the lock and look for other sessions which may have taken it
+
+      default:
+        window.location.reload();
```
dmsnell: I'm afraid that this will cause unexpected refreshes and maybe a refresh cycle.

at this point we have booted the app and acquired the lock at least once. we never expect this scenario but it could happen in at least a few rare circumstances:
 - our app slows down. maybe a very large not takes longer to update and delays the lock update. another session could have stolen the lock in that delay window.
 - someone directly messes with `localStorage` in the browser console.

should these things happen I don't know how best to respond but at this point we're taking the nuclear option and reloading everything, possibly interrupting someone who is working on a note and without alerting them to why.

the `window.alert()` was an attempt at putting up a modal they couldn't ignore that warns them of their predicament and gives them a choice to reload or not; to wait a second and see if they are willing to risk the bug in order to try and save what they had open.

thoughts?

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,78 @@
...
+        return false; // stop watching - the user can proceed at their own risk
+    }
+  });
+
+  window.addEventListener('beforeunload', function() {
+    keepGoing = false;
+    const [lastClient] = JSON.parse(localStorage.getItem('session-lock'));
+    console.log(lastClient, clientId);
```
dmsnell:
```js
debug(`last-known lock held by ${ lastClient }`);
```

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,78 @@
...
+  if (lastClient !== clientId && now - lastBeat < HEARTBEAT_DELAY * 5) {
+    return 'lock-unavailable';
+  }
+
+  // maybe nobody clearly has the lock, let's try and set it
+  localStorage.setItem('session-lock', JSON.stringify([clientId, now]));
+
+  // hard case - localStorage is shared mutable state across sessions
+  const [thisClient, thisBeat] = JSON.parse(
+    localStorage.getItem('session-lock') || emptyLock
```
dmsnell: `emptyLock` isn't JSON and `Infinity` collapses to `null` when encoded that way.

we can use a try/catch helper…

```js
const [ thisClient, thisBeat ] = parseJSON( localStorage.getItem( 'session-lock' ), emptyLock );

function parseJSON( s, fallbackValue ) {
	try {
		return JSON.parse( s );
	} catch ( e ) {
		return fallbackValue;
	}
}
```

dmsnell: or go one step further…

```js
const [ thisClient, thisBeat ] = load( 'session-lock', emptyLock );
store( 'session-lock', [ clientId, now ] );

function load( name, fallbackValue ) {
	try {
		return JSON.parse( localStorage.getItem( name ) );
	} catch ( e ) {
		return fallbackValue;
	}
}

function store( name, value ) {
	localStorage.setItem( name, JSON.stringify( value ) );
}
```

belcherj: It was never intended to be inside the JSON.parse.

belcherj: I think that is overkill. But that's me.

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,76 @@
...
+    if (!keepGoing) {
+      return false;
+    }
+    switch (grabSessionLock()) {
+      case 'lock-acquired':
+        return true; // keep updating the lock and look for other sessions which may have taken it
+
+      default:
+        window.alert(
+          "We've detected another session running Simplenote. This shouldn't be. Expect problems. A refresh will help."
```
dmsnell: cc: @SylvesterWilmott for copy text and the general UX question of "what do we do when we've been running fine and suddenly realize that another fighting copy of the app has loaded?"

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,76 @@
...
+        );
+        return false; // stop watching - the user can proceed at their own risk
+    }
+  });
+
+  window.addEventListener('beforeunload', function() {
+    keepGoing = false;
+    const [lastClient] = JSON.parse(localStorage.getItem('session-lock'));
```
dmsnell: this will throw a `TypeError` if the key has been removed and `JSON.parse()` returns `null`

did we forget to `|| emptyLock`?

SylvesterWilmott:
This looks fine to me to put out now. Is the intention to leave this as a stop-gap solution until we can find another way to manage it?

lib/components/boot-warning/index.jsx
SylvesterWilmott: Perhaps using `<p>` instead or even `<h3>`

Aside from that, the copy looks good, I'm not sure we need the word "currently", although I understand what we're trying to communicate it does make the sentence a bit more of a mouthful.

"Simplenote cannot be opened simultaneously in more than one tab or window per browser"

lib/utils/ensure-single-browser-tab-only.js
SylvesterWilmott: Just to clarify, am I right in assuming I would see this message in the first instance of Simplenote if I opened a new instance in another browser window?

belcherj:
Yes, this is intended as a stop gap.

lib/utils/ensure-single-browser-tab-only.js
dmsnell: This "should never occur" because if our lock is working then the second window will abort its boot and display the message "Can't open two tabs"

There are cases where this could occur, such as when processing note updates take exceptionally long, when an extension interferes in some unpredictable way, when someone clears `localStorage` in their browser, or when something else unexpected breaks our locking mechanism.

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -0,0 +1,77 @@
...
+  f() && setTimeout(() => loop(f, delay), delay);
+}
+
+function grabSessionLock() {
+  const [lastClient, lastBeat] =
+    JSON.parse(localStorage.getItem('session-lock')) || emptyLock;
+  const now = Date.now();
+  // easy case - someone else clearly has the lock
+  // add some hysteresis to prevent fighting between sessions
+  if (lastClient !== clientId && now - lastBeat < HEARTBEAT_DELAY * 5) {
```
dmsnell: this is fine, just noting a change - did we change the hysteresis from 1.5 to 5 intentionally? Totally fine if we did - I just want to make sure we intended on it.

belcherj: I made this change intentionally.  While testing I found that 1.5 was not sufficient as I getting through the check.

lib/utils/ensure-single-browser-tab-only.js
SylvesterWilmott: "We've detected another session running Simplenote, expect problems to occur. Please refresh the page."

SylvesterWilmott: "We've detected another session running Simplenote, this may cause problems while editing notes. Please refresh the page."

PR #1705, opened 2019-11-09
Search: Prioritize notes with search matches in their title

dmsnell:
We have been returning search results in no particular order. When we
get searches that return many notes though the noise can make the search
less valuable.

In this patch we're starting to prioritize search results where the
first line of the note contains the string searched for. This is in
effect saying that we're prioritizing the note title matches in
searches.

**Testing**

This is best tested on an account with many notes so that we can
compare the search results.

Find keywords that are both in the note titles of some notes and
also in the content of many notes.

In `develop` we'd expect that the search results will be muddled
by the prevalence of notes matching the query string, but in this
branch we'd expect title-matching notes to appear at the top.

**Before**
UnprioritizedSearch mov

**After**
TitlePrioritizedSearch mov

### DIFF
--- a/lib/utils/filter-notes.js
+++ b/lib/utils/filter-notes.js
@@ -59,7 +59,7 @@ export const searchPattern = filter => {
 const matchesTrashView = isViewingTrash => note =>
   isViewingTrash === !!get(note, 'data.deleted', false);
 
-const matchesTag = (tag, filter = '') => note => {
+const makeMatchesTag = (tag, filter = '') => note => {
   let filterTags = [];
   let match;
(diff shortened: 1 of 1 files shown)


belcherj (approved):
Im good with this if @codebykat thinks this is in line with the Search updates.

codebykat (approved):
I don't know how I missed this! This is great! Saves me some work 😇 I'd love to merge this before Monday so it can be considered part of the search milestone.

I tested on my personal account with a lot of notes and the result ordering is noticeably better. 🌟

PR #1713, opened 2019-11-12
Crash on missing platform support

dmsnell:
The Simplenote app, for better or worse, currently depends on a few
storage technologies that may be disabled through browser settings,
extensions, and manual configuration.

When these dependencies are missing we cannot function normally and
right now the app crashes or breaks in misleading ways.

In this patch we perform an up-front check when the app boots to verify
that we have some basic dependencies met. This is not an exhaustive
validation but it should catch the most common cases and give customers
an idea of why things may be broken.

SimplenoteMissingDeps mov

### DIFF
--- a/lib/utils/ensure-platform-support.js
+++ b/lib/utils/ensure-platform-support.js
@@ -1,0 +1,84 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { parse as parseCookie } from 'cookie';
+
+import BootWarning from '../components/boot-warning';
+
+const hasLocalStorage = () => {
+  try {
+    localStorage.setItem('__localStorageSentinel__', 'present');
+    localStorage.removeItem('__localStorageSentinel__');
+    return true;
+  } catch (e) {
+    return false;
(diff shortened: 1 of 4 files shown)


lib/utils/ensure-platform-support.js
```diff
@@ -0,0 +1,84 @@
...
+];
+
+const missingDeps = deps.filter(([, hasIt]) => !hasIt).map(([name]) => name);
+
+if (missingDeps.length) {
+  ReactDOM.render(
+    <BootWarning>
+      <p>
+        Simplenote depends on a few web technologies to operate. Please make
+        sure that you have all of the following enabled in your browser.
```
dmsnell: @SylvesterWilmott would you be willing to provide some copy-review on this?

This message is the only thing that would appear to a customer if they try to load Simplenote without having the necessary support enabled in the browser.

SylvesterWilmott: The copy looks good to me! Thanks @dmsnell

PR #1720, opened 2019-11-14
Conditionally watch event for unload to release session lock

belcherj:
iOS does not have a beforeunload event. The pagehide event for iOS causes
issues for Firefox. This conditionally watches the event based on if the
browser is iOS or not.

### Test
1. Open two tabs in Firefox, does the tab lock work correctly
2. Once this is merged and deployed to develop test on actual iPad

### Review

### DIFF
--- a/lib/utils/ensure-single-browser-tab-only.js
+++ b/lib/utils/ensure-single-browser-tab-only.js
@@ -44,7 +44,6 @@ if (!foundElectron) {
     keepGoing = false;
     const [lastClient] =
       JSON.parse(localStorage.getItem('session-lock')) || emptyLock;
-    console.log('session lock removed');
     lastClient === clientId && localStorage.removeItem('session-lock');
   }
 }
--- a/lib/utils/ensure-single-browser-tab-only.js
+++ b/lib/utils/ensure-single-browser-tab-only.js
@@ -36,14 +36,15 @@ if (!foundElectron) {
     }
   });
 
-  window.addEventListener('pagehide', unload);
-  window.addEventListener('beforeunload', unload);
-
+  const eventName = /iPad|iPhone|iPod/.test(navigator.userAgent)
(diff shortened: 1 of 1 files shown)


lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -36,14 +36,14 @@ if (!foundElectron) {
     }
   });
 
-  window.addEventListener('pagehide', unload);
-  window.addEventListener('beforeunload', unload);
-
+  const eventName = /iPad|iPhone|iPod/.test(navigator.userAgent)
+    ? 'pagehide'
+    : 'beforeunload';
```
dmsnell: would it be wise to perform feature-detection here instead of user-agent guessing?

```js
const eventName = 'undefined' !== typeof window.onbeforeunload
	? 'beforeunload'
	: 'pagehide'
```

lib/utils/ensure-single-browser-tab-only.js
```diff
@@ -36,14 +36,14 @@ if (!foundElectron) {
...
+  window.addEventListener(eventName, unload);
```
dmsnell: further, for the sake of making it easier to find this in the code with normal search tools, could we duplicate the hookup here?

```js
if ( 'undefined' !== typeof window.onbeforeunload ) {
	window.addEventListener( 'beforeunload', unload );
} else {
	window.addEventListener( 'pagehide', unload );
}
```

^^^ just a pet peeve of mine where we take expected string contents and hide them inside a variable.

dmsnell:
> Remove debugging

I hope that our _de-bugging_ will persist 😉

lib/utils/ensure-single-browser-tab-only.js
belcherj: done

PR #1723, opened 2019-11-15
Add script to deploy web app

belcherj:
### Fix
Losing GitHub Actions we made the call to make deployment local rather than
adding another level of CI

### Test
1. Checkout branch
2. `npm run deploy staging`
3. Does it push to webapp-staging?
4. Does staging.

### Review

### Release
 - Added script to deploy web app [#1723](https://github.com/Automattic/simplenote-electron/pull/1723)

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -63,7 +63,7 @@ fi
 git checkout -b $COMMIT_BRANCH
 
 # Delete files not needed in production
-find . -maxdepth 1 ! -name '.git' ! -name 'dist' ! -name 'vip' ! -name 'config-local.json' -delete
+find . -maxdepth 1 ! -name '.git' ! -name 'dist' ! -name 'vip' ! -name 'config-local.json' -exec rm -rf {} ';'
 
 # Copy vip files to base path
 cp -r vip/* ./
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -67,7 +67,9 @@ find . -maxdepth 1 ! -name '.git' ! -name 'dist' ! -name 'vip' -exec rm -rf {} \
 
 # Copy vip files to base path
 cp -r vip/* ./
+rm -rf vip
 
+# Add, commit, push
 touch .gitignore
 echo config-local.json >> .gitignore
 git add --all
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -1,0 +1,82 @@
+#!/bin/sh -l
+
+COMMIT_BRANCH=''
+WEB_APP_URL=''
+APP_ENGINE_URL=''
+
+if [ -z "$1" ]
+then
+  echo "You did not specify a deploy environment, your choices are: production, develop, staging"
+  exit 1
+fi
+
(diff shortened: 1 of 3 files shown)


bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
...
+  *)
+    exit 0
+    ;;
+esac
+
+# Install Node Modules
+npm install
+
+# Temporarily move config-local so it doesn't get overwritten
+mv config-local.json config-local.origional.json
```
codebykat: typo for "original"

bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Clean up old builds
+rm -rf dist
+
+# Build site
+NODE_ENV=production make build
+
+# Restore config-local
+rm config-local.json
+mv config-local.origional.json  config-local.json
```
codebykat: "original"

bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Checkout deploy branch
+if [ "$(git branch | grep $COMMIT_BRANCH)" ]
+then
+  git branch -D $COMMIT_BRANCH
+fi
+git checkout -b $COMMIT_BRANCH
+
+# Delete files not needed in production
+find . -maxdepth 1 ! -name '.git' ! -name 'dist' ! -name 'vip' -exec rm -rf {} \;
```
codebykat: It looks like this find is deleting the config .json files

bin/deploy.sh
belcherj: fixed

bin/deploy.sh
belcherj: fixed

bin/deploy.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Copy vip files to base path
+cp -r vip/* ./
+rm -rf vip
+
+# Add, commit, push
+touch .gitignore
+echo config-local.json >> .gitignore
+git add --all
+git commit -m "Build: $COMMIT_BRANCH [ci skip]"
```
dmsnell: if we're wanting to deploy a given branch why are we building this and making a commit to that branch?
do we not have a way to inject a different config for the different environments?

seems like we ought to be able to do it like this…

```bash
git checkout webapp-develop
git reset --hard develop
```

Does something in the build process push itself out to VIP Go?

belcherj: When the head of the webapp, webapp-staging, and webapp-develop branches change VIP Go grabs the contents of the repo and pushes it to the server.

We do not have a way to inject a different config for the different environments.

I am not sure what the code suggestion achieves.

dmsnell: the code suggestion would change the `HEAD` of those branches and should trigger web hooks to fire. the branches effectively would work like tags.

belcherj: VIP Go doesn't build the app it only grabs what's in the folder. https://github.com/Automattic/simplenote-electron/tree/webapp

PR #1726, opened 2019-11-16
Search: Add sort order to search results

### LINKED ISSUE #1625: Search: Add "sort by" to search results
## Updated designs:

Screen Shot 2020-12-23 at 10 28 55 PM

https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fd77f1846903987b1b40e75

Screen Shot 2020-12-23 at 10 35 01 PM

https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fd77f196e43637c2cb2889f

## Old designs below

See designs at p2XJRt-1k0-p2

Screen Shot 2019-10-04 at 1 23 44 PM

  
Screen Shot 2019-10-04 at 1 25 18 PM

Options: Relevance, alphabetically, created, modified

codebykat:
### Enhancement
This PR adds an ordering toggle to the bottom of the notes list when a search is active:

Screen Shot 2019-11-27 at 10 38 23 PM

Will fix #1625

### Test
1. Enter a search
2. Use the toggle to change the sort type and direction of the results.
3. Clear the search and verify that your global sort settings haven't been changed.
4. Start a new search and note that the sorting resets to the global setting.
5. Try in different browsers/OSes to make sure the dropdown styling is okay

*NOTE:* Title matches will get sorted to the top of the list as per #1705. So there's basically two separate sections: title and content. Within each section the notes should get sorted as per preferences. It's out of scope but I'd like to make this clearer in a separate PR.

### Dark Mode screenshots
Screen Shot 2019-11-27 at 10 37 51 PM
Screen Shot 2019-11-27 at 10 37 56 PM

### Release

> `RELEASE-NOTES.txt` was updated in 
932676d  with:
> 
> > Added sort order to search results [#1726](https://github.com/Automattic/simplenote-electron/pull/1726)

### DIFF
--- a/lib/sort-order-selector/index.jsx
+++ b/lib/sort-order-selector/index.jsx
@@ -22,6 +22,21 @@ export class SortOrderSelector extends Component {
     setSortType: PropTypes.func.isRequired,
   };
 
+  componentDidUpdate(prevProps) {
+    const { sortType, sortReversed, searchSortReversed, query } = this.props;
+    /* If there's a query and there wasn't before
+       (i.e., the search sort component is being newly displayed),
+       set the search sort to match the global setting.
+       This means whenever a user starts a new search, the notes will be filtered
+       in place, rather than being reordered to match the last selected search order. */
+    if (!prevProps.query && prevProps.query !== query) {
+      this.props.setSortType(sortType);
+      if (sortReversed !== searchSortReversed) {
+        this.props.toggleSortOrder();
+      }
+    }
+  }
+
   changeSort = event => {
     this.props.setSortType(event.currentTarget.value);
   };
--- a/lib/sort-order-selector/index.jsx
+++ b/lib/sort-order-selector/index.jsx
@@ -2,7 +2,10 @@ import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import appState from '../flux/app-state';
-import { toggleSortOrder, setSortType } from '../state/settings/actions';
+import {
+  toggleSearchSortOrder,
+  setSearchSortType,
+} from '../state/settings/actions';
 import classNames from 'classnames';
 import ArrowDownIcon from '../icons/arrow-down';
 
@@ -13,6 +16,8 @@ export class SortOrderSelector extends Component {
     query: PropTypes.string.isRequired,
     sortType: PropTypes.string.isRequired,
     sortReversed: PropTypes.bool.isRequired,
+    searchSortType: PropTypes.string.isRequired,
+    searchSortReversed: PropTypes.bool.isRequired,
     toggleSortOrder: PropTypes.func.isRequired,
     setSortType: PropTypes.func.isRequired,
   };
@@ -22,7 +27,7 @@ export class SortOrderSelector extends Component {
   };
 
   render() {
-    const { sortType, sortReversed, query } = this.props;
+    const { searchSortType, searchSortReversed, query } = this.props;
 
     const sortTypes = [
       {
@@ -45,14 +50,14 @@ export class SortOrderSelector extends Component {
           <div className="sort-order-selector">
             <div
               className={classNames('sort-order-reverse', {
-                'is-reversed': sortReversed,
+                'is-reversed': searchSortReversed,
               })}
               onClick={this.props.toggleSortOrder}
             >
               <ArrowDownIcon />
             </div>
             Sort by
-            <select value={sortType} onChange={this.changeSort}>
+            <select value={searchSortType} onChange={this.changeSort}>
               {sortTypes.map(type => (
                 <option key={type.id} value={type.id}>
                   {type.label}
@@ -69,6 +74,8 @@ export class SortOrderSelector extends Component {
 const mapStateToProps = ({ settings, appState: state }) => ({
   sortType: settings.sortType,
   sortReversed: settings.sortReversed,
+  searchSortType: settings.searchSortType,
+  searchSortReversed: settings.searchSortReversed,
   query: state.filter,
 });
 
@@ -80,8 +87,8 @@ function mapDispatchToProps(dispatch, { noteBucket }) {
     dispatch(actionCreators.loadNotes({ noteBucket }));
   };
   return {
-    setSortType: thenReloadNotes(setSortType),
-    toggleSortOrder: thenReloadNotes(toggleSortOrder),
+    setSortType: thenReloadNotes(setSearchSortType),
+    toggleSortOrder: thenReloadNotes(toggleSearchSortOrder),
   };
 }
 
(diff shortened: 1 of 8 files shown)


lib/state/settings/actions.js
```diff
@@ -52,6 +52,18 @@ export const setSortType = sortType => ({
...
+export const setSearchSortType = searchSortType => ({
+  type: 'setSearchSortType',
+  searchSortType,
+});
+
+export const toggleSearchSortOrder = () => (dispatch, getState) => {
+  dispatch({
+    type: 'setSearchSortReversed',
+    searchSortReversed: !getState().settings.searchSortReversed,
+  });
```
dmsnell: <s>here's a philosophical point of reflection: where is best to store the "toggle" state?

we can do it here in Redux middleware, which makes the UI components easier because they only need to call `toggle()`, but it complicates the Redux flows because the actions are no longer predictable.

we can do it in the UI and pass the next state, but that means we have to track it in the UI component and so complicates/duplicates it there.

one place I've come to accept is to use a hybrid of `mapStateToProps()` to get the existing state, `mapDispatchToProps()` to get the dispatch function, and occasionally, rarely, to use `mergeProps()` (also part of `connect()` but usually one we leave out) to keep the UI API clean and agnostic about the toggle state.

---

usually these complications are indications of a missing Eureka moment in a data structure somewhere. why is it hard for the UI to send the toggle state? (hint, boolean values should always raise our suspicion)

---

🤔 what if we stored `sortDirection` instead of `isReversed`?

```js
export const toggleSearchDirectionFrom = direction => ( {
  type: SEARCH_SET_DIRECTION,
  direction: direction === 'normal' ? 'reversed' : 'normal',
} )
```

### Manage it in the UI
not so bad because the state is simple enough and clear enough

```js
toggleSearchDirection = () => this.props.toggleSearchDirection( this.props.searchDirection )

<button onClick={ this.toggleSearchDirection } />

const mapStateToProps = ( { search: { direction } } ) => ( {
  searchDirection: direction,
} )

const mapDispatchToProps = dispatch => ( {
  toggleSearchDirection: direction => dispatch( toggleSearchDirectionFrom( direction ) ),
} )
```

### use `mergeProps` to remove the knowledge from the UI
more complicated `connect()` but more straightforward UI

```js
<button onClick={ this.props.toggleSearchDirection } />

const mapStateToProps = ( { search: { direction } } ) => ( {
  searchDirection: direction,
} )

const mapDispatchToProps = dispatch => ( {
  setSearchDirection: direction => dispatch( toggleSearchDirectionFrom( direction ) ),
} )

const mergeProps = ( stateProps, dispatchProps, ownProps ) => ( {
  ...ownProps,
  ...stateProps,
  ...dispatchProps,
  toggleSearchDirection: () => dispatchProps.toggleSearchDirection( ownProps.searchDirection ),
} )

connect( mapStateToProps, mapDispatchToProps, mergeProps )( Component );
```

Of course, that looks scary, especially if unfamiliar with it. I'm a fan of tracking this in the UI until it's clear we need more abstraction.  In any case, I still prefer keeping middleware behaviors outside of the action creators because of the pain I've had in trying to debug `redux thunk` stuff like this. In #1725 I recently proposed a new Redux middleware just for the purpose of executing the current search query (which is of course a bit more complicated because the search params are complicated). It's not that much more than Redux thunk actions but it preserves the predictability and audibility of our action sequence. [[adding middleware](https://github.com/Automattic/simplenote-electron/pull/1725/files#diff-e65c9a1430a959fb1c65a34167163c89R38), [defining middleware](https://github.com/Automattic/simplenote-electron/pull/1725/files#diff-d2dde7710fab6b285db73d1a1383dba5R13)]

### Middleware for predictability and introspection

```js
export const toggleSortOrder = () => ( {
  type: SORT_ORDER_TOGGLE,
} );

export const sortDirection = ( state = 'normal', { type, direction } ) =>
  SORT_ORDER_SET === 
```
</s>

**Eureka!**

See? it was way simpler than we were realizing. This is best handled in the reducer where all the data lives. 

```js
export const sortOrder = ( state = 'normal', { type } ) =>
  SORT_ORDER_TOGGLE === type
    ? ( state === 'normal' ? 'reverse' : 'normal' )
    : state;

export toggleSortOrder = () => ( {
  type: SORT_ORDER_TOGGLE,
} );

<button onClick={ this.props.toggleSortOrder } />

const mapStateToProps = ( { search: { sortOrder } } ) => ( {
  sortOrder,
} )

const mapDispatchToProps = ( {
  toggleSortOrder
} )
```

lib/state/settings/reducer.js
```diff
@@ -35,6 +37,10 @@ function reducer(state = initialState, action) {
       return { ...state, markdownEnabled: action.markdownEnabled };
     case 'setNoteDisplay':
       return { ...state, noteDisplay: action.noteDisplay };
+    case 'setSearchSortReversed':
+      return { ...state, searchSortReversed: action.searchSortReversed };
+    case 'setSearchSortType':
+      return { ...state, searchSortType: action.searchSortType };
```
dmsnell: we would have good value in using a new reducer in a search directory where we can store all the filtering and sort properties. these reducers used to be separate but they were merged and that can cause issues, especially with regards to handling more than one action type in a given reducer. it might be worth our time to decouple these as they once were but that would be a job for another PR

you can see what I was exploring in https://github.com/Automattic/simplenote-electron/pull/1725/files#diff-a36f9dcd572ee4de0417fd7a1f181542

dmsnell (changes requested):
@codebykat this is coming along and it looks like quite the challenge to figure out the details and ways that things are intertwined. good job and good luck on this journey!

SylvesterWilmott:
Looks good!

We can an icon we can use for the reordering results in the new icon pack

lib/sort-order-selector/index.jsx
```diff
@@ -0,0 +1,110 @@
...
+    searchSortReversed: PropTypes.bool.isRequired,
+    toggleSortOrder: PropTypes.func.isRequired,
+    setSortType: PropTypes.func.isRequired,
+  };
+
+  componentDidUpdate(prevProps) {
+    const { sortType, sortReversed, searchSortReversed, query } = this.props;
+    /* If there's a query and there wasn't before
+       (i.e., the search sort component is being newly displayed),
+       set the search sort to match the global setting.
```
dmsnell: Why do we have local and global search orders? Would it make sense to only have a single global sort order that changes when we click to change it?

That would simplify the amount of code and complexity we have to introduce to support both sorting systems

lib/sort-order-selector/index.jsx
```diff
@@ -0,0 +1,110 @@
...
+      },
+      {
+        label: 'Alphabetical',
+        id: 'alphabetical',
+      },
+    ];
+
+    return (
+      <Fragment>
+        {query.length > 0 && (
```
dmsnell: If we leave this check for the calling parent component then it will clean up this function and we can remove the `<Fragment>`

It means moving the check to the parent component but at least then what it leaves us is a component that can stand on its own a bit better.

codebykat:
Closing in favor of #2542

Issue #1578, opened 2019-09-20
Bold headings in markdown editor mode


vacocege:
### What
The macOS version of Simplenote bolds headings in the editor while writing markdown notes, making it easier to visualize sections of content. Can we add this feature to the electron app?

### How
Simplenote macOS screenshot
Notice the bold text of the ## h2.
![Simplenote macOS screenshot](https://simplenoteblog.files.wordpress.com/2018/02/screen-shot-2018-02-20-at-2-08-15-pm.png?w=1330)
We could add a `data-heading=true` prop to HTML heading elements. Currently there isn't a difference between the HTML of headings and regular text therefore you can't use CSS to change the font weight.
![HTML of a heading element in the markdown editor](https://user-images.githubusercontent.com/55606968/65364615-da5ccb00-dbc7-11e9-9ff8-8d06265e69f4.png)

roxihe7880:
PR #1727 was closed because of performance reasons. A large part of the app may have to be rewritten in order to implement this functionality.

dmsnell:
> PR #1727 was closed because of performance reasons.

Just for clarity sake @roxihe7880 - it wasn't closed for performance reasons. It was closed more because it takes a piecemeal approach to highlighting Markdown syntax which might end up making it harder to do full syntax highlighting/formatting.

wolisox428:
All or nothing.

Issue #944, opened 2018-10-21
Feature request : basic syntax highlighting when editing markdown


andaag:
I'd really like to see some basic syntax highlighting of the markdown editor when editing.

When everything is flat white it can be hard to see where the next header begins, compared to the previous one. As it's just a "#" in front of it.

alexanderte:
Totally agree. Formatting headings in bold and showing backticks with monospace would make Markdown preview unnecessary for me.

melsophos:
I also agree with this proposal.

leonardonormando:
I was about to request this feature when found this post. Even simple syntax highlighting  in markdown would be great.

Just to exemplify:
![image](https://user-images.githubusercontent.com/798716/61825347-930bc580-ae36-11e9-8c9a-0c1eaa674d3e.png)

belcherj:
Thanks! I’ve added this to our feature requests!

PR #1727, opened 2019-11-17
Render headings as rich text in editor mode

jivox92362:
screenshot-20191117-011902

### Fix
Render headings as rich text in editor mode.  Bold headings are a helpful visual component and already implemented in the MacOS version of the app.

### Test
1. Type a heading such as `# h1` or `## h2` or `### h3`
2. The text should transform into that element: the font-size should increase and the font-weight should change to bold

### Related issues
* #944
* #418
* #1578
* https://github.com/facebook/draft-js/issues/1198

#### Before/After
screenshot-20191117-012853
screenshot-20191117-011902

### DIFF
--- a/lib/components/heading/test.js
+++ b/lib/components/heading/test.js
@@ -1,0 +1,8 @@
+import Heading from './';
+import React from 'react';
+import { shallow } from 'enzyme';
+
+test.each([1, 2, 3, 4, 5, 6])('renders h%d', level => {
+  const wrapped = shallow(<Heading headingLevel={level}>Test</Heading>);
+  expect(wrapped.type()).toBe(`h${level}`);
+});
--- a/lib/editor/heading-decorator.js
+++ b/lib/editor/heading-decorator.js
@@ -1,0 +1,26 @@
+import SimpleDecorator from 'draft-js-simpledecorator';
+
+import Heading from '../components/heading';
+
+const headingRegex = /^([#]{1,6})(\s+)([^\s]+)/g;
+
+const headingDecorator = () => {
+  const strategy = (contentBlock, callback) => {
+    const text = contentBlock.getText();
+    let match;
+
+    while ((match = headingRegex.exec(text)) !== null) {
+      const fullText = text;
+      const headingLevel = match[1].length;
+
+      const start = match.index;
+      const end = start + fullText.length;
+
+      callback(start, end, { headingLevel, children: fullText });
+    }
+  };
+
+  return new SimpleDecorator(strategy, Heading);
+};
+
+export default headingDecorator;
--- a/lib/components/heading/index.jsx
+++ b/lib/components/heading/index.jsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Heading = ({ headingLevel, children }) => {
+  const elementType = `h${headingLevel}`;
+  const elementClass = `heading heading__${headingLevel}`;
+
+  return React.createElement(
+    elementType,
+    { className: elementClass },
+    children
+  );
+};
+
+Heading.propTypes = {
+  headingLevel: PropTypes.number.isRequired,
+  children: PropTypes.node.isRequired,
+};
+
+export default Heading;
--- a/lib/components/heading/style.scss
+++ b/lib/components/heading/style.scss
@@ -1,0 +1,7 @@
+/* .heading {} */
+/* .heading__h1 {} */
+/* .heading__h2 {} */
+/* .heading__h3 {} */
+/* .heading__h4 {} */
+/* .heading__h5 {} */
+/* .heading__h6 {} */
--- a/lib/editor/heading-decorator.test.js
+++ b/lib/editor/heading-decorator.test.js
@@ -1,0 +1,39 @@
+import headingDecorator from './heading-decorator';
+
+describe('headingDecorator', () => {
+  let contentBlock, callback;
+  let replaceRangeWithText;
+  let decorator;
+
+  beforeEach(() => {
+    contentBlock = {
+      getText: () => '# note title',
+      getKey: () => 'mock-key',
+    };
+    callback = jest.fn();
+    replaceRangeWithText = jest.fn();
+    decorator = headingDecorator(replaceRangeWithText);
+  });
+
+  it('passes the start and end offsets of the task prefix to the callback', () => {
+    decorator.strategy(contentBlock, callback);
+    expect(callback).toHaveBeenCalledWith(0, 12, expect.any(Object));
+
+    contentBlock.getText = () => '# updated note title';
+    decorator.strategy(contentBlock, callback);
+    expect(callback).toHaveBeenCalledWith(0, 20, expect.any(Object));
+  });
+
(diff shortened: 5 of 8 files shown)


jivox92362:
Will the change in line height break anything?

jivox92362:
Could I get help with this?

There are a few issues that I can't figure out how to fix:
* Changing an **h1** to an **h2**, or an **h2** to an **h3**, etc. does not live downgrade the heading
* After entering the first character of the heading, the cursor jumps to the beginning of the line `## h` `BREAKS HERE` `2`

jivox92362:
How can this functionality (rich live editing) be placed behind a settings toggle?

belcherj:
Hi, sorry we haven't gotten back to you. We've been pretty busy this week but I will take a deeper look Monday or Tuesday, my one concern is that the decorator process is already really slow for long notes and that this could make it slower.

jivox92362:
Thanks for the update. If it's too slow you can close the PR. I can use an alternate note app instead until this is implemented since I want the readability that bold headings or colored headings provide.

dmsnell:
Thanks for this submission! I too will try and get it in testing.

At first glance I have a couple thoughts to share:
 - in the past when we had explored this we had decided to keep the [same font size](https://github.com/Automattic/simplenote-electron/pull/418#issuecomment-262293114) for headings as for the regular text. I don't know how much that is still relevant or if @davewhitley has any new thoughts to share on that.
 - as you have noticed, we'll need to make sure this works smoothly before merging, meaning that we will have to be able to downgrade the heading level
 - we want to be suspicious about putting this in a settings toggle. if this change is good enough to change the editor's behavior then it should also be good enough to be there universally. if it's not good enough to deploy everywhere then that should at least make us wonder if we're not ready yet to make this change.

lib/components/heading/test.js
```diff
@@ -0,0 +1,12 @@
+import Heading from './';
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+it('renders the heading with the respective heading element for h1-h6', () => {
+  [1, 2, 3, 4, 5, 6].forEach(function(v) {
+    const tree = renderer
+      .create(<Heading headingLevel={v} content={`${v} content`} />)
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
dmsnell: snapshot testing here is a bit heavy-handed for what we want to test. that is, although the test _says_ that it's confirming the appropriate heading element it's really only making sure that the behavior doesn't change - right or wrong. the risk I'd like to avoid is failing a test when it's not wrong.

additionally, by running `forEach` inside of the test it seems like it can obscure the unit under test. how about using the `.each` test helper?

```js
test.each([1, 2, 3, 4, 5, 6])('renders h%d', level => {
  const wrapped = shallow(<Heading headingLevel={level} content="test" />);
  expect(wrapped.type()).toBe(`h${level}`);
});
```

lib/note-detail/heading/constants.js
```diff
@@ -0,0 +1,3 @@
+export const headingPrefixRegex = /^([#]{1,6)\s+[^\s]/g;
+
+export const headingRegex = /^([#]{1,6})(\s+)([^\s]+)/g;
```
dmsnell: These are specific enough and simple enough that we can keep them in the files where they are used.

lib/editor/heading-decorator.js
```diff
@@ -0,0 +1,26 @@
...
+    let match;
+
+    while ((match = headingRegex.exec(text)) !== null) {
+      const fullText = text;
+      const headingLevel = match[1].length;
+
+      const start = match.index;
+      const end = start + fullText.length;
+
+      callback(start, end, { headingLevel: headingLevel, content: fullText });
```
dmsnell: Did you consider passing the full text as children to the React component? Something about that feels more natural than mimicking it with new prop types…

```js
callback( start, end, { headingLevel, children: fullText } );
```

```js
const Heading = ({ headingLevel, children }) => {
   const elementType = `h${headingLevel}`;
   const elementClass = `heading heading__${headingLevel}`;

   return React.createElement(elementType, { className: elementClass }, children);
 };
```

```js
<Heading headingLevel={level}>Test</Heading>
```

dmsnell:
After some testing and review of the code I think we need to consider this less as a distinct decorator and more in the context of our overall Markdown approach. If we're going to style headers then I don't know why we would only do that and leave other Markdown things unstyled.

jivox92362:
Thanks for the review. It seems like the better approach would be to render items as their respective elements (`#` -> `h1`, `##` -> `h2`, `***` -> `ul`) but I'll address the issues you pointed out to make it easier for others to review.

Please note that I don't know React and though I like coding my only goal of this was to make it easier to read my notes! (by making h2s or h3s bold or a bright color in order to visually separate sections.) If someone wants to take this over or do it another way I don't mind :)

lib/components/heading/test.js
jivox92362: That would be a lot better. Fixed in 0414430.

lib/note-detail/heading/constants.js
jivox92362: Addressed in 67db72e49a9c2c5719c035ebdbbe79668296d410.

lib/editor/heading-decorator.js
jivox92362: These changes have been implemented. Thanks!

jivox92362:
@dmsnell I've implemented the suggested improvements. The editor still seems to be quick.

What do you want to do with this PR?

dmsnell:
> Please note that I don't know React

hey @jivox92362 - you did a pretty bang-up job then for not knowing React - well done.

> my only goal of this was to make it easier to read my notes!

I'd encourage you to use your fork and keep this in. I don't know if or when we'd likely get this kind of syntax highlighting in. You already saw how long we've tossed around the idea 🙃 - part of the problem is the multi-platform bit and Markdown support is already the least consistent part of the app.

We can talk about getting you running your fork with your actual notes if you like that idea.

> What do you want to do with this PR?

Let's hear @belcherj's thoughts on this. I'm not sure I would feel comfortable merging this in as it relates to the partial vs. full Markdown support.

---

It's exciting to review code like this. Please don't hesitate to push out more PRs. Sometimes we merge them; sometimes we don't; but I've seen some great ideas over the years that people sometimes keep around in their forks.

jivox92362:
@dmsnell Thanks for the encouragement. I can appreciate the reluctance to add extra features onto the app without serious consideration and I'm happy that things are simple.

I'll close this PR but I'll keep contributing. I won't need to maintain a fork though; I've switched to platform-specific notetaking apps on Linux and Android with external bidirectional syncing since bold headings are an essential feature for me.

jivox92362:
Bold headings were added to the macOS version with markdown support in https://github.com/Automattic/simplenote-macos/pull/169.

jivox92362:
@belcherj What are your thoughts on this?

roxihe7880:
Any update on whether this functionality is under consideration?

roxihe7880:
See #1578

belcherj:
This functionality is not currently under consideration.

Issue #1728, opened 2019-11-17
List of 'Collaborators' not correct


mkeyful:
When I share a note with friends, I do not appear in the list of 'collaborators' for them. 

Concrete example: 
I click 'Share' to share a note with a friend. 
I click 'Share' again and I can see my friend in the list of 'collaborators'.  
But when my friend goes to my note and clicks 'Share', the list of 'collaborators is empty. 
Thus, he has no clue who shared what with him after a while.

belcherj:
The owner of a note is not added as a tag. We will need to rework some things to make this work.  Going to leave open as I think this is more of a bug than a feature request.

dmsnell:
@belcherj this may not be an easy change and I wouldn’t consider it a bug so much as a part of the design that was always a bit odd.

We might be able to auto-add the account email as a collaborator tag but even then I don’t think it will all work the way we expect - showing the owner just wasn’t in the design.

(This may have been unfortunately obscured by moving the collaborators out of the tag view and into its own panel).

sandymcfadden:
It doesn't look like this is just the owner not showing in the list of collaborators. 

If the owner creates a note, adds two different collaborators they both show for the owner in the list.
Neither of the other collaborators seem to have a way to show that the note is being collaborated on or by who.

PR #1735, opened 2019-11-21
Updates collaborator email field validation

### LINKED ISSUE #233: Collaborate: Form validation for the email field
Maybe we can use the form validation we use on the log in screen.

screen shot 2016-03-18 at 10 06 27 am

### LINKED ISSUE #234: Collaborate: don't allow users to enter in their account email
I'm able to add my own email address, which is strange.

Related: #233 (form validation)

belcherj:
### Fix

Fixes #234 Fixes #233 

The email field for collaborators allowed for incomplete emails. When entering
an incomplete email it would add that partial as a tag on the note. This fix
also disallows the user from adding themselves as a collaborator.

### Test
1. Go to collaborators dialog
2. Add `foo@` see that it is not allowed
3. Add your email, see that it is not allowed
4. Goto a not that is shared with you, add your email, see that it is not allowed

This PR does not cover the case where a not is shared with someone and that user adds the owner as a collaborator. That will require more work as part of: https://github.com/Automattic/simplenote-electron/issues/1728

### Review

### Release

- Disallow partial emails and users own email from being adding to collaborators email field [#1735](https://github.com/Automattic/simplenote-electron/pull/1735)


lib/dialogs/share/index.jsx
```diff
@@ -116,12 +119,13 @@ export class ShareDialog extends Component {
                   onSubmit={this.onAddCollaborator}
                 >
                   <input
-                    ref={e => (this.collaboratorElement = e)}
-                    type="email"
-                    pattern="[^@]+@[^@]+"
                     className="settings-item-text-input transparent-input"
+                    pattern="^([^\x00-\x20\x22\x28\x29\x2c\x2e\x3a-\x3c\x3e\x40\x5b-\x5d\x7f-\xff]+|\x22([^\x0d\x22\x5c\x80-\xff]|\x5c[\x00-\x7f])*\x22)(\x2e([^\x00-\x20\x22\x28\x29\x2c\x2e\x3a-\x3c\x3e\x40\x5b-\x5d\x7f-\xff]+|\x22([^\x0d\x22\x5c\x80-\xff]|\x5c[\x00-\x7f])*\x22))*\x40([^\x00-\x20\x22\x28\x29\x2c\x2e\x3a-\x3c\x3e\x40\x5b-\x5d\x7f-\xff]+|\x5b([^\x0d\x5b-\x5d\x80-\xff]|\x5c[\x00-\x7f])*\x5d)(\x2e([^\x00-\x20\x22\x28\x29\x2c\x2e\x3a-\x3c\x3e\x40\x5b-\x5d\x7f-\xff]+|\x5b([^\x0d\x5b-\x5d\x80-\xff]|\x5c[\x00-\x7f])*\x5d))*(\.\w{2,})+$"
```
codebykat: Wow this is quite the regex 😅 

Can we add a comment summarizing it in English?

I found this on a web search as well: https://emailregex.com/

We could probably stand to make sure we're standardizing this across clients.

RELEASE-NOTES.txt
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## [v1.13.0]
+
+### Enhancements
+
+### Fixes
+
+- Disallow partial emails and users own email from being adding to collaborators email field [#1735](https://github.com/Automattic/simplenote-electron/pull/1735)
```
codebykat: users => user's

codebykat (changes requested):
Couple minor notes:

* If I input my own address and hit "Add email", it silently fails. It should fail validation instead.
* The error messages are different if I enter a string followed by @, versus two words separated by @. I'm not sure where the "Please match the requested format" bit is coming from but it shouldn't do that.

Screen Shot 2019-11-25 at 12 56 38 PM
Screen Shot 2019-11-25 at 12 56 59 PM

RELEASE-NOTES.txt
belcherj: fixed

lib/dialogs/share/index.jsx
dmsnell: what's our purpose in adding this here? that's probably the heart of the issue. email validators are notoriously flakey and likely to cause more headache than help.

if we want to prevent accidental submission then we might want to consider the most lenient of matchers…

```html
<input pattern="^[^@]+@.+" />
```

Because the server will confirm if it's valid or not. Just leaving some thoughts…

codebykat: So yeah I really more meant "can we add a comment that breaks down exactly what this regex is doing and where it came from". Is it a copy of something from the backend code? Or the same as what's used on the other clients? If not I would prefer we either use one of those instead of introducing something subtly different. Ideally this should be the same validation across the board.

dmsnell: I've seen these mega email RegExps many times and they usually cause more bugs than they resolve. The backend doesn't do what you might expect it to, and most don't because usually it's much more productive to resolve an issue when someone has used the wrong or invalid email address than it is to prevent them from doing so.

belcherj: The backed does not do any validation for emails.  If it is not an email it adds it as a tag. We have to have an email regex o this field.

dmsnell: > We have to have an email regex o this field.

Sure. I'm only suggesting that we relax the validation to the most basic form to prevent dealing with the validator refusing a bunch of legitimate emails _or_ having to maintain the pattern that nobody can read.

At a minimum, I concur with @codebykat that this needs an explanatory comment.

belcherj: I have switched to a simpler regex.  That regex will ensure `blah@blah.blah`. This regex will not ensure valid emails. If an invalid email is added to the field it will not be added as a collaborator, it will be added as a tag. We can revisit in another PR, I waa get this one merged.

lib/dialogs/share/index.jsx
```diff
@@ -116,12 +119,13 @@ export class ShareDialog extends Component {
                   onSubmit={this.onAddCollaborator}
                 >
                   <input
-                    ref={e => (this.collaboratorElement = e)}
-                    type="email"
-                    pattern="[^@]+@[^@]+"
                     className="settings-item-text-input transparent-input"
+                    // Regex to detect valid email
+                    pattern="^[^\s@]+@[^\s@]+\.[^\s@]+$"
```
dmsnell: I feel really bad commenting here again, but this pattern does relax some rules but probably rejects valid email address that the previous monster pattern allowed since emails can have more than one `@` sign. I checked and I don't think we have any existing accounts with more than one so it's probably safe to do, but it stood out to me so I asked about it.

lib/dialogs/share/index.jsx
dmsnell: >  This regex will not ensure valid emails.

Part of where I'm coming from is that no Regex will. Even if we get what looks like a valid email address it may have no inbox on the other side.

I went looking on the server and found [another insufficient RegExp](https://github.com/Simperium/simplenote-gae/commit/670aa58604d24cf185643391f6a192681755d126) 🙃 and already we can see that it's going to block domains like `.coffee`

:sigh:

Anyway this is why I like relaxed email validation 🤷‍♂ 

> We can revisit in another PR, I want get this one merged.

Just noting that I think this hits that longstanding issue where you don't know if you added the right email address and if your collaborator actually was able to open the note.

dmsnell (approved):
I left my comment about the pattern being overly-restrictive but I'm happy with the code as you wrote it if you want to get this out.

If you want to change it I'm fine with that - if you want to use what the backend uses that also seems meritorious.

```python
"^.+\\@(\\[?)[a-zA-Z0-9\\-\\.]+\\.([a-zA-Z]{2,4}|[0-9]{1,3\})(\\]?)$"
```

codebykat:
Ditto what @dmsnell said.

belcherj:
`^.+\\@\(\\[?)[a-zA-Z0-9\\-\\.]+\\.([a-zA-Z]{2,4}|[0-9]{1,3\})(\\]?)$`

belcherj:
The regex from the backend is not a valid regex. Just adding `<input pattern="^[^@]+@.+" />` and merging.

dmsnell:
> The regex from the backend is not a valid regex.

Ah, that's a Python pattern and it's written inside a normal string instead of a "raw" string literal, which can mess up the slashes.

```python
pattern1 = "^.+\\@(\\[?)[a-zA-Z0-9\\-\\.]+\\.([a-zA-Z]{2,4}|[0-9]{1,3\\})(\\]?)$"
pattern2 = r"^.+\@(\[?)[a-zA-Z0-9\-\.]+\.([a-zA-Z]{2,4}|[0-9]{1,3\})(\]?)$";

pattern1 == pattern2
```

The raw version should work in JS.

```js
const pattern = /^.+\@(\[?)[a-zA-Z0-9\-\.]+\.([a-zA-Z]{2,4}|[0-9]{1,3\})(\]?)$/gm;
```

Just to restate though, I support your choice in the PR. It looks like we should consider updating the server because it's going to be rejecting valid emails too; lots of them.

PR #1736, opened 2019-11-21
Fix settings URL so that it send the correct url param to GAE

### LINKED ISSUE #240: Confusing Edit Account flow
Mac OS X 
1.0.1
- Go to Preferences > Edit Account
- It goes to https://app.simplenote.com/settings
- I can see Display settings that only affect the web app

belcherj:
### Fix

Fixes: #240

We were directing users to /settings?source=blah which GAE automatically
redirected to /settings/ removing the url param. Additionally, the code always
returned 'electron' as the source which is not only incorrect but GAE only
understands the source as react. From the web app and from electron we should
send the source as react so that we hide settings that only apply to the old
web app.

### Test
1. Goto settings and click edit account
2. Are display settings visible?
3. DO this in both the electron and web app

### Review

### Release

- Only display view settings when coming from the old web app [#1736 ](https://github.com/Automattic/simplenote-electron/pull/1736)


codebykat (approved):
👍

PR #1741, opened 2019-11-22
Fixes bug where window is dragged when trying to select a note revision

### LINKED ISSUE #898: Click-dragging the pointer on the history slider moves the window
**Steps to reproduce**: Follow the steps below to reproduce the issue. 

- Get the Simplenote Electron app. 
- Create a note. 
- Create another note, for no reason.
- Come back to the first note and modify the note contents.
- Click on the history button at the top and see the slider appear over there.
- Click-drag the pointer on the slider.

#### What I expected

The pointer on the slider moves.

#### What happened instead

The Simplenote Electron app window moves.

**Note:** Clicking at select stages works fine though. 

#### OS version

macOS 10.14 - Mac OS Mojave

#### Screenshot / Video

[Here is a recording of this behavior](https://cloudup.com/cVc_-CFqVSb).

[This is similar to the report here](https://github.com/Automattic/simplenote-electron/issues/454), but that one was for the buttons.

belcherj:
### Fix
Fixes: https://github.com/Automattic/simplenote-electron/issues/898

There was a div that was draggable underneath the revision selector.
Additionally, the revision selector was draggable. In remedying this we are no
longer rendering the note toolbar header while the revision panel is open. We
are also not rendering the revision panel when it is not open. The showing and
hiding of these panels is overly complex and should be looked at in the future
PR.

### Test
1. Build electron `npm run dev`
2. Open revision selector
3. Can you slide revision?

### Review

### Release

- Revision selector is now usable without dragging the entire window [#1741](https://github.com/Automattic/simplenote-electron/pull/1741)


lib/revision-selector/index.jsx
```diff
@@ -125,6 +125,10 @@ export class RevisionSelector extends Component {
       'is-visible': isViewingRevisions,
     });
 
+    if (!isViewingRevisions) {
```
codebykat: I suppose we can't move this higher in the function because linting? It would be nice to short-circuit some of the above logic sooner.

codebykat (approved):
Verified the original issue and the fix 👍 

One tiny code style question but feel free to ignore it.

lib/revision-selector/index.jsx
belcherj: Done, not a linting issue.

PR #1743, opened 2019-11-25
Adds a Clipboard Button component

### LINKED ISSUE #232: Publish: Change "copy" to "copied"  after click
It would be nice if "Copy" changed to "Copied" after clicking it. After a second or two, it should revert back to "Copy".

copy

belcherj:
### Fix

Closes: https://github.com/Automattic/simplenote-electron/issues/232

The inline button component to copy the publish url was lacking an inidcation
that the text was copied. Additionally, the current setup was using a copy
method that can have issue in some browsers. This commit switches to use of the
clipboard npm repo, moves code into a component, and adds an indication that
the text was copied.

### Test
1. Build project
2. Have note that is published
3. Click copy button
4. Observe that text is copied
5. Observe that text of button changes to: "Copied!" for 4000ms

### Review

### Release

- Added indication that publish url has been copied [#1743](https://github.com/Automattic/simplenote-electron/pull/1743)

### DIFF
--- a/lib/components/clipboard-button/index.jsx
+++ b/lib/components/clipboard-button/index.jsx
@@ -1,22 +1,21 @@
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import ReactDom from 'react-dom';
 import Clipboard from 'clipboard';
 
-function ClipboardButton({ text, disabled }) {
-  const buttonRef = React.useRef();
-
-  const textCallback = React.useRef();
-  const successCallback = React.useRef();
+function ClipboardButton({ text }) {
+  const buttonRef = useRef();
+  const textCallback = useRef();
+  const successCallback = useRef();
 
   const onCopy = () => {
     setCopied(true);
   };
 
-  const [isCopied, setCopied] = React.useState(false);
+  const [isCopied, setCopied] = useState(false);
 
   // toggle the `isCopied` flag back to `false` after 4 seconds
-  React.useEffect(() => {
+  useEffect(() => {
     if (isCopied) {
       const confirmationTimeout = setTimeout(() => setCopied(false), 4000);
       return () => clearTimeout(confirmationTimeout);
@@ -24,15 +23,14 @@ function ClipboardButton({ text, disabled }) {
   }, [isCopied]);
 
   // update the callbacks on rerenders that change `text` or `onCopy`
-  React.useEffect(() => {
+  useEffect(() => {
     textCallback.current = () => text;
     successCallback.current = onCopy;
   }, [text, onCopy]);
 
   // create the `Clipboard` object on mount and destroy on unmount
-  React.useEffect(() => {
-    const buttonEl = ReactDom.findDOMNode(buttonRef.current);
-    const clipboard = new Clipboard(buttonEl, {
+  useEffect(() => {
+    const clipboard = new Clipboard(buttonRef.current, {
       text: () => textCallback.current(),
     });
     clipboard.on('success', () => successCallback.current());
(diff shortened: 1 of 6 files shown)


codebykat (changes requested):
This is working well 👍 

It's possible to click "Copy" before the link has been generated, we should disable that.

I did notice there's some kind of race condition if you quickly toggle the Published toggle off and on a few times... sometimes if it's off and you turn it on, it will flash the "on" state and then quickly toggle back to "off". I have reproduced that on develop as well so it wasn't introduced by this PR.

belcherj:
The button is disabled unless there is a publish url: https://github.com/Automattic/simplenote-electron/pull/1743/files#diff-5e84e31e46424a0777d56ad7fc5009b3R193

We should handle the race condition in a separate PR.

dmsnell:
> I did notice there's some kind of race condition if you quickly toggle the Published toggle off and on a few times... sometimes if it's off and you turn it on, it will flash the "on" state and then quickly toggle back to "off". I have reproduced that on develop as well so it wasn't introduced by this PR.

What's happening is that when we hit publish all we do is set a property on the note; we add a system tag saying "this note should be published" and Simperium sync's the data. When the server sees the changes going through it adds the publish links.

In between our clicks we are probably merging the local changes with the confirmation from the server of our previous clicks. If you can get the timing just right (may need to slow down your network connection) then you can probably tap a pattern into the toggle, wait a moment, and watch it mysteriously repeat the pattern back to you as the Simperium change confirmations return.

codebykat:
@belcherj I think my original comment was a bit confusing. Can we fix the button to not SEEM clickable until the note is published? Right now it still lets you click it and changes color:

Untitled

lib/components/clipboard-button/index.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactDom from 'react-dom';
+import Clipboard from 'clipboard';
+
+function ClipboardButton({ text, disabled }) {
```
dmsnell: I noticed that we have this prop but we never set it. Should we just leave it out for now?

lib/components/clipboard-button/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+  const buttonRef = React.useRef();
+
```
dmsnell: nitpicky here - is there a reason for the extra space?

codebykat:
This works well for me now 👍

lib/components/clipboard-button/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+  // update the callbacks on rerenders that change `text` or `onCopy`
+  React.useEffect(() => {
+    textCallback.current = () => text;
+    successCallback.current = onCopy;
+  }, [text, onCopy]);
+
+  // create the `Clipboard` object on mount and destroy on unmount
+  React.useEffect(() => {
+    const buttonEl = ReactDom.findDOMNode(buttonRef.current);
```
dmsnell: Maybe I'm misunderstanding but I don't think we should ever need `findDOMNode()` with refs - I noted we are using `useRef()` instead of `createRef()` and learned that `useRef()` exists and should be used in these situations so that it doesn't create a new `ref` on every render.

```js
// bail if we've unmounted or something by the time this gets called
if ( ! buttonRef.current ) {
  return;
}

const clipboard = new Clipboard( buttonRef.current, {
  …
} );

…
```

dmsnell:
After writing this with React hooks how are you feeling about them @belcherj? It took me more time than usual to understand what we were trying to accomplish with them and I'm curious how you felt about thinking in terms of hooks vs. writing a traditional React class for this component.

belcherj:
Hooks are way cleaner. It may take you longer to grok now but once we are using hooks on the regular it will seem more comfortable.

Issue #626, opened 2017-09-07
App icon is not visible on Xfce Desktop on Linux


ajyotirmay:
There is no application icon visible, either in the panel, the title bar, or the application switcher (alt+tab)

#### OS version
Xubuntu 16.04.3

#### Screenshot
![image](https://user-images.githubusercontent.com/8706789/30177436-627afbb4-93f5-11e7-8d41-6f18e87a7af9.png)

![image](https://user-images.githubusercontent.com/8706789/30177485-84a7ac6e-93f5-11e7-91cc-d42605db5e9a.png)

sujaataarya:
It's great if you can turn the panel into the system tray!

kungpaogao:
How did you fix this? I am also running Xubuntu, and this happens for multiple Electron-based programs, including Simplenote.

nichcz:
+1

ajyotirmay:
@kungpaogao my bad, I was using KDE and it was working just fine there, so I thought this issue was fixed. When I installed the latest build on my Xubuntu Xenial, this issue was still existing.

I'm reopening this issue, as it seems like this one still exists on Xubuntu/Xfce-Desktop.

ajyotirmay:
The following code snippet is from a Facebook Messenger application that I was creating for my desktop. I'm able to display a proper icon for my application, even on Xfce desktop.

`window = new BrowserWindow({
        width: 1000,
        height: 600,
        icon: '/home/apurv/.applications/fb-messenger/app/images/96x96.png',
        webPreferences: {
          nodeIntegration: false
        }
    });
`

From what I think, the following code will be able to display an icon for Simplenote:
`icon: '/home/apurv/.applications/fb-messenger/app/images/96x96.png'
`

But this needs to be adapted according to this project's needs. If any one can help with this, it will be great. :)

facosta0787:
Is there any solution for this issue ?

sandymcfadden:
This is an old issue. After testing in the most recent LTS version of Xubuntu 20.04 the icon shows in the appropriate places including the panel at the top and application switcher. Closing.

PR #1745, opened 2019-11-25
[WIP] Fixes system icon so that it displays properly on Linux

belcherj:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/626

The system icon does not appear properly on Linux. This commit uses the Electron
nativeImage function to properly pass the image.

### Test
***(Required)*** List the steps to test the behavior.  For example:
> 1. Go to...
> 2. Tap on...
> 3. See error...

### Review
***(Required)*** Add instructions for reviewers.  For example:
> Only one developer and one designer are required to review these changes, but anyone can perform the review.

### Release
***(Required)*** Add a concise statement to `RELEASE-NOTES.txt` if the changes should be included in release notes. Include details about updating the notes in this section. For example:
> `RELEASE-NOTES.txt` was updated in d3adb3ef with:
> 
> > Added markdown support

If the changes should not be included in release notes, add a statement to this section. For example:

> These changes do not require release notes.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -6,6 +6,7 @@ const {
   ipcMain,
   shell,
   Menu,
+  nativeImage,
   session,
 } = require('electron');
 
@@ -58,7 +59,7 @@ module.exports = function main() {
       height: mainWindowState.height,
       minWidth: 370,
       minHeight: 520,
-      icon: iconPath,
+      icon: nativeImage.createFromPath(iconPath),
       titleBarStyle: 'hidden',
       show: false,
       webPreferences: {


desktop/app.js
```diff
@@ -58,7 +59,7 @@ module.exports = function main() {
       height: mainWindowState.height,
       minWidth: 370,
       minHeight: 520,
-      icon: iconPath,
+      icon: nativeImage.createFromPath(iconPath),
```
dmsnell: According to the docs it looks like this should result in the same image as just passing the string path, as before.

> In Electron, for the APIs that take images, you can pass either file paths or NativeImage instances. An empty image will be used when null is passed. -[Electron API Docs](https://electronjs.org/docs/api/native-image)

Did this change make the icon work?

belcherj:
This does not fix the problem. Closing PR will have to revisit later.

PR #1754, opened 2019-12-03
Undo unintentional changes introduced into master

dmsnell:
## Summary

In this patch we're manually undoing unintentional changes to `master`.
Our intention is to merge this into `master` directly and to do so before 1.12

---

At different times in the past we inadvertently added these changes into
`master` while resolving merge conflicts. Unfortunately they have caused
`master` to live a different history than we thought it would and their
continued existence makes every release difficult.

In this patch we're manually "undoing" these changes so that they don't
interfere with the release merges. This will change `master` but none of
these changes _in_ `master` should be there anyway as they don't have a
corresponding commit in `develop`. That is, we have probably been
releasing builds where we were unaware that these changes were there.
Anyhoo they aren't that critical and so nothing should have been broken
because of it.

 - `README.md` was updated inline in [ff503fe0] on Oct. 22 in order to
   manually trigger a re-deployment of the app
 - `actions/`, `lib/search-field/index.jsx`, and `webpack.config.js`
   were brought in during the merge of `1.7.0` in #1508 on August 12
   This lived on and came back when the `1.8.0` release branch was merged
   into the `1.9.0` release branch, presumably due to a stale copy of
   `master` on a local machine, though the true reason is lost to history.
 - `lib/note-list/style.scss` during the confusing conflict-resolution
   that was the [1.9.0 release]

After this patch we should be back into a place where we can merge the
release branch into `master` without any headache. After this patch we
should have nothing come over if we were to merge `master` into
`release/1.12`. The same patch should be legitimate onto `master` as
well as onto `develop`

[ff503fe0]: https://github.com/Automattic/simplenote-electron/commit/ff503fe093b42f4343f134b6f7c5285111517b92
[1.9.0 release]: https://github.com/Automattic/simplenote-electron/commit/7235c093f0ac4a91a01024194eed8581ab5b31aa#diff-7f4631b1fc04a0dcacda9baf3992abefR128


bummytime:
Thank you for tackling this @dmsnell 👍

dmsnell:
@belcherj what will happen when we merge this into `master`? Will our web app redeploy?

None fo the changes should break anything, but I just want to know before we do it.

belcherj:
Nothing will happen.  We have to manually deploy the web app now.

PR #1761, opened 2019-12-05
Added GenericName field for app on Linux

SaeedBaig:
Added "GenericName" field to the desktop option for Linux (so that its icon displays a useful subtitle when hovered over in KDE panel).
Since the convention is a short descriptive name of what the app is, I used "Note Taking Application".

### Fix
A short description of the app ("Note Taking Application") is displayed when the mouse hovers over Simplenote's icon in the KDE panel and in other search-results on KDE.

**Pic1**: Screenshot of old icon in KDE panel:

before_panel

**Pic2**: Screenshot after change:

after_panel

**Pic3**: Example of what other apps on Linux look like:

Screenshot_20191206_015515

**Pic4**: Screenshot of old icon in KDE Application Launcher:

before_launcher

**Pic5**: Screenshot after change:

after_launcher

This brings its behaviour/appearance in line with other apps on Linux (e.g. Google Chrome uses "Web Browser", GIMP uses "Image Editor", Atom uses "Text Editor", etc).

### Test
Mostly a UX improvement, so not much in the way of tests. Just:
1. Build Simplenote on a Linux distro that uses KDE (e.g. Kubuntu). A reboot of the system may be necessary to force it to register the changes (only required once).
2. Search for "Simplenote" from the Application Launcher, right-click it, and select "Pin to Task Manager".
3. Hover the mouse over the Simplenote icon in the panel.
4. A tooltip displaying the title "Simplenote" and a subtitle "Note Taking Application" should appear over it, as in Pic2.

### Review
Only one designer is required to review these changes, but anyone can perform the review.
For the reviewer, ensure that this change does not cause unwanted/ugly changes to the appearance/UX of the app elsewhere throughout the system (e.g. when appearing in search results on the desktop).

### Release
This change is fairly minor so probably does not warrant release notes.

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -93,6 +93,7 @@
     "desktop": {
       "Name": "Simplenote",
       "Comment": "Simplenote for Linux",
+      "GenericName": "Note Taking Application",
       "Type": "Application",
       "Icon": "simplenote",
       "StartupNotify": "true",


belcherj:
Thanks for the PR!

PR #1763, opened 2019-12-06
Fix focus-mode transition to small screen

### LINKED ISSUE #1760: Responsive: Resizing with sidebar hidden should start in note focus
A user in #2538788-zen was confused by this, when a note is displayed full-screen (with the sidebar hidden) and you then resize the window to a narrow width, it changes the view to the note list.

In this situation it would be better to change directly to the note view (i.e., the note list is still hidden).

qualitymanifest:
### Fix
Closes https://github.com/Automattic/simplenote-electron/issues/1760

When resizing to small screen, if focus mode is enabled, keep the note open

### Review
See steps above

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 ## [v1.13.0]
 
 ### Enhancements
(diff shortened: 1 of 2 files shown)


codebykat:
This is great, thanks so much for the fix @qualitymanifest !

PR #1767, opened 2019-12-09
Remove code that stops settings from scrolling on shorter screens

belcherj:
### Fix
At some point, some code was added that allowed the tabs to scroll. In adding
this it caused the settings tab to display at full height and not be contained
by the scroll box in the modal. This fixes
https://href.li/?https://github.com/Automattic/simplenote-electron/issues/1268

This is not a long term solution as it is still not a great user experience.  We need to drop the modals for full-page settings.

### Test
1. Open app.simplenote.com with dev tools open
2. Open settings to the display tab
3. Enable mobile responsive mode
4. Set the height to less than 640px
5. See that it doesn't scroll
6. Do the same on this branch
7. Observe that it does scroll

Old, doesn't scroll:

old

After, scrolls:

Dec-09-2019 14-28-09

### Review

### Release

- Makes settings scrollable on shorter smaller view ports [#1767](https://github.com/Automattic/simplenote-electron/pull/1767)

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,7 @@
 - Keep note open when transitioning to small screen in focus mode [#1763](https://github.com/Automattic/simplenote-electron/pull/1763)
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


codebykat (approved):
Looks reasonable, fixes the issue 👍

PR #1773, opened 2019-12-10
Explore: Add e2e-testing with Spectron

dmsnell:
In this patch we're creating the most basic form of end-to-end testing
using the Spectron framework. This boots the app and verifies that it
opens.

The files aren't in the right spot and the structure of the tests need
to change but this gives us an idea of how to run these tests.

## Testing
Checkout this branch and run the tests setting the username/password
for some test account.

```bash
TEST_USERNAME="simplenote@domain.invalid" TEST_PASSWORD="blarg" npm test
```

You should see the app appear and the credentials entered automatically
into the login form.

## Notes

 - We can't run the tests with the DevTools open or they will fail
 - This was pretty simple to get setup
 - We need to see if we can get these running in CI for us so that they can approve or reject a PR based on the results of the test
 - According to the docs we have to run a specific version of Spectron corresponding to the version of Electron in use

## Links

 - [Spectron](https://github.com/electron-userland/spectron#application-api)
 - [webdriver-io API](https://webdriver.io/docs/api.html)


codebykat (approved):
This is awesome! Seems like a great start. Not sure what kind of review you want at this point but I'll approve to say the code looks good and it works.

dmsnell:
Thanks @codebykat - were you able to get the tests to run on your computer?

I think I'd like to let this sit for another day or two and give people the chance to comment, but after that I'd like to get it merged even with all its shortcomings so that we can start.

I'm thinking that for now we don't need to have these automated as it would suffice for now if we had an end-to-end test suite to run it locally as a verification step before making a release. That could save us considerable headache.

belcherj:
This is causing CI to fail. Once that is resolved ping me for a review.

codebykat:
The tests did run for me, yes.

dmsnell:
When I get back around to this I'd like to move the e2e testing out of CI for now and let it be a manual process. We can iterate on it later and taking it out of the automated tests is surely going to make it easier in the beginning.

belcherj:
Sounds good!

PR #1774, opened 2019-12-11
Sync note after restoration to update the state

### LINKED ISSUE #1772: History: Restoring a note breaks the note list
### Expected

I expect that I can restore a past revision of a note and continue using the app.

### Observed

After restoring a past revision of a note, I can't open any other notes in the note list. I can click on them and they are highlighted in the list, but they don't open in the editor.

The issue persists until I relaunch the app.

### Reproduced

1. Open the app.
2. Select a note that has a history (past revisions).
3. Open the revision selector.
4. Select and restore a past revision.
5. Select a different note in the notes list and notice that it doesn't open in the editor.

-|-|-|-
Windows|VirtualBox VM|Win10|1.13.0-beta1
Mac|Macbook Pro (2017)|macOS 10.15.1|1.13.0-beta1

I can also reproduce the issue on macOS 10.15.1 in Chrome 78.0.3904.108 at https://app.simplenote.com/ (running version 1.12.0 according to the About section).

cc @belcherj

belcherj:
### Fix
Fixes #1772

After restoring a note the editor was stuck o the selected revision. This PR
clears that revision from the flux state. Additionally, that restoration of a
note was not causing a sync to occur.  This triggers a note sync that updates
everything and cements the restored version as the current version.

### Test
1. Open a note with history
2. Open revision panel
3. Select a version
4. Restore a version
5. Refresh
6. Is note set to revision?
7. Redo steps above, without refresh
8. Select another note
9. Select note that you accepted revision on
10. Is the note still on the revision?

### Review

### Release


lib/app.jsx
```diff
@@ -353,8 +356,8 @@ export const App = connect(
       // a note. The NoteEditor will notify via props when
       // it's time to sync via Simperium
       const { noteBucket } = this.props;
-
       noteBucket.update(note.id, updatedNote.data, {}, { sync: false });
+      this.syncNote(note.id);
```
dmsnell: I think this new line defeats the purpose of the lines above it.

we want to call `syncNote()` when we select a revision, but not on every change otherwise we'd have a mass of granular edits in the revision.

dmsnell (changes requested):

lib/app.jsx
belcherj: I have a local change that addresses this, still working o the PR. Not quite ready for review.

lib/app.jsx
```diff
@@ -348,13 +351,11 @@ export const App = connect(
...
       };
 
-      // update the bucket but don't force sync right away
-      // as this happens per keystroke when the user is editing
-      // a note. The NoteEditor will notify via props when
-      // it's time to sync via Simperium
       const { noteBucket } = this.props;
-
-      noteBucket.update(note.id, updatedNote.data, {}, { sync: false });
+      noteBucket.update(note.id, updatedNote.data, {}, { sync });
```
belcherj: I would think passing true here to sync would cause a sync but it doesn't.

dmsnell: What indicators are telling you that it isn't?

If it's not then we're probably seeing some other problem.

Also, since we can't get rid of `syncNote()` anyway where do you see the tradeoffs between adding the `sync` param here and calling `syncNote()` in the `revision-selector`?

belcherj: > What indicators are telling you that it isn't?

The note list isn't updating, if you refresh before adding more text to the note it resets back to before the revision was restored.  

> Also, since we can't get rid of syncNote() anyway where do you see the tradeoffs between adding the sync param here and calling syncNote() in the revision-selector?

I'm not really sure what you are asking.

dmsnell: > The note list isn't updating, if you refresh before adding more text to the note it resets back to before the revision was restored.

This could go back to our funny data flow. Simperium operations really ought to be watching app operations and updating independently as side-effects, likewise should probably call some app action to update a note. It's not clear throughout who owns the data - Simperium, or our local copy in app state.

> I'm not really sure what you are asking.

In this PR we're adding a new boolean parameter to `onUpdateContent` but another function already exists to do what we want: `this.syncNote()`

My question was why we choose to do that instead of call `syncNote()` when restoring a revision. I can make a guess that it's because we'd have to pass `syncNote()` down the component tree which would involve more work but I didn't know if you had another conclusion I didn't see.

belcherj: > This could go back to our funny data flow. Simperium operations really ought to be watching app operations and updating independently as side-effects, likewise should probably call some app action to update a note. It's not clear throughout who owns the data - Simperium, or our local copy in app state.

I don't think we can address this in this PR. I'd really like to get this in so the app isn't broken.

> My question was why we choose to do that instead of call syncNote() when restoring a revision. I can make a guess that it's because we'd have to pass syncNote() down the component tree which would involve more work but I didn't know if you had another conclusion I didn't see.

1. I didn't want to have to pass it down
2. The whole pass it down thing is a pretty terrible practice in general
3. The revision selector component shouldn't really have to worry about what to do when a revision is restored. I didn't want to add additional complexity down stream.

dmsnell: > 1. I didn't want to have to pass it down
> 2. The whole pass it down thing is a pretty terrible practice in general

I agree, though I've been staying in line with our practice of passing things down for the sake of consistency until we get a chance to do real refactoring in the app. I personally want to see these pass-downs all replaced with predictable and semantically-oriented Redux action calls so that we can get rid of the real culprit - complicated interplay with side-effects.

> 3. The revision selector component shouldn't really have to worry about what to do when a revision is restored. I didn't want to add additional complexity down stream.

Given the way the app works I don't agree as much on this one. The tags, for example, immediately sync once we add them or remove them. The editor is the odd-ball maybe and perhaps it should hold the exception, but there's good reason in my mind to make an explicit "sync now" when replacing the editor text with a revision vs "sync soon" when editing text directly.

If it were me writing this PR I think that I'd pass down `syncNote()` and that's about a 65%-75% confidence. It's not me though and I have no strong objection here, meaning that I support your choices. Mainly I wanted to get on the same page and talk about it, because it stood out to me as a bit unexpected here (plus it's introducing a new boolean function argument which I like to avoid).

Don't think too much about it.

belcherj:
Merging.  We can revisit if need be.

lib/app.jsx
dmsnell: In `<NoteDetail />` we see where we control `syncNote()` separate to enforce an update after checking a todo item. It's not inconsistent with this PR so it'd be good if we _either_ updated the PR code to match the existing pattern in the codebase _or_ updated the `toggleTask()` so match the new patterns introduced here.

PR #1780, opened 2019-12-16
Updates the prettier command and applies prettier fixes

belcherj:
### Fix
The prettier command was not properly checking all files under code revision.
It was only checking files that were one layer deep. This updates the command
so that it runs on all files and applies formatting to those files

### Test
Does app run?

### Review

### Release

- Applied prettier formatting to all files [#1780](https://github.com/Automattic/simplenote-electron/pull/1780)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,14 +1,16 @@
 {
   "version": "0.2.0",
-  "configurations": [{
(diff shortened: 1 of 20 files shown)


dmsnell (approved):
Good update!

PR #1782, opened 2019-12-16
Lint Fixes

belcherj:
### Fix
Clean code offending the linter

### Test
1. `npm test`
2. Click around the app to ensure nothing breaks

These changes could have an impact on many parts so now is a good time to make them.

### Review

### Release

### DIFF
--- a/desktop/updater/lib/Updater.js
+++ b/desktop/updater/lib/Updater.js
@@ -21,7 +21,7 @@ class Updater extends EventEmitter {
   ping() {}
 
   onDownloaded(event) {
-    console.log('Update downloaded');
+    console.log('Update downloaded'); // eslint-disable-line no-console
 
     // electron-updater provides us with new version
     if (event.version) {
@@ -32,11 +32,11 @@ class Updater extends EventEmitter {
   }
 
   onNotAvailable() {
-    console.log('Update is not available');
+    console.log('Update is not available'); // eslint-disable-line no-console
   }
 
   onError(event) {
-    console.log('Update error', event);
+    console.log('Update error', event); // eslint-disable-line no-console
   }
 
   onConfirm() {}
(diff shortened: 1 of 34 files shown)


desktop/updater/lib/Updater.js
```diff
@@ -94,7 +94,7 @@ class Updater extends EventEmitter {
     let text = originalText;
 
     for (const key in macros) {
-      if (macros.hasOwnProperty(key)) {
+      if (Object.prototype.hasOwnProperty.call(macros, key)) {
```
dmsnell: what did the linter not like about this?

did we even write this code or have we pulled it in from a dependency?

lib/components/boot-warning/index.jsx
```diff
@@ -1,9 +1,14 @@
...
 import './style';
 
 const BootWarning = ({ children }) => (
   <h3 className="boot-warning__message">{children}</h3>
 );
 
+BootWarning.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
```
dmsnell: Do you feel the same way about `PropTypes` that the linter does?

In my experience with Calypso they had a tendency to make things just as bad as they helped in other places: they get out of sync (and the linter don't care) and require the extra maintenance.

Moving to Typescript helped out with prop types because we got some support from the compiler, but I think we made some efforts over the years to remove `propTypes`

desktop/updater/lib/Updater.js
belcherj: https://eslint.org/docs/rules/no-prototype-builtins

lib/components/boot-warning/index.jsx
belcherj: I'm open to changing the rule. I am also open to Typescript.

belcherj: And by open, I think we should add the ability to use Typescript. I know it isn't quite as powerful when bolted on but the code base isn't that large.

dmsnell: it's workable easily at our size. I've explored replacing Babel with TypeScript too, which could be a nice move.

desktop/updater/lib/Updater.js
dmsnell: I can't recall ever having had bugs from this violation and it looks like it makes the code significantly more noisy to provide that safety. Also, in this case that it fixes we're absolutely _not_ violating any safety concerns as is possible to verify by inspection.

Do you have strong feelings about this rule? I'd be inclined to turn it off.

PR #1783, opened 2019-12-16
Deletes the util for ensuring a single tab

belcherj:
### Fix
The replication bug needed this util to prevent it from happening. Thankfully
we were able to resolve it via another fix. This tab restriction is no longer
needed.

### Test
1. Make sure app builds

### Review

### Release

- Delete unused tab restriction util [#1783](https://github.com/Automattic/simplenote-electron/pull/1783)


codebykat (approved):
:shipit:

dmsnell (commented):
Thanks for pulling this out!

PR #1784, opened 2019-12-16
Transition TransitionDelayEnter to React hooks

belcherj:
### Fix
This is a good small example of how React hooks are used.  Nothing changes
except we are now using a functional component and using hooks over lifecycle
methods.

### Test
1. Clear site data, reload
2. Sign in
3. Observe Simplenote log appears then disappears

Note: The value of this component is really really low. I'm thinking about deleting it.

### Review

### Release

- Transition TransitionDelayEnter to React hooks [#1784](https://github.com/Automattic/simplenote-electron/pull/1784)

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -15,7 +15,7 @@
 
 - Updated dependencies [#1759](https://github.com/Automattic/simplenote-electron/pull/1759)
 - Applied prettier formatting to all files [#1780](https://github.com/Automattic/simplenote-electron/pull/1780)
-- Transition TransitionDelayEnter to React hooks [#1784](https://github.com/Automattic/simplenote-electron/pull/1784)
+- Migrate TransitionDelayEnter to React hooks [#1784](https://github.com/Automattic/simplenote-electron/pull/1784)
 
 ## [v1.13.0]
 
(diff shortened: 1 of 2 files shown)


RELEASE-NOTES.txt
```diff
@@ -15,6 +15,7 @@
 
 - Updated dependencies [#1759](https://github.com/Automattic/simplenote-electron/pull/1759)
 - Applied prettier formatting to all files [#1780](https://github.com/Automattic/simplenote-electron/pull/1780)
+- Transition TransitionDelayEnter to React hooks [#1784](https://github.com/Automattic/simplenote-electron/pull/1784)
```
codebykat: I would recommend a verb other than "Transition" as "Transition TransitionDelayEnter" is needlessly confusing. Maybe "Migrate"? But, does this need a release note at all?

codebykat: I would recommend a verb other than "Transition" as "Transition TransitionDelayEnter" is needlessly confusing. Maybe "Migrate"? But, does this need a release note at all?

codebykat (approved):
Thank you, this is helpful for me! I was wondering what fixing those warnings would look like.

It tests out fine. Just one little thought about the release note.

PR #1786, opened 2019-12-17
Deletes context-menu.js as it is unused

belcherj:
### Fix
This file is not included anywhere and does not appear to be used.

### Test
1. Build the electron app.
2. Does everything work properly?

### Review

### Release

- Deleted unused context-menu file [#1786](https://github.com/Automattic/simplenote-electron/pull/1786)


dmsnell:
The `ContextMenu` was removed in #821 when we pulled in spell-checking as a module.

PR #1789, opened 2019-12-18
Add eslint plugin for React Hooks

belcherj:
### Fix
Add eslint plugin for React Hooks

### Test
1. `npm install` 
2. ` npm runn lint`

### Review

### Release

- Added React Hooks Eslinnt Plugin [#1789](https://github.com/Automattic/simplenote-electron/pull/1789)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,7 @@
     "prettier",
     "prettier/react"
   ],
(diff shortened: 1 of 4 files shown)


dmsnell (approved):
I welcome the lint rules which can catch subtle and complicated nuances of React Hooks, though I hope that we can be very cautious and guarded about adopting hooks generally. Let's chat in Slack about it.

PR #1793, opened 2019-12-23
Refactor state: Move editorMode into a separate reducer

dmsnell:
**Status**: This is a collector branch where I have been exploring. After several derivative patches have been applied I'll come back and trim this down and request a real review on it.

Working on making the smallest change to move functionality out of the complicated `app-state` and into separate reducers.

This will drive towards two ends:
 - clarify the update sequence to that updates are atomic and predictable
 - focus Redux actions on app semantics

Maybe more. Either way it should help us predict and eliminate a number of bugs associated with immutable update sequences that get interrupted and are non atomic.

Even in `editorMode` we have a strange situation where the call to `dispatch(newNote(…))` then calls a nested `dispatch('setEditorMode')` and a few other `dispatch` calls that may end up out of sequence based on the network activity in Simperium.

### DIFF
--- a/lib/flux/app-state.ts
+++ b/lib/flux/app-state.ts
@@ -7,6 +7,7 @@ import analytics from '../analytics';
 
 import { AppState, State } from '../state';
 import * as T from '../types';
+import actions from '../state/actions';
 
 const debug = Debug('appState');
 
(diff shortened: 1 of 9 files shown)


dmsnell:
Closing because it could be better recreated from scratch and some of the ideas explored have been otherwise merged already.

PR #1795, opened 2019-12-23
Refactor: Remove "pinned" exceptions

dmsnell:
To understand this change we have to undestand some core data models in
the app and some of the app's history.

Most data structures are composed of two parts: the Simperium `Entity`
which includes Simperium information of id, version, and wrapped data;
and the app-specific `data` which comes from Simplenote. That is, a
`note` is composed of the `Note` type which includes things like
`content` and `tags` and the `NoteEntity` which is the wrapped version.

We store all of the Simperium entities in `indexedDB` and in the
inception of this app we told `indexedDB` to create a search index based
off of whether or not a note was pinned. This was to make it fast to
list the notes with pinned notes at the top.

In order to facilitate the `pinned` index we had to create a fake
top-level property called `pinned` becuase we can't tell the browser:
"build an index on whether the `"pinned"` value is in the
`data.systemTags` array."

However, in [229a76b] we removed this index and moved pinned-sorting
into the application itself. When we did that we left behind the relic
of this top-level `pinned` value.

In this patch we're removing that relic. Why? Because it creates another
level of complexity by special-casing the `pinned` value which we have
to know about and reference when looking at the pinned status of a note.
By undoing this exception we can rely on the more semantically-native
means to check if or set whether a note is pinned.

This patch was triggered by what appeared to be an error when
introducing TypeScript types. We were looking at `note.pinned` which
shouldn't exist on the Simperium `Entity<Note>`. It _wasn't_ an error in
the app because the value existed but it wasn't clear that it _should_
exist.

I have preferred to undo the exception rather than to adjust the type.

## Testing

Since this change affects pinning we need to test the pinning flows:
 - Open the app and make sure that pinned notes are at the top
 - Pin a note and make sure that it jumps to the top
 - Unpin a note and make sure that it returns to its former spot in the list
 - Pin a note from a remote device and make sure that it updates accordingly here
 - Do the same as the previous step but for unpinning
 - Export the notes and make sure that the `pinned` property comes through to the exported format appropriately.
 - Import notes with `pinned` properties and make sure that they import properly.

[229a76b]: 229a76b4e6992728c20746cb5b6d0d3d32d27571

### DIFF
--- a/lib/flux/test.js
+++ b/lib/flux/test.js
@@ -51,9 +51,9 @@ describe('appState action reducers', () => {
       it('should load the newest version of the selected note', () => {
         const oldState = {
           notes: [{}],
-          note: { id: 'foo', data: 'old' },
+          note: { id: 'foo', data: { content: 'old', systemTags: [] } },
         };
-        const newNote = { id: 'foo', data: 'new' };
+        const newNote = { id: 'foo', data: { content: 'new', systemTags: [] } };
         const newNoteArray = [newNote];
         const newState = notesLoaded(oldState, {
           notes: newNoteArray,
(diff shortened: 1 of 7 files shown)


lib/flux/test.js
```diff
@@ -70,8 +70,14 @@ describe('appState action reducers', () => {
...
         };
-        const firstNote = { id: 'foo', data: 'first' };
-        const newNoteArray = [firstNote, {}];
+        const firstNote = {
+          id: 'foo',
+          data: { content: 'first', systemTags: [] },
+        };
+        const newNoteArray = [
+          firstNote,
+          { id: 'bar', data: { content: 'boo', systemTags: [] } },
```
codebykat: 👻

codebykat (approved):
Great catch on all this stuff that isn't necessary anymore! This tests out fine.

PR #1796, opened 2019-12-23
App-state: Remove unused `indexCtr`

dmsnell:
In the earlier stages of the app we relied on counting index operations
to defer updating the internal note list. This was to prevent thrashing
in the app when receiving updates for an account with lots of notes.

In #89 we moved the `indexCtr` value out of the controlled app state and
turned it into an instance variable on the `actionMap` object. This
change was made because we couldn't wait for a Redux update cycle to
update that value - the instance property updated immediately and
imperatively.

In #987 we removed the index-counting operation entirely and instead
relied on a more natural `debounce()` on `bucket.on('index')`. This
removed the need for the work-around but didn't remove the `indexCtr`
property.

In this patch we're finally removing it to finish the work from #987.
The property is currently unused and unreferenced in the app.

## Testing

There should be no functional or visual changes in this patch.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -23,6 +23,7 @@
 - Added git hooks to run format, lints, and tests [#1790](https://github.com/Automattic/simplenote-electron/pull/1790)
 - Added React Hooks ESLint Plugin [#1789](https://github.com/Automattic/simplenote-electron/pull/1789)
 - Added end-to-end testing with Spectron [#1773](https://github.com/Automattic/simplenote-electron/pull/1773)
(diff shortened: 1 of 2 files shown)


codebykat (approved):
👍 

As an aside, do we really need release notes for very minor code clean-up that doesn't create any user-facing changes? My instinct would have been to skip it but I don't know if we have a hard-and-fast rule about what changes "require" release notes. cc @bummytime

dmsnell:
> do we really need release notes for very minor code clean-up

My plan was to incorporate all of the random cleanups into this one line. We might need to do a PR to cleanup the release notes before the next release though.

PR #1797, opened 2019-12-23
Fix: Possibly unknown note when creating note.

dmsnell:
When creating a new note we know that the operation might fail but we
keep on treating it as a success.

In this patch we send a debug message if the note bucket fails to create
our note and we don't continue processing as if it had.

## Testing

This is going to be a hard one to test unless you mangle Simperium, so
focus on auditing the change itself and consider the possible side-effects
of it. Make sure that we can still create notes and that those new notes
get synchronized with other devices.

This change is introducing a safety guard where none previously existed.


codebykat (approved):
As you said, I can't easily create the conditions for this failure but the code looks fine and a smoke-test didn't surface any issues. Do we want to handle this error in a user-facing way someday too?

dmsnell:
> Do we want to handle this error in a user-facing way someday too?

Yes, and I would have here but I didn't know the answer to how we ought to nor know how to quickly estimate what we'd have to do, plus it'd introduce more complexity in the state model I'm trying to undo (the action-creators in action-map), so for now I thought that a `debug()` call was the least we could do and that it's still better than it is now.

I'd guess that we'd have reports if this happened frequently enough for it to be a problem.

PR #1798, opened 2019-12-27
Tags: Treat comma as a submit key

### LINKED ISSUE #1670: Tag entry allows commas, empty tag
Typing a comma in the tag field results in an empty tag being added, and the comma remaining in the text field. The empty tag then shows up in the tag list and can be removed by clicking on the empty tag bubble, but not by backspacing. Typing further commas clutters up the tag field with commas:

Screen Shot 2019-10-24 at 10 21 32 PM

### Expected
* Empty tags should be filtered out.
* Commas should be deleted after processing tag entry.

Tested on OSX/Firefox.

codebykat:
In the tag input, typing a comma acts like a submit (it gets stripped out and the tag is submitted). This fixes an issue where you could type a single comma and have a blank tag added.

Fixes #1670 

### Fix
Having the comma act as a "submit" was the intent of the original code, but there were bugs with it. Instead of testing to see if the value `endsWith` a comma and then attempting to strip it out, let's just go all-in and intercept the comma as if it was an Enter.

### Test
1. Go to a note
2. Focus on the "Add a tag..." text input at the bottom
3. Type a comma by itself. Note that it does not appear.
4. Type a tag name followed by a comma. Verify that the tag is added to the note.

### Release
`RELEASE-NOTES.txt` was updated in 414beb22ec6f5a1aaae0b3e0a61a6fcb6f038a28 with:
 
> Fixed an issue where typing a comma in the tag input would insert an empty tag [#1798](https://github.com/Automattic/simplenote-electron/pull/1798)

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -12,6 +12,7 @@
 - Added indication that publish url has been copied [#1743](https://github.com/Automattic/simplenote-electron/pull/1743)
 - Disallow partial emails and user's own email from being adding to collaborators email field [#1735](https://github.com/Automattic/simplenote-electron/pull/1735)
 - Fixed keyboard shortcut to toggle markdown preview [#1788](https://github.com/Automattic/simplenote-electron/pull/1788)
+- Fixed an issue where typing a comma in the tag input would insert an empty tag [#1798](https://github.com/Automattic/simplenote-electron/pull/1798)
 
 ### Other Changes
 
--- a/lib/tag-input/index.jsx
+++ b/lib/tag-input/index.jsx
@@ -5,6 +5,7 @@ import { get, identity, invoke, noop } from 'lodash';
 const KEY_TAB = 9;
 const KEY_ENTER = 13;
 const KEY_RIGHT = 39;
+const KEY_COMMA = 188;
 
 const startsWith = prefix => text =>
   text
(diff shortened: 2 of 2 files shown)


dmsnell (approved):
On iOS the comma is allowed as part of the tag name and it pops in when you hit space.
On macOS the comma submits the tag just as proposed in this PR.

It makes sense to use comma as a submission token. If we're already typing then submitting a tag early and starting to type a new one is going to be roughly equivalent to typing a series of comma-separated tags and then submitting all at once. I think that in any case we'll be getting earlier feedback on what someone is actually doing because they aren't waiting until the end to realize that it splits on the comma.

Nice work.

RELEASE-NOTES.txt
```diff
@@ -12,6 +12,7 @@
 - Added indication that publish url has been copied [#1743](https://github.com/Automattic/simplenote-electron/pull/1743)
 - Disallow partial emails and user's own email from being adding to collaborators email field [#1735](https://github.com/Automattic/simplenote-electron/pull/1735)
 - Fixed keyboard shortcut to toggle markdown preview [#1788](https://github.com/Automattic/simplenote-electron/pull/1788)
+- Fixed an issue where typing a comma in the tag input would insert an empty tag [#1798](https://github.com/Automattic/simplenote-electron/pull/1798)
```
dmsnell: This change feels more like a change than a fix, even though it does fix. "Insert a tag when pressing comma" or something like that seems to capture the customer-impact of the change.

codebykat: Well, previously it inserted a tag when pressing comma, it just also allowed empty strings and failed to clean up the input field after doing so.

PR #1801, opened 2019-12-31
Re-render immediately when system theme is changed

codebykat:
### Fix
In #1581 we added a setting for System theme, however the app currently only checks the system theme on load. This PR adds a listener to the browser shell that stores the system theme in app state, allowing the app to re-render immediately when the system theme is changed.

### Test
1. Load the app
2. Change the system theme
3. Observe that the theme changes without requiring a reload

### Review

### Release
 `RELEASE-NOTES.txt` was updated in ea483769292cace8825a2730ea0604b50c94c6ba with:
 
> When system theme is selected in Settings, changing the system theme is now immediately reflected in the app [#1801](https://github.com/Automattic/simplenote-electron/pull/1801)


dmsnell (approved):
Good find on using the `BrowserShell` and using the `matchMedia().addListener()`. I think this will work great as part of the imperative Browser bits.

PR #1802, opened 2020-01-02
Update Dependencies

belcherj:
### Fix
Updates Dependencies

### Test
Run app and click around.

### Review

### Release

- Updated dependencies [#1802](https://github.com/Automattic/simplenote-electron/pull/1802)


package.json
```diff
@@ -66,9 +66,9 @@
     "@types/redux": "3.6.0",
     "@types/redux-localstorage": "1.0.8",
     "@types/rimraf": "2.0.3",
-    "@types/webdriverio": "4.13.3",
-    "@typescript-eslint/eslint-plugin": "2.12.0",
-    "@typescript-eslint/parser": "2.12.0",
+    "@types/webdriverio": "5.0.0",
```
dmsnell: we need to leave `webdriverio` and `spectron` alone unless we're updating `electron` as the versions need to correspond. I added a note in the README but looking back on it now it doesn't seem so clear on that point.

package.json
```diff
@@ -107,21 +107,21 @@
     "react-test-renderer": "16.12.0",
     "rimraf": "3.0.0",
     "sass-loader": "8.0.0",
-    "spectron": "6.0.0",
-    "style-loader": "1.0.1",
-    "typescript": "3.7.3",
+    "spectron": "9.0.0",
```
dmsnell: see note above: we need to leave this alone outside of an electron update

dmsnell (changes requested):
I know that we're updating electron in another PR, but for this one can we leave `webdriverio` and `spectron` as-is so we don't introduce a commit with mismatched testing libraries?

package.json
belcherj: ahh, completely missed that readme update

PR #1804, opened 2020-01-03
Use name attribute not the deprecated getter function

belcherj:
### Fix
https://electronjs.org/docs/all#appgetname is deprecated,
https://electronjs.org/docs/all#appname should be used instead. This PR
replaces uses of `.getName()` with `.name`.

### Test
Open built dmg from ci and ensure the App name in the context menu is correct.


belcherj:
Well, this doesn't work. Getting undefined in the menu from the app built by cci

PR #1806, opened 2020-01-03
Sets babel compact option to false is development

belcherj:
### Fix
Babel was outputting warnings about lodash and react-dom that it wasn't
removing white space. This sets babel compact to false in order to stop those
warnings. We also don't need babel to compact in development.
https://babeljs.io/docs/en/options#compact

### Test
Does the app build from `npm run dev`?

### Review

### Release

- Fixed build warning [#1806](https://github.com/Automattic/simplenote-electron/pull/1806)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -18,6 +18,9 @@ module.exports = function(api) {
     '@babel/plugin-syntax-dynamic-import',
   ];
   const env = {
+    development: {
+      compact: false,
+    },
     test: {
       plugins: ['dynamic-import-node'],
     },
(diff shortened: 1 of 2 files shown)


babel.config.js
```diff
@@ -18,6 +18,9 @@ module.exports = function(api) {
     '@babel/plugin-syntax-dynamic-import',
   ];
   const env = {
+    development: {
+      compact: false,
+    },
     test: {
       plugins: ['dynamic-import-node'],
```
dmsnell: should it be here too?

what if we turned it off but added it in `production`? is there any place other than production to turn it on?

belcherj: My feeling is that we should run the tests on compact.

PR #1808, opened 2020-01-04
Refactor(NoteList): Use already assigned list iterator

dmsnell:
In the existing code we're recomputing the current list item in the note list
when we've already assigned it above. In this patch we're using the
already-assigned value.

exploratory PR #1807.

## Testing

Smoke test with some note searches.
Open the settings panel and change the list display mode.
Make sure that the notes list displays correctly.

This is a very minor change and should be able to be confirmed
through code inspection and auditing.


belcherj (approved):
Ship it!

PR #1809, opened 2020-01-04
Refactor(NoteList): Use numbers for tabIndex instead of strings

dmsnell:
Previously we have been (successfully) passing string values of `0` into the `tabIndex` property on parts of the note list preview. While acceptible this is the wrong datatype. In this patch we're passing the `{0}` value instead through React so that the types unify.

exploratory PR #1807.

## Testing

Open Simplenote and make sure that when tabbing you don't stop on the note
pinner or the div surrounding the preview. Confirm that the behavior matches
that in `develop`

This is a very minor change and should be able to be confirmed
through code inspection and auditing.


belcherj (approved):
Tested and works. We should revisit tabbing at some point. So much potential for improvement.

dmsnell:
Updated to rebase `RELEASE-NOTES` against `develop`

PR #1810, opened 2020-01-04
Refactor(NoteList): Reshape composite note list data structure

dmsnell:
When adding tag-suggestions in the note list we create a "composite list" of notes and placeholders. The placeholders designate where headers will go, where the tag suggestions list will go, and where a message will go if there are no matching notes.

There were two issues with this structure:
 - it was difficult to type because of how its structure overlapped a `NoteEntity` but didn't have a key field for discriminating the type. everything works just fine but the type signatures become harder to write.
 - we were storing display information in the `data` property and that, as a render concern, wasn't necessary

Further, we were directly mutating `notes` which came from `filterNotes()` and the Redux state. We want to avoid directly mutating values we receive from other places in order to keep data flows clear and obvious.

In this patch we are replacing the object data type for placeholders with simpler string counterparts, moving the label text into the `render` function, and exchanging a mutating operation for a non-mutating one.

exploratory PR #1807.

## Testing

This affects the tag suggestions auto-completer and the note list when searching.
Smoke test by searching in various ways: with filter text, with a tag from the tag list, with `tag:something` in the filter text, with the trash selected, with no notes matching the search, etc…

In early testing at one point I had the notes header and tag-suggestions header backwards, for example.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -25,7 +25,7 @@
 - Added React Hooks ESLint Plugin [#1789](https://github.com/Automattic/simplenote-electron/pull/1789)
 - Added end-to-end testing with Spectron [#1773](https://github.com/Automattic/simplenote-electron/pull/1773)
 - Removed a workaround for indexing note pinned status [#1795](https://github.com/Automattic/simplenote-electron/pull/1795)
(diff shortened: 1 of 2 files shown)


belcherj (approved):
Much cleaner!

dmsnell:
Updated to rebase `RELEASE-NOTES` with recent changes from `develop`

PR #1811, opened 2020-01-04
Refactor(NoteList): Use `createRef()` instead of ref function

dmsnell:
Keeping up with React standards. This patch replaces the use of a ref function
with a `createRef()`. This will end up making our types easier.

exploratory PR #1807.

## Testing

This is a small change. Smoke test by moving around the notes list.


belcherj (approved):
Good change. Tested and works

dmsnell:
Updated to rebase `RELEASE-NOTES` with updates from `develop`

dmsnell:
Rebased to catch file rename in `develop` from #1813

PR #1813, opened 2020-01-04
TypeScript: Rename all files in `lib/` as `.ts`/`.tsx`

dmsnell:
We've been carving out individual files in separate PRs but haven't seen any
issues with converting to TypeScript files. In this patch we're doing it in
bulk so we can stop conflating changes with renames and so that we'll benefit
from the additional type-checking going on throughout the app.

**Do you see any reason why we shouldn't do this?**

## Testing

There should be no visual or functional changes in this PR.
In fact, the output JS should be identical to what it was before.

Verify that the app builds and runs, verify that the built bundles run.


belcherj:
Looks like you are way ahead of me:  https://github.com/Automattic/simplenote-electron/pull/1812#pullrequestreview-338851182

dmsnell:
Rebased against `develop` to catch changes in `lib/note-list/` which would have been merge conflicts

PR #1814, opened 2020-01-04
Add missing decorators in note list previews

dmsnell:
When showing each item in the note list we run the content decorators on them
in order to show checkboxes instead of the markdown syntax for a checkbox and
we also highlight the matching text from the search box.

Previously we have only been showing the first match for each decorator because
of a couple small bugs:
 - the task prefix matcher wasn't using the `m` multi-line flag in its RegExp
   pattern
 - the search filter text was only replacing the first occurrance, see JS
   "replaceAll()" proposal

In this patch we're rewritten `decorateWith` to automatically create a global
RegExp pattern for the search text and the task prefix RegExp pattern has been
corrected.

After this patch you can now see all of the decorated items in the note list;
for example, if you had two checklist items in a note then you can now see
multiple checkboxes in the note list whereas before you'd only see the first
one and then the second one would appear as ` - [ ]` or `- [x]`

## Testing

Open an account with notes having multiple repeated text in the first few lines.

For example…(`one` appears in each line)
```
# A short poem
One river,
Two stones,
In the zone.
```

Search for the repeated pattern, such as `one`

You should find that in **develop** you'll only see the first result highlighted
but in this branch all of the `one`s should be highlighted.

Repeat the experiment with a checklist…
```
# Groceries
 - [ ] Eggs
 - [ ] Chicken
 - [ ] Milk
```

In **develop** only the first item will render as a checklist but in this branch
all three should, if they are visible in the preview.

Verify that if you search for regular-expression-y patterns then you'll only
get results matching the literal text and not the pattern itself.

```
Patterns I use:

4 a number - /\d/
```

If you search for `\d` then the `\d` in the note should be highlighted, not the `4`

## Before
missing-decorators mov

## After
all-the-decorators mov

### DIFF
--- a/lib/note-detail/toggle-task/constants.ts
+++ b/lib/note-detail/toggle-task/constants.ts
@@ -3,6 +3,6 @@ export const selectors = {
   markdownRoot: '[data-markdown-root]',
 };
 
-export const taskPrefixRegex = /^(\s*)(-[ \t]+\[[xX\s]?\])/g;
+export const taskPrefixRegex = /^(\s*)(-[ \t]+\[[xX\s]?\])/gm;
 
 export const taskRegex = /^(\s*)(-[ \t]+\[[xX\s]?\])(.+)/gm;
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -14,6 +14,7 @@
 - Fixed keyboard shortcut to toggle markdown preview [#1788](https://github.com/Automattic/simplenote-electron/pull/1788)
 - Fixed an issue where typing a comma in the tag input would insert an empty tag [#1798](https://github.com/Automattic/simplenote-electron/pull/1798)
 - When system theme is selected in Settings, changing the system theme is now immediately reflected in the app [#1801](https://github.com/Automattic/simplenote-electron/pull/1801)
+- Show all checkboxes and search results in note list [#1814](https://github.com/Automattic/simplenote-electron/pull/1814)
 
 ### Other Changes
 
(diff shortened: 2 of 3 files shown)


belcherj (approved):
Works great.

dmsnell:
Rebased to catch renames from #1813 in `develop`

lib/note-detail/toggle-task/constants.ts
```diff
@@ -3,6 +3,6 @@ export const selectors = {
   markdownRoot: '[data-markdown-root]',
 };
 
-export const taskPrefixRegex = /^(\s*)(-[ \t]+\[[xX\s]?\])/g;
+export const taskPrefixRegex = /^(\s*)(-[ \t]+\[[xX\s]?\])/gm;
```
dmsnell: @mirka can you remember if there was any specific reason not to include all search results here by adding the `m`? I didn't see any mention of it in the commit log but I don't want to accidentally introduce a bug you intentionally avoided.

mirka: It seems like between the time I added this regex and now, the behavior of the `Comfy`/`Expanded` note list views have changed. Before, it showed the title and only the first line of the body, with `Expanded` showing more of the first line if it was a long paragraph. Now, it's concatenating multiple lines of the body if the lines are short.

So back then we didn't need the `m` flag, simply because a note body preview would never show more than one task item line. (In that sense, it's actually the `g` flag that was superfluous!)

dmsnell: ah! that makes sense. we did update the preview generation so that it would show the start of the first few lines of the note instead of collapsing them. we missed the impact here then when we made that change.

thanks for chiming in and helping @mirka!

PR #1818, opened 2020-01-07
Remove prop types warnings

belcherj:
### Fix
Remove prop types warnings from the linter as we will be removing them with the use of Typescript types

### Test
No testing needed

### Review


belcherj:
It actually wants "off", change has been made.

Issue #1525, opened 2019-08-15
Toggle Sidebar button non-responsive until dev console is opened


rachelmcr:
### Expected

I expect that clicking the "Toggle Sidebar" button will expand/collapse the sidebar.

### Observed

Clicking the "Toggle Sidebar" button did nothing.

When I opened the dev console, the button suddenly started working and continued working after I closed the console.

### Reproduced

1. Log in to the app if needed.
2. Click the "Toggle Sidebar" button. Note that it is unresponsive.
3. Open the dev console (under Help > Advanced > Debugging Console).
4. Try again to toggle the sidebar.

Reproduced on both Windows 10 VM and macOS 10.14.5, with Simplenote 1.8-beta1

codebykat:
This has happened to me twice now and I can't figure out how to consistently reproduce it. In both cases it did not start working when I opened the dev console, but did start working after I resized the window to be narrow enough that the toggle went away, and then back to full width.

codebykat:
Three times and it seems to always happen when I first load up the dev environment in a browser, so some race condition with the cache possibly? Once I've loaded it in a browser logging out, reloading, and stopping and restarting the dev server fail to produce the bug again.

PR #1819, opened 2020-01-07
Fixes focus mode toggle

belcherj:
### Fix
Fixes: #1525

When the app loads there is potential for the `is-note-open` css class to be
missing on app-layout. This removes the reliance on isNoteOpen as it is note
needed and should be independent. Now focus mode class will remain on all
screen widths but will only change the layout for widths greater than 750px. We
should look at why `is-note-open` is not applied to app layout and how the layout
is set up in general.

### Test
This is hard to test since the original issue is hard to replicate. The best thing you can do is make the function that adds `is-note-open` not add it.

1. Open app on widescreen
2. Toggle focus mode several times
3. Open app
4. Toggle focus mode
5. Make width less than 750px
6. Click back and forth between note list and note view
7. Open menus etc.

### Review

### Release

- Fixes focus mode toggle sometimes not functioning [#1819](https://github.com/Automattic/simplenote-electron/pull/1819)

### DIFF
--- a/lib/app-layout/style.scss
+++ b/lib/app-layout/style.scss
@@ -3,16 +3,27 @@
   flex: 1 0 auto;
   width: 100%;
 
-  &.is-focus-mode.is-note-open {
+  &.is-focus-mode {
     .app-layout__source-column {
       overflow: hidden;
       opacity: $fade-alpha;
       width: 0;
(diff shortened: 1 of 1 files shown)


belcherj:
Yes, we never really hide anything.

PR #1820, opened 2020-01-07
WIP: Explore pushing the types through Redux all the way

dmsnell:
**DO NOT MERGE**: This is only intended to serve as a place to explore and discuss.

How far can we go with autocomplete and compile-time guarantees?

I'm exploring different approaches to doing that with several goals:
 - get full auto-complete on all calls to dispatch, all state values, all action types
 - move runtime overhead into the type system
 - reduce the number of required imports around the app
 - remove existing clutter and minimize syntax required for typing


dmsnell:
Obviated by #1855

PR #1822, opened 2020-01-08
Render unfinished lists correctly

### LINKED ISSUE #676: Markdown (preview) mode lists problem
Hi,

I've faced strange behavior while creating a lists in Simplenote 1.1.0. Wasn't sure if that's expected behavior for the markdown, but I tested it later in online markdown tools and they didn't act the same as Simplenote.

#### Steps to reproduce

1. Create a new note and paste the following text (make sure to have markdown for that note enabled):

```
# Test

- test 1
- test 2
- test 3
- 
```

2. Switch from Edit to Preview mode.

#### What I expected

I expected consistently styled list without the bold parts.

#### What happened instead

Last list item turned into bold (maybe even heading?) and the last `-` character disappeared.

screenshot_289

#### OS version
Ubuntu 16.10

belcherj:
### Fix

Fixes: #676

When a list has a trailing `-` the markdown renderer (showdown) will make the
previous list element an h2. Technically this is valid markdown and is doing
what it should but this is usually not the intention of the user. Because we
auto-add  a `-` after a carriage return users run into this issue regularly.

This issue is resolved by enabling the `smoothLivePreview` option in showdown.

You can replicate this with the following markdown:

```
- test 1
- test 2
- test 3
-
```
Currently, this will output:

```
<ul>
<li>test 1</li>
<li>test 2</li>
</ul>
<h2 id="test3">- test 3</h2>
```
After the addition of the `smoothLivePreview` option it will return:
```
<ul>
<li>test 1</li>
<li>test 2</li>
<li>test 3</li>
-
</ul>
```

### Test
1. Use the sample markdown from above
2. Toggle preview to ensure output is correct

### Review

### Release

- Fixed markdown render issue with lists with hypehn only on new line [#1822](https://github.com/Automattic/simplenote-electron/pull/1822)

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -16,7 +16,7 @@
 - When system theme is selected in Settings, changing the system theme is now immediately reflected in the app [#1801](https://github.com/Automattic/simplenote-electron/pull/1801)
 - Show all checkboxes and search results in note list [#1814](https://github.com/Automattic/simplenote-electron/pull/1814)
 - Fix ol numbering in markdown preview [#1823](https://github.com/Automattic/simplenote-electron/pull/1823)
-- Fixed markdown render issue with lists with hypehn only on new line [#1822](https://github.com/Automattic/simplenote-electron/pull/1822)
+- Prevents weird effects in live previews due to incomplete input [#1822](https://github.com/Automattic/simplenote-electron/pull/1822)
 
 ### Other Changes
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -16,6 +16,7 @@
 - When system theme is selected in Settings, changing the system theme is now immediately reflected in the app [#1801](https://github.com/Automattic/simplenote-electron/pull/1801)
 - Show all checkboxes and search results in note list [#1814](https://github.com/Automattic/simplenote-electron/pull/1814)
 - Fix ol numbering in markdown preview [#1823](https://github.com/Automattic/simplenote-electron/pull/1823)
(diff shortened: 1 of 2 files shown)


RELEASE-NOTES.txt
```diff
@@ -15,6 +15,7 @@
 - Fixed an issue where typing a comma in the tag input would insert an empty tag [#1798](https://github.com/Automattic/simplenote-electron/pull/1798)
 - When system theme is selected in Settings, changing the system theme is now immediately reflected in the app [#1801](https://github.com/Automattic/simplenote-electron/pull/1801)
 - Show all checkboxes and search results in note list [#1814](https://github.com/Automattic/simplenote-electron/pull/1814)
+- Fixed markdown render issue with lists with hypehn only on new line [#1822](https://github.com/Automattic/simplenote-electron/pull/1822)
```
dmsnell: There's a typo here on `hypehn` but also I feel like the option may fix some other stylistic issues. We might want to borrow from `showdown` itself

```
 - Prevents weird effects in live previews due to incomplete input [#1822](…)
```

dmsnell (approved):
Thanks for this update. I left a note about the README which might be worth updating before merging but it's your call.

As an earmark I think we can come back to this and refactor how we instantiate `showdown`. Several of these options may be possible to set in the initial `new showdown.Converter(…)` call and it looks like we may have more options than we used to, for example, the `tasklists` option.

RELEASE-NOTES.txt
belcherj: Done

belcherj:
> Thanks for this update. I left a note about the README which might be worth updating before merging but it's your call.

Di you mean Release notes? I do not see a note about the readme.

Issue #1531, opened 2019-08-16
Flickers between two notes


gubrist:
### Expected
Open Simplenote, select one note and read/edit it

### Observed
The whole app flickers between two (oder more?) notes , see video below
You are not able to do anything.

### Reproduced
Happened on various Windows PC's independently
- on Electron (days before) and 
- on Web Version (today)

### Video:
https://youtu.be/h5OtylKxNwo

### Make|Model|OS Version|App Version
- Electron version before 1.7.0
- Web Version (Firefox)
...both Windows 10

bummytime:
Bump. Seen by a user ~2 weeks ago → Chrome Version 79.0.3945.88 on Windows

msilbers:
"When I use simplenote on the web (chrome) the screen flashes, flipping between notes, i can't select anything. it's usually between one of hte pinned notes & whichever one I am trying to actually read."

both Chrome and Internet Explorer
reported internally #2638774-zen

msilbers:
Windows app and web app, Simplenote, v1.12.0
Flashing between notes while the notes are loading. seems to stop flashing once they are all loading in the browser

"after some experimentation, it seems like the period where this can happen is while notes are loading in."

reported in #2635070-zen

msilbers:
"The browser version isn’t working at all now. It "flashes” and all I can make out is “no notes” and then goes blank. On different browsers it’s the same or at first it appears to be ok - I’m seeing the proper notes - but it’s very slow to switch between notes then will get stuck on a note and redisplay it rapidly and at that point it’s unresponsive."

reported in Safari and Brave, in #2626756-zen

belcherj:
This should have already been resolved and is a different issue from #2085

belcherj:
If you are still experiencing this issue please comment and I will reopen.

PR #1825, opened 2020-01-09
Fixes thrashing between notes

belcherj:
### Fix
Fixes: #1531

We have reports of users who has seen the web and desktop apps thrashing
between two notes. What happens is that the user will get into a state where
for some reason, the selected note is repeatedly changing back and forth between
two notes.

I believe that this was happening because the componentDidUpdate on the note
list was setting the selected note id if the note id was different than
previous props. My suspicion is that it was putting the selected note id into
an infinite loop of changing back and forth between two notes.

I have not been able to replicate this bug but this seems like the proper
solution.

### Test
1. Click around app
2. Make sure the note editor is displaying the note you think it should
3. Do search
4. Open trash
5. Trash a note
6. Restore a note

### Review

### Release

### DIFF
--- a/lib/flux/app-state.ts
+++ b/lib/flux/app-state.ts
@@ -206,6 +206,19 @@ export const actionMap = new ActionMap({
       });
     },
 
+    searchAndSelectFirstNote: {
+      creator({ filter }: { filter: T.TagEntity }) {
+        return (dispatch, getState) => {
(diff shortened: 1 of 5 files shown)


belcherj:
resolved in: https://github.com/Automattic/simplenote-electron/pull/1851

PR #1826, opened 2020-01-10
Search / Tag suggestions: Escape regex characters in user input

codebykat:
### Fix
Typing a regex special character into the search field messed up the tag suggestions. At best this would show irrelevant suggestions. At worst it would crash the app.

### Test
1. Type `?` into the search field. Note that it shows only tags and notes that actually contain "?"s.
2. Type `??`. Note that the app does not crash and also shows only tags and notes that match.
3. Try other regex special characters, such as `* / \ ( ) $ ^ ~ { } + &` and verify nothing weird happens.

### Release
`RELEASE-NOTES.txt` was updated in 03b0ab93536de41e5e634ef89afecf70c550a049 with:
> Properly escape special characters from the search field when matching tags [#1826](https://github.com/Automattic/simplenote-electron/pull/1826)


codebykat:
Abandoning; will be fixed by #1827

PR #1828, opened 2020-01-10
Filter notes: Make tag match inclusive of unicode characters

### LINKED ISSUE #1770: Search: No results when doing a tag search with non-English characters
### Expected

I expect to be able to search with the `tag:` syntax for any of my tags.

### Observed

Some of my tags include non-English characters, and those tag searches return no search results. Here are two tags I have that I observed this issue with:

* `سلام`
* `español`

### Reproduced

1. Create a tag with a non-English character, such as `español`, and add it to at least one note.
2. In the tag list, select the tag `español`.
3. Notice that the note list shows the notes with that tag.
4. Go back to the All Notes view.
5. Search for just the tag name `español`.
6. Notice that under "Search by Tag" you see `tag:español`.
7. Select `tag:español`.
8. Notice that the note list says "No Notes" instead of showing the note(s) with that tag.

simplenote-electron-nonENtags

-|-|-|-
Windows|VirtualBox VM|Win10|1.13.0-beta1
Mac|Macbook Pro (2017)|macOS 10.15.1|1.13.0-beta1

cc @codebykat

codebykat:
### Fix
This fixes #1770, a bug where tags with non-alphanumeric characters (other languages, emoji, etc) showed "No Notes" when typed into the search bar or chosen from tag suggestions, even when there were matching notes.

### Test
1. Follow the testing instructions at #1770 and confirm that non-alphanumeric tags properly filter the note results. 
2. Confirm that you can still search for multiple tags (e.g. `tag:one tag:two`).

### Release
`RELEASE-NOTES.txt` was updated in 46a0dfc726a7f8717f39f0e790ec15ccff8eca9d with:
> Fixed a bug where searching for a tag containing non-alphanumeric characters erroneously returned no notes [#1828](https://github.com/Automattic/simplenote-electron/pull/1828)

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -17,6 +17,7 @@
 - Show all checkboxes and search results in note list [#1814](https://github.com/Automattic/simplenote-electron/pull/1814)
 - Fix ol numbering in markdown preview [#1823](https://github.com/Automattic/simplenote-electron/pull/1823)
 - Prevents weird effects in live previews due to incomplete input [#1822](https://github.com/Automattic/simplenote-electron/pull/1822)
+- Fixed a bug where searching for a tag containing non-alphanumeric characters erroneously returned no notes [#1828](https://github.com/Automattic/simplenote-electron/pull/1828)
 
 ### Other Changes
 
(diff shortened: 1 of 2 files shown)


dmsnell (approved):
As discussed in Slack this seems to be a good move. I can't remember entirely why I added the `(?!\B)` at the end originally but we agreed that even when it matches things like periods and punctuation, those are valid tag names. The only thing constraining tag names are the characters that the apps use to split them into separate tags.

PR #1829, opened 2020-01-10
Note List: Replace custom height cache with one provided by react-virtualized

dmsnell:
Follows exploratory work in #1807

For a long time we have had our own custom note list height cache. This was
used to tell `react-virtualized` how tall each item in the list is so that it
can determine how many and which ones to render. That cache was complicated and
sensitive due to the caching and performance-oriented nature of the code.

Whether `CellMesurerCache` existed when we first put in `react-virtualized` I
don't know but it does do what we want it to do and with less guessing. It
keeps track of the heights for the rendered components and provides a clean API
to interface with the `AutoSizer`.

In this patch we're removing our custom height cache and instead relying on the
one provided by `react-virtualized`.

## Testing

Many app flows interact with the note list. Please vet as many of them as you can
find. A failure in this code would likely lead to skewed or obscured views in the list.

There are some existing quirks in the way that the tag suggestions and list filtering
interact; this patch does not attempt to address those.

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -100,11 +100,15 @@ const renderNote = (
 ): ListRowRenderer => ({ index, key, parent, style }) => {
   const note = notes[index];
 
-  if (
-    'tag-suggestions' === note ||
-    'notes-header' === note ||
-    'no-notes' === note
-  ) {
+  if ('no-notes' === note) {
+    return (
+      <div className="note-list is-empty" style={{ ...style, height: 200 }}>
+        <span className="note-list-placeholder">No Notes</span>
+      </div>
+    );
+  }
+
+  if ('tag-suggestions' === note || 'notes-header' === note) {
     return (
       <CellMeasurer
         cache={heightCache}
@@ -115,12 +119,8 @@ const renderNote = (
       >
         {'tag-suggestions' === note ? (
           <TagSuggestions />
-        ) : 'notes-header' === note ? (
-          <div className="note-list-header">Notes</div>
         ) : (
-          <div className="note-list is-empty">
-            <span className="note-list-placeholder">No Notes</span>
-          </div>
+          <div className="note-list-header">Notes</div>
         )}
       </CellMeasurer>
     );
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -1,23 +1,15 @@
-/**
- * @module NoteList
- *
- * This module includes some ugly code.
- * The note list display is a significant source of
- * visual re-render lag for accounts with many notes.
- * The trade-offs in this file reflect a decision to
- * optimize heavy inner loops for performance over
- * code burden.
- *
- * Any changes to this code which could affect the
- * row height calculations should be double-checked
- * against performance regressions.
- */
 import React, { Component, Fragment, createRef } from 'react';
 import PropTypes from 'prop-types';
-import { AutoSizer, List } from 'react-virtualized';
+import {
+  AutoSizer,
+  CellMeasurer,
+  CellMeasurerCache,
+  List,
+  ListRowRenderer,
+} from 'react-virtualized';
 import PublishIcon from '../icons/feed';
 import classNames from 'classnames';
-import { debounce, get, isEmpty } from 'lodash';
+import { get } from 'lodash';
 import { connect } from 'react-redux';
 import appState from '../flux/app-state';
 import analytics from '../analytics';
@@ -36,6 +28,7 @@ import * as T from '../types';
 import { getTerms } from '../utils/filter-notes';
 
 type OwnProps = {
+  isSmallScreen: boolean;
   noteBucket: T.Bucket<T.Note>;
 };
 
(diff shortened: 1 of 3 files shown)


dmsnell:
@codebykat because of how coupled the tag suggestions and the note list are I have folded _moving tag suggestions into state_ into this PR where we can fully test the flow.

note that now tag suggestions are locked with the note filtering and occur on a debounce while someone is changing the search filter. note too that the wishy-washy memoization from `tag-suggestions` is also gone on account of this.

there are directions to move after this. in my own play here I was trivially able to modify `filterTags()` to return not just the first five tags matching the query but also to return them in order of how many notes in the current filter also have that tag. I backed out that change because it had some usability issues and belongs on its own, but now that we've centralized the filtering and are no longer creating a new `compositeNoteList()` on every call to `NoteList::mapStateToProps` we have much more room to improve the quality of the search results.

it's my expectation that it will now be possible to move search filtering into a `WebWorker` after this change where we can do much better quality searching with far less impact on the main UI thread.

codebykat:
Tests failing on:

> noteTitleAndPreview › should complete in a reasonable amount of time

SO I guess it wasn't just my imagination :sob:

I'm testing this out now!

codebykat:
I'm still seeing a noticeable lag and gap in the filtering :(

dmsnell:
> Okay so now the noticeable lag is between typing a result (and seeing search matches highlighted) and the tag matches (and headers) showing up. It's slow/jumpy enough that it feels broken :/

This should now synchronize with the note list filtering, which to me was the bigger problem before with them appearing at different times. We can play around a bit with the debounce delay but we won't be able to filter notes as fast as we have been getting our tag suggestions, no matter how we cut it, just because note filtering is much slower and takes up the CPU (locking the UI).

If we start looking at #1835 we can think more about faster search updates since it won't hold up the main thread. I'm seeing "note thrashing" pretty bad in that branch though so I wouldn't recommend moving forward until we have that solved.

what are your thoughts about this? you can try changing the delay in `middleware.js` to 200ms and see if that feels better.

codebykat:
Okay so. 200ms feels better! BUT. Something is getting backed up? It starts out zippy enough but the more I type stuff in the filter, the slower it gets until eventually I can't type at all without waiting a few seconds for any characters to show up. Do we need to bail on the debounce somewhere?

I tried to record performance but I'm not sure how to read this. There are definitely some "DOM events" (apparently event handlers?) that are taking forever. I saved the perf recording too.

Screen Shot 2020-01-16 at 3 44 44 PM

dmsnell:
> Something is getting backed up?

Actually I bet they are doing just that. If the search is still going we might queue up another one when indeed we ought to clear the queue…

dmsnell:
One thing I noticed @codebykat is that when searching it would pull up a big note with lots of links and `draft-js` would bring the app to a halt. This could be related to our thrashing issue. Maybe we need to solve that in this PR too.

It's becoming quite big but I attribute that to the deep coupling existing within the app.

codebykat:
Yeah, I noticed a lag with large notes too (I have a few on my test account due to the duplication bug) but this was an additional one that seemed to accumulate over time regardless of which notes were selected.

belcherj:
Is this ready for another review?

dmsnell:
Rebuilt based on dependent changes made along the way. This should be ready for a full re-review.

lib/note-list/index.tsx
```diff
@@ -59,244 +52,126 @@ type DispatchProps = {
...
-    case 'no-notes':
-      return (
-        <div key={key} style={style} className="note-list is-empty">
-          <span className="note-list-placeholder">No Notes</span>
-        </div>
-      );
+        ) : 'notes-header' === note ? (
+          <div className="note-list-header">Notes</div>
+        ) : (
+          <div className="note-list is-empty">
```
codebykat, line 120: I think we need to keep the style argument here so that the empty note list gets the correct height? (But just trying that by itself did not fix the issue)

lib/note-list/index.tsx
```diff
@@ -1,23 +1,15 @@
-/**
- * @module NoteList
- *
- * This module includes some ugly code.
```
codebykat, line 4: 👏

lib/note-list/index.tsx
```diff
@@ -59,244 +52,126 @@ type DispatchProps = {
...
-};
-
-/** @type {Number} height of a single header in list rows */
-const HEADER_HEIGHT = 28;
-
-/** @type {Number} height of a single tag result row in list rows */
-const TAG_ROW_HEIGHT = 40;
-
-/** @type {Number} height of a the empty "No Notes" div in the notes list */
-const EMPTY_DIV_HEIGHT = 200;
```
codebykat, line 110: Ah, we needed this in order to prevent the placeholder getting squished up against the top

lib/note-list/index.tsx
```diff
@@ -59,244 +52,126 @@ type DispatchProps = {
...
-  const note = notes[index];
-
-  // handle special sections
-  switch (note) {
-    case 'notes-header':
-      return HEADER_HEIGHT;
-    case 'tag-suggestions':
-      return HEADER_HEIGHT + TAG_ROW_HEIGHT * tagResultsFound;
-    case 'no-notes':
-      return EMPTY_DIV_HEIGHT;
```
codebykat, line 153: ☝️

codebykat:
I also tested with a variety of lengths of tag suggestions and that all seems to be fine, it's just the empty placeholder that's wonky now.

lib/note-list/index.tsx
codebykat: So I know what the fix for this *should* be, and it's:

```
  style={{  ...style, height: 200 }}
```
But that yields THIS nonsense, for which I have no explanation whatsoever except that I clearly do not understand anything:

Screen Shot 2020-04-13 at 10 34 30 PM

codebykat:
Proposed fix for the above: https://github.com/Automattic/simplenote-electron/pull/1829/commits/54fbe95e56ef28c7fc93756d469073080255d01b

PR #1833, opened 2020-01-13
Allow width attribute on img tags

### LINKED ISSUE #1832: Add image resize ability to markdown
### What
Please add image resize ability in markdown mode like joplin has.

### Why
Some images are too big or too small for viewing in a notes app. It would be nice to be able to resize them.

### How
You already support ``, but the width/height attributes are ignored unfortunately. Please implement them. ``, for example, should behave just like the HTML tag where it sets the width to 50 pixels and proportionally adjusts the height.

qualitymanifest:
Closes https://github.com/Automattic/simplenote-electron/issues/1832

### Fix
Allow width attribute on `img` elements

### Test
1. Create an `img` element
2. Give it a width attribute
3. Success

### Review
Only one developer is required to review these changes

### Release
Updated release notes in bdfd05c


belcherj (approved):
Great change!

PR #1834, opened 2020-01-14
Fix tag rename functionality

belcherj:
### Fix

There is a typo in the variable rename function that is causing tags to be set
as undefined on tag rename.

### Test
1. Add tag to a site
2. Change the tag name
3. Observe that tag is renamed o the ote

### Review

### Release

- Fixed tag rename functionality [#1834](https://github.com/Automattic/simplenote-electron/pull/1834)


codebykat:
👏

PR #1835, opened 2020-01-15
WIP: Search: Offload processing to WebWorker and free up main thread

dmsnell:
Compare against #1829 (off of which it's based) instead of using the "Files changed" tab here.

[Real files changed in this branch ⑂](https://github.com/Automattic/simplenote-electron/compare/refactor/remove-custom-note-list-height-cache...search/offload-filtering)

## Status

 - the app is notably more interactive in this branch while searching than in `develop`
 - the "thrashing note" (being worked on in #1825) appears to be even more frequently triggered here and I'm not sure why. there is a lifecycle-method loop in the `NoteList` and also in the `NoteContentEditor`. In the `NoteList` it's changing the selected note when that should happen in the middleware when filtering notes. ⭕️ Move `closeNote()` into the middleware
 - Apart from testing to make sure this doesn't break things it seems fairly robust and there's nothing I would be strongly opposed to merging
 - **DO NOT MERGE** and definitely not before #1829 is merged. Since this is a branch on a branch it's likely to be that we'll have to rebase and update and resolve conflicts before this is ready to be considered for inclusion.

## Description

As long as our `NoteList` has been bound up in re-computing the fitlered notes on every render and on every call to `mapStateToProps` and as long as it has been calculating note indices based on that list it calculates we have been in a pickle when it comes to rendering performance whil searching.

We have taken a few steps towards mitigating the CPU bottleneck:
 - debounced the search filter
 - debounced searching when updating the filter
 - moved the filtered notes and tags into Redux and only
   compute them when a debounce times out after events change
   the filtering: e.g. a search or a remote update

In this patch we're doing what we've long waited to do which is move the search filtering out of the main thread and into a WebWorker. By doing this we not only eliminate one of the biggest performance bottlenecks for accounts with lots of data but we have opened ourselves up to dramatically increasing the quality of our search results since the costs of performing the searching and indexing can be overlooked as the main UI thread remains responsive.

This specific patch only takes the first step which is reproducing the search as-is. It also trades some of the performance gain by passing all of the notes and tags in the account to the WebWorker on every update. However, as this is a relatively fast operation it shouldn't pose any significant obstacle.

If we develop this design for search indexing one area we can start improving is to stop sending the full contents of all the notes and tags on every fitlering event. It's enough to pass the search filter, whether the trash is open, and if there's a selected tag. We can do this by trapping more actions in the middleware and maintaining a copy of the notes in the WebWorker.

We can probably do even better by maintaining our copy of the notes independently, by pulling from `indexedDB`, though that may carry with it the higher risk of further coupling our Ghost and Bucket implementations to a shared-state system we are trying to avoid.

Once the data-passing and synchronization is figured out we can start to perform more robust indexing of the notes and rank search results.

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -19,7 +19,16 @@ import {
   checkboxDecorator,
   makeFilterDecorator,
 } from './decorators';
-import TagSuggestions, { getMatchingTags } from '../tag-suggestions';
+import TagSuggestions from '../tag-suggestions';
+
+import { State } from '../state';
+import * as T from '../types';
+
+export type NoteListItem =
+  | T.NoteEntity
+  | 'tag-suggestions'
+  | 'notes-header'
+  | 'no-notes';
 
 AutoSizer.displayName = 'AutoSizer';
 List.displayName = 'List';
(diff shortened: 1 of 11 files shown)


dmsnell:
Closing in favor of #1941

PR #1848, opened 2020-01-20
Updates Dependencies to latest

belcherj:
### Fix
Updates dependencies to latest

### Test
1. Run app, click around, look for errors

### Review

### Release

- Updated Dependencies [#1848](https://github.com/Automattic/simplenote-electron/pull/1848)


package.json
```diff
@@ -88,29 +88,29 @@
...
     "rimraf": "3.0.0",
-    "sass-loader": "8.0.0",
+    "sass-loader": "8.0.2",
     "spectron": "6.0.0",
-    "style-loader": "1.1.2",
-    "typescript": "3.7.4",
-    "wait-on": "3.3.0",
+    "style-loader": "1.1.3",
+    "typescript": "3.7.5",
+    "wait-on": "4.0.0",
```
dmsnell: do we even need this? this is a major update so I checked to see what we'd need to change and it's not referenced anyway. it came in #859 with `misty` and I'm not sure if it was meant to be a transitive/peer dependency. I haven't tried building without it

belcherj: https://github.com/Automattic/simplenote-electron/blob/develop/misty.yml#L6

belcherj: It is used.

belcherj: I don't think the API changed but they rewrote the code thus the full point bump.

PR #1851, opened 2020-01-21
Refactor selected note state from flux to redux

belcherj:
### Fix
This PR moves the selected note state from flux to redux. It also eliminates
the need for selectedNoteId in appState.

### Test
1. Test the note selection flow
2. Tets the remote note change flows
3. Test note share flow
4. Test note publish
5. Test trash flow
6. Test tag selection
7. Test search

### Review

### Release

- Refactored selected note state [1851](https://github.com/Automattic/simplenote-electron/pull/1851)


lib/note-editor/index.tsx
```diff
@@ -161,13 +162,15 @@ export class NoteEditor extends Component {
...
-const mapStateToProps = ({ appState: state, settings }) => ({
+const mapStateToProps = ({ appState: state, settings, ui: { note } }) => ({
   allTags: state.tags,
   filter: state.filter,
   fontSize: settings.fontSize,
   editorMode: state.editorMode,
   isEditorActive: !state.showNavigation,
+  note,
   revision: state.revision,
+  tags: get(note, 'data.tags', []),
```
dmsnell: _note_ that with our TypeScript annotations we can relax our runtime safety here - we can call `note.data.tags` directly. If we try and call that by accident on `Note` instead of `NoteEntity` it will alert us.

if `tags` is possibly missing then we'll need to update our type or make sure to add it on entry into the app

belcherj: Noice, changed

lib/state/ui/middleware.ts
```diff
@@ -39,7 +48,46 @@ export default store => {
         }
         break;
     }
-
+    switch (action.type) {
+      case 'App.closeNote':
+      case 'App.showAllNotes':
+      case 'App.selectTrash':
+      case 'App.selectTag':
+        store.dispatch(setSelectedNote(null));
```
dmsnell: these right here need not be in middleware. they can be left in the reducer to keep the complexity out of the system.

```ts
const setSelectedNote = (
   state = defaultNote,
   { type, note }: { type: string; note: T.NoteEntity[] }
 ) => {
  switch (type ) {
    case 'App.closeNote':
    case 'App.showAllNotes':
    case 'App.selectTrash':
    case 'App.selectTag':
      return null;
    case SET_SELECTED_NOTE:
      return note;
    default:
      return state;
  }
 };
```

lib/state/ui/middleware.ts
```diff
@@ -39,7 +48,46 @@ export default store => {
...
+                ? action.notes.find(
+                    note => note.id === store.getState().ui.note.id
+                  )
+                : store.getState().ui.note
+            )
+          );
+        } else {
+          store.dispatch(
+            setSelectedNote(action.notes.length ? action.notes[0] : null)
+          );
```
dmsnell: similarly we can move this into the reducer as well since `action.notes` is available to us in the reducer

```ts
const setSelectedNote = (
   state = defaultNote,
   { type, note }: { type: string; note: T.NoteEntity[] }
 ) => {
  switch (type ) {
    …
    case 'App.notesLoaded': {
      if ( ! state ) {
        return action.notes[0] || null;
      }

      const note = action.notes.find( ( { id } ) => id === state.id );
      return note || state;
    }
    …
}
```

fun fact, although the entire `switch` block shares variable name context, if we add curlies to a `case` then we get a new context and can create variable names without conflict with other `case` blocks

dmsnell (changes requested):
You picked a hard one to start with 😀

I know we have some poor examples in the code already where our action creators are imperative and they suggest that we should move more logic into the middleware but I don't think there's any need in this PR to make any changes to the middleware - it looked like all of the code there could exist solely in the `selectedNote` reducer.

If that doesn't make sense I'd be happy to delve further into it but would you mind trying to make that change here? It will probably feel better after that.

lib/state/ui/reducer.ts
```diff
@@ -22,4 +26,38 @@ const visiblePanes = (state = defaultVisiblePanes, { type, show }) => {
   return state;
 };
 
-export default combineReducers({ filteredNotes, visiblePanes });
+const defaultNote: T.NoteEntity | null = null;
```
dmsnell: what's the reason for creating `defaultNote` here? I know we have a pattern occasionally of creating an empty list or object (since things like array literals are always reference-unique) but `null` is always `null` no matter which `null` it is.

lib/state/ui/reducer.ts
```diff
@@ -22,4 +26,38 @@ const visiblePanes = (state = defaultVisiblePanes, { type, show }) => {
...
+const setSelectedNote = (state = defaultNote, action) => {
+  switch (action.type) {
+    case 'App.notesLoaded': {
+      if (!state) {
+        return action.notes[0] || null;
+      }
+
+      const note = action.notes.find(({ id }) => id === state.id);
+      return note || state;
```
dmsnell: the code as written is fine but now that I'm looking at it there's also the collapsed version

```ts
return action.notes.find(({ id }) => id === state.id) || null;
```

though if what you wrote earlier is the case and we need to check for `notes.length` obviously we'll need more code.

if we're safe (which we can verify through the type system)…

```ts
case 'App.notesLoaded':
  return state
    ? (action.notes.find(({id}) => id === state.id) || state)
    : (action.notes[0] || null);
```

lib/state/ui/reducer.ts
belcherj: Good catch, not sure what I was doing there.

lib/state/ui/middleware.ts
belcherj: done

lib/state/ui/middleware.ts
belcherj: done

lib/dialogs/share/index.tsx
```diff
@@ -207,9 +205,9 @@ export class ShareDialog extends Component {
...
 
-export default connect(
-  state => ({
-    settings: state.settings,
-  }),
-  { updateNoteTags }
-)(ShareDialog);
+const mapStateToProps = ({ settings, ui: { note } }) => ({
+  settings,
+  note: note || {},
```
dmsnell: given that we know this will be `T.NoteEntity | null` do we need the `|| {}` here? I'm guessing that other code in this file was easier to keep if we defaulted to an empty object, but at the same time `{}` seems more obscure than `null` to indicate that it's missing.

don't worry about it if we're matching existing behavior or this is the difference between a small change and a huge change. it would be nice though if we let those types flow smoothly through the module.

lib/note-editor/index.tsx
dmsnell: if it was changed it doesn't appear to have made it here

as a benefit of removing `get()` - since this is `mapStateToProps` we will trigger re-renders as long as we return `[]` instead of some array that is reference-equal to the one before it (such as `note.data.tags` when the tags don't change).

lib/note-list/index.tsx
```diff
@@ -348,11 +348,6 @@ export class NoteList extends Component {
       this.recomputeHeights();
     }
 
-    // Ensure that the note selected here is also selected in the editor
-    if (selectedNoteId !== prevProps.selectedNoteId) {
-      onSelectNote(selectedNoteId);
-    }
```
dmsnell: does this PR also fix the thrashing issue?

lib/note-list/index.tsx
```diff
@@ -496,14 +491,12 @@ const { recordEvent } = tracks;
...
+  ui: { filteredNotes, note },
   settings: { noteDisplay },
 }) => {
   const tagResultsFound = getMatchingTags(state.tags, state.filter).length;
-
-  const noteIndex = Math.max(state.previousIndex, 0);
-  const selectedNote = state.note ? state.note : filteredNotes[noteIndex];
-  const selectedNoteId = get(selectedNote, 'id', state.selectedNoteId);
+  const selectedNote = note;
+  const selectedNoteId = get(selectedNote, 'id', null);
```
dmsnell: we should be free to use optional-chaining, but I haven't confirmed

```ts
const selectedNoteId = selectedNote?.id;
```

if it matters that it be `null` in the absence of a `selectedNote` then we only need add that default

```ts
const selectedNoteId = selectedNote?.id || null;
```

we could use the nullish coalescing operator to prevent collapsing a falsey value `…?.id ?? null` but since we know that `T.EntityId` is a string and an empty string is an invalid entity id we can leave the `||`

lib/note-editor/index.tsx
dmsnell: I see now too that `note` may be `null`, though we can employ the optional-chaining trick from below. I see why you did this now.

```ts
tags: note?.data.tags ?? []
```

lib/note-list/index.tsx
```diff
@@ -552,7 +545,7 @@ const mapStateToProps = ({
     prevNote,
     selectedNotePreview,
     selectedNoteContent: get(selectedNote, 'data.content'),
-    selectedNoteId,
+    selectedNoteId: get(selectedNote, 'id'),
```
dmsnell: did we update `selectedNoteId` above and then not use it here?

lib/note-toolbar/index.tsx
```diff
@@ -171,4 +172,8 @@ export class NoteToolbar extends Component {
   };
 }
 
-export default NoteToolbar;
+const mapStateToProps = ({ ui: { note } }) => ({
+  note,
+});
```
dmsnell: it's so good to see changes like this where we're moving from passing props _through_ components to getting the props from the global Redux store 👍

lib/state/action-types.ts
```diff
@@ -1,3 +1,4 @@
 export const AUTH_SET = 'AUTH_SET';
 export const FILTER_NOTES = 'FILTER_NOTES';
+export const SET_SELECTED_NOTE = 'SET_SELECTED_NOTE';
```
dmsnell: as you work here please be thinking about how best to represent our Redux store, actions, and selectors.

since we have a very minimal store to start with _and_ we're 100% TypeScript there are ways we can think about changing this to be more helpful in the IDE. if you haven't had the chance you can play around with some of the ideas in #1820. some things in there I like and others I don't.

lib/state/index.ts
```diff
@@ -68,7 +67,8 @@ export const store = createStore(
         [path]: omit(state[path], 'focusModeEnabled'),
       }),
     }),
-    applyMiddleware(thunk, uiMiddleware)
+    applyMiddleware(thunk, uiMiddleware),
+    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
```
dmsnell: can we pull this into another PR? no need to rush, but before merging it would be helpful to have it separated

lib/state/ui/middleware.ts
```diff
@@ -1,5 +1,5 @@
 import { AnyAction } from 'redux';
-import { filterNotes as filterAction } from './actions';
+import { filterNotes as filterAction, setSelectedNote } from './actions';
```
dmsnell: this can go, as can the removed newline below

lib/state/ui/reducer.ts
```diff
@@ -22,4 +26,42 @@ const visiblePanes = (state = defaultVisiblePanes, { type, show }) => {
...
+    case 'App.notesLoaded': {
+      return state
+        ? action.notes.find(({ id }) => id === state.id) || state
+        : action.notes[0] || null;
+    }
+    case 'App.noteUpdatedRemotely':
+    case 'App.selectNote': {
+      return {
+        ...action.note,
+        hasRemoteUpdate: action.hasRemoteUpdate,
```
dmsnell: did you check if this is used throughout the rest of the application?

if it is then we will want to add it to the type of `NoteEntity`. I'm guessing that we may not need it but in any case I'd rather have `hasRemoteUpdate?: boolean` on the type so we can rely on the type-checker to guard us

lib/state/ui/reducer.ts
```diff
@@ -22,4 +26,42 @@ const visiblePanes = (state = defaultVisiblePanes, { type, show }) => {
...
+    case 'App.showAllNotes':
+    case 'App.selectTrash':
+    case 'App.selectTag':
+      return null;
+    case SET_SELECTED_NOTE:
+      return action.note;
+    case 'FILTER_NOTES':
+      if (
+        action.notes.length &&
+        (!state || !action.notes.some(note => note.id === state.id))
```
dmsnell: it looks like we'll always have `notes` as an array…this shouldn't need the check…

```ts
import { FILTER_NOTES } from '../action-types'

case FILTER_NOTES:
  return state && action.notes.some(({id}) => id === state.id)
    ? state
    : action.notes[0] || null;
```

lib/note-list/index.tsx
belcherj: It should if my hypothesis that this code is the cause. I am unable to replicate so it is a bit of a shot in the dark. I wasn't able to figure out any other reason for the thrashing to occur. I think once we get more things into redux we can reduce the number of events that modify the selected note this eliminating any additional cause.

lib/dialogs/share/index.tsx
belcherj: I did that for a reason however I don't think we need the empty object.

lib/note-toolbar/index.tsx
belcherj: For sure.  As we can start deconstructing components and moving logic out to utils the codebase will be more manageable.

lib/state/action-types.ts
belcherj: What do you think of this?  I still don't have my brain fully wrapped on TS.

https://redux.js.org/recipes/usage-with-typescript/#type-checking-actions--action-creators

lib/state/index.ts
belcherj: Yeah, for sure.

lib/state/action-types.ts
dmsnell: We'll discuss it beyond this PR, but an earlier incantation of #1820 looked closer to your link. I changed it to try and minimize the number of named imports and improve the auto-complete while writing reducers and `connect()` functions.

Although I've grown accustomed to inferring the types of `mapStateToProps` I have come to like separating it into separate `OwnProps`, `DispatchProps`, and `StateProps` because of how they ease the typing of `mapStateToProps`, `mapDispatchToProps`, and `connect()`.

lib/flux/app-state.ts
```diff
@@ -396,47 +371,20 @@ export const actionMap = new ActionMap({
       },
     },
 
-    selectNote(state: AppState, { note, hasRemoteUpdate }) {
+    selectNote(state: AppState) {
```
dmsnell: right here it's evident how instead of a Redux model we built global setters in the application. we're asking "what data does this change" instead of "what changes this data"

lib/note-info/index.tsx
```diff
@@ -188,9 +188,7 @@ function characterCount(content) {
 
 const { markdownNote, pinNote, toggleNoteInfo } = appState.actionCreators;
 
-const mapStateToProps = ({ appState: state, ui: { filteredNotes } }) => {
-  const noteIndex = Math.max(state.previousIndex, 0);
-  const note = state.note ? state.note : filteredNotes[noteIndex];
+const mapStateToProps = ({ ui: { note } }) => {
```
dmsnell: excellent example of where our data can get out of sync, where we're using separate references to try and refer to the same data

lib/state/ui/reducer.ts
```diff
@@ -22,4 +26,42 @@ const visiblePanes = (state = defaultVisiblePanes, { type, show }) => {
   return state;
 };
 
-export default combineReducers({ filteredNotes, visiblePanes });
+const setSelectedNote = (state = null, action) => {
```
dmsnell: in Redux fashion we'd rather name this reducer `selectedNote` or even just `note` since it describes a data type and the ways that data changes vs. describing a _setter_ which updates data.

we can see how each reducer acts as an independent OO actor governing its own state and responding to messages received from "the outside world" by means of the actions

lib/note-list/index.tsx
belcherj: Ohhhhhhhh wow.  Optional chaining is my fav new feature no more `obj && obj.prop`. I so happy this is a thing!!

belcherj: Stacking: `let customerCity = customer.details?.address?.city;`!!!!

dmsnell: just know that it's not the same as `a && a.b && a.b.c` because it's only checking for `null` and `undefined` and not any falsey value. this is stronger in terms of safety but weaker in terms of what we usually do

lib/types.ts
```diff
@@ -18,6 +18,7 @@ export type Note = {
   content: string;
   creationDate: SecondsEpoch;
   deleted: boolean;
+  hasRemoteUpdate?: boolean;
```
dmsnell: According to what's written above, and as your IDE should point out if we type `action.note`, this is the wrong spot for `hasRemoteUpdate` since we added it to a `NoteEntity`.

We could:
 - change `Entity` to capture the possibility
 - change `NoteEntity = Entity<Note> & { hasRemoteUpdate?: boolean }`
 - deprecate the need for the property

I'd vote the second or third option.

belcherj: deprecate the need for the property, I think we can do this eventually but right now it is a complex change to make.

dmsnell: that's fine. I didn't know if it was or wasn't.

package.json
```diff
@@ -47,6 +47,7 @@
     "@babel/core": "7.7.7",
     "@babel/plugin-proposal-class-properties": "7.7.4",
     "@babel/plugin-proposal-object-rest-spread": "7.7.7",
+    "@babel/plugin-proposal-optional-chaining": "7.8.3",
```
dmsnell: let's get this in a pre-PR

lib/state/index.ts
```diff
@@ -34,7 +34,6 @@ export type AppState = {
   previousIndex: number;
   revision: T.NoteEntity | null;
   searchFocus: boolean;
-  selectedNoteId: T.EntityId | null;
```
dmsnell: `note?: T.NoteEntity` above should go too, no?

package.json
belcherj: It was added by default to preset env: https://github.com/babel/babel/releases/tag/v7.8.0

We just need to rebase in [this PR](https://github.com/Automattic/simplenote-electron/pull/1848) after it is merged

dmsnell (changes requested):
It looks like we need to rebase this ontop of the changes introduced in #1855.

Specifically noting the need in `action-types` where we're introducing a new string action-type export.

lib/state/action-types.ts
dmsnell: after the merge of #1855 and #1856 we didn't remove the existing constants but we don' need to add new ones here either. I'll try and get those existing ones removed in a separate PR so they don't stick around and confuse us.

belcherj:
@dmsnell This is ready for more review.

lib/app-layout/index.tsx
```diff
@@ -76,25 +74,20 @@ export const AppLayout: FunctionComponent<Props> = ({
...
           <NoteEditor
             isSmallScreen={isSmallScreen}
-            note={note}
             noteBucket={noteBucket}
             onNoteClosed={onNoteClosed}
             onUpdateContent={onUpdateContent}
             syncNote={syncNote}
-            tags={
-              get(note, 'data.tags', []) /* flattened to trigger re-render */
-            }
```
dmsnell: very good - more split-data going away 🎆

lib/note-editor/index.tsx
dmsnell: it doesn't look like we ever use `props.tags` - did you try removing them? we reference `props.allTags` but otherwise we get `tags` from `this.props.revision || this.props.note`

lib/state/ui/actions.ts
```diff
@@ -15,6 +15,13 @@ export const toggleSimperiumConnectionStatus: A.ActionCreator<A.ToggleSimperiumC
   simperiumConnected,
 });
 
+export const setSelectedNote: A.ActionCreator<A.SetSelectedNote> = (
+  note: T.NoteEntity | null
+) => ({
+  type: 'SET_SELECTED_NOTE',
+  note,
+});
```
dmsnell: after these changes is there any reason to `setSelectedNote(null)`?

lib/app.tsx
```diff
@@ -85,6 +86,7 @@ function mapDispatchToProps(dispatch, { noteBucket }) {
       dispatch(actionCreators.setSearchFocus({ searchFocus: true })),
     setSimperiumConnectionStatus: connected =>
       dispatch(toggleSimperiumConnectionStatus(connected)),
+    setSelectedNote: id => dispatch(setSelectedNote(id)),
```
dmsnell: The use looks correct but this signature looks misleading…did we mean to put `setSelectedNote: note => dispatch(setSelectedNote(note))`?

dmsnell:
@belcherj I added a new commit to try out the changes I proposed.

 - Added type annotations where we're introducing new things
 - Renamed `setSelectedNote` to `selectNote` to focus more on the action we're doing rather than the data we're changing
 - Removed the `tags` prop
 - Removed lingering references to `noteUpdatedRemotely`

I'll be thoroughly testing this later on but I need a break from it at the moment

lib/app.tsx
belcherj: ooops

lib/note-editor/index.tsx
belcherj: removed in: https://github.com/Automattic/simplenote-electron/pull/1851/commits/d14a6e9141ecbfc237edf5ec2c2abd50a5c2ad05

lib/app.tsx
belcherj: removed in: d14a6e9

lib/state/ui/actions.ts
belcherj: also resolved in https://github.com/Automattic/simplenote-electron/pull/1851/commits/d14a6e9141ecbfc237edf5ec2c2abd50a5c2ad05

belcherj:
I have tested this a bunch today. I think it is ready to roll.

dmsnell (approved):
Thanks for everything here; it's really exciting to see state finally cleaning up. According to your own testing please merge at your convenience.

Issue #1852, opened 2020-01-22
Tags: Tag can't be "unselected" and can be deleted unexpectedly


rachelmcr:
### Expected

When I am using the keyboard only, I expect that a tag will only be deleted if I backspace over it twice (once to select it and a second time to delete it).

### Observed

If I backspace over a tag once, it is selected and there is no way to deselect it. If I then start entering a new tag and hit the backspace key at any point, the previously selected tag is deleted.

### Reproduced

1. On a note, enter a tag (e.g. `test`).
2. Hit the backspace key once and notice that the tag is selected (highlighted in blue).
3. Hit the tab key and notice that the cursor appears again but the tag is still selected.
4. Start typing a new tag (e.g. `testing`).
5. Hit the backspace key and notice that the last letter you typed is deleted but also the previously selected tag (e.g. `test`) is also deleted unexpectedly.

This quick screencast shows this behavior and also shows that if a tag hasn't been selected, deleting text works as expected:

![electron-funwithtags](https://user-images.githubusercontent.com/8658164/72912225-392b8000-3d33-11ea-9e02-20b953708bed.gif)

Make|Model|OS Version|App Version
-|-|-|-
Macbook Pro|macOS|10.15.2|Simplenote 1.12 & 1.14-beta2

PR #1853, opened 2020-01-23
Allow unselecting tags by keyboard

qualitymanifest:
### Fix
Closes https://github.com/Automattic/simplenote-electron/issues/1852
If a tag is selected and the right arrow key is pressed, the tag is unselected and the focus is moved back into the tag field.

### Test
See the issue for details to reproduce, but use right arrow to unselect

### Review

### Release
Release notes updated in bde11cf

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 ### Enhancements
 - Stop erasing the copy buffer if copying empty editor selections [#1847](https://github.com/Automattic/simplenote-electron/pull/1847)
+- Allow for unselecting a tag by pressing right arrow [#1853](https://github.com/Automattic/simplenote-electron/pull/1853)
 
 ### Fixes
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,7 +4,7 @@
 
 ### Enhancements
 - Stop erasing the copy buffer if copying empty editor selections [#1847](https://github.com/Automattic/simplenote-electron/pull/1847)
-- Allow for unselecting a tag by tab or right arrow [#1853](https://github.com/Automattic/simplenote-electron/pull/1853)
+- Allow for unselecting a tag by tab or right arrow [#1853](https://github.com/Automattic/simplenote-electron/pull/1853) @qualitymanifest
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


qualitymanifest:
Just realized I forgot about the tab key! I'll add that in in a second. I figure I'll leave the right arrow control in as well, that's the way that I would naturally think to exit the tag chip.

qualitymanifest:
Good point, I didn't see the key handlers in `tag-input`

It looks like `tag-input`'s `onChange` does called for regular keypresses though, inside `onInput` (a few times - might be worth looking into?), and `this.state.tagInput` does get updated. Here's a screenshot of two keypresses (red line separating them) with a bunch of console.logs peppered in:
snote

I think the reason why `tag-input` doesn't intercept keyboard events when a tag is selected is because when a tag is "selected", the focus is actually on `hidden-tag` which is a child of `tag-field` and unrelated to `tag-input`

qualitymanifest:
Actually now that I re-read your comment I'm confused regarding the `interceptKeys` comment, mainly because `tag-field` and `tag-input` both have an `interceptKeys` method so I'm not sure which one you're referring to.

`tag-input`'s tab and and right-arrow handlers are for completing a suggestion, not for handling keyboard events when a tag-chip (really, `hidden-tag`) is "selected".

`tag-field`'s only current key handler, backspace, is for deleting a tag-chip when it is "selected" - so I think it makes sense to also put the other tag-chip key handlers here.

I could definitely be missing something though!

dmsnell:
@qualitymanifest what I was seeing at a cursory glance was that we only seem to be sending the updated tag text from `tag-input` to `tag-field` when we select a tag or accept the tag suggestion. because of this, when the `tag-field` intercepts the key presses it's making an assertion about the current `tag-input` text that's "out of date."

in other words, both components should agree that the input text is `compu` when we are typing `computer` but haven't yet finished typing. I believe that the parent component only sees `` until at the end of the typing it gets told "choose `computer`"

qualitymanifest:
@dmsnell From my understanding, `tag-field`'s state is getting updated on each keypress. `tag-input`'s `inputObserver` is triggered on keypress and kicks off `onInputMutation` -> `onInput` -> `onChange` -> `tag-field`'s `storeTagInput` where it is updated

That being said - it seems unrelated to the issue of being able to key out of a tag-chip. `tag-field`'s current `interceptKeys` handler is only for selecting and deleting a tag chip, and there is no key handler to unselect that tag-chip

qualitymanifest:
Made the changes, reads a lot better with the constants @dmsnell

PR #1855, opened 2020-01-23
Types: Add core types to support Redux refactors

dmsnell:
Introduces a basic design for managing our Redux store in the typesystem
and establishes norms for how to create new action types and reducers as
well as how to add types to our React components and interact with Redux

There should be no functional or visible changes in this PR - this only adds
new types and documentation to the app and should have no impact on the
code produced.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,220 @@
+## Redux and Typescript
+
+The Redux store shall be fully-typed to provide safety and to enable auto-completion and automated bug-discovery in an IDE.
+Please follow these examples when creating new parts of app state and when connecting through `react-redux`.
+
+### Importing Types
+
+Try to keep type imports consistent across different modules.
+
+`action-types` contains the list of possible action types, not surprisingly, as well as the helper types for writing action creators and reducers.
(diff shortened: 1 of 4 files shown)


dmsnell:
Testing:

⭕️ set auto-hide menu bar - _wasn't sure how to test_
⭕️ had trouble unsetting `Share analytics` - in turning it off it would immediately turn back on automatically. this seems to only happen after seeing the `get-note-titles` issue below
⭕️ seeing `Could not find note with id '…'` from `get-note-titles.ts` when opening tag drawer
✅ Toggling markdown for a note changes the mode and toggles the preview icon
✅ Changing the list display mode in settings instantly updates the note list view
✅ Changing the sort type and direction instantly updates
⭕️ Alphabetical sort puts `'Something` above `Apple`, lexically the quote appears before `a`
✅ Changing the theme instantly updated the display of the app
⭕️ There doesn't seem to be a way to go back to `system` setting for theme
✅ Using the search bar filters the notes as I expect
✅ Using the search bar with a tag selected also filters as expected
✅ Toggling the tag draw opens and closes it as expected

PR #1856, opened 2020-01-24
Refactor: Add types to legacy Redux modules

dmsnell:
In this patch we're adopting the changes in #1855 and updating legacy
reducers, action types, and action creators. This serves both as a guide
for adding new Redux pieces while improving the overall type-safety and
completeness of the Redux subsystem in the app.

## Testing

Except for the authentication status there should be no functional or
visual changes in this PR. Please test the affected flows anyway and
report if you find any discrepancies.

Because we're changing from using `Symbol`s for authentication
status verify that you can login, logout, fail to login, etc…

### DIFF
--- a/lib/state/action-types.ts
+++ b/lib/state/action-types.ts
@@ -1,3 +1,7 @@
+import * as T from '../types';
+
+import { AuthState } from './auth/constants';
+
 export const AUTH_SET = 'AUTH_SET';
 export const FILTER_NOTES = 'FILTER_NOTES';
 export const TAG_DRAWER_TOGGLE = 'TAG_DRAWER_TOGGLE';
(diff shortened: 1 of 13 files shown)


dmsnell:
Testing:

⭕️ set auto-hide menu bar - _wasn't sure how to test_
⭕️ had trouble unsetting `Share analytics` - in turning it off it would immediately turn back on automatically. this seems to only happen after seeing the `get-note-titles` issue below
⭕️ seeing `Could not find note with id '…'` from `get-note-titles.ts` when opening tag drawer
✅ Toggling markdown for a note changes the mode and toggles the preview icon
✅ Changing the list display mode in settings instantly updates the note list view
✅ Changing the sort type and direction instantly updates
⭕️ Alphabetical sort puts `'Something` above `Apple`, lexically the quote appears before `a`
✅ Changing the theme instantly updated the display of the app
⭕️ There doesn't seem to be a way to go back to `system` setting for theme
✅ Using the search bar filters the notes as I expect
✅ Using the search bar with a tag selected also filters as expected
✅ Toggling the tag draw opens and closes it as expected

dmsnell:
Failed to reproduce `get-note-title` bug in `develop`, switched back to this branch, unable to reproduce so I'm going to okay it as a glitch that we can look for.

Sort and theme issues exist in `develop`

PR #1861, opened 2020-01-24
Fix: Added change notes to wrong version (#1860)

dmsnell:
In #1831 we added the change notes to v1.14.0's notes but this fix did not make
it into the v1.14.0 release. In this patch we're moving the notes into the
future v1.15.0 release's notes.

As this only touches the release notes there should be no need for testing.


dmsnell:
Merging without finished tests since no code changes are involved.

PR #1866, opened 2020-01-27
Move editor mode to redux

belcherj:
### Fix
This PR moves the editor mode state from flux to redux.

### Test
1. Open app
2. Click into note with markdown
3. Does the note display in edit mode?
4. Click preview button
5. Does note display in preview mode
6. Click between notes
7. Does preview mode remain enabled
8. Turn off preview mode
9. Does preview mode turn off
10. Turn preview back on
11. Turn off markdown in the info panel for a note
12. Does preview mode turn off?

### Review

### DIFF
--- a/lib/flux/app-state.ts
+++ b/lib/flux/app-state.ts
@@ -33,7 +33,6 @@ const toggleSystemTag = (
 };
 
 const initialState: AppState = {
-  editorMode: 'edit',
   filter: '',
   previousIndex: -1,
   notes: null,
@@ -47,7 +46,6 @@ const initialState: AppState = {
   editingTags: false,
   dialogs: [],
   nextDialogKey: 0,
-  shouldPrint: false,
   searchFocus: false,
   unsyncedNoteIds: [], // note bucket only
 };
@@ -137,12 +135,6 @@ export const actionMap = new ActionMap({
       });
     },
 
-    setEditorMode(state: AppState, { mode }: { mode: T.EditorMode }) {
-      return update(state, {
-        editorMode: { $set: mode },
-      });
-    },
-
     showDialog(state: AppState, { dialog }) {
       const { type, multiple = false, title, ...dialogProps } = dialog;
 
@@ -211,10 +203,6 @@ export const actionMap = new ActionMap({
 
           if (state.showTrash) {
             dispatch(this.action('showAllNotes'));
-          }
-
-          if (settings.markdownEnabled) {
-            dispatch(this.action('setEditorMode', { mode: 'edit' }));
           }
 
           // insert a new note into the store and select it
@@ -325,12 +313,6 @@ export const actionMap = new ActionMap({
     setIsViewingRevisions(state: AppState, { isViewingRevisions }) {
       return update(state, {
         isViewingRevisions: { $set: isViewingRevisions },
-      });
-    },
-
-    setShouldPrintNote(state: AppState, { shouldPrint = true }) {
-      return update(state, {
-        shouldPrint: { $set: shouldPrint },
       });
     },
 
--- a/lib/note-editor/index.tsx
+++ b/lib/note-editor/index.tsx
@@ -5,15 +5,20 @@ import appState from '../flux/app-state';
 import TagField from '../tag-field';
 import { property } from 'lodash';
 import NoteDetail from '../note-detail';
+import { toggleEditMode } from '../state/ui/actions';
 
 import * as S from '../state';
 import * as T from '../types';
+
+type DispatchProps = {
+  toggleEditMode: () => any;
+};
 
 type StateProps = {
   note: T.NoteEntity | null;
 };
 
-type Props = StateProps;
+type Props = DispatchProps & StateProps;
 
 export class NoteEditor extends Component<Props> {
   static displayName = 'NoteEditor';
@@ -21,7 +26,6 @@ export class NoteEditor extends Component<Props> {
   static propTypes = {
     allTags: PropTypes.array.isRequired,
     closeNote: PropTypes.func.isRequired,
-    editorMode: PropTypes.oneOf(['edit', 'markdown']),
     isEditorActive: PropTypes.bool.isRequired,
     isSmallScreen: PropTypes.bool.isRequired,
     filter: PropTypes.string.isRequired,
@@ -30,12 +34,10 @@ export class NoteEditor extends Component<Props> {
     onNoteClosed: PropTypes.func.isRequired,
     onUpdateContent: PropTypes.func.isRequired,
     revision: PropTypes.object,
-    setEditorMode: PropTypes.func.isRequired,
     syncNote: PropTypes.func.isRequired,
   };
 
   static defaultProps = {
-    editorMode: 'edit',
     note: {
       data: {
         tags: [],
@@ -73,11 +75,7 @@ export class NoteEditor extends Component<Props> {
       'p' === key.toLowerCase() &&
       this.markdownEnabled
     ) {
-      const prevEditorMode = this.props.editorMode;
-      const nextEditorMode = prevEditorMode === 'edit' ? 'markdown' : 'edit';
-
-      this.props.setEditorMode({ mode: nextEditorMode });
-
+      this.props.toggleEditMode();
       event.stopPropagation();
       event.preventDefault();
       return false;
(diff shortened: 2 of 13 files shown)


lib/flux/app-state.ts
```diff
@@ -146,11 +145,7 @@ export const actionMap = new ActionMap({
       });
     },
 
-    setEditorMode(state: AppState, { mode }: { mode: T.EditorMode }) {
-      return update(state, {
-        editorMode: { $set: mode },
-      });
-    },
+    setEditorMode(state: AppState, { mode }: { mode: T.EditorMode }) {},
```
belcherj: We still need this in the interim  while we untangle more of this.

lib/state/ui/reducer.ts
```diff
@@ -1,12 +1,26 @@
...
 
 import * as A from '../action-types';
 import * as T from '../../types';
 
 const defaultVisiblePanes = ['editor', 'noteList'];
 const emptyList: unknown[] = [];
 
+const editorMode: A.Reducer<T.EditorMode> = (
+  state = 'edit',
+  action: AnyAction
```
dmsnell: you shouldn't need to import `AnyAction` or provide the type to `action` as it's covered by `A.Reducer`

belcherj: The linter was yelling at me.

belcherj: Screen Shot 2020-01-27 at 7 13 08 PM

App.setEditorMode is not typed in action-types.ts.

Should we go ahead and type all the legacy "actions" out while we move everything over to redux?

dmsnell: I don't know if it's worth typing all the existing actions. That would be a good PR to write and see if it's practical. I'm more comfortable leaving in the type errors though than doing something we wouldn't want to do in order to satisfy the linter.

belcherj: I removed the type

belcherj:
@dmsnell @codebykat ready for more review

lib/state/ui/reducer.ts
belcherj: I removed the type, will explore creating a PR, but I think that I'll be through this sooner than later making that work obsolete.

belcherj:
@dmsnell @codebykat please rereview

lib/types.ts
```diff
@@ -62,7 +62,7 @@ export type Bucket<T> = {
 // Application Types
 ///////////////////////////////////////
 
-export type EditorMode = 'edit' | 'markdown' | 'preview';
```
dmsnell: Good catch. I wonder why I saw `preview` before, because I thought I had… 🤔

lib/flux/app-state.ts
```diff
@@ -145,11 +144,7 @@ export const actionMap = new ActionMap({
       });
     },
 
-    setEditorMode(state: AppState, { mode }: { mode: T.EditorMode }) {
-      return update(state, {
-        editorMode: { $set: mode },
-      });
-    },
+    setEditorMode() {},
```
dmsnell: In line 225/220 below we do this…

```ts
if (settings.markdown) {
  dispatch(this.action('setEditorMode', { mode: 'edit' });
}
```

…which we can tell is superfluous and causes another dispatch cycle.

instead we can watch for `App.newNote` in our `editorMode` reducer and make sure that whenever we create a new note we reset to the edit mode.

```ts
const editorMode: A.Reducer<T.EditorMode> = (state = 'edit', action) => {
   switch (action.type) {
     case 'SET_EDITOR_MODE':
     case 'App.setEditorMode':
       return action.mode;
     case 'App.newNote':
       return 'edit'; 
     default:
       return state;
   }
 };
```

lib/note-toolbar-container.ts
```diff
@@ -129,7 +129,7 @@ const mapDispatchToProps = dispatch => ({
   deleteNoteForever: args => dispatch(deleteNoteForever(args)),
   noteRevisions: args => dispatch(noteRevisions(args)),
   restoreNote: args => dispatch(restoreNote(args)),
-  setEditorMode: args => dispatch(setEditorMode(args)),
+  setEditorMode: mode => dispatch(setEditorMode(mode)),
```
dmsnell: No need to do all the work in this PR, but these actions somewhat subvert our normal Redux pattern. All we're doing is passing them to the child component, so how about moving them properly while we're fixing the other aspect of this one Redux state type.

That is, can we eliminate this action from the `<NoteToolBar />` container and instead put it in `<NoteToolbar />` where it's used?

In that file it looks like we're using state-based logic as well when all we care about doing is toggling. If we wanted, we could further remove the duplicated logic and create a second new action-type and let the reducer centralize and isolate that logic…

```ts
// action-types
export type SetEditorMode = Action<'SET_EDITOR_MODE', { mode: T.EditorMode }>;
export type ToggleEditorMode = Action<'TOGGLE_EDITOR_MODE'>;

// reducer
const editorMode: A.Reducer<T.EditorMode> = (state = 'edit', action) => {
   switch (action.type) {
     case 'TOGGLE_EDITOR_MODE':
       return state === 'edit' ? 'markdown' : 'edit';
     case 'SET_EDITOR_MODE':
     case 'App.setEditorMode':
       return action.mode;
     case 'App.newNote':
       return 'edit';
     default:
       return state;
   }
 };
```

lib/note-editor/index.tsx
```diff
@@ -68,7 +68,7 @@ export class NoteEditor extends Component {
       const prevEditorMode = this.props.editorMode;
       const nextEditorMode = prevEditorMode === 'edit' ? 'markdown' : 'edit';
 
-      this.props.setEditorMode({ mode: nextEditorMode });
+      this.props.setEditorMode(nextEditorMode);
```
dmsnell: referring to my comment from below this place also appears to be asking for a `TOGGLE_EDITOR_MODE` action and were to add that we could remove the duplicated logic for setting the `nextEditorMode`

```ts
  closeNote,
  toggleEditorMode,
}

handleShortcut = event => {
  …

  if ( … ) {
    this.props.toggleEditorMode();
    event.stopPropagation();
    event.preventDefault();
    return false;
  }
}
```

dmsnell:
It looks like if we handle `App.newNote` in the `editorMode` reducer we can eliminate the old method altogether in `app-state`. For some reason I remember having an issue with that, maybe because `newNote` is a `creator`? If that's the case we might find it more appealing in the short-term to stub out a `NEW_NOTE` or `CREATE_NOTE` (I like "create" since it seems less ambiguous to receiving a new note from the server) action and trap on that

```ts
// action-types
export type CreateNote = Action<'CREATE_NOTE'>;

// editorMode reducer
case 'CREATE_NOTE':
  return 'edit';
```

Later on that action type could be filled-in with additional necessary properties when we remove it from `app-state`

lib/flux/app-state.ts
belcherj: Screen Shot 2020-01-30 at 2 56 44 PM

App.newNote does not fire on creating a new note.

dmsnell: > For some reason I remember having an issue with that, maybe because newNote is a creator? 

That must be the problem I found. In that case do you have thoughts on creating a new proper action type (which at this point need not include any parameters or reducers) for `CREATE_NOTE` which we could then listen for?

We won't get rid of the double-dispatch but at least it would preserve the semantic of the triggering action - user created a note - which we could listen for, and which we can expand as we go to fill in (that new action).

lib/state/ui/reducer.ts
```diff
@@ -1,11 +1,24 @@
 import { difference, union } from 'lodash';
 import { AnyAction, combineReducers } from 'redux';
+
 import * as A from '../action-types';
 import * as T from '../../types';
 
 const defaultVisiblePanes = ['editor', 'noteList'];
 const emptyList: unknown[] = [];
 
+const editMode: A.Reducer<boolean> = (state = true, action) => {
```
dmsnell: changing from named states to boolean seems like a semantic regression.

does `true` or `false` mean "preview"?

I can follow the idea if we're saying "edit mode" _is_ the mode, though particularly since we talk about edit vs. preview when discussing this I personally find `'edit'` and `'preview'` clearer in the code

belcherj: I reasoned about this for a while.  I'd love to chat about it but I came to the opposite conclusion. We are in either editor more or preview.  If edit mode is true then we are in preview.  Just as is the tag panel open or closed, or are we in focus mode or is the note list open.

PR #1869, opened 2020-01-28
Move print state from flux to redux

belcherj:
### Fix
We use a shouldPrint boolean in AppState to determine if we should print. When shouldState becomes true we initiate a print command. This was all happening in AppState.

This PR moves this state to state/ui/print. Due to the way that Electron menu bar actions interact with the app I have included a switch statement on L265 of lib/app.tsx. This allows us to start routing actions to redux. This will eventually make the code preceding it obsolete.

### Test
1. Open electron app
2. Ensure the app is in focus not the dev tools
3. Under the file menu item click print
4. Does the print dialog come up?
5. Print
6. Follow steps again
7. Click cancel
8. Follow steps again
9. Does dialog pop up?

### Review

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -30,6 +30,7 @@ import {
   pick,
   values,
 } from 'lodash';
+import { toggleShouldPrint } from './state/ui/actions';
 
 import * as settingsActions from './state/settings/actions';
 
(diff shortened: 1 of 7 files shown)


dmsnell:
If I'm understanding this correctly then we can get rid of this entire bit of state, create a new empty Action type, and send the print command in middleware.

In fact, since we're not even doing anything but `window.print()` I don't understand why we even have this state.

What happens if you take all the `shouldPrint` state out and then in `onAppCommand` we do this?

```ts
if ('setShouldPrint' === command.action) {
  window.print();
}
```

If that works we could use a better name in the Electron menu than `setShouldPrint`, such as `print` or `printNote`

dmsnell:
In 0eb023f5ed612a620df56774466a0cf5c8c25dae I do see what may explain this: always printing the rendered markdown view.

If that's the case and the only reason this exists then I could imagine a more clean way of implementing this by having `onAppCommand()` force that explicitly instead of relying on detecting `shouldPrint`

```ts
if ('printNote' === command.action) {
  const { editorMode, note, setEditorMode } = this.props;

  setTimeout( () => {
    setEditorMode( 'preview' );
    setTimeout( () => {
      window.print();
      setEditorMode( editorMode );
    }, 0 );
  }, 0 );
}
```

This isn't good either, and looks pretty bad, but at least it's in one spot and decouples an imperative - print the note - from a declarative render of the app. The timeouts are to make sure the app re-renders because I think that even though we dispatch the change the state will update and this function will finish before the app re-renders. Setting a timeout of 0 should schedule the print to occur as soon as possible after the other events flush. This may be wrong. It may still happen before re-render.

Anyway I think it's worth trying to eliminate the state here if we can do it even with a solution as bad as this if we can make it reliable because of how it will remove that re-render-loop inside the components.

Thoughts?

belcherj:
closing in favor of https://github.com/Automattic/simplenote-electron/pull/1872

PR #1870, opened 2020-01-28
Move simperium connection state from flux to redux

belcherj:
### Fix
When Simperium connects and disconnects we display different text and a different icon in the sync status popover. This PR moves that state from flux to the redux state.

### Test
1. Observe the sync status in the hamburger menu
2. Does it say that it is connnected?

I have not been able to get the websocket to send a disconnection message. I am not sure how to test the disconnection.

### Review

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -175,12 +175,8 @@ export const App = connect(
         .on('message', setLastSyncedTime)
         .on('message', this.syncActivityHooks)
         .on('send', this.syncActivityHooks)
-        .on('connect', () => {
-          this.props.setSimperiumConnectionStatus(true);
-        })
-        .on('disconnect', () => {
-          this.props.setSimperiumConnectionStatus(false);
-        });
+        .on('connect', () => this.props.setSimperiumConnectionStatus(true))
+        .on('disconnect', () => this.props.setSimperiumConnectionStatus(false));
 
       this.onLoadPreferences(() =>
         // Make sure that tracking starts only after preferences are loaded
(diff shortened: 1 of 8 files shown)


lib/app.tsx
```diff
@@ -166,18 +169,18 @@ export const App = connect(
...
         .on('authorized', this.onAuthChanged)
         .on('unauthorized', this.onAuthChanged)
         .on('message', setLastSyncedTime)
         .on('message', this.syncActivityHooks)
         .on('send', this.syncActivityHooks)
-        .on('connect', () => setConnectionStatus({ isOffline: false }))
-        .on('disconnect', () => setConnectionStatus({ isOffline: true }));
+        .on('connect', () => {
+          this.props.setSimperiumConnectionStatus(true);
+        })
```
dmsnell: fun fact! we don't need the curlies

```ts
.on('connect', () => this.props.setSimperiumConnectionStatus(true))
```

belcherj: ohh, true. I think I had console.logs in there for testing.

lib/components/sync-status/index.tsx
```diff
@@ -1,16 +1,19 @@
...
-class SyncStatus extends Component {
-  static propTypes = {
-    isOffline: PropTypes.bool.isRequired,
-    unsyncedNoteIds: PropTypes.array.isRequired,
-  };
+import * as S from '../../state';
+
+type StateProps = {
+  simperiumConnected: boolean;
+  unsyncedNoteIds: string[];
```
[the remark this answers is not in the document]
belcherj: made that change most likely while you were in the middle of your review.

belcherj:
I did find it odd that this was in UI but figured it would be easy to move later.

PR #1871, opened 2020-01-29
Move unsyncednoteids state from flux to redux

belcherj:
### Fix
Moves unsyncednoteids state from flux to redux

### Review
1. Have unsynced notes, you can probably create a few by deleting notes and then trashing permanently
2. Toggle hamburger menu
3. Hover over unsynced notes at bottom
4. Does everything display the same as production

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -30,7 +30,10 @@ import {
   pick,
   values,
 } from 'lodash';
-import { toggleSimperiumConnectionStatus } from './state/ui/actions';
+import {
+  setUnsyncedNoteIds,
+  toggleSimperiumConnectionStatus,
+} from './state/ui/actions';
 
 import * as settingsActions from './state/settings/actions';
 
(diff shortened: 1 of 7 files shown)


lib/app.tsx
```diff
@@ -30,7 +30,10 @@ import {
   pick,
   values,
 } from 'lodash';
-import { toggleSimperiumConnectionStatus } from './state/ui/actions';
+import {
+  setUnsyncedNoteIds,
+  toggleSimperiumConnectionStatus,
+} from './state/ui/actions';
```
dmsnell: should be able to import them all in one go if you find it easier

```ts
import * as actions from './state/actions';

dispatch( actions.ui.setUnsyncedNoteIds );
```

belcherj: I like having them separated. If there were more than five or we had multiple /actions imports this would make sense to me.

lib/components/sync-status/index.tsx
```diff
@@ -62,19 +62,17 @@ class SyncStatus extends Component<StateProps> {
...
 }
 
 const mapStateToProps: S.MapState<StateProps> = ({
-  appState: state,
-  ui: { simperiumConnected },
+  ui: { simperiumConnected, unsyncedNoteIds },
 }) => ({
   simperiumConnected,
-  unsyncedNoteIds: state.unsyncedNoteIds,
+  unsyncedNoteIds,
```
dmsnell: this can come out entirely, no? since the `<SyncStatusPopover />` not gets it itself?

PR #1872, opened 2020-01-29
Remove print state and call print directly

belcherj:
### Fix
This PR calls the print function directly from app.tsx rather than setting up state. By not having to set editor/preview mode we decrease the complexity of this functionality while giving the user the option to print the editor view OR the markdown styled mode.  

Before this change, the print function was intended to only print the formatted markdown mode but the functionality is broken. This PR removes the broken code.

### Test
1. Open electron
2. From file menu click print
3. Observe it prints the note
4. Turn on markdown mode
5. Print
6. Observe that it prints formatted markdown

### Review

### DIFF
--- a/desktop/menus/file-menu.js
+++ b/desktop/menus/file-menu.js
@@ -28,7 +28,7 @@ const submenu = [
   {
     label: '&Print…',
     accelerator: 'CommandOrControl+P',
-    click: appCommandSender({ action: 'setShouldPrintNote' }),
+    click: appCommandSender({ action: 'printNote' }),
   },
 ];
 
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -30,7 +30,6 @@ import {
   pick,
   values,
 } from 'lodash';
-import { setEditorMode } from './state/ui/actions';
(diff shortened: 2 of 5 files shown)


dmsnell:
Actually we need not overcomplicate this: in testing in `develop` it only prints the current view regardless of the editor mode, which means that I think we can skip all the waiting routines and just use the most basic form:

```ts
if ('printNote' === command.action) {
  return window.print();
}
```

dmsnell:
> Before this change, the print function would only print the formatted markdown mode.

In my testing the behavior is the same before and after. It's possible that the same timing issues we ran into in this PR were already present in `develop` though. To the best of a quick search I found that people were resorting to strange means in order to figure out when React had finished updating the DOM.

belcherj:
I think removing the functionality and giving the user more choice is better all around!

belcherj:
The code that is being removed by this PR tried to switch the mode to markdown before printing. That code is broken.  This PR removes it.

dmsnell:
Formatted-only print removed in #1013
Printing introduced in 0eb023f5ed612a620df56774466a0cf5c8c25dae

PR #1874, opened 2020-01-30
Tools: Change icon for Export Notes from arrow to download icon

### LINKED ISSUE #1394: Exporting Notes: Save As dialog appears multiple times
When exporting your notes, the save dialog appears multiple times depending on how many times you click the export button:

export_issue

To repro:
1. On any OS (mac, Linux, Windows) open the Settings→Tools menu
2. Click the export button multiple times
3. Notice the "Save As" dialog appears multiple times

codebykat:
### Fix
To eliminate confusion about the Export Notes button triggering a file download, this PR changes the right pointing arrow icon to a download icon.

Fixes #1394 

### Test
1. Go to Settings -> Tools
2. Check out the new icon
3. Make sure you can still download the notes.zip

### Screenshots

**Before**
Screen Shot 2020-01-29 at 8 52 26 PM

**After**
Screen Shot 2020-01-29 at 8 46 53 PM

### Review
❓ Should we also change the wording? Right now import/export has a nice symmetry, but I wonder if "Download notes" would be clearer?

### Release
`RELEASE-NOTES.txt` was updated in 758b7d0fbd99c73cfff779a8129c546e45bdad18 with:
 
> Changed Export Notes icon in Settings to indicate that it's a file download [#1874](https://github.com/Automattic/simplenote-electron/pull/1874)

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -13,11 +13,9 @@
 
 ### Other Changes
 
-- Added types to state/ui/actions [#1849](https://github.com/Automattic/simplenote-electron/pull/1849)
-
-### Other Changes
-
 - Updated Dependencies [#1848](https://github.com/Automattic/simplenote-electron/pull/1848)
+- Added types to state/ui/actions [#1849](https://github.com/Automattic/simplenote-electron/pull/1849)
+- Changed Export Notes icon in Settings to indicate that it's a file download [#1874](https://github.com/Automattic/simplenote-electron/pull/1874)
 
 ## [v1.14.0]
 
--- a/lib/dialogs/button-group/index.tsx
+++ b/lib/dialogs/button-group/index.tsx
@@ -10,7 +10,7 @@ const ButtonGroup = props => {
         <li key={item.slug} className="button-group__item theme-color-border">
           <button type="button" onClick={() => onClickItem(item)}>
             {item.name}
-            <div class="icon">{item.icon ? <item.icon /> : '→'}</div>
+            <div className="icon">{item.icon ? <item.icon /> : '→'}</div>
           </button>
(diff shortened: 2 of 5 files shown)


SylvesterWilmott:
Taking a look at this again, I'm not sure if we're on the right track.

The arrow indicator is presumably there to represent going deeper into settings, which we do on import but we open a second modal which typically isn't the interaction we would expect from having an arrow/chevron. Export simply triggers a download so an arrow isn't needed there either.

>Should we also change the wording? Right now import/export has a nice symmetry, but I wonder if "Download notes" would be clearer?

I like the symmetry there, and I would also expect symmetrical icons, something representing in and something for out.

I think we should undo the icon change and leave it as it is (both with forward-facing arrows) because this change implies "icons that represent an action go on the right-hand side" which I'm not certain is what we want. The more I look at these settings, the more I feel it could do with a system of where to include action-representative icons and settings controls. We should include something like this in our UI project phase 2.

codebykat:
@SylvesterWilmott Yeah I don't love it either, but was trying to think of an easy way (before we redesign the whole UI) to set an expectation that clicking on "Export" triggers a file download instead of being a navigation link. Any other suggestions? We could put the icon next to the "Export Notes" and leave the right-pointing one on the right but that seems weird too... 🤔

SylvesterWilmott:
>was trying to think of an easy way (before we redesign the whole UI) to set an expectation that clicking on "Export" triggers a file download instead of being a navigation link

How about removing the arrow on the export option?

belcherj:
I have removed the web group as reviewers when you are ready for dev review please re-add.

codebykat:
Abandoning this for now, not high priority and we can revisit as part of UI refresh

PR #1876, opened 2020-01-31
Fix: Restore markdown preview button

dmsnell:
In #1851 we refactored the selected note state but during testing
missed a reference to `state.appState.note` which should have been
updated to `state.ui.note`. Nothing crashed in the app because it
was being used for its truthiness but the result was that when
checking if Markdown was enabled for any given note the answer
would always end up as "no."

In this patch we're restoring the state reference which fixes that
bug and restores the preview button in the note toolbar.


bummytime:
Thank you @dmsnell!

PR #1880, opened 2020-02-04
Fix: Update selected note when making changes in the editor view

dmsnell:
Resolves #1878
Resolves #1879
Resolves #1882 

It appears that we removed a flow in #1851 which updated app state in response
to updates from the note editor[1]. While this didn't impact the operation of
the editor it did present itself as a bug when previewing markdown-rendered
notes. The editor is stateful on its own and so continued to function
independently of the state of the selected note's content. The preview,
however, grabs the contents of the selected note whenever it's activated and
then renders those into a DOM node. This means that it showed the old contents
of the note unless you navigated to another note and back, which swapped out
the selected note with the new updated contents.

In this patch, which currently is probably wrong in some way, we're updating
the selected note when calling `onChangeContent` so that we don't get out of
sync with the data sources. Hopefully at the end of our state cleanup and
stabilization project this will disappear but for now this fix catches the
gap that left split data sources.

Probably there were other bugs we didn't discover as a result of this.

Mainly we need to test and audit for traces of cycles/loops in the update
so that we don't start spinning or overwriting data we just entered. In my
brief testing the easy cases worked as I expected them to simply and singly.

**In `develop`**
brokenPreview mov

**In this branch**
fixedPreview mov

Props to @surfer0627 for reporting

[1]: https://github.com/Automattic/simplenote-electron/pull/1851/files#diff-3c87750a7cc32e0127447ffc78f5c0d5L300

PR #1881, opened 2020-02-05
Move Search state to redux

belcherj:
### Fix
This moves the search state to Redux from flux.

- Removed [ref](https://github.com/Automattic/simplenote-electron/pull/1881/files#diff-911dea0a0678832102ff0b94927cddc8L225) prop that wasn't being used
- Renamed filter to searchQuery which better describes what it is and increases the ability to find instances of where it is used in the future.
- Removed props from components that only pass them through

### Test
1. Tests search very heavily.

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -192,7 +192,7 @@ const getRowHeight = rowHeightCache(computeRowHeight);
  * @see react-virtual/list
  *
  * @param {Object[]} notes list of filtered notes
- * @param {String} filter search filter
+ * @param {String} searchQuery search searchQuery
  * @param {String} noteDisplay list view style: comfy, condensed, expanded
  * @param {Number} selectedNoteId id of currently selected note
  * @param {Function} onSelectNote used to change the current note selection
@@ -202,7 +202,7 @@ const getRowHeight = rowHeightCache(computeRowHeight);
 const renderNote = (
   notes,
   {
-    filter,
+    searchQuery,
     noteDisplay,
     selectedNoteId,
     onNoteOpened,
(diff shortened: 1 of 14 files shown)


lib/state/ui/middleware.ts
```diff
@@ -30,9 +30,9 @@ export const middleware: Middleware<{}, S.State, Dispatch> = store => {
 
       // on updating the search field we should delay the update
       // so we don't waste our CPU time and lose responsiveness
-      case 'App.search':
+      case 'SEARCH':
         clearTimeout(searchTimeout);
-        if ('App.search' === action.type && !action.filter) {
+        if ('SEARCH' === action.type && !action.searchQuery) {
```
dmsnell: due to the type narrowing we don't need to re-verify the action type. it suffices to test `if (!action.searchQuery)`

dmsnell: Simplified in 7f0a867

lib/note-list/index.tsx
```diff
@@ -357,7 +356,7 @@ export class NoteList extends Component<Props> {
     }
 
     // Deselect the currently selected note if it doesn't match the search query
-    if (filter !== prevProps.filter) {
+    if (searchQuery !== prevProps.searchQuery) {
```
dmsnell: looks like this PR needs a rebase against the more recent changes to the `selectedNote`

dmsnell: I take it back. looks like we need to circle back here and make sure we don't need this any longer. right now it appears to update the `previousIndex` which eventually should be decoupled from things like the selected note IMO

dmsnell (changes requested):
Looks like we do need to rebase against `develop` to pull in changes from #1880 and then test this. I didn't test yet because of the rebase/update. I didn't want to test and catch the same bugs the other PRs have addressed

lib/note-list/index.tsx
belcherj: You are very right, this code is no longer needed and could have been the cause of some bugs in searching, good catch!

PR #1884, opened 2020-02-06
Types: Add some more types to boot and some utils

dmsnell:
Only real change should be some optional-chaining.
Discovered a bug in `resetAuth` that needs to be fixed outside of this branch.

## Testing

Should be no real code changes, no functional or visual changes.

PR #1885, opened 2020-02-06
Fix: Actually reset auth when authorizing without acess token

dmsnell:
Discovered via TypeScript in #1884

Previously when authorizing an account which yields no `access_token`
we have been attempting to reset the authorization state by dispatching
`resetAuth`. Unfortunately we haven't been dispatching the function
call but rather the function/action-creator itself leading to no
actionable dispatch.

In this patch we're fixing the call by calling `resetAuth()` which
then passes a real action to the dispatcher.

This should fix an unsurfaced but during invalid authentication flows.

## Testing

It would be good to recreate the initial login bug in order to reproduce this
and confirm the fix. We know that the code is wrong but I'm not sure _how_
that impacts the app's behavior or what defects surface because of it.

### DIFF
--- a/lib/boot.ts
+++ b/lib/boot.ts
@@ -169,7 +169,7 @@ let props = {
       .then(user => {
         resetStorageIfAccountChanged(username);
         if (!user.access_token) {
-          return store.dispatch(resetAuth);
+          return store.dispatch(resetAuth());
         }
 
         store.dispatch(setAccountName(username));
--- a/lib/boot.ts
+++ b/lib/boot.ts
@@ -135,7 +135,7 @@ let props = {
       .then(user => {
         resetStorageIfAccountChanged(username);
         if (!user.access_token) {
-          return store.dispatch(resetAuth);
+          return store.dispatch(resetAuth());
         }
 
         store.dispatch(setAccountName(username));


dmsnell:
> Could we modify the code, removing the bang on L140 to determine if the reset works?

Absolutely but I'm nervous about what that would actually tell us. I didn't have the time last night to investigate how we could get to that point without an `access_token`. I'd like to try and follow that flow, especially for the additional benefit of knowing why it exists and how we might eventually add an automated test for it.

dmsnell:
Thankfully I left some notes in #213 when I added the test for the missing `auth_token`. In examining Automattic/node-simperium it's clear that the call to authenticate always returns inside the `then()` even if the authentication fails.

Looks like I introduced the defect in #409 in switching to Redux but the bug didn't present itself to the app user because of the [guard added against blank login fields](https://github.com/Automattic/simplenote-electron/pull/213/files#diff-f041a427c603cb2a78907c18cf90d341R84) paired with the fact that we return in this situation vs. continuing along the auth process. That is, we have layers of protection in the auth code that prevented this defect from causing any visible problem.

I'm now confident in knowing that there is a reason we didn't see this cause any problems, that we can't test it because of the fact that the trigger for the bug is avoided, and that this is the right course of action to take in order to make the code right.

That leaves one final note: is this code necessary? Plausibly there could still be a reason an authentication attempt could fail to return an `auth_token` and if we're missing that we straight-up shouldn't continue. It makes sense to keep this in for the very same layered defense that protected us from the impact of this bug.

PR #1886, opened 2020-02-06
Types: Add types to <EditableList /> component

dmsnell:
As we changed some code here, specifically around initializing instance
variables, we need to be extra careful in testing to prevent regressions.

### DIFF
--- a/lib/editable-list/index.tsx
+++ b/lib/editable-list/index.tsx
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import SmallCrossOutlineIcon from '../icons/cross-outline-small';
 import ReorderIcon from '../icons/reorder';
@@ -31,10 +30,6 @@ export class EditableList extends Component<Props> {
   reorderingClientY: number = 0;
   reorderingOffsetY: number = 0;
   reorderingTranslateY: number = 0;
-
-  static propTypes = {
-    editing: PropTypes.bool.isRequired,
-  };
 
   static defaultProps = {
     getItemKey: (item: T.TagEntity) => item.id,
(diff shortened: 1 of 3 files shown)


lib/types.ts
```diff
@@ -68,3 +68,12 @@ export type ListDisplayMode = 'expanded' | 'comfy' | 'condensed';
 export type SortType = 'alphabetical' | 'creationDate' | 'modificationDate';
 export type Theme = 'system' | 'light' | 'dark';
 export type TranslatableString = string;
+
+///////////////////////////////////////
+// Language and Platform
+///////////////////////////////////////
+
+export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
```
belcherj: I feel like these need some explanation as to what they are.

lib/editable-list/index.tsx
```diff
@@ -5,22 +5,39 @@ import classNames from 'classnames';
...
   static displayName = 'EditableList';
 
+  reorderingElement: HTMLElement | null = null;
+  reorderingClientY: number = 0;
+  reorderingOffsetY: number = 0;
+  reorderingTranslateY: number = 0;
+
   static propTypes = {
-    className: PropTypes.string,
     editing: PropTypes.bool.isRequired,
```
belcherj: Why leave this?

dmsnell: huh. missed that!

lib/types.ts
dmsnell: can do and will add. I feel like `XOR` should be in the language. it's here because the `|` is only really useful when we have a property on which to discriminate alternative types.

```ts
type Colors = { name: 'red', value: 0xff000 } | { name: 'green', value: 0x00ff00 };
```

This fails if our types our different

```ts
type Size = { radius: number } | { width: number; height: number };
```

It _should_ be able to do this but it can't. When we don't have that property we have to rely on more advanced typing. In this case it's a mapped type called `XOR` which does exactly what `|` seems like it should do. `Without` is part of it and gets the difference of keys between them. Here I've used `XOR` because `MouseEvent` and `TouchEvent` don't have the discriminating property.

dmsnell:
Testing in batch with other PRs merging today.

PR #1888, opened 2020-02-07
Fix keyboard shortcuts on Windows Chrome

belcherj:
### Fix
Fixes the ctrl + shift + j/k keyboard shortcut on Windows Chrome and Windows Electron.

Adds the same logic to other shortcuts to ensure bugs of the same sort don't pop up:
- ctrl + shift + t - open tag list
- ctrl + shift + n - open note list
- ctrl + t - move between tag editor and note editor

### Test
1. In windows chrome
2. Click a note in the note list
3. Use ctrl + shift + j/k to navigate between notes
4. Use ctrl + shift + up/down arrow keys to navigate between notes
5. Use ctrl + shift + t to open close the tag list
6. In OSX Electron make Simplenote small so only note list or note is visible
7. Open a note
8. Use ctrl + shift + n to go back to note list
9. Open note
10. Use ctrl + t to move back and forth between tag editor and note editor

Note:
ctrl + shift + n and ctrl + t do not work in many places do to Chrome not allowing us to override the built in keyboard shortcuts. https://bugs.chromium.org/p/chromium/issues/detail?id=33056

### Release
- Fixes ctrl + shift + j/k keyboard shortcut on Windows [#1888](https://github.com/Automattic/simplenote-electron/pull/1888)

### DIFF
--- a/lib/note-editor/index.tsx
+++ b/lib/note-editor/index.tsx
@@ -84,7 +84,12 @@ export class NoteEditor extends Component<Props> {
     }
 
     // open note list - shift + n
-    if (this.props.isSmallScreen && cmdOrCtrl && shiftKey && 'n' === key) {
+    if (
+      this.props.isSmallScreen &&
+      cmdOrCtrl &&
+      shiftKey &&
+      'n' === key.toLowerCase()
+    ) {
       this.props.closeNote();
       this.props.onNoteClosed();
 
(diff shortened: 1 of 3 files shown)


lib/note-editor/index.tsx
```diff
@@ -84,7 +84,13 @@ export class NoteEditor extends Component<Props> {
     }
 
     // open note list - shift + n
-    if (this.props.isSmallScreen && cmdOrCtrl && shiftKey && 'n' === key) {
+    console.log(ctrlKey, key, metaKey, shiftKey);
```
dmsnell: oops: need to remove this

PR #1889, opened 2020-02-07
Add types and reduce function context

belcherj:
### Fix
We pass an object to a isDevConfig but it only uses the one property.  This changes the function to only accept that property and types it as a boolean. I updated the one place where the function is called. I also updated the tests.

### Test
1. Run Simplenote with console open.
2. In dev you should see: `Simperium config: Development`
3. In prod you should see: `Simperium config: Production`


belcherj:
Do we need this code? Is this helpful information? Could we combine it with the dev badge that gets added to the bottom of the note editor?

belcherj:
Agreed, I want to come back and delete the function all together.

PR #1894, opened 2020-02-11
Refactor getNoteTitles

belcherj:
### Fix
This PR refactors getNoteTiles to make the code a little cleaner, add types to
function, and removes lodash dependency.

### Test
1. Have unsynced note ids
2. Toggle the hamburger menu causing the popover to load

### DIFF
--- a/lib/components/sync-status/get-note-titles.ts
+++ b/lib/components/sync-status/get-note-titles.ts
@@ -8,7 +8,7 @@ type NoteTitle = {
 };
 
 const getNoteTitles = (
-  ids: String[] = [],
+  ids: T.EntityId[] = [],
   notes: T.NoteEntity[] = [],
   limit: number = Infinity
 ) => {
--- a/lib/components/sync-status/get-note-titles.ts
+++ b/lib/components/sync-status/get-note-titles.ts
@@ -9,10 +9,13 @@ type NoteTitle = {
 };
 
 const getNoteTitles = (
-  ids: T.EntityId[] = [],
-  notes: T.NoteEntity[] = [],
+  ids: T.EntityId[],
+  notes: T.NoteEntity[] | null,
   limit: number = Infinity
-) => {
+): NoteTitle[] => {
+  if (!notes) {
+    return [];
+  }
   const wantedIds = new Set(ids);
   const wantedNotes = filterAtMost(
     notes,
--- a/lib/components/sync-status/get-note-titles.ts
+++ b/lib/components/sync-status/get-note-titles.ts
@@ -12,8 +12,8 @@ const getNoteTitles = (
   notes: T.NoteEntity[] = [],
   limit: number = Infinity
 ) => {
-  const matchedNotes = ids.reduce((acc: NoteTitle[], id: String) => {
-    const note = notes.find((thisNote: T.NoteEntity) => thisNote.id === id);
+  const matchedNotes = ids.reduce((acc: NoteTitle[], id) => {
+    const note = notes.find(thisNote => thisNote.id === id);
     return note
       ? [...acc, { id, title: noteTitleAndPreview(note).title }]
       : acc;
--- a/lib/components/sync-status/get-note-titles.ts
+++ b/lib/components/sync-status/get-note-titles.ts
@@ -3,8 +3,8 @@ import noteTitleAndPreview from '../../utils/note-utils';
 import * as T from '../../types';
 
 type NoteTitle = {
-  id: String;
-  title: String;
+  id: T.EntityId;
+  title: string;
 };
 
 const getNoteTitles = (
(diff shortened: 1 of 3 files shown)


lib/components/sync-status/get-note-titles.ts
```diff
@@ -1,24 +1,25 @@
...
-    }
+type NoteTitle = {
+  id: String;
+  title: String;
+};
 
-    return { id, title: noteTitleAndPreview(note).title };
-  });
+const getNoteTitles = (
+  ids: String[] = [],
```
dmsnell: it's safe to use `T.EntityId` here since we're indexing the `notes` with it. `String` works because `EntityId` is an alias for `string` but we might as well leave that more specific.

Also note that in a confusing way `string` and `String` are different types. `String` should generally be reserved for places where we create `new String("something")` and the same goes for the other capitalized types.

lib/components/sync-status/get-note-titles.ts
```diff
@@ -1,24 +1,25 @@
...
+  notes: T.NoteEntity[] = [],
+  limit: number = Infinity
+) => {
```
dmsnell: This is a philosophical question more than an issue but we can add `NoteTitle` as the return type of this function.

On the plus side that enforces the contract such that it can prevent accidentally changing the output type. For instance, as-is type inference determines the return type. We could add `[0]` to the end of the function to return a single result and the inference would auto-update. In a fully-typed codebase we'd catch the errors, but here that will slip by.

If we say it must return `NoteTitle[]` then it will prevent us from unknowingly changing the return value

lib/components/sync-status/get-note-titles.ts
```diff
@@ -1,24 +1,25 @@
...
+  const matchedNotes = ids.reduce((acc: NoteTitle[], id: String) => {
+    const note = notes.find((thisNote: T.NoteEntity) => thisNote.id === id);
```
dmsnell: on the other side we can safely leave the annotations out of these two lines for `id` and `thisNote` because they follow from the types of `ids` and `notes`. if I leave out those annotations and hover over the identifiers my IDE shows the right types.

```ts
const matchedNotes = ids.reduce((acc: NoteTitle[], id) => {
  const note = notes.find(thisNote => thisNote.id === id);
  return note
    ? [ ...acc, { id, title: noteTitleAndPreview(note).title } ]
    : acc;
}, []);
```

lib/components/sync-status/get-note-titles.ts
```diff
@@ -1,24 +1,25 @@
...
+    return note
+      ? [...acc, { id, title: noteTitleAndPreview(note).title }]
+      : acc;
+  }, []);
 
-  return compact(matchedNotes);
+  return matchedNotes.slice(0, limit);
```
dmsnell: It looks like we're doing two things in this PR: (a) fixing a bug due to not checking for `null`; and (b) refactoring the `getNoteTitles` function.

If we want to focus on the bug it might be easier to do that first and then come back for a refactor so that we're not fixing a defect _and_ changing behaviors in the same patch (as we're doing with the move from `null` + `compact()` to the `reducer()`)

If we want to go ahead and refactor we might consider a variation of the function that doesn't involve the `reduce()`. Inside that reducer we're iterating across notes and filtering and mapping all at once, and it will get progressively slower as we get more notes because of the loop-inside-a-loop. We could separate the duties and do one quick setup to remove the inner loop. I'm borrowing `filterAtMost()` from `tag-suggestions/index` but it's comparable to `.filter().slice()` and just skips processing any more after we get our limit.

```ts
const getNoteTitles = (
  ids: T.EntityId[] = [],
  notes: T.NoteEntity[] = [],
  limit: number = Infinity
): NoteTitle[] => {
  const wantedIds = new Set(ids);
  const wantedNotes = filterAtMost(notes, ({ id }) => wantedIds.has(id), limit);

  return wantedNotes.map(note => ({
    id: note.id,
    title: noteTitleAndPreview(note).title,
  }));
};
```

Creating the `Set()` turns our list-scan into a single lookup per id and should only incur the cost of the first iteration through the `notes` array. If we have two or more `notes` then the cost of doing that should be offset (if we only have one then it doesn't really matter anyway).

`filterAtMost()` saves on an intermediate data structure and only iterates across ten `notes` if that's the limit even if we have 10,000 notes in the account.

lib/components/sync-status/get-note-titles.ts
belcherj: Duh, I shoulda know that.

Also, was planning to inquire about `string` and `String`

belcherj: Fixed in fffcc43

lib/components/sync-status/get-note-titles.ts
belcherj: Fixed in: 95ea31d

lib/components/sync-status/get-note-titles.ts
belcherj: I agree with that, resolved in a0f831f

lib/components/sync-status/get-note-titles.ts
belcherj: Created another PR for Fix here: https://github.com/Automattic/simplenote-electron/pull/1898

belcherj: Changed in: febd61e

I very much dislike the added complexity of filterAtMost but agreed that it would be more performance.

PR #1895, opened 2020-02-11
Refactors closeNote to Redux

belcherj:
### Fix

Refactors `closeNote` moving the function from the flux actions to Redux. In doing this it required the removal of the note list open/closed that was being stored in the component state of app.tsx.

### Test
1. Have SImplenote open with a small width so only the note list appears
2. Click a note
3. Click the back button at the top
4. Click note again
5. Trash it, does note list appear?
6. Open Trashed note list
7. Click into note
8. Click back button at top
9. Permanently trash note, should go back to trashed note list
10. Empty trash, should see blank note editor
11. Go back to all notes
12. Create new note
13. Play with search

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -32,6 +32,7 @@ import {
 } from 'lodash';
 import {
   createNote,
+  closeNote,
   setUnsyncedNoteIds,
   toggleSimperiumConnectionStatus,
 } from './state/ui/actions';
@@ -50,6 +51,7 @@ export type OwnProps = {
 
 export type DispatchProps = {
   createNote: () => any;
+  closeNote: () => any;
   selectNote: (note: T.NoteEntity) => any;
 };
 
@@ -95,6 +97,7 @@ const mapDispatchToProps: S.MapDispatch<
       ]),
       dispatch
     ),
+    closeNote: () => dispatch(closeNote()),
     loadTags: () => dispatch(loadTags()),
     setSortType: thenReloadNotes(settingsActions.setSortType),
     toggleSortOrder: thenReloadNotes(settingsActions.toggleSortOrder),
@@ -158,10 +161,6 @@ export const App = connect(
       onSignOut: () => {},
     };
 
-    state = {
-      isNoteOpen: false,
-    };
-
     UNSAFE_componentWillMount() {
       if (isElectron()) {
         this.initializeElectron();
@@ -217,24 +216,10 @@ export const App = connect(
     }
 
     componentDidUpdate(prevProps) {
-      const { settings, isSmallScreen } = this.props;
+      const { settings } = this.props;
 
       if (settings !== prevProps.settings) {
         ipc.send('settingsUpdate', settings);
-      }
-
-      // If note has just been loaded
-      if (prevProps.appState.note === undefined && this.props.appState.note) {
-        this.setState({ isNoteOpen: true });
-      }
-
-      if (isSmallScreen !== prevProps.isSmallScreen) {
-        this.setState({
-          isNoteOpen: Boolean(
-            this.props.appState.note &&
-              (settings.focusModeEnabled || !isSmallScreen)
-          ),
-        });
       }
     }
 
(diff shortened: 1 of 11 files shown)


lib/state/ui/reducer.ts
```diff
@@ -52,24 +51,48 @@ const simperiumConnected: A.Reducer<boolean> = (state = false, action) =>
...
+        newSet.delete('noteList');
+        return newSet;
+      }
+    case 'TAG_DRAWER_TOGGLE': {
+      if (action.show) {
+        return new Set(state).add('tagDrawer');
+      } else {
+        const newSet = new Set(state);
+        newSet.delete('tagDrawer');
+        return newSet;
```
dmsnell: as noted in Slack, we're going to remove `visiblePanes` right?

belcherj: yes, I hadn't updated yet.

lib/note-toolbar/index.tsx
```diff
@@ -180,10 +180,10 @@ const mapStateToProps: S.MapState<StateProps> = ({
...
-const mapDispatchToProps: S.MapDispatch<DispatchProps> = dispatch => ({
-  closeNote: () => dispatch(closeNote()),
-  toggleEditMode: () => dispatch(toggleEditMode()),
-  toggleNoteInfo: () => dispatch(toggleNoteInfo()),
-});
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
+  closeNote,
+  toggleEditMode,
+  toggleNoteInfo,
+};
```
dmsnell: 😎

lib/app.tsx
```diff
@@ -217,25 +216,11 @@ export const App = connect(
...
-      }
-
-      if (isSmallScreen !== prevProps.isSmallScreen) {
-        this.setState({
-          isNoteOpen: Boolean(
-            this.props.appState.note &&
-              (settings.focusModeEnabled || !isSmallScreen)
-          ),
-        });
-      }
```
dmsnell: Earmarking to test…

lib/state/ui/reducer.ts
```diff
@@ -39,6 +37,19 @@ const listTitle: A.Reducer<T.TranslatableString> = (
...
+  switch (action.type) {
+    case 'CLOSE_NOTE': {
+      return true;
+    }
+    case 'App.selectNote':
+      return false;
+
+    default:
+      return state;
+  }
```
dmsnell: For actions like `deleteNoteForever` and `restoreNote` and the like is this now handled automatically by some other code, that the note closes?

lib/state/ui/reducer.ts
```diff
@@ -52,19 +63,6 @@ const simperiumConnected: A.Reducer<boolean> = (state = false, action) =>
...
-  action
-) => {
-  if ('TAG_DRAWER_TOGGLE' === action.type) {
-    return action.show
-      ? union(state, ['tagDrawer'])
-      : difference(state, ['tagDrawer']);
-  }
-
-  return state;
-};
```
dmsnell: Thanks for removing this. Probably would have been nice to do it in a separate PR with its own traceable history, or at least a mention in the `git` commit, since now we don't have any explanation for why this was removed in the code or `git` history - it just disappears.

lib/flux/app-state.ts
```diff
@@ -395,12 +389,10 @@ export const actionMap = new ActionMap({
         note: T.NoteEntity;
         previousIndex: number;
       }) {
-        return dispatch => {
+        return () => {
           if (note) {
             note.data.deleted = true;
             noteBucket.update(note.id, note.data);
-
-            dispatch(this.action('closeNote', { previousIndex }));
```
dmsnell: Are we removing the functionality that the `previousIndex` provided?

dmsnell: Looks like this was broken in #1851 so it won't appear in testing of this PR

lib/state/ui/reducer.ts
dmsnell: Noted in testing: the note closes automatically due to the filtering middleware which implicitly unselects it _as it no longer exists_.

PR #1896, opened 2020-02-12
State Refactor: Move note info into Redux

codebykat:
### Fix
This moves the `showNoteInfo` boolean and associated toggles into Redux state.

### Test
1. Click the note details pane and make sure note details is shown
2. Click the X or outside the pane and make sure it toggles closed

### Review
Code review: Please check to make sure I did this right and didn't miss anything! I used #1881 as an example.

I moved `showNoteInfo` from `appState.showNoteInfo` to `state.ui.showNoteInfo`, added new actions and a new reducer, and removed the legacy action.

In the reducer I am definitely eyeing `VisiblePanes` as potentially a more logical place to put this (along with the Navigation, see below) but this PR seemed complicated enough already so I decided to stick with a statewide boolean for now.

Another thing of note is that the former incarnation of this action also sets `showNavigation` and `editingTags` to `false` when the notes panel is shown. I was unable to figure out what edge cases that was meant to handle, so I'm not sure if it is still necessary, and it seems it will be somewhat complicated without growing the PR to also move those two state variables into Redux. (See my line comments on the diff below.)

### DIFF
--- a/lib/flux/app-state.ts
+++ b/lib/flux/app-state.ts
@@ -39,7 +39,6 @@ const initialState: AppState = {
   revision: null,
   showTrash: false,
   showNavigation: false,
-  showNoteInfo: false,
   isViewingRevisions: false,
   editingTags: false,
   dialogs: [],
@@ -70,7 +69,6 @@ export const actionMap = new ActionMap({
 
       return update(state, {
         showNavigation: { $set: true },
-        showNoteInfo: { $set: false },
       });
     },
 
@@ -487,20 +485,6 @@ export const actionMap = new ActionMap({
     noteRevisionsLoaded(state: AppState, { revisions }) {
       return update(state, {
         revisions: { $set: revisions },
-      });
-    },
-
-    toggleNoteInfo(state: AppState) {
-      if (state.showNoteInfo) {
-        return update(state, {
-          showNoteInfo: { $set: false },
-        });
-      }
-
-      return update(state, {
-        showNoteInfo: { $set: true },
-        showNavigation: { $set: false },
-        editingTags: { $set: false },
       });
     },
 
--- a/lib/note-info/index.tsx
+++ b/lib/note-info/index.tsx
@@ -1,17 +1,23 @@
 import React, { Component } from 'react';
+import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import onClickOutside from 'react-onclickoutside';
 import { includes, isEmpty } from 'lodash';
+import format from 'date-fns/format';
+
+import appState from '../flux/app-state';
 import PanelTitle from '../components/panel-title';
 import ToggleControl from '../controls/toggle';
-import format from 'date-fns/format';
 import CrossIcon from '../icons/cross';
-import { connect } from 'react-redux';
-import appState from '../flux/app-state';
-import { setMarkdown } from '../state/settings/actions';
+
+import actions from '../state/actions';
 
 import * as S from '../state';
 import * as T from '../types';
+
+type OwnProps = {
+  noteBucket: T.Bucket<T.Note>;
+};
 
 type StateProps = {
   isMarkdown: boolean;
@@ -19,9 +25,15 @@ type StateProps = {
   note: T.NoteEntity | null;
 };
 
-type Props = StateProps;
+type DispatchProps = {
+  onOutsideClick: () => any;
+};
+
+type Props = OwnProps & StateProps & DispatchProps;
 
 export class NoteInfo extends Component<Props> {
+  static displayName = 'NoteInfo';
+
   static propTypes = {
     markdownEnabled: PropTypes.bool,
     onPinNote: PropTypes.func.isRequired,
@@ -194,7 +206,7 @@ function characterCount(content) {
   );
 }
 
-const { markdownNote, pinNote, toggleNoteInfo } = appState.actionCreators;
+const { markdownNote, pinNote } = appState.actionCreators;
 
 const mapStateToProps: S.MapState<StateProps> = ({ ui: { note } }) => ({
   note,
@@ -202,13 +214,16 @@ const mapStateToProps: S.MapState<StateProps> = ({ ui: { note } }) => ({
   isPinned: !!note && note.data.systemTags.includes('pinned'),
 });
 
-const mapDispatchToProps = (dispatch, { noteBucket }) => ({
+const mapDispatchToProps: S.MapDispatch<DispatchProps, OwnProps> = (
+  dispatch,
+  { noteBucket }
+) => ({
   onMarkdownNote: (note, markdown = true) => {
     dispatch(markdownNote({ markdown, note, noteBucket }));
     // Update global setting to set markdown flag for new notes
-    dispatch(setMarkdown(markdown));
+    dispatch(actions.settings.setMarkdown(markdown));
   },
-  onOutsideClick: () => dispatch(toggleNoteInfo()),
+  onOutsideClick: () => dispatch(actions.ui.toggleNoteInfo()),
   onPinNote: (note, pin) => dispatch(pinNote({ noteBucket, note, pin })),
 });
 
--- a/lib/note-toolbar/index.tsx
+++ b/lib/note-toolbar/index.tsx
@@ -12,13 +12,16 @@ import RevisionsIcon from '../icons/revisions';
 import TrashIcon from '../icons/trash';
 import ShareIcon from '../icons/share';
 import SidebarIcon from '../icons/sidebar';
+
 import { toggleEditMode } from '../state/ui/actions';
+import { toggleNoteInfo } from '../state/ui/actions';
 
 import * as S from '../state';
 import * as T from '../types';
 
 type DispatchProps = {
   toggleEditMode: () => any;
+  toggleNoteInfo: () => any;
 };
 
 type StateProps = {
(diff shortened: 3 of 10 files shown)


lib/flux/app-state.ts
```diff
@@ -74,7 +73,7 @@ export const actionMap = new ActionMap({
 
       return update(state, {
         showNavigation: { $set: true },
-        showNoteInfo: { $set: false },
+        showNoteInfo: { $set: false }, // @todo?
```
codebykat: I think this is safe to remove (how do you activate the navigation panel without also triggering the note info "click outside" action?) but if not this needs to figure out how to set state.ui.showNoteInfo

lib/state/ui/reducer.ts
```diff
@@ -34,6 +34,21 @@ const visiblePanes: A.Reducer<string[]> = (
...
+    return action.show;
+
+    // @todo?
+    // originally if note info toggles in,
+    // showNavigation and editingTags were set to false
+    //     showNoteInfo: { $set: true },
+    //     showNavigation: { $set: false },
+    //     editingTags: { $set: false },
+  }
+
```
codebykat: Again, I believe we don't need to handle these cases explicitly, but I hope someone can verify that

lib/state/ui/reducer.ts
```diff
@@ -34,6 +34,21 @@ const visiblePanes: A.Reducer<string[]> = (
   return state;
 };
 
+const showNoteInfo: A.Reducer<boolean> = (state = false, action) => {
+  if ('NOTE_INFO_TOGGLE' === action.type) {
+    return action.show;
```
dmsnell: considering that our only user-initiated actions seem to be _toggling_ and our action creator is even named `toggleNoteInfo` we can probably simplify this by arranging our state into a toggle paradigm and skip the associated boolean operator.

```
const showNoteInfo: A.Reducer<boolean> = (state = false, action) => {
  switch (action.type) {
    case 'NOTE_INFO_TOGGLE':
      return !state;

    case 'App.toggleNavigation': // or whatever
      return false;

    default:
      return state;
  }
}
```

It's in this manner that we can move from asking "what data does this change?" as seen in `app-state.ts` to "what changes this data?" where the reducers only need to govern their own type. if we run into problems with things like the fact that `toggleNavigation` might turn _on_ the note info panel then we just need to be creative, but as in many places so in this PR, I think that it's safe to say that `toggleNavigation` should _always_ disable the note info panel because _either_ it's disabling it while opening navigation _or_ it would have already been disabled when closing navigation

lib/note-toolbar/index.tsx
```diff
@@ -184,4 +188,8 @@ const mapStateToProps: S.MapState<StateProps> = ({ ui: { note } }) => ({
   note,
 });
 
-export default connect(mapStateToProps)(NoteToolbar);
+const mapDispatchToProps = (dispatch, {}) => ({
+  onToggleNoteInfo: () => dispatch(toggleNoteInfo(true)),
+});
```
dmsnell: if we remove the `show` parameter as discussed in the `reducer` file below then this simplifies

```ts
  onToggleNoteInfo: toggleNoteInfo,
}
```

lib/note-toolbar/index.tsx
```diff
@@ -134,7 +138,7 @@ export class NoteToolbar extends Component<Props> {
           <div className="note-toolbar__button">
             <IconButton
               icon={<InfoIcon />}
-              onClick={this.props.onShowNoteInfo}
+              onClick={() => onToggleNoteInfo()}
```
dmsnell: nitpick: we don't need to create a new function here to wrap the existing one. we can pass `onToggleNoteInfo` directly into it and that will save us down the line (since as-is here we're creating a new function on every call to `render()`)

```ts
onClick={onToggleNoteInfo}
```

note that in some cases this isn't possible because the calling function supplies arguments we don't want or expect. in those cases we need to manually prevent that by creating the wrapper. technically you've done that in `mapDispatchToProps` when you wrote `() => dispatch(toggleNoteInfo(true))` because any arguments passed by `onClick` would be ignored there.

lib/note-info/index.tsx
```diff
@@ -208,7 +212,7 @@ const mapDispatchToProps = (dispatch, { noteBucket }) => ({
     // Update global setting to set markdown flag for new notes
     dispatch(setMarkdown(markdown));
   },
-  onOutsideClick: () => dispatch(toggleNoteInfo()),
+  onOutsideClick: () => dispatch(toggleNoteInfo(false)),
```
dmsnell: see below note about `show` and this action

lib/flux/app-state.ts
```diff
@@ -74,7 +73,7 @@ export const actionMap = new ActionMap({
 
       return update(state, {
         showNavigation: { $set: true },
-        showNoteInfo: { $set: false },
+        //showNoteInfo: { $set: false }, // @todo?
```
dmsnell: as we'll capture this in the reducer when listening for `App.toggleNavigation` we can go ahead and remove it here

dmsnell:
Lookin' good @codebykat - I hope you found the examples and documentation helpful. Left a few notes which I think might make your job easier. If not, don't hesitate to reach out and challenge or ask.

lib/state/ui/reducer.ts
codebykat: Yeah, I thought about that. Do we need to worry about race conditions (which is usually the only reason to be explicit about passing something to a toggle)?

It's also sort of nice that it's very explicit -- the "open" action passes `true` and the "close" action passes `false` -- but not really necessary.

dmsnell: > Do we need to worry about race conditions

The nice thing about simple reducers vs. imperative updates or _Redux thunks_ is that there are no race conditions. Each reducer updates its own data only in response to the actions initiated by the user _or_ by some network activity or timer event.

For the race to occur we'd need to be able to dispatch and update state after _some_ reducers have updated but before _this_ reducer updates, which isn't possible.

belcherj:
Looks great, I didn't see anything outside of what Dennis came up with!

belcherj:
Does look like you have a linting error. https://circleci.com/gh/Automattic/simplenote-electron/9628?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

codebykat:
Cool, that all makes sense! Fixed some nits and up for review again. Thanks @dmsnell :)

One thing is I couldn't figure out how to remove the dispatch argument from MapDispatchToProps in `note-info` ([here](https://github.com/Automattic/simplenote-electron/pull/1896/commits/a4ebbbd48349e139843bf1785b62d563b3f27c1f#diff-18a172cafe49ad20599008c9b8ce1733L209)):

Screen Shot 2020-02-12 at 10 07 21 AM

It works in `note-toolbar` ([here](https://github.com/Automattic/simplenote-electron/pull/1896/commits/a4ebbbd48349e139843bf1785b62d563b3f27c1f#diff-e98743febf9f62311da7008a480dad44L191)) and I'm not sure what the difference is:

Screen Shot 2020-02-12 at 10 05 54 AM

There is some other stuff that needs to be moved to Redux in `note-detail`, and I don't think I understand the `onOutsideClick` situation, so maybe that's why?

lib/note-info/index.tsx
```diff
@@ -19,9 +22,15 @@ type StateProps = {
   note: T.NoteEntity | null;
 };
 
-type Props = StateProps;
+type DispatchProps = {
+  onToggleNoteInfo: () => any;
```
dmsnell: We've added this type but then our contract breaks in `mapDispatchToProps` (and I'm seeing that error in my IDE) because the component as written expects and is provided `onOutsideClick`

We need to _either_ rename this here to `onOutsideClick` _or_ rename all of the `onOutsideClick` references to `onToggleNoteInfo`

lib/note-toolbar/index.tsx
```diff
@@ -184,4 +188,8 @@ const mapStateToProps: S.MapState<StateProps> = ({ ui: { note } }) => ({
   note,
 });
 
-export default connect(mapStateToProps)(NoteToolbar);
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
+  onToggleNoteInfo: toggleNoteInfo,
```
dmsnell: Since we're creating the interface and the names let's abandon `on_` nomenclature - stick to the active voice and indicate what we want, not _when_ a thing happens.

```ts
  toggleNoteInfo
}
```

lib/state/ui/reducer.ts
dmsnell: we can verify it!

dmsnell: in testing I can see that it's more of a choice to _change_ the behavior or keep it. if we take this out then the note info stays open when we flip the navigation panel.

codebykat:
Rebased and tested both this and the edit mode toggle (which was entangled).

lib/state/ui/reducer.ts
codebykat: I haven't been able to make that happen, what are the steps? If I open note info, there's no way to open the navigation panel (any click is an "outside click" that closes the note info).

codebykat: From Slack: "Use the keyboard shortcut Cmd + t and flip the tag drawer/nav panel open and closed"

I tried this and it does seem to close the info panel, so I'm gonna say.. it's fine.

dmsnell: sorry. I should have clarified that - obviously this code closes the note info panel when toggling the navigation panel. _however_, when I removed the `case 'App.toggleNavigation'` as it seemed like you were suggesting would be possible, then the note info would stay open on navigation.

That is, this is an answer to the question "can we skip handling these cases explicitly" and that answer is "if we want and decide to."

PR #1898, opened 2020-02-12
Fixes error where having unsynced note ids but no notes can crash the app

belcherj:
### Fix
When you have 0 notes and have unsynced changes you can cause the function to
crash the app. In my testing, I was not able to create these test conditions.
You will see in the modified code that we ensure that we at very least have an
empty array before attempting array functions. This eliminates the existing
bug.

### Test
1. Have no notes in state but have unsynced note ids
2. Toggle the hamburger menu causing the popover to load

### Release
- Fixes error where having unsynced note ids but no notes can crash the app

### DIFF
--- a/lib/components/sync-status/get-note-titles.ts
+++ b/lib/components/sync-status/get-note-titles.ts
@@ -1,7 +1,7 @@
 import { compact } from 'lodash';
 import noteTitleAndPreview from '../../utils/note-utils';
 
-const getNoteTitles = (ids, notes, limit = Infinity) => {
+const getNoteTitles = (ids = [], notes = [], limit = Infinity) => {
   const matchedNotes = ids.map((id, i) => {
     if (i >= limit) {
       return;


lib/components/sync-status/get-note-titles.ts
```diff
@@ -1,7 +1,7 @@
 import { compact } from 'lodash';
 import noteTitleAndPreview from '../../utils/note-utils';
 
-const getNoteTitles = (ids, notes, limit = Infinity) => {
+const getNoteTitles = (ids = [], notes = [], limit = Infinity) => {
```
dmsnell: Can we avoid having multiple optional arguments following each other? This opens up a very odd situation where it looks like we can skip `ids` or `notes` when in reality it would just mush them together and break.

```ts
getNoteTitles(10); // ?
getNoteTitles([{id, data, version}, {id, data, version}]); // ?
```

it seems fair to enforce `ids` as `T.EntityId[]` and/or define `notes` to be `T.NoteEntity[]`. if we demand `notes: T.NoteEntity[]` though we should conditionally call this function. since `notes` in app state can be null it also makes sense to leave that open here…

```ts
const getNoteTitles = (ids: T.EntityId[], notes: T.NoteEntity[] | null, limit = Infinity) => {
  if (!notes) {
    return [];
  }

  …
}
```

belcherj: Probably something to handle over here: https://github.com/Automattic/simplenote-electron/pull/1894

belcherj: Handled in this commit: https://github.com/Automattic/simplenote-electron/pull/1894/commits/03b84ecc9a60b438a40a6075704724e77703e7a1

belcherj:
This PR is no longer needed, fixed in #1894

PR #1901, opened 2020-02-12
Move showTrash from appState to Redux

belcherj:
### Fix
Requires: https://github.com/Automattic/simplenote-electron/pull/1866 for `CREATE_NOTE`

### Test
1. Open menu
2. Click trash
3. Restore note
4. Delete note
5. Trash all notes
6. Click 'All Notes'
7. Try to click new note button but realize it is disabled

We do need `CREATE_NOTE` since you can create a new note from the system menu in Electron

### DIFF
--- a/lib/state/action-types.ts
+++ b/lib/state/action-types.ts
@@ -69,6 +69,7 @@ export type ToggleEditMode = Action<'TOGGLE_EDIT_MODE'>;
 export type ToggleRevisions = Action<'REVISIONS_TOGGLE'>;
 export type ToggleTagDrawer = Action<'TAG_DRAWER_TOGGLE', { show: boolean }>;
 export type ToggleTagEditing = Action<'TAG_EDITING_TOGGLE'>;
+export type ToggleTrash = Action<'TRASH_TOGGLE'>;
 export type SelectNote = Action<'SELECT_NOTE', { note: T.NoteEntity }>;
 
 export type ActionType =
@@ -98,7 +99,8 @@ export type ActionType =
   | ToggleRevisions
   | ToggleSimperiumConnectionStatus
   | ToggleTagDrawer
-  | ToggleTagEditing;
+  | ToggleTagEditing
+  | ToggleTrash;
 
 export type ActionCreator<A extends ActionType> = (...args: any[]) => A;
 export type Reducer<S> = (state: S | undefined, action: ActionType) => S;
--- a/lib/state/action-types.ts
+++ b/lib/state/action-types.ts
@@ -69,7 +69,6 @@ export type ToggleEditMode = Action<'TOGGLE_EDIT_MODE'>;
 export type ToggleRevisions = Action<'REVISIONS_TOGGLE'>;
 export type ToggleTagDrawer = Action<'TAG_DRAWER_TOGGLE', { show: boolean }>;
 export type ToggleTagEditing = Action<'TAG_EDITING_TOGGLE'>;
-export type ToggleTrash = Action<'TRASH_TOGGLE'>;
 export type SelectNote = Action<'SELECT_NOTE', { note: T.NoteEntity }>;
 
 export type ActionType =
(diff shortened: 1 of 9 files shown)


lib/state/ui/reducer.ts
```diff
@@ -21,17 +20,41 @@ const simperiumConnected: A.Reducer<boolean> = (state = false, action) =>
...
   action
 ) => {
-  if ('TAG_DRAWER_TOGGLE' === action.type) {
-    return action.show
-      ? union(state, ['tagDrawer'])
-      : difference(state, ['tagDrawer']);
+  switch (action.type) {
+    case 'App.selectTrash': {
+      return new Set(state).add('trash');
+    }
```
dmsnell: we only need the curlies if we want to create a new scope/if we want to create variables. they won't be necessary here.

lib/state/ui/reducer.ts
```diff
@@ -21,17 +20,41 @@ const simperiumConnected: A.Reducer<boolean> = (state = false, action) =>
...
+      if (action.show) {
+        return new Set(state).add('tagDrawer');
+      } else {
+        const newSet = new Set(state);
+        newSet.delete('tagDrawer');
+        return newSet;
+      }
+    }
+    default:
+      return state;
```
dmsnell: How do you feel about moving the inclusion/exclusion operations into a callbable function? besides becoming symmetrical we can eliminate the need to manually create new scopes and temporary variables…

```ts
case 'TRASH_TOGGLE':
  return toggle(state, 'trash');

case 'TAG_DRAWER_TOGGLE':
  return action.show
    ? include(state, 'tagDrawer' )
    : exclude(state, 'tagDrawer' );
```

dmsnell (commented):
Trash is an odd view. It's not globally boolean, since obviously the tag drawer needs to be open in order to see it. It's not a tag even though it acts like one. It's not really a pane so it's surprising to see it in `visiblePanes`

Did you consider putting it anywhere else? It's somewhat of a global mode since it governs the buttons too. What do you think would be the tradeoffs of keeping it as a global boolean?

lib/state/ui/reducer.ts
```diff
@@ -52,17 +50,20 @@ const simperiumConnected: A.Reducer<boolean> = (state = false, action) =>
...
 
-const visiblePanes: A.Reducer<string[]> = (
-  state = defaultVisiblePanes,
-  action
-) => {
-  if ('TAG_DRAWER_TOGGLE' === action.type) {
-    return action.show
-      ? union(state, ['tagDrawer'])
-      : difference(state, ['tagDrawer']);
+const trash: A.Reducer<boolean> = (state = false, action) => {
```
dmsnell: can we call this something other than `trash`? maybe even keep `showTrash` or `viewingTrash` or `trashOpen`?

I find `trash` (especially in `mapStateToProps`) to be a wee bit confusing

belcherj: done

lib/state/action-types.ts
```diff
@@ -67,6 +67,7 @@ export type ToggleSimperiumConnectionStatus = Action<
 >;
 export type ToggleEditMode = Action<'TOGGLE_EDIT_MODE'>;
 export type ToggleTagDrawer = Action<'TAG_DRAWER_TOGGLE', { show: boolean }>;
+export type ToggleTrash = Action<'TRASH_TOGGLE'>;
```
dmsnell: This doesn't matter really, but it's odd that everywhere else we use `toggle` and then `trash` but here we use for the action name `trash` then `toggle`.

Calypso has some standards setup for organization that push the noun/receiver first but that was also in part because of a lack of a type system to prevent overlap of names. Here we have fewer actions plus we have a type system so if there's a chance we feel obliged to hold Calypso style I wouldn't worry about it.

belcherj: I deleted this code in the next commit.  Yes, we should have a standard.

PR #1902, opened 2020-02-12
Move list title to redux

belcherj:
### Fix
Move list title to redux. The list title is used as the placeholder for the search field.

### Test
1. Should be 'All Notes' when page loads
2. Open trash, should be 'Trash'
3. Select tag, should be the tags name.

### DIFF
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -11,14 +11,17 @@ const filteredNotes: A.Reducer<T.NoteEntity[]> = (
   action
 ) => ('FILTER_NOTES' === action.type ? action.notes : state);
 
-const listTitle: A.Reducer<string> = (state = 'All Notes', action) => {
+const listTitle: A.Reducer<T.TranslatableString> = (
+  state = 'All Notes',
+  action
+) => {
   switch (action.type) {
     case 'App.showAllNotes':
       return 'All Notes';
     case 'App.selectTrash':
       return 'Trash';
     case 'App.selectTag':
-      return action?.tag?.data?.name;
+      return action.tag.data.name;
     default:
       return state;
   }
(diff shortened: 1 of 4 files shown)


lib/state/ui/reducer.ts
```diff
@@ -11,6 +11,19 @@ const filteredNotes: A.Reducer<T.NoteEntity[]> = (
   action
 ) => ('FILTER_NOTES' === action.type ? action.notes : state);
 
+const listTitle: A.Reducer<string> = (state = 'All Notes', action) => {
```
dmsnell: We can ignore it, but I used `T.TranslatableString` to start keeping track of things we'll need to translate later. Might be handy to keep around

lib/state/ui/reducer.ts
```diff
@@ -11,6 +11,19 @@ const filteredNotes: A.Reducer<T.NoteEntity[]> = (
...
+  switch (action.type) {
+    case 'App.showAllNotes':
+      return 'All Notes';
+    case 'App.selectTrash':
+      return 'Trash';
+    case 'App.selectTag':
+      return action?.tag?.data?.name;
```
dmsnell: Can this be `undefined`/`null`? I did some searching and couldn't find anywhere that it's called without a `T.TagEntity`

dmsnell (commented):
Nice update. I like how this pulls together the places that impact the placeholder. Mentioning the placeholder it earmarks some future work where maybe this should be less a string value and more a mode that gets translated (eventually) in `render()`

This might very well be a key part to understanding the role of `trash` and fit into state-bound URL paths.

Except for the optional-chaining which I didn't think was necessary you have my ✅

lib/state/ui/reducer.ts
belcherj: I guess now that I have tools I just started using them.  You are right I went too far here.

lib/state/ui/reducer.ts
belcherj: Done

lib/state/ui/reducer.ts
belcherj: Done

PR #1903, opened 2020-02-12
Move viewing revisions logic to redux

belcherj:
### Fix
Move viewing revisions logic to redux.

### Test
1. open revisions
2. cancel revisions
3. restore revisions
4. when revisions panel is open click outside of the revisions panel

### DIFF
--- a/lib/note-toolbar/index.tsx
+++ b/lib/note-toolbar/index.tsx
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { connect } from 'react-redux';
+import { connect, DispatchProp } from 'react-redux';
 import PropTypes from 'prop-types';
 import { noop } from 'lodash';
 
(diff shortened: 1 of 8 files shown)


lib/note-toolbar/index.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { connect } from 'react-redux';
+import { connect, DispatchProp } from 'react-redux';
```
dmsnell: This inclusion was probably unintended.

lib/state/ui/reducer.ts
```diff
@@ -86,7 +86,15 @@ const showNoteInfo: A.Reducer<boolean> = (state = false, action) => {
 
     case 'App.toggleNavigation':
       return false;
+    default:
+      return state;
+  }
```
dmsnell: maybe belongs in another PR? It looks like we introduced this bug and didn't notice it because `undefined` collapses into `false`

PR #1904, opened 2020-02-12
Refactor: Visible Panes as Set

dmsnell:
This patch introduces a new set utility library for adding, removing,
and toggling values in a set to make a working "option bag."

It also changes the `visiblePane` state from a list of strings into this
new option bag, removing the reliance on the `lodash` equivalent
functions for using the array.

This is meant to be precursor work for #1901 and #1903 should it be useful.

### DIFF
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -1,9 +1,10 @@
-import { difference, union } from 'lodash';
-import { AnyAction, combineReducers } from 'redux';
+import { combineReducers } from 'redux';
+
+import * as set from '../../utils/set-utils';
+
 import * as A from '../action-types';
 import * as T from '../../types';
 
-const defaultVisiblePanes = ['editor', 'noteList'];
 const emptyList: unknown[] = [];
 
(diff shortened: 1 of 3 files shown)


dmsnell:
I think we're completely ignoring the existing `visiblePane` state so maybe we ought to remove it?

belcherj:
I think we are headed down the wrong path. This a good time to assess how we want to handle things like this.

Including a utility function to run the reducer adds a layer of abstraction that I am not comfortable with.

These visible panes should be their own state with bool values:

```
{ ui: {
  trash: bool,
  editor: bool,
  noteList: bool,
  navigation: bool
  noteInfo: bool
} }
```

Overall, I do not see the value of pushing them all to visible panes and then having to use a util manage the way sets work. I would be okay with a sub or completely different state tree which I think would accomplish the same thing. Additionally using sets requires more logic in the component.

```
state.ui.visiblePanes.has('trash')
```
VS
```
state.ui.trash
```

dmsnell:
Closing since `visiblePanes` were removed in #1895

PR #1905, opened 2020-02-13
Tag suggestions: Matches should be case insensitive

codebykat:
### Fix
Tag suggestions should be case insensitive. Quick fix by lowercasing all the comparison strings before comparing (since `startsWith` doesn't have a flag for case sensitivity and moving to regex seemed overly complex). Added a test for this too.

### Test
1. Search for the name of a tag but with different case.
2. Verify tag suggestions suggest the tag.
3. Smoke test for other regressions involving search and tag suggestions.

### Release

> Fixed tag suggestions to be case insensitive

### DIFF
--- a/lib/tag-suggestions/index.tsx
+++ b/lib/tag-suggestions/index.tsx
@@ -98,6 +98,7 @@ export const filterTags = (tags: T.TagEntity[], query: string) => {
   // but without knowing where the cursor is it's maybe the best we can do
   const tagTerm = query
     .trim()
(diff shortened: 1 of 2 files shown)


dmsnell:
For the time being I don't see it being a problem to call `toLowerCase()` on every tag because we're going to have to do some kind of string op on every one each time we change the search field text.

if/when we move searching into a `WebWorker` this will all change. see #1835 for an example

PR #1907, opened 2020-02-16
State refactor: Move toggle navigation to Redux

codebykat:
### Fix
Move `showNavigation` and `toggleNavigation` to Redux state. Adds action `NAVIGATION_TOGGLE`.

### Test
1. Clickie the hamburger
2. Notification panel should appear
3. It should go away with ESC or clicking anywhere
4. Key combo cmd+T should also toggle it in and out
5. Try to break other things too

### Review
See comment about `editingTags` state

### Release

> TBD

### DIFF
--- a/lib/navigation-bar/index.tsx
+++ b/lib/navigation-bar/index.tsx
@@ -14,7 +14,23 @@ import { viewExternalUrl } from '../utils/url-utils';
 import appState from '../flux/app-state';
 import DialogTypes from '../../shared/dialog-types';
 
-export class NavigationBar extends Component {
+import { toggleNavigation } from '../state/ui/actions';
+
+import * as S from '../state';
+import * as T from '../types';
+
+type StateProps = {
+  showNavigation: boolean;
+};
+
+type DispatchProps = {
+  onOutsideClick: () => any;
+  toggleNavigation: Function;
+};
+
+type Props = StateProps & DispatchProps;
+
+export class NavigationBar extends Component<Props> {
   static displayName = 'NavigationBar';
 
   static propTypes = {
@@ -22,12 +38,10 @@ export class NavigationBar extends Component {
     dialogs: PropTypes.array.isRequired,
     isElectron: PropTypes.bool.isRequired,
     onAbout: PropTypes.func.isRequired,
-    onOutsideClick: PropTypes.func.isRequired,
     onSettings: PropTypes.func.isRequired,
     onShowAllNotes: PropTypes.func.isRequired,
     selectTrash: PropTypes.func.isRequired,
     selectedTag: PropTypes.object,
-    showNavigation: PropTypes.bool.isRequired,
     showTrash: PropTypes.bool.isRequired,
   };
 
@@ -127,11 +141,15 @@ export class NavigationBar extends Component {
   }
 }
 
-const mapStateToProps = ({ appState: state, settings, ui: { showTrash } }) => ({
+const mapStateToProps = ({
+  appState: state,
+  settings,
+  ui: { showNavigation, showTrash },
+}) => ({
   autoHideMenuBar: settings.autoHideMenuBar,
   dialogs: state.dialogs,
   selectedTag: state.tag,
-  showNavigation: state.showNavigation,
+  showNavigation,
   showTrash,
 });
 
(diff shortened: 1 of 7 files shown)


lib/navigation-bar/index.tsx
```diff
@@ -127,22 +143,21 @@ export class NavigationBar extends Component {
...
 
 const {
   showAllNotesAndSelectFirst,
   selectTrash,
   showDialog,
-  toggleNavigation,
 } = appState.actionCreators;
 
-const mapDispatchToProps = {
+const mapDispatchToProps: S.MapDispatch<DispatchProps, OwnProps> = {
```
dmsnell: note that when we don't use or rely on `ownProps` then we don't have any need to include it in the type signature. in fact, _any_ time we use the object form of `mapDispatchToProps` (vs. the function form `(dispatch, ownProps) => …`) then we don't need to include the `OwnProps` in the signature.

on a deeper level, if you inspect the type you can see that it uses a default parameter much the same way a normal functions does…

```ts
type MapDispatch<DispatchProps, OwnProps = {}> = { … }
```

lib/search-bar/index.tsx
```diff
@@ -14,33 +14,42 @@ import NewNoteIcon from '../icons/new-note';
...
   onNewNote,
-  onToggleNavigation,
   searchQuery,
   showTrash,
 }) => (
   <div className="search-bar theme-color-border">
-    <IconButton icon={<MenuIcon />} onClick={onToggleNavigation} title="Menu" />
+    <IconButton
+      icon={<MenuIcon />}
+      onClick={() => toggleNavigation()}
```
dmsnell: should be safe to pass just the function itself instead of creating a new one. I'm pretty sure you'd catch this but I wanted to point it out since I saw it.

```ts
onClick={toggleNavigation}
```

lib/search-bar/index.tsx
```diff
@@ -59,15 +68,21 @@ const mapStateToProps: S.MapState<StateProps> = ({
   showTrash,
 });
 
-const mapDispatchToProps = (dispatch, { noteBucket, onNoteOpened }) => ({
+const mapDispatchToProps: S.MapDispatch<DispatchProps, OwnProps> = (
+  dispatch,
+  { noteBucket, onNoteOpened }
```
dmsnell: perfect use of `OwnProps` in the type signature since we're relying on these two passed-in props

lib/state/ui/reducer.ts
```diff
@@ -70,14 +71,39 @@ const showNoteInfo: A.Reducer<boolean> = (state = false, action) => {
...
+
+const showNavigation: A.Reducer<boolean> = (state = false, action) => {
+  console.log(action);
+  switch (action.type) {
+    case 'NAVIGATION_TOGGLE':
+      return !state;
+
+    // @todo -- should be implemented with editingTags actually
+    // case 'App.editingTags':
+    // return false;
```
dmsnell: when you get to it I'd expect the `@todo` items to appear like so…

```ts
case 'App.selectTag':
case 'App.selectTrash':
case 'App.showAllNotes':
	return false;
```

lib/search-bar/index.tsx
codebykat: This was an artifact of me not being able to get it to call the dang thing. I just figured out that I need to pass it in above.. but I don't understand why...

lib/state/ui/reducer.ts
codebykat: The one I'm stuck on is that it was initially setting `editingTags` to false as well, and... we don't have that in the reducers yet...

codebykat:
IT'S ALIVE

ALIVE

There's one @todo left, which is that `editingTags` in the old state needs to get set to false... somehow...

codebykat:
`editingTags` is done!! thanks @belcherj ! This is ready for :eyes: now

lib/navigation-bar/index.tsx
```diff
@@ -14,20 +14,34 @@ import { viewExternalUrl } from '../utils/url-utils';
...
+import * as S from '../state';
+import * as T from '../types';
+
+type StateProps = {
+  showNavigation: boolean;
+};
+
+type DispatchProps = {
+  onOutsideClick: () => any;
+  toggleNavigation: Function;
```
dmsnell: this is fine, though I have preferred the `() => any` style both for consistency but also for one more thing: `Function` doesn't convey that this is a "fire-and-forget" action the same way that `() => any` does to me.

`Function` doesn't specify the argument types while `() => any` communicates "this function expects no arguments" and sometimes that's handy to know, especially if jumping into the code and trying to work with what's there.

since these are right next to each other, did you have a reason to keep them separate from each other?

codebykat: No, literally just copy and pasting things without understanding the difference. I was going to ask which is preferred and why!

lib/state/ui/reducer.ts
dmsnell: not sure what you are saying there. if we're getting it from state though and we haven't set it in a reducer it will probably be `undefined` which casts to `false` in boolean conditions

PR #1908, opened 2020-02-17
State Refactor: move setSearchFocus to two actions, focusSearch and blurSearch

codebykat:
### Fix
This moves `setSearchFocus` to Redux state

### Test
1. Make sure search/filter behaves as expected
2. Searching for a tag also should focus search when a tag suggestion is clicked
2. Also test the keybinding (cmd+F) in Electron to focus the search field

### Review
The state boolean was originally called `searchFocus`, I renamed it to `searchFocused`.

The corresponding action was originally `set` with `true` and `false` arguments. I experimented with breaking it out instead into a `focusSearch` and `blurSearch` action. I still think it's somewhat confusing that the search field calls `onSearchFocus` to unfocus search... but we can refactor it into oblivion once we move everything into Redux. 

### Release
> TBD

### DIFF
--- a/lib/state/action-types.ts
+++ b/lib/state/action-types.ts
@@ -51,15 +51,13 @@ export type SetWPToken = Action<'setWPToken', { token: string }>;
 /*
  * Normal action types
  */
+export type BlurSearch = Action<'BLUR_SEARCH'>;
 export type CreateNote = Action<'CREATE_NOTE'>;
 export type CloseNote = Action<'CLOSE_NOTE'>;
 export type FilterNotes = Action<'FILTER_NOTES', { notes: T.NoteEntity[] }>;
+export type FocusSearch = Action<'FOCUS_SEARCH'>;
 export type Search = Action<'SEARCH', { searchQuery: string }>;
 export type SetAuth = Action<'AUTH_SET', { status: AuthState }>;
-
-export type SetSearchFocus = Action<'SET_SEARCH_FOCUS'>;
-export type UnsetSearchFocus = Action<'UNSET_SEARCH_FOCUS'>;
-
 export type SetUnsyncedNoteIds = Action<
   'SET_UNSYNCED_NOTE_IDS',
   { noteIds: T.EntityId[] }
@@ -76,10 +74,12 @@ export type ToggleTagEditing = Action<'TAG_EDITING_TOGGLE'>;
 export type SelectNote = Action<'SELECT_NOTE', { note: T.NoteEntity }>;
 
 export type ActionType =
+  | BlurSearch
   | CreateNote
   | CloseNote
   | LegacyAction
   | FilterNotes
+  | FocusSearch
   | Search
   | SelectNote
   | SetAccountName
@@ -90,7 +90,6 @@ export type ActionType =
   | SetLineLength
   | SetMarkdownEnabled
   | SetNoteDisplay
-  | SetSearchFocus
   | SetSortReversed
   | SetSortTagsAlpha
   | SetSortType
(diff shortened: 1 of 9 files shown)


lib/state/action-types.ts
```diff
@@ -55,6 +55,10 @@ export type CreateNote = Action<'CREATE_NOTE'>;
 export type FilterNotes = Action<'FILTER_NOTES', { notes: T.NoteEntity[] }>;
 export type Search = Action<'SEARCH', { searchQuery: string }>;
 export type SetAuth = Action<'AUTH_SET', { status: AuthState }>;
+
+export type SetSearchFocus = Action<'SET_SEARCH_FOCUS'>;
+export type UnsetSearchFocus = Action<'UNSET_SEARCH_FOCUS'>;
```
dmsnell: I like this change but `SetSearchFocus` and `UnsetSeachFocus` read a bit awkward to me. Did you have thoughts about making it active such as with `FocusSearch` and `UnfocusSearch` or `BlurSearch` (if we wanted to stick with HTML terminology)

lib/state/ui/actions.ts
```diff
@@ -31,6 +31,14 @@ export const search: A.ActionCreator<A.Search> = (searchQuery: string) => ({
   searchQuery,
 });
 
+export const setSearchFocus: A.ActionCreator<A.SetSearchFocus> = () => ({
+  type: 'SET_SEARCH_FOCUS',
+});
+
+export const unsetSearchFocus: A.ActionCreator<A.UnsetSearchFocus> = () => ({
+  type: 'UNSET_SEARCH_FOCUS',
+});
```
dmsnell: Note that we can leave the action untouched but create two action creators. This is typical as a kind of helper to the state being updated.

```ts
export type SetSearchFocus = Action<'SET_SEARCH_FOCUS', { shouldHaveFocus: boolean; }>;
```

```ts
export const focusSearch: A.ActionCreator<A.SetSearchFocus> = () => ({
  type: 'SET_SEARCH_FOCUS',
  shouldHaveFocus: true,
});

export const unfocusSearch: A.ActionCreator<A.SetSearchFocus> = () => ({
  type: 'SET_SEARCH_FOCUS',
  shouldHaveFocus: false,
});
```

```ts
const searchFocused: A.Reducer<boolean> = (state = false, action) => {
  switch (action.type) {
    case 'SET_SEARCH_FOCUS':
      return action.shouldHaveFocus;

    default:
      return state;
  }
}
```

codebykat: > This is probably all a lot to say that this never should have been in state 🙃

Yeah, I was trying to comprehend the point of this thing and why Search calls SetSearchFocus and it's logically nonsense.

lib/state/action-types.ts
codebykat: Fine by me, I will update it

lib/state/ui/actions.ts
dmsnell: It's fine to leave it in for now and replicate the `app-state` behavior. _Changes_ can come more easier _after_ the refactor 👍

codebykat: Hmm... I like it better as two, personally.

codebykat:
Updated and rebased 👍

lib/state/index.ts
```diff
@@ -35,7 +35,7 @@ export type AppState = {
   preferences?: T.Preferences;
   previousIndex: number;
   revision: T.NoteEntity | null;
-  searchFocus: boolean;
+  searchFocused: boolean;
```
belcherj: I think you want to delete this.

codebykat: Whoops good catch!

codebykat:
See #1914 - closing for now

PR #1913, opened 2020-02-18
e2e tests: improve reliability of login test

dmsnell:
Previously the tests were a bit flakey, working some times and not others.  In
this patch we've added the `--runInBand` option to `jest` so that the tests run
sequentially in the same process as the test runny. It may have been that
`jest` was attempting to run the tests in parallel even though the nature of
testing Electron/Spectron doesn't support this and so we would sometimes fail
based on concurrency conflicts.

A few new tests have been added in order to verify that the new updates are
working and to demonstrate how we can start adding testable behaviors into the
app. We still need to improve how the tests are organized and setup to minimize
the noise in the code and isolate the tests.

## Review: Signaling from the app to the tests

We would ideally like to avoid timeouts as a way of progressing through our
tests. Timeouts are inherently flakey because they represent race conditions
on uncontrollable factors. That means we prefer to wait for specific events
and send those events from the running app to the test runner.

In this PR I've created a system that pushes new entries to a global array in
order to signal an event.

```ts
// when we only need to indicate that an event occurred
window.testEvents.push('notesLoaded');

// when we want to send data out as well
window.testEvents.push(['noteSelected', noteId, note.content]);
```

This looks similar to our Redux actions though may occur at different times
and circumstances. These are to be used when we can't wait for the presence
of specific DOM nodes via the builtin `webdriverio` function `waitForExist()`.

Through the use of a webpack `DefinePlugin` entry we can do this with a simple
conditional which will be stripped entirely from `development` and `production`
builds, purely because it's discernible at compile-time if the condition is true.

```ts
__TEST__ && window.testEvents.push('someEvent');
```

## Testing

Since this _only_ affects the in-progress e2e tests it should have zero impact on
the resultant app code. Test by running the e2e test suite.

**It is necessary to build the app with `NODE_ENV=test` before running these tests.**
I hope to cure this at some point but it's just an inconvenience at the moment.

```bash
NODE_ENV=test npm run build
npm run test-e2e
```

### DIFF
--- a/e2e/test.ts
+++ b/e2e/test.ts
@@ -14,21 +14,39 @@ const el = (app: Application, selector: string) => app.client.$(selector);
 const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
 const waitFor = async (app: Application, selector: string, msTimeout = 10000) =>
   expect(await app.client.waitForExist(selector, msTimeout)).toBe(true);
-const waitForWindowProp = async (
+const waitForEvent = async (
   app: Application,
-  propName: string,
+  eventName: string,
   msTimeout = 10000
-) => {
+): Promise<any[]> => {
   const tic = Date.now();
 
   return new Promise(async (resolve, reject) => {
     const f = async () => {
-      const result = await app.client.executeAsync(function(prop, done) {
-        return done(window[prop]);
-      }, propName);
+      const result = await app.client.execute(function() {
+        var events = window.testEvents;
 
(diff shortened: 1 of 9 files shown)


dmsnell:
@belcherj updated so that tests now properly fail

belcherj:
Looks like you have a lint error.

dmsnell:
> Looks like you have a lint error.

@belcherj doesn't look like it anymore

PR #1915, opened 2020-02-19
Fix revisions loading on revisions panel toggle

belcherj:
### Fix
I broke the loading of revisions with this removal: https://github.com/Automattic/simplenote-electron/pull/1903/files#diff-e98743febf9f62311da7008a480dad44L60

This adds it back.  This is a terrible way of loading revisions but it gets us back to a good state.

### Test
1. Load app
2. Open revisions panel
3. Revert note

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -42,7 +42,7 @@ class NoteContentEditor extends Component<Props> {
     content: PropTypes.shape({
       text: PropTypes.string.isRequired,
       hasRemoteUpdate: PropTypes.bool.isRequired,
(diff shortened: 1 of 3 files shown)


belcherj:
@codebykat fixed the prop type.  I think this has bee wrong for a long time.

codebykat:
Hmmmmm. I'm seeing a weird... glitch... where when I try to drag the revision slider it gets "stuck" in the middle. Like if you drag from the right to left (or left to right) on a note with more than 3 revisions, it stops responding right at the center.

If you scrub off the left, it also stops responding, but this is also true on production. :/

PR #1916, opened 2020-02-19
Fix: Console error on loading Settings->Display dialog

codebykat:
### Fix
This fixes the following console error which is thrown when opening Settings and the Display tab:

```Warning: Failed prop type: The prop `className` is marked as required in `CheckboxControl`, but its value is `undefined`.
    in CheckboxControl (created by SettingsGroup)
    in SettingsGroup (created by DisplayPanel)```

Fixed by making `className` an optional prop of the child component.

### Test
1. Open Settings
2. Click Display tab
3. Note lack of console errors

### Release

### DIFF
--- a/lib/controls/checkbox/index.tsx
+++ b/lib/controls/checkbox/index.tsx
@@ -14,7 +14,7 @@ function CheckboxControl({ className, ...props }) {
 }
 
 CheckboxControl.propTypes = {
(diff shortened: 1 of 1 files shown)


codebykat:
Yeah, I chose to remove the requirement since it didn't seem like there was any reason it should be required to pass in a class name.

PR #1920, opened 2020-02-20
Move tag selection to redux

belcherj:
### Fix
Moves selected tag state from app state to Redux. 

### Test
1. Select a tag
2. Does list filter
3. Select show all notes
4. Does list filter back to all
5. Select tag
6. Select trash
7. Does it show all notes in trash
8. Select tag
9. Does it show notes with that tag in the trash


lib/flux/app-state.ts
```diff
@@ -161,9 +139,7 @@ export const actionMap = new ActionMap({
               modificationDate: timestamp,
               shareURL: '',
               publishURL: '',
-              tags: ([] as T.TagName[]).concat(
```
belcherj: I had no idea that if you created a new note while on a tag that it would add that tag. I think we should delete that and add it back as we move newNote over to Redux.

dmsnell: the idea originally was that since you were "inside" the tag you would be creating a note inside of it too. indicating which tab we're inside has always been awkward. I'd be more inclined to keep the behavior until we make a decision to change it because otherwise we're going to have to come back and undo this refactor if we don't like the behavioral change.

is it difficult to maintain the existing behavior?

lib/state/ui/middleware.ts
```diff
@@ -19,7 +19,7 @@ export const middleware: S.Middleware = store => {
       case 'App.deleteNoteForever':
       case 'App.notesLoaded':
       case 'App.restoreNote':
-      case 'App.selectTag':
+      case 'SELECT_TAG':
```
dmsnell: Considering the semantics of what we're doing I would love to entertain drawing a clearer distinction between what we're doing here and what we do when we click on a tag inside the note editor.

For instance, could this be `OPEN_TAG` while clicking on or selecting the tag in the note editor would be `SELECT_TAG`? I doubt we'd make that a global Redux action but for the sake of our own confusion when discussing it maybe something like `OPEN_` could remove the ambiguity?

lib/flux/app-state.ts
belcherj: Added back, done.

lib/state/ui/middleware.ts
belcherj: Done.

lib/flux/app-state.ts
```diff
@@ -69,34 +69,12 @@ export const actionMap = new ActionMap({
...
-              notes: getState().appState.notes,
-            })
-          );
-        };
-      },
-    },
-
-    selectTag(state: AppState, { tag }: { tag: T.TagEntity }) {
-      return update(state, {
-        tag: { $set: tag },
```
dmsnell: so nice removing this triple-dispatch

lib/flux/app-state.ts
```diff
@@ -69,34 +69,12 @@ export const actionMap = new ActionMap({
...
-        return (dispatch, getState) => {
-          dispatch(this.action('selectTag', { tag }));
-          dispatch(
-            this.action('notesLoaded', {
-              notes: getState().appState.notes,
-            })
-          );
-        };
-      },
-    },
```
dmsnell: so nice removing this triple-dispatch

lib/flux/app-state.ts
```diff
@@ -162,7 +143,7 @@ export const actionMap = new ActionMap({
               shareURL: '',
               publishURL: '',
               tags: ([] as T.TagName[]).concat(
-                state.tag ? state.tag.data.name : []
+                selectedTag ? selectedTag.data.name : []
```
dmsnell: no need to do it here in this PR but I hope you're starting to scream "middleware!" when you see code like this 😄

lib/state/action-types.ts
```diff
@@ -93,6 +94,7 @@ export type ActionType =
   | Search
   | SelectNote
   | SelectRevision
+  | SelectTag
```
dmsnell: Let's match names here with the action type? Or do you see value in keeping this `SelectTag`?

lib/state/action-types.ts
```diff
@@ -198,8 +200,6 @@ type LegacyAction =
   | Action<'App.notesLoaded', { notes: T.NoteEntity[] }>
   | Action<'App.onNoteBeforeRemoteUpdate', { noteId: T.EntityId }>
   | Action<'App.preferencesLoaded', { analyticsEnabled: boolean }>
-  | Action<'App.selectTag', { tag: T.TagEntity }>
-  | Action<'App.selectTagAndSElectFirstNote'>
```
dmsnell: I realize now that this list obviates the issue tracking these

lib/state/ui/reducer.ts
```diff
@@ -175,7 +187,7 @@ const note: A.Reducer<T.NoteEntity | null> = (state = null, action) => {
     case 'App.emptyTrash':
     case 'App.showAllNotes':
     case 'App.selectTrash':
-    case 'App.selectTag':
+    case 'OPEN_TAG':
```
dmsnell: this is a job for another day/PR but I don't think we need all these anymore since `FILTER_NOTES` takes care of them. maybe it's still best. what's odd is setting it to `null` and then immediately setting it to some other note

dmsnell:
@belcherj I added a commit to push the rename all the way through the app. now there's "open tag" when talking about the tag drawer but "selected tag" when talking about highlighting a single tag in the tag list / tag field / tag line inside the note editor.

you can back out of this commit if you think it goes too far. my preference is to go 100% or 0% so we don't end up with crossed terms on the new actions (e.g. `SelectTag = Action<'OPEN_TAG'>`)

Also I added some types and we can take those out if we want but I found them useful to make sure what I was changing was proper.

There's a merge conflict with or without my commit but it was pretty trivial to resolve. If you want I'd be happy to rebase over `develop`

dmsnell:
## Testing

- [ ] Logout
- [ ] Login with wrong password fails
- [ ] Login with correct password
- [ ] Create new note appears in other device
- [ ] Changes to new note sync to other device
- [ ] Changes to new note sync _from_ other device (after debounce delay)
- [ ] New tag immediately syncs to other device
- [ ] New tag immediately syncs _from_ other device
- [ ] Removed tag immediately syncs to other device
- [ ] Removed tag immediately syncs _from_ other device
- [ ] Note publishes with link
- [ ] Note unpublishes
- [ ] Note publish change syncs _from_ other device (visible with dialog open)
- [ ] Markdown setting syncs _from_ other device
- [ ] Preview mode disappears/reappears when receiving remote changes to markdown setting
- [ ] Note pinning syncs immediately from either direction
- [ ] Note pinning works regardless if clicking in list view or from note info
- [ ] Viewing history on one device leaves note unchanged on other device
- [ ] Restoring history immediately syncs note from both directions
- [ ] Restoring history revisions updates pinned status
- [ ] Restoring history revisions updates markdown status
- [ ] Restoring history revisions updates publish status
- [ ] Can view trashed notes by clicking on `Trash`
- [ ] Can delete note forever from trash screen
- [ ] Can restore note from trash screen
- [ ] Can trash note
- [ ] Can preview markdown with 👁 button
- [ ] Can flip to edit mode with 👁 button
- [ ] Markdown rendered in note list when markdown enabled
- [ ] Markdown syntax unrendered in note list when markdown disabled
- [ ] Can filter by tag when clicking on tag in tag drawer
- [ ] Can add tag to note and have it appear in filtered tag view when previously not in filter
- [ ] Searching in the search field highlights matches in note list
- [ ] Searching in the search field highlights matches in the note editor
- [ ] Searching in the search field highlights matches in the note editor during preview
- [ ] Clearing the search field immediately updates filtered notes
- [ ] Clicking on different tags or `All Notes` or `Trash` immediately updates filtered notes
- [ ] Can search by keyword, filtering debounced
- [ ] Tag auto-completes appear when typing in search field
- [ ] Typing `tag:` does not result a list of all tags in the autocompleter
- [ ] Typing `tag:` and something else, like `tag:te` results in autocomplete results starting with that something else, e.g. `test`
- [ ] Tag suggestions suggest tags regardless of case
- [ ] Search field updates with results of `tag:test` format search string
- [ ] Can toggle sidebar
- [ ] Syncs when introducing sequential surrogate pairs sharing the same high surrogate, e.g. `🅰🅱` to `🅰🅰🅱`
- [ ] When disabling network connectivity, changes start counting in unsync'ed changes counter
- [ ] When going back online changes sync and counter resets to `All chagnes synced`
- [ ] Can change analytics sharing setting
- [ ] Changing `Note Display` mode immediately updates and reflects in note list
- [ ] Changing `Sort Type` mode immediately updates and reflects in note list
- [ ] Toggling `Sort Order` immediately updates and reflects in note list for each sort type
- [ ] For each sort type the pinned notes appear first in the note list
- [ ] Changing `Theme` immediately updates app for desired color scheme
- [ ] Using the `Insert checklist` item from the format menu inserts a checklist
- [ ] "Undo" undoes the last edit
- [ ] Typing `- [x]` creates a checked checklist item
- [ ] Typing `- [ ]` created an unchecked checklist item
- [ ] Typing `-` creates a list
- [ ] Typing _tab_ in a list item underneath another list item indents item
- [ ] Typing _shift tab_ in the same spot unindents the list
- [ ] Changing `-` to `+` changes the list item bullet, also for `*` and `•` (`\u2022`)
- [ ] All list bullet types render to markdown lists
- [ ] Added URL is linkified
- [ ] When clicking on link it opens in new window
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + <kbd>p</kbd> toggles preview mode
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + (<kbd>k</kbd> or <kbd>up</kbd>) selects next note above current note, stops at top of list
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + (<kbd>j</kbd> or <kbd>down</kbd>) selects next note below current note, stops at bottom of list
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>t</kbd> toggles tab list
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>n</kbd> creates new note
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + <kbd>n</kbd> _when in small screen mode_ navigates to note list
- [ ] Can print note in plaintext view
- [ ] Can print note in Markdown-preview view

PR #1927, opened 2020-02-24
Fixes tag list editing mode

belcherj:
### Fix
Fixes note list editing. editingTags is now on the ui state branch not appState.

### Test
1. Can you enter edit mode for the tag list?


dmsnell:
For reference: linking to #1899 where this was introduced

PR #1941, opened 2020-02-28
Search/offload filtering 2

dmsnell:
Second version #1835

## Highlights

 - offloads search into a `WebWorker` which means search doesn't block the main thread
 - on my heaviest accounts and worst searches it's still finishing within 1ms and completely responsive. the effect is huge when comparing delayed results to instant results


lib/search/worker/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    const searchTags = new Set<T.TagName>();
+    let match;
+    while ((match = tagPattern.exec(query)) !== null) {
+      searchTags.add(match[1].toLocaleLowerCase());
+    }
+    return searchTags;
+  };
+
+  const updateFilter = (scope = 'quickSearch') => {
+    const tic = performance.now();
```
codebykat: "tick" probably?

lib/search/worker/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    const matches = new Set<T.EntityId>();
+
+    for (const [noteId, note] of notes.values()) {
+      // return a small set of the results quickly and then
+      // queue up another search. this improves the responsiveness
+      // of the search and it gives us another opportunity to
+      // receive inbound messages from the main thread
+      // in testing this was rare and may only happen in unexpected
+      // circumstances such as when performing a garbage-collection
+      const toc = performance.now();
```
codebykat: and "tock" idk but "tic" is not the best :)

lib/search/index.ts
```diff
@@ -0,0 +1,106 @@
+import SearchWorker from 'worker-loader!./worker';
```
codebykat: Just curious, what is this syntax?

lib/search/worker/index.ts
dmsnell: Common practice I picked up in my MATLAB days: tic and toc - l like their symmetry

lib/search/index.ts
dmsnell: That is a way to get a webpack loader working without having to configure webpack to handle the file type. In this case it runs the import through the webworker loader so that it turns it into something we can load in the worker frame

lib/search/worker/index.ts
codebykat: It's just that "tic" has another meaning and it's not a good one..

codebykat: I don't really feel strongly about it though

dmsnell:
Thanks both of you for the review. If you don't have strong feeling about tic/toc I'd like to keep them; maybe it's just my way of putting a mark on the world. Maybe there's something about the computational world that led them this way, but [MATLAB], [R], and [octave] all use tic/toc, and some other languages adopted that same nomenclature as well.

now you have me wondering what the history of its origin is 🤔

[MATLAB]: https://www.mathworks.com/help/matlab/ref/tic.html
[R]: https://collectivemedia.github.io/tictoc/tic.html
[octave]: https://octave.org/doc/v4.0.0/Timing-Utilities.html#XREFtic

PR #1958, opened 2020-03-13
Fix application signing to generate appx for windows store

belcherj:
### Fix
Windows App store requires that the appx be unsigned. This PR adds an additional job to build the appx for release to the store.

### Test
Run the build on a tag and upload the built asset to the Windows store.

### Release
No release notes needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,11 +179,10 @@ jobs:
           command: |
             # Workaround for Sign Tool "private key filter" bug in Circle's Windows image.
             # Ref: https://travis-ci.community/t/codesigning-on-windows/
-            # 
-            # Fix: Import .p12 into the local certificate store. Sign Tool will use 
+            #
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -179,18 +179,49 @@ jobs:
...
             - release\*.appx
             - release\*.yml
+  windowsstore:
+    executor:
+      name: win/default
+    working_directory: C:\Users\circleci\simplenote-electron
+    environment:
+      CSC_LINK: ''
+      WIN_CSC_LINK: ''
+    steps:
```
nsakaimbo: Instead of the somewhat unintuitive filtering in the "workflow" section, it might work to use Circle config's "when" syntax here:

```
windowsstore:
  when:
    condition: << pipeline.git.tag >>
      steps:
        // ...
```

dmsnell:
Closing in favor of #1961

PR #1919, opened 2020-02-20
Move previous index store to Redux

belcherj:
### Fix

Moves the previousIndex state from app state to redux. This implementation is mimicking what we were doing in app state and should not be considered a final implementation of how this should be done.  One app state is migrated we can take a better look at how to implement this properly.

Also a fix for:

Selects note above recently-trashed note ⭕️ selects top note in list 5
Selects note above recently-restored note in trash view ⭕️ selects top note in list 5
Selects note above recently-deleted-forever note in trash view ⭕️ selects top note in list 5

In #1851 we inadvertently broke this behavior and then in #1895 removed the mechanism that had previously enabled it.

Taken from: https://github.com/Automattic/simplenote-electron/issues/1911 Thanks @dmsnell 

### Test
1. Test that selects note above recently-trashed note
2. Test that selects note above recently-restored note in trash view
3. Tets that selects note above recently-deleted-forever note in trash view

PR #1966, opened 2020-03-17
Fix: Remember opened tag in search and send previous index

dmsnell:
In #1941 we introduced two regressions and this patch fixes them.

 - search needs to remember the opened tag and limit results to those
   matching within that tag. this was overlooked when updating the
   search query and updating the `filterTags` _from_ that search query

 - when passing the `filterNotes` action back to the application we
   have to make sure to pass the previous index. this was lost during
   the merge of the PR because because the additions from #1919, which
   fixed the previous regression for it, were deleted as part of
   moving `ui/middleware` into `search/index`

With these changes in place the two regressions should be fixed again.

## Testing

Apart from the normal checklist note that #1941 introduced regressions
to the following items. Verify that they are fixed in this branch.

### DIFF
--- a/lib/search/index.ts
+++ b/lib/search/index.ts
@@ -17,16 +17,17 @@ export const middleware: S.Middleware = store => {
 
   searchProcessor.onmessage = event => {
     switch (event.data.action) {
-      case 'filterNotes':
+      case 'filterNotes': {
+        const { appState, ui } = store.getState();
         store.dispatch(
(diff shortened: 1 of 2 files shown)


dmsnell:
Introduced a big performance regression when searching: since this selects the previous note on ever search we can end up rendering and re-rendering slow notes in the draft-js editor. This kills interactivity.

PR #1975, opened 2020-03-20
Move analytics preference to redux

belcherj:
### Fix
This PR moves the analytics and preferences state from App State to redux. Analytics is the only thing that uses the preferences bucket so moving them together makes sense.

TODO: Send application opened event

### Test
1. Slip a console.log in here: https://github.com/Automattic/simplenote-electron/blob/develop/lib/analytics/index.ts#L44

### Release
Not added: `Move analytics preference to redux`


dmsnell:
We might be making it harder on ourselves if we try and carry over the imperative nature of fetching preferences. Inherently we know that we shouldn't be showing any analytics value until we've gotten a positive value on the setting; likewise we shouldn't be enabling the toggle until that time. That is, if we previously received an `on` or `off` message for it then we can continue to show that, but we never _fetch_ the setting. The setting comes to us.

If we consider the analytics option as a tri-state toggle that can be `true`, `false`, or `null` then we can adequately model this very thing.

```ts
const sendAnalytics: S.Reducer<boolean | null> = ( state = null, action ) => {
  switch (action.type) {
    case 'TOGGLE_ANALYTICS':
      return !state;

    case 'ANALYTICS_REMOTE_UPDATE':
      return action.sendAnalytics;

    case 'LOGOUT':
      return null;

    default:
      return state;
  }
}
```

This formulation should end up simplifying much of the code in this PR. In `app.jsx` for instance we can do something like this…

```ts
preferencesBucket.on('index', () => {
  preferencesBucket.get('preferences-key', ( error, preferences ) => {
    if ( error || ! preferences.data.hasOwnProperty( 'analytics_enabled' ) ) {
      return;
    }

    this.props.remotePreferencesUpdate( preferences.data.analytics_enabled );
  } )
} );

preferencesBucket.on('update', (entityId, data) => {
  if ('preferences-key' !== entityId || !data.hasOwnProperty('analytics_enabled')) {
    return;
  }

  this.props.remotePreferencesUpdate( data.analytics_enabled );
} );
```

The middleware can handle the `window` globals and the updates…

```ts
const result = next( action );
const nextState = store.getState();

…

case 'ANALYTICS_REMOTE_UPDATE':
  window.analyticsEnabled = actions.sendAnalytics;
  break;

case 'TOGGLE_ANALYTICS':
  preferencesBucket.update('preferences-key', { analytics_enabled: nextState.sendAnalytics } );
  break;

…
```

lib/dialogs/settings/panels/account.tsx
```diff
@@ -67,9 +80,17 @@ const AccountPanel = props => {
...
-export default AccountPanel;
+const mapStateToProps: S.MapState<StateProps> = ({
+  settings: { analyticsEnabled },
+}) => ({
+  analyticsEnabled,
+});
+
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
+  toggleAnalyticsEnabled: () => actions.settings.toggleAnalyticsEnabled(),
+};
```
dmsnell, line 94: small thing, but I think it's worth keeping `mapDispatchToProps` in the simpler object form until we need to use the function form so that we can avoid some of the scenarios where it's so complex that it's confusing

```ts
	toggleAnalyticsEnabled: actions.settings.toggleAnalyticsEnabled,
}
```

lib/state/simperium/middleware.ts
```diff
@@ -60,6 +92,13 @@ export const middleware: S.Middleware = store => {
       case 'REVISIONS_TOGGLE':
         fetchRevisions(store, nextState);
         break;
+
+      case 'TOGGLE_ANALYTICS_ENABLED':
+        setAnalyticsEnabled(nextState);
+        break;
+      case 'FETCH_PREFERENCES':
+        fetchPreferences(store);
```
dmsnell, line 100: there should be no need to carry-over `fetchPreferences`

there's only a few cases we need to cover, even if it does take some care to really nail it

| | `state` | Display | User Action | App Response |
|---|---|---|---|---|
| Unfetched, changed by user | `null` | Indeterminate display but toggle-able | Toggles | Update state, middleware sends update to Simperium |
| Unfetched, received `index` or `update` | `null` | Flips to value of state | | `state` becomes `true` \| `false` |
| Fetched | `true` \| `false` | Normal toggle | Toggles | Update state, middleware sends update to Simperium |

What this boils down to is that there are really only two state: indeterminate and toggled. Either a remote update or a user action can set it into toggled mode. It can _only_ be indeterminate if we haven't heard from Simperium yet or for some strange reason there's no preference at all in the account.

Simperium will _always_ "fetch" the initial value and all we need to do is faithfully render that.

lib/state/simperium/middleware.ts
```diff
@@ -60,6 +92,13 @@ export const middleware: S.Middleware = store => {
       case 'REVISIONS_TOGGLE':
         fetchRevisions(store, nextState);
         break;
+
+      case 'TOGGLE_ANALYTICS_ENABLED':
+        setAnalyticsEnabled(nextState);
```
dmsnell, line 97: one value of passing `doEnable` as a property on toggle-able actions is that we can reuse the same action instead of creating separate ones for on/off/toggle

```ts
case 'TOGGLE_ANALYTICS_ENABLED':
	setAnalyticsEnabled( action.doEnable ?? ! state.settings.analyticsEnabled );
	break;
```

lib/state/simperium/middleware.ts
```diff
@@ -43,6 +45,36 @@ const fetchRevisions = (store: S.Store, state: S.State) => {
...
+
+const setAnalyticsEnabled = (state: S.State) => {
+  const objectKey = 'preferences-key';
+  buckets.preferences.get(objectKey, (e, preferences) => {
+    buckets.preferences.update(objectKey, {
+      ...preferences.data,
+      analytics_enabled: state.settings.analyticsEnabled,
+    });
+  });
+};
```
dmsnell, line 76: it might be helpful to leave `state` out of here and focus on the semantic of what we're doing specifically. if we were to stick to one Redux action we could even inline the call.

```ts
const { preferences } = buckets;

case 'TOGGLE_ANALYTICS':
	preferences.get( 'preferences-key', ( e, { data } ) => {
		if ( e ) {
			debug('Failed to get previous preferences data');
			break;
		}

		const analytics_enabled = action.doEnable ?? ! state.settings.analyticsEnabled;

		preferences.update( 'preferences-key', { ...data, analytics_enabled } );
	} );
	break;
```

lib/dialogs/settings/panels/account.tsx
belcherj: For sure.

dmsnell:
This is really coming together. One thing that I think is now more clear that would improve the legibility and data flow is if we created a new reducer simply for preferences and tied that to the bucket updates. This entire thing is complicated by the fact that we're abusing the way we interact with buckets, I think in order to support one or both of the mobile apps. Instead of having a `preferences` bucket with an entity named `preferences-key` (super weird, right?) we would rather have a `preferences` bucket with separate entities for each preference - then we could just update `send_analytics` instead of updating the entire bucket.

Here's how that could look…

```ts
// state/settings/reducer.ts

type Preferences = {
  analytics_enabled: boolean | null
}

const uninitializedPreferences = {
  analytics_enabled: null,
}

const preferences: A.Reducer<Preferences> = ( state = uninitializedAnalytics, action ) => {
  switch (action.type) {
    case 'TOGGLE_ANALYTICS':
      return { ...state, analytics_enabled: action.sendAnalytics ?? !state.analytics_enabled };

    case 'REMOTE_PREFERENCES_UPDATE':
      return action.preferences;

    default:
      return state;
  }
}
```

```ts
// simperium/middleware
case 'TOGGLE_ANALYTICS':
  buckets.preferences.update('preferences-key', nextState.settings.preferences );
  break;
```

lib/state/simperium/middleware.ts
```diff
@@ -60,6 +62,14 @@ export const middleware: S.Middleware = store => {
       case 'REVISIONS_TOGGLE':
         fetchRevisions(store, nextState);
         break;
+
+      case 'TOGGLE_ANALYTICS':
+        buckets.preferences.update('preferences-key', nextState.preferences);
```
dmsnell, line 67: we will want to update `window.analyticsEnabled` here as well so that it occurs immediately, otherwise we introduce a window of time in which we send analytics even though someone has opted out

lib/state/settings/actions.ts
```diff
@@ -4,6 +4,10 @@ import * as A from '../action-types';
 
 const ipc = getIpcRenderer();
 
+export const fetchPreferences: A.ActionCreator<A.FetchPreferences> = () => ({
+  type: 'FETCH_PREFERENCES',
+});
+
```
dmsnell, line 10: this can go entirely now

lib/state/preferences/reducer.ts
```diff
@@ -0,0 +1,30 @@
...
+  action
+) => {
+  switch (action.type) {
+    case 'TOGGLE_ANALYTICS':
+      return {
+        ...state,
+        analytics_enabled:
+          typeof action.sendAnalytics !== 'undefined' ||
+          !state.analytics_enabled,
+      };
```
dmsnell, line 22: I'm confused on what we're trying to accomplish here. It reads "if the action specifies whether or not to turn on the analytics then turn them on regardless, otherwise flip what we previously had"

should it be this?

```ts
typeof action.sendAnalytics !== 'undefined' ? action.sendAnalytics || !state.analytics_enabled
```

or this?

```ts
action.hasOwnProperty( 'sendAnalytics' ) ? action.sendAnalytics : !state.analytics_enabled
```

lib/state/preferences/reducer.ts
```diff
@@ -0,0 +1,30 @@
...
+    case 'ANALYTICS_REMOTE_UPDATE':
+      return { ...state, analytics_enabled: action.sendAnalytics };
```
dmsnell, line 24: two points here:

we don't have to accept this snake-case property in our app. we think about the properties as camelCase. we can preserve that by making a parsing step whenever the data comes in, and whenever the data goes out.

```ts
return { ...state, analyticsEnabled: action.sendAnalytics };
```

secondly, and explaining why some of this is confusing, it looks like we're trying to operate at the `analyticsEnabled` level _and_ at the `preferences` level at the same time. this is confusing.

_If_ we are going to treat this like any normal Simperium object we should swap it out here…

```ts
case 'PREFERENCES_REMOTE_UPDATE':
	return action.preferences;
```

…but otherwise we might want to just treat it as a simpler object, just the analytics…

```ts
const sendAnalytics: A.Reducer<boolean | null> = (state = null, action) => {
	switch (action.type) {
		case 'TOGGLE_ANALYTICS':
			return action.sendAnalytics ?? !state;

		case 'PREFERENCES_REMOTE_UPDATE':
			return action.data.sendAnalytics;
	}
}
```

We don't have to care about all the preferences object, but we should make sure we don't handicap ourselves from the onset.

```ts
const parsePreferences = ( { send_analytics, ...preferences } ) => ( {
	...preferences,
	sendAnalytics: send_analytics,
} );

const unparsePreferences = ( { sendAnalytics, ...preferences } ) => ( {
	...preferences,
	send_analytics: sendAnalytics
} );

// app.tsx
preferencesBucket.on('index', () => {
    preferencesBucket.get('preferences-key', (error, preferences) => {
        if (error || !preferences.data.hasOwnProperty('analytics_enabled')) {
            return;
        }

        this.props.remotePreferencesUpdate( parsePreferences( preferences.data ) );
    });
});

preferencesBucket.on('update', (entityId, data) => {
	if (
		'preferences-key' !== entityId ||
		!data.hasOwnProperty('analytics_enabled')
	) {
		return;
	}

	this.props.remotePreferencesUpdate( parsePreferences( data ) );
});

// middleware.ts
case 'TOGGLE_ANALYTICS':
	buckets.preferences.update( 'preferences-key', unparse( nextState.preferences ) );
```

dmsnell: the naming thing isn't an issue with the `note` bucket because the names are already in the form we expect, otherwise we would do that there too. eventually we will probably want this parsing step anyway so we can stop doing things like `if ( note.data.systemTags.includes( 'pinned' ) )` and instead reliably rely on `if ( note.isPinned )`

these are enhancements not necessary yet, but in the case of preferences something felt more pertinent to me when I reviewed them

codebykat (commented):
I have two not-very-important comments here. One is that I also dislike mixing snakes 🐍 🐍 🐍  in with our camels 🐫 🐫 🐫 .

The second is that we should consider using set/unset terminology when passing a boolean and "toggle" for actions that do not pass an argument, but flip a thing to the opposite of whatever it's set to. I believe this isn't the first time this has come up but we are not at all consistent about it in the codebase right now and I think it has the potential to be confusing.

Anyway these are both pretty nitpicky and in a big picture sense, this looks fine and seems to work. I like the preferences in its own object/reducer.

codebykat:
BTW we can probably take this out of draft status no?

codebykat:
On second thought, why are we adding preferences when we already have settings?

belcherj:
Closing as this will be addressed in another refactor.

PR #1981, opened 2020-03-28
State Refactor: Move dialog renderer to redux

codebykat:
### Fix
This moves showDialog and closeDialog into Redux state, in the process refactoring `DialogRenderer` to be slightly less overly complicated. This includes passing a simple string to `showDialog` instead of a full `DialogType` object, eliminating the unique key (`closeDialog` will always just close the topmost dialog) and eliminating the `multiple` option, which we weren't using.

### Test
1. Test all dialog types (Settings, Import, About, Share)
2. Test using menu items in Electron
3. Test stuff the dialogs do (toggling settings, publishing note, importing notes, etc.)

### Release

TBD


lib/app.tsx
```diff
@@ -258,10 +262,6 @@ export const App = connect(
         return window.print();
       }
 
-      if ('focusSearchField' === command.action) {
-        return this.props.focusSearchField();
-      }
-
```
dmsnell: why is the import added above and then this block removed from here?

lib/types.ts
```diff
@@ -61,6 +61,7 @@ export type Bucket<T = unknown> = {
 ///////////////////////////////////////
 // Application Types
 ///////////////////////////////////////
+export type DialogType = 'ABOUT' | 'IMPORT' | 'SETTINGS' | 'SHARE';
```
dmsnell: Wonderful glorious typer you are!

lib/dialog-renderer/index.tsx
```diff
@@ -1,64 +1,96 @@
...
+export class DialogRenderer extends Component<Props> {
+  static displayName = 'DialogRenderer';
+
+  static propTypes = {
+    appProps: PropTypes.object.isRequired,
+    buckets: PropTypes.object,
+    themeClass: PropTypes.string,
+    isElectron: PropTypes.bool.isRequired,
+    isMacApp: PropTypes.bool.isRequired,
+  };
```
dmsnell: Considering that we appear to have added these here we might as well make then _types_ instead of `propTypes` since it will remove some code from the runtime but enhance our safety on the compile-time

```ts
type OwnProps = {
  appProps: object;
  buckets: Record<'noteBucket' | 'tagBucket' | 'preferencesBucket', T.Bucket>;
  themeClass: string;
  isElectron: boolean;
  isMacApp: boolean;
}
```

We could also just make `appProps` and `buckets` type `object` which would satisfy enough. looking at these props highlights that we're passing too much down through the components here and while I don't think it would be good to muddle this PR with it, I think a follow-up PR is warranted where we extract the props.

looking at the different dialogs it seems like we need three types of props and we can probably eliminate all of them:
 - redux actions which formerly had to be passed down but which can be imported on the component that uses them (`closeDialog()` for instance)
 - Simperium buckets for directly manipulating and inspecting the Simperium data, which probably would be better handled in the Simperium middleware (such as the warning not to sign-out if there are pending changes). This very well could be one source of our headaches when occasionally we can't logout.
 - Props that aren't used. Some of the old Simperium props appear to be left in place even while updating has been moved to Redux instead.

lib/dialog-renderer/index.tsx
```diff
@@ -1,64 +1,96 @@
...
-    themeClass,
-    closeDialog,
-    dialogs,
-    isElectron,
-    isMacApp,
-  } = props;
-
-  const renderDialog = dialog => {
-    const { key, title } = dialog;
     const DialogComponent = Dialogs[dialog.type];
```
dmsnell: This little bit of magic shouldn't work and probably only does because of the fact that `webpack` isn't entirely doing what it's supposed to be doing. We can fix the imported `../dialogs` file but probably the best thing to do would be to `case` this for now, which would also remove that additional layer of implicit logic or string-based magic.

```ts
import * as Dialogs from '../dialogs'

const getDialogComponent = type => {
  switch (type) {
    case 'About': return Dialogs.About;
    case 'Import': return Dialogs.Import;
    case 'Settings': return Dialogs.Settings;
    case 'Share': return Dialogs.Share;
  }
}
```

we could even make the function look like the variable itself…

```ts
const DialogComponent = type => {
  switch (type) {
    case 'About': return Dialogs.About;
    case 'Import': return Dialogs.Import;
    case 'Settings': return Dialogs.Settings;
    case 'Share': return Dialogs.Share;
  }
}

<DialogComponent(dialogType) buckets=… />
```

lib/dialog-renderer/index.tsx
```diff
@@ -1,64 +1,96 @@
...
 
     if (DialogComponent === null) {
       throw new Error('Unknown dialog type.');
     }
 
-    const closeThisDialog = () => closeDialog({ key });
```
dmsnell: yay - less accounting on our part by choosing a more appropriate data structure

lib/dialog-renderer/index.tsx
```diff
@@ -1,64 +1,96 @@
...
     );
-  };
+  }
 
-  return <Fragment>{dialogs.map(renderDialog)}</Fragment>;
-};
+  render() {
+    return (
+      <Fragment>{this.props.dialogs.map(this.renderDialog, this)}</Fragment>
+    );
```
dmsnell: It seems like our rendering function isn't doing much and in general we can inline it here instead of abstracting it into a function.

```ts
render() {
  const { appProps, buckets, dialogs, isElectron, isMacApp, themeClass } = this.props;

  return (
    <Fragment>
      { dialogs.map( dialog => (
        <Modal … >
          { 'About' === dialog.type ? (
            <AboutDialog /> // no props needed
          ) : 'Import' === dialog.type ? (
            <ImportDialog buckets={buckets} isElectron={isElectron} />
          ) : 'Settings' === dialog.type ? (
            <SettingsDialog buckets={buckets} isElectron={isElectron} isMacApp={isMacApp} {...appProps} />
          ) : 'Share' === dialog.type ? (
            <ShareDialog /> // also I don't think any props are needed here
          ) : null
          }
        </Modal>
      ) ) }
    </Fragment>
  );
}
```

This also seems to suggest that the `dialogs` objects are unnecessary and your new `DialogType` should suffice as a string itself. The only use that I saw was that the dialogs render their title as `dialog.title`, which doesn't make sense since we have no other title for the `About` dialog than what it itself wants to render. Might as well take that out.

lib/note-detail/index.tsx
```diff
@@ -238,7 +239,7 @@ const mapStateToProps: S.MapState<StateProps> = ({
   ui,
   settings,
 }) => ({
-  dialogs: state.dialogs,
+  dialogs: ui.dialogs,
```
dmsnell: based on the usage inside the component I think it would be an appropriate update to change this from passing in the list of opened dialogs so that it passed in whether or not dialogs are open.

```ts
const mapStateToProps…
isViewingDialog: ui.dialogs.length > 0,
```

lib/state/action-types.ts
```diff
@@ -63,6 +64,7 @@ export type RemoteNoteUpdate = Action<
   { noteId: T.EntityId; data: T.Note }
 >;
 export type RestoreNote = Action<'RESTORE_NOTE', { previousIndex: number }>;
+export type ShowDialog = Action<'SHOW_DIALOG', { dialog: string }>;
```
dmsnell:
```ts
{ dialog: T.DialogType }
```

lib/state/ui/reducer.ts
```diff
@@ -7,6 +7,22 @@ import * as T from '../../types';
 
 const emptyList: unknown[] = [];
 
+const dialogs: A.Reducer<T.DialogType[]> = (state = [], action) => {
```
dmsnell: if we clear up the `{dialog: T.DialogType}` in the `action-types` file then the TypeScript error here goes away too

lib/state/ui/reducer.ts
```diff
@@ -7,6 +7,22 @@ import * as T from '../../types';
...
+  switch (action.type) {
+    case 'CLOSE_DIALOG':
+      return state.slice(0, -1);
+
+    case 'SHOW_DIALOG':
+      return [...state, action.dialog];
+
+    case 'App.authChanged':
+      return [];
```
dmsnell: this case looks weird. is this a result of signing out?
is there no sign-out action that's run?

lib/state/ui/reducer.ts
```diff
@@ -162,8 +178,8 @@ const showNavigation: A.Reducer<boolean> = (state = false, action) => {
     case 'SELECT_TRASH':
     case 'SHOW_ALL_NOTES':
       return false;
-    case 'App.showDialog':
-      if (action.dialog && action.dialog.type === 'Settings') {
+    case 'SHOW_DIALOG':
+      if (action.dialog && action.dialog === 'SETTINGS') {
```
dmsnell: we can remove these safety checks now because TypeScript will warn if we reference a property that might be missing. in fact, it autocompletes for me purely based on the action type

autoCompleteDialog mov

lib/app.tsx
codebykat: it doesn't need to be special-cased if it's explicitly imported. I think having the special case here block was a mistake

codebykat: (also, sorry, this is a fix unrelated to the dialogs, I just saw it in passing 🤷‍♀ )

lib/dialog-renderer/index.tsx
codebykat: I toyed with removing the `closeDialog` function passing, but it isn't quite as simple as just adding it to the `dialog` component. The import dialog also sets that to null while an import is happening, and allows closing again after it completes. I wasn't sure how to handle that case without prop passing and it was seeming increasingly tangential to the purpose... do you think `dialog` should handle it, and if so, what should `import` do to temporarily disable closing?

dmsnell: > The import dialog also sets that to null while an import is happening, and allows closing again after it completes.

Yeah I missed this. It's something I think will be clearer after this merges. I think that if we do what I mentioned about making the dialog choice explicit/cased then that will help because for the simple cases we can leave the prop out and let the more complicated case sit for its own work.

lib/state/action-types.ts
codebykat: THANK YOU lol I looked for it and couldn't find the one I missed!

lib/state/ui/reducer.ts
codebykat: I actually don't know, I moved it over from here: https://github.com/Automattic/simplenote-electron/blob/develop/lib/flux/app-state.ts#L29

lib/dialog-renderer/index.tsx
codebykat: Update: less magic 😿

lib/note-detail/index.tsx
codebykat: I don't actually even know how to hit this code path

lib/dialog-renderer/index.tsx
codebykat: Just wanted to note I didn't add the props here, I just moved them: https://github.com/Automattic/simplenote-electron/blob/develop/lib/dialog-renderer/index.tsx#L54

I turned the `DialogRenderer` into a component, I'm not sure what it was before.

codebykat:
Ooooookay. Major update complete. As per https://github.com/Automattic/simplenote-electron/pull/1981#discussion_r400566383 I axed the `DialogTypes` from `DialogRenderer` and circumvented the need for it to import `closeDialog` by moving the `<Modal>` into the individual dialogs (which seems to make more sense anyhow). I did end up having to pass around `themeClass` though and moving that into state might be more appropriate for a follow-up PR.

I'm not sure if it makes sense for the `Dialog` component to import `closeDialog` instead of having it passed in, given that the parent will have to load the function as well?

This was pretty major surgery so all dialogs need a thorough smoke-test again (in both web and Electron).

lib/app.tsx
dmsnell: > I think having the special case block here was a mistake. (The following lines check to see if it's a callable action and then call it.)

ah that makes sense. it _wasn't_ a mistake - I was rebelling against the next few lines by trying to make it explicit what actions we intend on responding to so we can eventually get rid of that `canCall` thing. too many headaches trying to understand the control flow of actions from Electron

lib/dialog-renderer/index.tsx
dmsnell: I missed them at the end of the file. would you mind changing them to the `Props` type here anyway? we're still splitting it up into the typed variants for state and it's just a few lines. no biggie if you don't want to - we can do it later

lib/state/ui/reducer.ts
dmsnell: yeah okay there it is - https://github.com/Automattic/simplenote-electron/commit/0c033413fec2dc050c68eec8ac849a764978068f

so I guess that what happens is you sign out, and if we don't clear it, then the app re-opens with the dialogs still open. I'm ear-marking that for a better action-type like `SIGN_OUT`, which I presume we'll get when we refactor `App.authChanged`

lib/dialog-renderer/index.tsx
```diff
@@ -1,64 +1,87 @@
...
-          requestClose={closeThisDialog}
-          isElectron={isElectron}
-          isMacApp={isMacApp}
-          {...appProps}
-        />
-      </Modal>
+      <Fragment>
+        {this.props.dialogs.map(dialog =>
+          'ABOUT' === dialog ? (
+            <AboutDialog key="dialog-renderer__about" themeClass={themeClass} />
```
dmsnell: keys _only_ need to be unique up to the closest parent, so we are free to drop the prefix here - I think, unless it yelled at you because `<Fragment>` isn't a real parent.

```ts
key="about"
```

lib/dialogs/about/index.tsx
```diff
@@ -1,125 +1,151 @@
...
+    const { closeDialog, themeClass } = this.props;
     const thisYear = new Date().getFullYear();
 
     return (
-      <div className="about">
-        <Dialog hideTitleBar onDone={requestClose} title={dialog.title}>
-          <div className="about-top">
-            <SimplenoteLogo />
+      <Modal
+        key="dialog-renderer__content__about"
```
dmsnell: shouldn't need the `key` in here since this we're not inside a list

lib/dialogs/about/index.tsx
```diff
@@ -1,125 +1,151 @@
...
   };
 
   render() {
-    const { dialog, requestClose } = this.props;
+    const { closeDialog, themeClass } = this.props;
     const thisYear = new Date().getFullYear();
 
     return (
-      <div className="about">
-        <Dialog hideTitleBar onDone={requestClose} title={dialog.title}>
```
dmsnell: oh who'da'thunk it? I didn't think about moving the `Modal` back into the dialogs. did this end up feeling better to you than wrapping them in the `<Modal>`?

codebykat: It's a bit repetitive, BUT means the parent doesn't have to load the close function or determine the modal titles, so on the whole, yes. But I'm open to differing opinions.

lib/dialog-renderer/index.tsx
codebykat: Yep NBD

lib/dialogs/about/index.tsx
codebykat: haha clearly I don't understand the rules for keys

lib/app.tsx
codebykat: Oh!!! Okay, should I revert it? Right now since it's the only special-cased one it seemed wrong.

codebykat: Reverted!

lib/dialogs/about/index.tsx
codebykat: OK I lied I think I like this better https://github.com/Automattic/simplenote-electron/pull/1981/commits/b98f9606cd71bf987fb4f3b627910e80e136f921

lib/dialog-renderer/index.tsx
```diff
@@ -1,64 +1,95 @@
...
-  themeClass: PropTypes.string,
-  closeDialog: PropTypes.func.isRequired,
-  dialogs: PropTypes.array.isRequired,
-  isElectron: PropTypes.bool.isRequired,
-  isMacApp: PropTypes.bool.isRequired,
-};
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = dispatch => ({
+  closeDialog: () => {
+    dispatch(closeDialog());
+  },
```
dmsnell, line 92: just a nitpick, but we can collapse this one of two ways, or in two steps:

```ts
// the extra curlies aren't necessary
const mapDispatchToProps: S.MapDispatch<DispatchProps> = dispatch => ({
   closeDialog: () => dispatch(closeDialog())
});
```

```ts
// and we don't need to create a function here largely because our action-creators
// don't depend on additional parameters
const mapDispatchToProps: S.MapDispatch<DispatchProps> = { closeDialog };
```

at least the first step is nice. the second one is more for your choosing.

lib/dialogs/settings/panels/tools.tsx
```diff
@@ -43,9 +48,8 @@ ToolsPanel.propTypes = {
 };
 
 const mapDispatchToProps = dispatch => {
-  const { showDialog } = appState.actionCreators;
   return {
-    showImportDialog: () => dispatch(showDialog({ dialog: IMPORT })),
+    showImportDialog: () => dispatch(showDialog('IMPORT')),
```
dmsnell, line 52: we're in a type-mismatch here that doesn't show because we haven't added a type signature to `mapDispatchToProps`.

If we were to do that…

```ts
const mapDispatchToProps: S.MapDispatch<DispatchProps> = …
```

then the compiler would warn us that we're not sending `showDialog`

We can either update `ToolsPanel` to call `showDialog` and update this reference in `mapDispatchToProps` or we can rename the type in `DispatchProps` to `showImportDialog`

The final piece is adding the type to the component itself…

```ts
import React, { FunctionComponent } from 'react';

type Props = DispatchProps;

const ToolsPanel: FunctionComponent<Props> = ( … ) => …
```

codebykat: Thanks! I couldn't tell what this... WAS. Or is. There are a number of these kicking around the codebase that aren't defined as either Components or FunctionComponents. They normally have a signature like this one and a PropTypes definition at the end.

codebykat:
Updated again! Think I got all the types fixed 🤞

dmsnell:
## Testing

- [ ] Logout
- [ ] Login with wrong password fails
- [ ] Login with correct password
- [ ] Create new note appears in other device
- [ ] Changes to new note sync to other device
- [ ] Changes to new note sync _from_ other device (after debounce delay)
- [ ] New tag immediately syncs to other device
- [ ] New tag immediately syncs _from_ other device
- [ ] Removed tag immediately syncs to other device
- [ ] Removed tag immediately syncs _from_ other device
- [ ] Note publishes with link
- [ ] Note unpublishes
- [ ] Note publish change syncs _from_ other device (visible with dialog open)
- [ ] Markdown setting syncs _from_ other device
- [ ] Preview mode disappears/reappears when receiving remote changes to markdown setting
- [ ] Note pinning syncs immediately from either direction
- [ ] Note pinning works regardless if clicking in list view or from note info
- [ ] Viewing history on one device leaves note unchanged on other device
- [ ] Restoring history immediately syncs note from both directions
- [ ] Restoring history revisions updates pinned status
- [ ] Restoring history revisions updates markdown status
- [ ] Restoring history revisions updates publish status
- [ ] Can view trashed notes by clicking on `Trash`
- [ ] Can delete note forever from trash screen
- [ ] Can restore note from trash screen
- [ ] Can trash note
- [ ] Can preview markdown with 👁 button
- [ ] Can flip to edit mode with 👁 button
- [ ] Markdown rendered in note list when markdown enabled
- [ ] Markdown syntax unrendered in note list when markdown disabled
- [ ] Can filter by tag when clicking on tag in tag drawer
- [ ] Can add tag to note and have it appear in filtered tag view when previously not in filter
- [ ] Searching in the search field highlights matches in note list
- [ ] Searching in the search field highlights matches in the note editor
- [ ] Searching in the search field highlights matches in the note editor during preview
- [ ] Clearing the search field immediately updates filtered notes
- [ ] Clicking on different tags or `All Notes` or `Trash` immediately updates filtered notes
- [ ] Can search by keyword, filtered instantly
- [ ] Tag auto-completes appear when typing in search field
- [ ] Typing `tag:` does not result a list of all tags in the autocompleter
- [ ] Typing `tag:` and something else, like `tag:te` results in autocomplete results starting with that something else, e.g. `test`
- [ ] Tag suggestions suggest tags regardless of case
- [ ] Search field updates with results of `tag:test` format search string
- [ ] Can toggle sidebar
- [ ] Syncs when introducing sequential surrogate pairs sharing the same high surrogate, e.g. `🅰🅱` to `🅰🅰🅱`
- [ ] When disabling network connectivity, changes start counting in unsync'ed changes counter
- [ ] When going back online changes sync and counter resets to `All chagnes synced`
- [ ] Can change analytics sharing setting
- [ ] Changing `Note Display` mode immediately updates and reflects in note list
- [ ] Changing `Sort Type` mode immediately updates and reflects in note list
- [ ] Toggling `Sort Order` immediately updates and reflects in note list for each sort type
- [ ] For each sort type the pinned notes appear first in the note list
- [ ] Changing `Theme` immediately updates app for desired color scheme
- [ ] Using the `Insert checklist` item from the format menu inserts a checklist
- [ ] "Undo" undoes the last edit
- [ ] Typing `- [x]` creates a checked checklist item
- [ ] Typing `- [ ]` created an unchecked checklist item
- [ ] Typing `-` creates a list
- [ ] Typing _tab_ in a list item underneath another list item indents item
- [ ] Typing _shift tab_ in the same spot unindents the list
- [ ] Changing `-` to `+` changes the list item bullet, also for `*` and `•` (`\u2022`)
- [ ] All list bullet types render to markdown lists
- [ ] Added URL is linkified
- [ ] When clicking on link it opens in new window
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + <kbd>p</kbd> toggles preview mode
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + (<kbd>k</kbd> or <kbd>up</kbd>) selects next note above current note, stops at top of list
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + (<kbd>j</kbd> or <kbd>down</kbd>) selects next note below current note, stops at bottom of list
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>t</kbd> toggles tab list
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>n</kbd> creates new note
- [ ] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + <kbd>n</kbd> _when in small screen mode_ navigates to note list
- [ ] Can print note in plaintext view
- [ ] Can print note in Markdown-preview view

PR #1982, opened 2020-04-02
Search: Move out of WebWorker and back into main thread

dmsnell:
Fixes #1941
Supplants #1966
Supplants #1979

It turns out that having search inside a WebWorker isn't as necessary
as we thought and at the same time presents one major obstacle: search
must be asynchronous. This causes a few problems with the app is
currently designed, namely that the interaction between `previousIndex`,
trash/delete/restore operations, and the note list.

The primary goal of the larger state-refactor project has been to
eliminate non-atomic state updates and the WebWorker's asyncronous
mandate means that there is no way to synchronously update state, which
means that there's no way to run the trashing actions at the same time
that we update the search filter. This leaves an awkward rendered state
where the note in the note list is trashed and the toolbar above it
shows the trash "Delete forever" and "Restore" buttons but we're still
looking at the note list.

In order to resolve these bugs and eliminate further issues I have
brought the search back into the main thread. Why? Won't this destroy
all that we gained in terms of performance?" you might ask. No, actually
most of the performance gain came from changes I made to the search
mechanism _while_ moving it into a WebWorker. I had wanted it to be in
a WebWorker because it was slow, but now that we can see that it's
very very fast and shouldn't be a UI blocker we don't have the same
motivation to get it out of the main thread.

In summary, this patch moves the search back into the main thread and
exposes the `updateFilter()` function so that it can be called synchronously
for actions which demand immediate UI updates.

## Testing

Run the testing checklist.
Pay attention to the trashing/restoring/deleting actions as we discovered a
regression from search work when testing #1981


codebykat:
## Testing

- [x] Logout
- [x] Login with wrong password fails
- [x] Login with correct password
- [x] Create new note appears in other device
- [x] Changes to new note sync to other device
- [x] Changes to new note sync _from_ other device (after debounce delay)
- [x] New tag immediately syncs to other device
- [x] New tag immediately syncs _from_ other device
- [x] Removed tag immediately syncs to other device
- [x] Removed tag immediately syncs _from_ other device
- [x] Note publishes with link
- [x] Note unpublishes
- [x] Note publish change syncs _from_ other device (visible with dialog open)
- [x] Markdown setting syncs _from_ other device
- [x] Preview mode disappears/reappears when receiving remote changes to markdown setting
- [x] Note pinning syncs immediately from either direction
- [x] Note pinning works regardless if clicking in list view or from note info
- [x] Viewing history on one device leaves note unchanged on other device
- [x] Restoring history immediately syncs note from both directions
- 🚫 Restoring history revisions updates pinned status
- 🚫 Restoring history revisions updates markdown status
- 🚫 Restoring history revisions updates publish status
- [x] Can view trashed notes by clicking on `Trash`
- [x] Can delete note forever from trash screen
- [x] Can restore note from trash screen
- [x] Can trash note
- [x] Selects note above recently-trashed note
- [x] Selects note above recently-restored note in trash view
- [x] Selects note above recently-deleted-forever note in trash view
- [x] Can preview markdown with 👁 button
- [x] Can flip to edit mode with 👁 button
- 🚫 Markdown rendered in note list when markdown enabled
- [x] Markdown syntax unrendered in note list when markdown disabled
- [x] Can filter by tag when clicking on tag in tag drawer
- [x] Can add tag to note and have it appear in filtered tag view when previously not in filter
- [x] Can search by keyword with tag selected
- [x] Searching in the search field highlights matches in note list
- [x] Searching in the search field highlights matches in the note editor
- 🚫 Searching in the search field highlights matches in the note editor during preview
- [x] Clearing the search field immediately updates filtered notes
- [x] Clicking on different tags or `All Notes` or `Trash` immediately updates filtered notes
- [x] Can search by keyword, filtered instantly
- [x] Tag auto-completes appear when typing in search field
- [x] Typing `tag:` does not result a list of all tags in the autocompleter
- [x] Typing `tag:` and something else, like `tag:te` results in autocomplete results starting with that something else, e.g. `test`
- [x] Tag suggestions suggest tags regardless of case
- [x] Search field updates with results of `tag:test` format search string
- [x] Can toggle sidebar
- [x] Syncs when introducing sequential surrogate pairs sharing the same high surrogate, e.g. `🅰🅱` to `🅰🅰🅱`
- [x] When disabling network connectivity, changes start counting in unsync'ed changes counter
- [x] When going back online changes sync and counter resets to `All changes synced`
- [x] Can change analytics sharing setting
- [x] Changing `Note Display` mode immediately updates and reflects in note list
- [x] Changing `Sort Type` mode immediately updates and reflects in note list
- [x] Toggling `Sort Order` immediately updates and reflects in note list for each sort type
- [x] For each sort type the pinned notes appear first in the note list
- ❓ Changing `Theme` immediately updates app for desired color scheme **System theme offered but not working in browser**
- [x] Using the `Insert checklist` item from the format menu inserts a checklist
- [x] "Undo" undoes the last edit
- [x] Typing `- [x]` creates a checked checklist item
- [x] Typing `- [ ]` created an unchecked checklist item
- [x] Typing `-` creates a list
- [x] Typing _tab_ in a list item underneath another list item indents item
- [x] Typing _shift tab_ in the same spot unindents the list
- ❓ Changing `-` to `+` changes the list item bullet, also for `*` and `•` (`\u2022`) -- **doesn't change following list items, what does this mean?**
- [x] All list bullet types render to markdown lists
- ❓ Added URL is linkified **Only in preview**
- [x] When clicking on link it opens in new window
- 🚫  <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + <kbd>p</kbd> toggles preview mode **not in browser, it opens a private window**
- [x] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + (<kbd>k</kbd> or <kbd>up</kbd>) selects next note above current note, stops at top of list
- [x] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + (<kbd>j</kbd> or <kbd>down</kbd>) selects next note below current note, stops at bottom of list
- [x] <kbd>CmdOrCtrl</kbd> + <kbd>t</kbd> toggles tab list
- 🚫  <kbd>CmdOrCtrl</kbd> + <kbd>n</kbd> creates new note **in browser this is a new window**
- [x] <kbd>CmdOrCtrl</kbd> + <kbd>shift</kbd> + <kbd>n</kbd> _when in small screen mode_ navigates to note list
- [x] Can print note in plaintext view
- [x] Can print note in Markdown-preview view

dmsnell:
> System theme offered but not working in browser

This worked for me in macOS, I had to activate Windows to test it so I'll have to try once I reboot into my real Windows install, and I wasn't sure how or if Ubuntu/Gnome/Linux in general supports Light/Dark mode.

My suspicion is that we can do this with [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) but maybe we need to investigate more.

PR #1983, opened 2020-04-02
Add keyboard shortcuts explainer

### LINKED ISSUE #700: Shortcut Documentation

### LINKED ISSUE #798: Ctrl+Shift+Up/Down should do text selection, not go up/down the notes list
#### Steps to reproduce
1. Open a note
2. Select text using Ctrl+Shift+Left/Right keys, it works
3. Now try selecting text using Ctrl+Shift+Up/Down keys, doesn't work

#### What I expected
The text selection would select the entire line as in how all office programs, text editors and note apps function.

#### What happened instead
The notes list got navigated instead using the said key combination which is not in line with any office program, text editor or other notes apps.

#### OS version
Ubuntu 18.04

### LINKED ISSUE #985: Keyboard shortcut for preview on / off

### LINKED ISSUE #1167: Keyboard shortcuts are not documented
Keyboard shortcuts are not documented at all, for the Linux (Electron?) app and the web app.

I would like them to be visible somewhere in the app.

"Preview markdown", which is useful during writing a note, would be the most important for me. I have found out here: https://github.com/Automattic/simplenote-electron/issues/37 that this function exists, but that shortcut doesn't work.

I have found a list of working shortcuts there: https://github.com/Automattic/simplenote-electron/issues/700 but "preview markdown" is missing.

### LINKED ISSUE #1222: Shortcuts for switching mode
#### What I expected

It should have shortcuts to switch between edit and preview mode.

#### Simplenote version

version 1.4.0

#### OS version

Ubuntu 18.04

dmsnell:
Follows #603
Resolves #700
Resolves #798
Resolves #985
Resolves #1167
Resolves #1222

This application has had keyboard shortcuts since #603 but they have
been undocumented making them hard to discover and remember.

In this patch we're picking up the work started in #700 and adding
a keyboard shortcut overylay, opened through the Help menu or by
pressing `CmdOrCtrl` + `?`. This dialog shows a list of the
app-specific keyboard shortcuts and what they do.

Oh yeah, and we're changing the "Find" shortcut to <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>F</kbd>
in accordance with earlier design decisions to differentiate "search within
this note" vs. "search all notes" and also removes the <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Up</kbd>/<kbd>Down</kbd>
from note navigation so that it stops interfering with text selection.

Screen Shot 2020-04-01 at 10 30 54 PM

keyboardExplainer mov

Screen Shot 2020-04-16 at 7 42 38 PM

## Testing

Run through the testing checklist.
Pay particular attention to keyboard shortcuts.
Try to break them and use them in unintended circumstances.

### DIFF
--- a/desktop/menus/edit-menu.js
+++ b/desktop/menus/edit-menu.js
@@ -33,17 +33,17 @@ const buildEditMenu = settings => {
         label: '&Select All',
         role: 'selectall',
       },
+      { type: 'separator' },
       {
-        type: 'separator',
+        label: '&Trash Note',
+        click: appCommandSender({ action: 'trashNote' }),
       },
+      { type: 'separator' },
       {
         label: 'Search &Notes…',
-        accelerator: 'CommandOrControl+Shift+F',
         click: appCommandSender({ action: 'focusSearchField' }),
       },
-      {
-        type: 'separator',
-      },
+      { type: 'separator' },
       {
         label: 'C&heck Spelling',
         type: 'checkbox',
--- a/desktop/menus/edit-menu.js
+++ b/desktop/menus/edit-menu.js
@@ -38,7 +38,7 @@ const buildEditMenu = settings => {
       },
       {
         label: 'Search &Notes…',
-        accelerator: 'CommandOrControl+F',
+        accelerator: 'CommandOrControl+Shift+F',
         click: appCommandSender({ action: 'focusSearchField' }),
       },
       {
--- a/desktop/menus/file-menu.js
+++ b/desktop/menus/file-menu.js
@@ -5,7 +5,6 @@ const { appCommandSender } = require('./utils');
 const submenu = [
   {
     label: '&New Note',
-    accelerator: 'CommandOrControl+N',
     click: appCommandSender({ action: 'newNote' }),
   },
   { type: 'separator' },
--- a/desktop/menus/format-menu.js
+++ b/desktop/menus/format-menu.js
@@ -3,7 +3,6 @@ const { appCommandSender } = require('./utils');
 const submenu = [
   {
     label: 'Insert &Checklist',
-    accelerator: 'CommandOrControl+Shift+C',
     click: appCommandSender({ action: 'insertChecklist' }),
   },
 ];
--- a/desktop/menus/help-menu.js
+++ b/desktop/menus/help-menu.js
@@ -4,11 +4,17 @@ const menuItems = require('./menu-items');
 const platform = require('../detect/platform');
 const build = require('../detect/build');
 
+const { appCommandSender } = require('./utils');
+
 const submenu = [
   {
     label: 'Help && &Support',
     accelerator: platform.isLinux() ? 'F1' : null,
     click: () => shell.openExternal('https://simplenote.com/help'),
+  },
+  {
+    label: '&Keyboard Shortcuts',
+    click: appCommandSender({ action: 'showDialog', dialog: 'KEYBINDINGS' }),
   },
   { type: 'separator' },
   {
--- a/lib/app-layout/index.tsx
+++ b/lib/app-layout/index.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent, Suspense } from 'react';
+import React, { Component, Suspense } from 'react';
 import classNames from 'classnames';
 import { connect } from 'react-redux';
 
@@ -8,6 +8,7 @@ import RevisionSelector from '../revision-selector';
 import SearchBar from '../search-bar';
 import SimplenoteCompactLogo from '../icons/simplenote-compact';
 import TransitionDelayEnter from '../components/transition-delay-enter';
+import actions from '../state/actions';
 
 import * as S from '../state';
 import * as T from '../types';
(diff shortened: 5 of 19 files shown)


SylvesterWilmott:
This is great!

codebykat:
Very cool! I have a few comments:

* The height is a bit awkward (it's just... sliiiiiightly too tall for the content). Especially on browser it looks ugly that it winds up with a scrollbar.

Screen Shot 2020-04-02 at 5 14 37 PM

* Loooooots of these don't actually work in the browser, so we should fix them or consider removing them.

* It's also not ideal that there's no way to view this dialog in the browser without already knowing the magical key combination. In related news, the key combo itself is not in the dialog? If you're in the app and open it via the menu, useful info -- and especially useful info because now you know how to open it in the browser, in fact the only way to open it in the browser.

* A few of them work with either Cmd or Ctrl (Mac/Firefox), which is odd.

codebykat:
Things that don't work in Firefox/OSX:

`ctrl + shift + F` appears to highlight the next character. `cmd +shift + f` is full screen
`ctrl +`, `ctrl -`, and `ctrl 0` do nothing -- with `cmd` instead of `ctrl`, they do adjust font size in the browser
`ctrl + N` goes to next line in editor, `cmd + N` is a new browser window
`ctrl + P` does nothing, `cmd + P` is the browser print dialog
`ctrl + shift + C` and `cmd + shift + C` do nothing
`ctrl + shift + N` obviously only does anything if you're in a small window

dmsnell:
@codebykat I have reviewed all the shortcuts and made some adjustments. I'll post more later, but I tested in Firefox and Safari and Electron.

dmsnell:
> It's also not ideal that there's no way to view this dialog in the browser without already knowing the magical key combination. In related news, the key combo itself is not in the dialog?

Yeah I don't know what to do about discoverability but I did add it to the list. Maybe we can put it somewhere in the note list or in settings?

> A few of them work with either Cmd or Ctrl (Mac/Firefox), which is odd.

I believe that's the legacy of the cross-platform support. Any ideas? We can probably check `isMacOS` where we listen for these but it seemed fine to allow some hotkeys to work across all platforms.

> Loooooots of these don't actually work in the browser, so we should fix them or consider removing them.

Should be fixed as of 7b90418 - I removed tips from the browser that only work in Electron and removed keys already handled by the browser, like font zooming and printing.

dmsnell:
We need some more hotkeys…

## Essential

 - Open trash view
 - Open all notes view
 - Navigate tags in tag list

## Nice to have

 - Restore note
 - Delete forever
 - Pin note
 - Toggle Markdown Enabled

belcherj:
Are all of Kat's comments addressed? Is it time for another review here?

dmsnell:
@belcherj they are all addressed and yes it's time for a new review.

I expect some discussion about how I addressed them but I tended to each point in the way that seemed best as I was doing it

dmsnell:
> Delete note shortcut doesn't seem to work anywhere, with any modifier key. Tested in Electron and Firefox.

Updated in a816e4d to accept backspace or delete.

> The shortcut to open this is given as modifier + ?, but it's actually modifier + / 

Used <kbd>/</kbd> in 6cd7beb

> Settings isn't dismissed by hitting the same hotkey again, only by ESC. However the keyboard shortcuts panel is toggled by its hotkey, so I think Settings should exhibit the same behavior.

I'd like to hold off on this for now due to questions about the change. It's not hard, but I'm not sure it's a positive change.

> The keyboard shortcuts panel stays open if you open e.g. Settings on top of it; it also allows you to toggle navigation elements in the background (e.g. "toggle tag list"). That seems basically fine but wanted to make sure it was intentional.

It was intentional to allow opening the keyboard shortcuts from anyway. Basically it was the refactoring of the dialog logic that gave me the idea. It wasn't intentional to make it non-modal, but I don't know if that's something we want to tackle here either. I think it would take a lot of work to prevent other interactions and I'm not sure they are problematic in the first place. Thoughts?

> In Electron it's showing my modifier key as CMD, in Firefox as CTRL. In general this is fine because that's the one that, indeed, works with the most combinations. There are still some inconsistencies between the two but those are mostly fine.

Any idea how to resolve this? I left some permissive, where either modifier works, so that people jumping between platforms can use their learned keyboard shortcuts. Balancing that out I didn't want to avoid the platform norms for the sake of consistency in our app.

> The weirdest one is "insert checklist" which ONLY works with CMD in Electron, but works with both in Firefox. It seems like it's the only one that's not a browser-specific behavior (i.e. print, font size, other stuff that is in menus) that isn't working with both modifiers.

Good catch that's apparently a result of trapping the key through Electron vs. in the app itself. Replaced Electron keybinding in 891b65c

> Should we also include TAB / SHIFT+TAB to indent / unindent list items?

They have their own issues and I'm nervous to broadcast that functionality even more right now because it takes about a paragraph to explain that you have to have a list bullet with an empty line after it in order for those to work.

dmsnell:
> As far as I'm concerned this is good to go whenever you feel like it's ready.

Thanks for all the review. It made the outcome here much better than when it started. With all the remaining deficiencies I'm going to merge. I'd like to continue iterating on keyboard functionality to make some quirks better, but this PR accomplishes quite a bit already:
 - improve consistency of keybindings across the platforms and in consistency with the native platform on which they run
 - add tooltips for shortcuts on buttons that have corresponding shortcuts
 - add keybinding map with menu option / sidebar indicator
 - fix broken shortcuts

PR #1989, opened 2020-04-09
Refactor: Remove previousIndex

dmsnell:
We have been tracking `previousIndex` when trashing, restoring, and deleting notes forever in order to properly find the next note to select after these operations. This process required tracking the `previuosIndex` through multiple contexts, some of which only sometimes tracked it. The result was a confusing system we couldn't touch without breaking.

In this patch we're eliminating the `previousIndex` altogether and moving the logic for selecting a note into the `note-list` where it holds the notion of a previous or a next note, a note above or a note below the current selection.

While this may ultimately be a pit-stop on the road to refactoring how `selectedNote` or `note` is stored in state it removes the state tracking that was so cumbersome and fragile.

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -209,9 +209,7 @@ const createCompositeNoteList = (
 export class NoteList extends Component<Props> {
   static displayName = 'NoteList';
 
-  state = {
-    selected: { noteId: null, index: null },
-  };
+  state = { selectedIndex: null };
 
   list = createRef<List>();
 
@@ -238,9 +236,7 @@ export class NoteList extends Component<Props> {
       showTrash,
       tagResultsFound,
     } = nextProps;
-    const {
-      selected: { noteId, index },
-    } = this.state;
+    const { selectedIndex } = this.state;
 
     if (
       noteDisplay !== this.props.noteDisplay ||
@@ -252,9 +248,9 @@ export class NoteList extends Component<Props> {
     }
 
     if (
-      index &&
+      selectedIndex &&
       selectedNote &&
-      notes[index]?.id === selectedNote.id &&
+      notes[selectedIndex]?.id === selectedNote.id &&
       showTrash === this.props.showTrash &&
       openedTag === this.props.openedTag
     ) {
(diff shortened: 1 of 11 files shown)


lib/note-list/index.tsx
```diff
@@ -356,13 +359,52 @@ export class NoteList extends Component<Props> {
...
+    const { notes, selectedNote, showTrash } = nextProps;
+    const {
+      selected: { noteId, index },
+    } = this.state;
+
+    if (
+      index &&
+      selectedNote &&
+      index < notes.length &&
+      notes[index].id === selectedNote.id
```
dmsnell, line 372:
```ts
const a = index < notes.length && notes[index].id === selectedNote.id
const b = notes[index]?.id === selectedNote.id
```

^^^ these two, for our purposes, will be identical statements

lib/note-list/index.tsx
```diff
@@ -375,15 +417,29 @@ export class NoteList extends Component<Props> {
...
-      this.props.onSelectNote(this.props.nextNote.id);
+      if (-1 === highlightedIndex) {
+        return true;
+      }
+
+      if (index < 0 || !notes[index - 1]?.id) {
+        return true;
+      }
+
+      this.props.onSelectNote(notes[index - 1].id);
```
dmsnell, line 442: we can also rely on that `?.` operator and collapse this check into an assertion…

```ts
const noteIdAbove = notes[index - 1]?.id;

if (noteIdAbove) {
  this.props.onSelectNote( noteIdAbove );
}
```

lib/note-list/index.tsx
```diff
@@ -291,6 +336,46 @@ export class NoteList extends Component<Props> {
...
+
+    const noteAt = notes.findIndex(item => item?.id === selectedNote?.id);
+
+    if (selectedNote && noteAt > -1) {
+      return noteAt;
+    }
+
+    if (selectedNote) {
+      return Math.min(index, notes.length - 1); // different note, same index
+      // throw new Error('selected note not in list');
```
codebykat, line 372: we can delete this commented line

lib/note-list/index.tsx
```diff
@@ -301,23 +386,26 @@ export class NoteList extends Component<Props> {
...
       searchQuery,
       tagResultsFound
     );
 
+    const specialRows = compositeNoteList.length - notes.length;
+
     const renderNoteRow = renderNote(compositeNoteList, {
       searchQuery,
+      highlightedIndex:
+        highlightedIndex !== null ? highlightedIndex + specialRows : null,
```
codebykat, line 405: This is okay but it might be clearer to make it a variable (maybe instead of `specialRows`)

lib/note-toolbar-container.ts
```diff
@@ -24,46 +24,32 @@ type NoteChanger = {
   note: T.NoteEntity;
 };
 
-type ListChanger = NoteChanger & { previousIndex: number };
-
```
codebykat, line 28: yay, this is a positive code smell. a code fragrance?

PR #1991, opened 2020-04-13
Move tagsLoaded to redux

belcherj:
### Fix
Moves tagsLoaded to redux from appState. This is the most minor change that I could make getting it ported over. The tags functionality in state/domain will need to be refactored to use the middleware architecture but this PR gets tags out of appState.

### Test
1. We should run the full checklist on this one.
2. Open the sidebar, do tags load?
3. Delete a tag, reload page, do tags load as expected
4. Manually sort tags, reload, do they stay sorted
5. Change the settings for sort tags alpha
6. Open side bar, are they alpha sorted
7. Reload page, are they alpha sorted?

### Release
Not updated: Moved tagsLoaded to redux

### DIFF
--- a/lib/state/tags/reducer.ts
+++ b/lib/state/tags/reducer.ts
@@ -9,8 +9,8 @@ const tags: A.Reducer<T.TagEntity[]> = (state = [], action) => {
         // Sort tags alphabetically by 'name' value
         sortedTags.sort((a: T.TagEntity, b: T.TagEntity) => {
           return a?.data?.name
-            .toLowerCase()
-            .localeCompare(b.data?.name.toLowerCase());
+            .toLocaleLowerCase()
+            .localeCompare(b.data?.name.toLocaleLowerCase());
         });
       } else {
         // Sort the tags by their 'index' value
--- a/lib/state/tags/reducer.ts
+++ b/lib/state/tags/reducer.ts
@@ -1,0 +1,29 @@
+import * as A from '../action-types';
+import * as T from '../../types';
+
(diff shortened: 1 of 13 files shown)


lib/note-list/index.tsx
```diff
@@ -500,10 +500,11 @@ const { emptyTrash, loadAndSelectNote } = appState.actionCreators;
 
 const mapStateToProps: S.MapState<StateProps> = ({
   appState: state,
+  tags,
   ui: { filteredNotes, note, searchQuery, showTrash },
   settings: { noteDisplay },
 }) => {
-  const tagResultsFound = getMatchingTags(state.tags, searchQuery).length;
+  const tagResultsFound = getMatchingTags(tags, searchQuery).length;
```
dmsnell, line 507: also touched in #1995 so one of this or that will need to rebase

lib/state/tags/reducer.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const tags: A.Reducer<T.TagEntity[]> = (state = [], action) => {
+  switch (action.type) {
+    case 'TAGS_LOADED': {
+      const sortedTags = action.tags.slice();
+      if (action.sortTagsAlpha) {
+        // Sort tags alphabetically by 'name' value
+        sortedTags.sort((a: T.TagEntity, b: T.TagEntity) => {
+          return a?.data?.name
+            .toLowerCase()
```
dmsnell, line 12: while we're at it, this should be `toLocaleLowerCase()`

PR #1993, opened 2020-04-13
Update Dependencies

belcherj:
### Fix
Update most dependencies

### Test
1. Need to run the checklist

### Release
Note updated: Updated dependencies


dmsnell:
At some point we should circle back to the `html-webpack-plugin` updates because it looks like we can create a React component to produce that, which might make some things easier to maintain, especially if we want to ship an empty app shell vs. an empty HTML page

PR #1997, opened 2020-04-14
Updates prettier and runs it

belcherj:
### Fix
Updates prettier and runs the code fixes.

There are two changes here:
Adds a space after the function keyword. [Here is the explanation.](https://prettier.io/blog/2020/03/21/2.0.0.html#always-add-a-space-after-the-function-keyword-3903httpsgithubcomprettierprettierpull3903-by-j-f1httpsgithubcomj-f1-josephfrazierhttpsgithubcomjosephfrazier-sosukesuzukihttpsgithubcomsosukesuzuki-thorn0httpsgithubcomthorn0-7516httpsgithubcomprettierprettierpull7516-by-bakkothttpsgithubcombakkot)
`function()`
vs
`function ()`

Puts parens around single arguments in arrow functions. [Here is the explanation.](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev)
`const foo = bar => {}`
vs
`const foo = (bar) => {}`

### Test
1. Should just be formatting changes

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -38,9 +38,7 @@
     "prettier/prettier": [
       "error",
       {
-        "bracketSpacing": true,
-        "singleQuote": true,
-        "trailingComma": "es5"
+        "singleQuote": true
       }
     ],
     "react/display-name": "warn",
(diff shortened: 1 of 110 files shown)


dmsnell:
This seems fine to me after recreating it from a recent `develop` commit.

PR #1998, opened 2020-04-14
Change single quotes to double quotes

belcherj:
### Fix
This PR is branched off of #1997 Once it is merged, I will rebase

I'd like to use the stock Prettier configuration. The only way we differ from the default is single vs double quotes on strings. [Here is the rationale from Prettier](https://prettier.io/docs/en/rationale.html#strings)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -35,12 +35,7 @@
     "no-lonely-if": "error",
     "no-shadow": "warn",
     "no-spaced-func": "error",
-    "prettier/prettier": [
-      "error",
-      {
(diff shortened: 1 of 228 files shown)


dmsnell:
If we're voting my vote would be against this. For one we're diverging from how WordPress and Automattic formats JavaScript and two, this is a massive PR that will wipe out a lot of readily-accessible history in the gutter for no ready benefit. I'd much rather see us use `--single-quote`

belcherj:
Closing as we will not be merging this one.

PR #2000, opened 2020-04-14
Moves notesLoaded from appState to redux

belcherj:
### Fix
Moves notesLoaded from appState to redux. A pretty cut and dry one.

### Test
1. Do notes load
2. Can you search
3. Can you trash note

### Release
`Moved notesLoaded from appState to Redux`

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -446,12 +446,7 @@ export const App = connect(
     syncActivityHooks = data => {
       activityHooks(data, {
         onIdle: () => {
-          const {
-            appState: { notes },
-            client,
-            noteBucket,
-            setUnsyncedNoteIds,
-          } = this.props;
+          const { notes, client, noteBucket, setUnsyncedNoteIds } = this.props;
 
           nudgeUnsynced({ client, noteBucket, notes });
           setUnsyncedNoteIds(getUnsyncedNoteIds(noteBucket));
@@ -475,7 +470,6 @@ export const App = connect(
 
     render() {
       const {
-        appState: state,
         authIsPending,
         isAuthorized,
         isDevConfig,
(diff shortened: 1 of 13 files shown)


codebykat (approved):
Needs a rebase but code looks good and didn't see any issues in testing 👍

dmsnell:
This PR _does_ move some functionality from `app-state` to new Redux files, but I think it's mostly a _change_ vs. tackling some issue with our old state design.

`state.notes` doesn't hold the current value of notes. it holds an odd copy of notes as they were when they loaded during the first `index` operation from Simperium. if we're going to go ahead and move this into a new reducer then I think we need to make sure that we listen to `update` actions (`REMOTE_NOTE_UPDATE`) and also to operations locally within the app.

the real value in "Reduxifying" `notes` is making it representative of the app state and removing the data-flow cycles and duplications existent already in the app. to that point I think it's very deceiving how small this looks if all we do is recreate the logic from `app-state` into new and separate files.

if we want we can merge this, but at that point I'm a bit puzzled at its purpose. we can always come back and work on it later, but why not now when its fresh in our heads?

what is the point of `notes` in state?
 - it should canonically represent the current state of all notes in an account from the perspective of this local device/app instance
 - it should receive all updates from Simperium to keep remote changes in sync
 - it should be updated when locally modifying a note to keep local changes in sync

to that end we should probably move the pinned partitioning into `note-list`, probably, or into the `filteredNotes` reducer. right now it appears like it's coincidence that pinned notes appear at the top of the list and the `notes` reducer needlessly recreates the `notes` value because of what it's doing.

this also overlaps the work going on with `previousIndex` and `selectedIndex` and `selectedNote` though that probably belongs separately so you can ignore this comment other than to note that we should be thinking about them together.

I'd suggest trying to see how much it would take to at least update in response to remote and local note edits and to also move the pinned logic outside of this state value.

We're currently relying on remote note updates anyway to update the state value. How much would it take to change that so that, for example, the `notes` reducer listens to actions like `App.emptyTrash` - that's where the real substance of a Redux refactor lies.

belcherj:
Closing in favor of THE refactor.

PR #2001, opened 2020-04-15
[WIP] Refactor: Try middleware for analytics

codebykat:
### Fix
My attempt at centralizing the tracks calls to one place in the app

### Test
Add a `console.log` in `recordEvent` and make sure the correct events are firing in response to actions taken throughout the app.

### Release
TBD

### DIFF
--- a/lib/state/analytics/middleware.ts
+++ b/lib/state/analytics/middleware.ts
@@ -8,35 +8,45 @@ import * as T from '../../types';
 export const middleware: S.Middleware = store => {
   return next => (action: A.ActionType) => {
     const result = next(action);
-    const nextState = store.getState();
 
-    // @todo events throughout the app that don't have a corresponding action
-    // analytics.tracks.recordEvent('user_account_created');
-    // analytics.tracks.recordEvent('editor_checklist_inserted');
-    // analytics.tracks.recordEvent('importer_import_completed', {
-    // analytics.tracks.recordEvent('editor_note_collaborator_added');
-    // analytics.tracks.recordEvent('editor_note_collaborator_removed');
-    // analytics.tracks.recordEvent('editor_note_edited');
-    // analytics.tracks.recordEvent('list_tag_deleted');
+    // @todo uncomment to ship, this breaks debugging :)
+    // if (!window.sendAnalytics) {
+    //   return result;
+    // }
+
+    // const nextState = store.getState();
+
+    // @todo events that don't have a corresponding action and aren't in redux
+    // - editor_checklist_inserted
+    // - importer_import_completed
 
     switch (action.type) {
+      /* catch-all action used by redux components for these events:
+         - editor_note_collaborator_added
+         - editor_note_collaborator_removed
+         - editor_note_edited
+         - list_tag_deleted
+         - user_account_created
+      */
+      case 'RECORD_EVENT':
+        if (action.eventName) {
+          analytics.tracks.recordEvent(action.eventName);
+        }
+
+      /* events that map to an action directly */
       case 'AUTH_SET':
         if (action.status === 'authorized') {
-          // @todo called twice for some reason
           analytics.tracks.recordEvent('user_signed_in');
         }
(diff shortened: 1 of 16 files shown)


belcherj (commented):
This is really promising. I think that we should get these actions added. Most of them will be useful as we move state to redux and the simperium middleware.

lib/state/analytics/middleware.ts
```diff
@@ -0,0 +1,82 @@
...
+    // analytics.tracks.recordEvent('importer_import_completed', {
+    // analytics.tracks.recordEvent('editor_note_collaborator_added');
+    // analytics.tracks.recordEvent('editor_note_collaborator_removed');
+    // analytics.tracks.recordEvent('editor_note_edited');
+    // analytics.tracks.recordEvent('list_tag_deleted');
+
+    switch (action.type) {
+      case 'AUTH_SET':
+        if (action.status === 'authorized') {
+          // @todo called twice for some reason
```
dmsnell, line 25: which is called twice? `AUTH_SET`, or `AUTH_SET` _with_ `status === 'authorized'`? because `AUTH_SET` should be called multiple times in succession throughout the sign-in process

lib/state/analytics/middleware.ts
```diff
@@ -0,0 +1,82 @@
+import analytics from '../../analytics';
+import actions from '../actions';
+
+import * as A from '../action-types';
+import * as S from '../';
+import * as T from '../../types';
+
+export const middleware: S.Middleware = store => {
+  return next => (action: A.ActionType) => {
+    const result = next(action);
```
dmsnell, line 10: it should be handled in our analytics code itself, but this is a great place to check for the analytics toggle and skip the middleware if analytics are turned off

```ts
const result = next( action );

if ( ! window.sendAnalytics ) {
	return result;
}
```

lib/state/analytics/middleware.ts
```diff
@@ -0,0 +1,82 @@
...
+    const nextState = store.getState();
```
dmsnell, line 11: note that unless we need to read data from state this shouldn't be necessary

lib/state/analytics/middleware.ts
```diff
@@ -0,0 +1,82 @@
...
+        break;
+      case 'RESTORE_NOTE':
+        analytics.tracks.recordEvent('editor_note_restored');
+        break;
+      case 'SEARCH':
+        if (action.searchQuery) {
+          // @todo this gets called for each character typed,
+          // possibly also happening in develop?
+          analytics.tracks.recordEvent('list_notes_searched');
+        }
```
dmsnell, line 55: we can trap `FILTER_NOTES` instead if we prefer. we could either get the search query from state or we could add the searched-for query into the `FILTER_NOTES` action itself

codebykat: FILTER_NOTES seems even worse, as it gets called in lots of places not triggered by searching 🤔

lib/state/analytics/middleware.ts
codebykat: huh it WAS being called twice with status authorized, but maybe no longer..

dmsnell: we should probably find out why if it's being called with with `authorized` - that wouldn't likely have been changed in years

lib/state/analytics/middleware.ts
dmsnell: ah okay. we're specifically trying to see every time someone enters a character into the search field…

```
s
so
som
some
somet
someth
somethi
somethin
something
```

this could be a decent place for a debounce possibly unless we want the numbers to be grossly inflated, which of course I'm guessing they already are…

codebykat: yeah, that was my "possibly also happening in develop?" comment. I am pretty sure this is, intentionally or not, what this event is already tracking

codebykat:
Done in #2223

PR #2002, opened 2020-04-15
Add GitHub action to deploy the app to develop.simplenote.com

belcherj:
Reverts Automattic/simplenote-electron#1718

GitHub Actions should be working again. This PR enables auto-deployment to develop.simplenote.com when something gets merged to develop.

I removed the auto-deploy to production until we are comfortable with auto deployments.  I would also like to explore auto-deploying any new tags to staging.


codebykat (commented):
Can we test this in any way before merging it? I mean, it looks fine.

.github/actions/deploy/entrypoint.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+git checkout -b $COMMIT_BRANCH
+
+# Delete files not needed in production
+find . -maxdepth 1 ! -name '.git' ! -name 'dist' ! -name 'vip' -exec rm -rf {} \;
+
+# Copy vip files to base path
+cp -r vip/* ./
+
+# Delete uneeded vip folder
```
codebykat, line 52: typo: missing n unneeded

belcherj:
Fixed misspelling. How do I get Grammarly working in vs code and the command? I have become completely reliant on spell-check. :)

dmsnell:
Besides auto-deploying to `develop.simplenote.com` what benefit does this change bring us? Could we auto-deploy with our existing CI config?

belcherj:
This PR auto-deploys to develop.simplenote.com and adds the capability to auto-deploy to staging and production. We could auto-deploy in CircleCi but the configuration is pretty messy in dealing with GitHub tokens.  GH Actions are faster and we don't run into queuing.

dmsnell:
> We could auto-deploy in CircleCi but the configuration is pretty messy in dealing with GitHub tokens.

If they are the ones at `settings/keys` that makes sense. Do we need to remove Circle configs in this PR so they don't coexist?

belcherj:
> If they are the ones at settings/keys that makes sense. Do we need to remove Circle configs in this PR so they don't coexist?

This code only auto-deploys it doesn't do anything that CircleCi does. Im not sure what you are asking here.

dmsnell:
> This code only auto-deploys it doesn't do anything that CircleCi does. Im not sure what you are asking here.

We have [bin/deploy.sh](https://github.com/Automattic/simplenote-electron/blob/develop/bin/deploy.sh) already, which is called from `npm run deploy` so on that part maybe I'm asking more "should we be calling `npm run deploy`" than about CircleCI.

Also though we have [the setup code](https://github.com/Automattic/simplenote-electron/blob/c49de3747abc0935b265d26978f15edbeea9b1a2/.circleci/config.yml#L74) duplicated in the deploy Dockerfile. Should we be reusing that Dockerfile in both places? is the duplication necessary? the long specific list of Linux dependencies particularly catches my attention as places that are likely to get out of sync and produce subtle differences depending on how we build this.

belcherj:
You are right, should definitely use the existing `npm run deploy`.

The Linux dependencies are different for the two environments. I think we work on deduping as we move the process off of CircleCi.

.github/actions/deploy/deploy.yml
```diff
@@ -0,0 +1,23 @@
...
+  GITHUB_TOKEN:
+    description: 'GitHub Token'
+    requried: true
+  BRANCH:
+    description: 'Branch'
+    required: true
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
+  entrypoint: './.github/actions/deploy/entrypoint.sh'
```
dmsnell, line 19: the [docs](https://help.github.com/en/actions/building-actions/dockerfile-support-for-github-actions#entrypoint) indicate that this is here to override the `ENTRYPOINT` in the Dockerfile. since we're creating the Dockerfile and not overwriting anything I don't understand why we need it, or why we need the `entrypoint.sh` file at all.

it seems like if it works then things could be kept simpler if we directly embedded the call to `npm run deploy`

if it's the case that we need this external script in order to pass the `BRANCH` argument then I think a couple things would help:
 - remove `ENTRYPOINT` from the `Dockerfile`
 - add an explanatory comment why we're taking several steps in what seems like it should only take one

the docs don't seem incredibly clear to me. it would be worth verifying if we have access to the `ENV` vars like `BRANCH` from inside the Dockerfile without external helpers.

```Dockerfile
RUN dpkg…

CMD ["sh", "-c", "npm run deploy $BRANCH"]
```

.github/actions/deploy/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:10.16.3-jessie
+
+RUN dpkg --add-architecture i386
+RUN apt update
+RUN apt -y install python make libxkbfile-dev libxkbfile-dev:i386 libx11-dev libx11-dev:i386 libxss-dev gcc-multilib g++-multilib rpm
```
dmsnell, line 5: fun fact, not the biggest deal here, but each of these lines creates a new "layer" in the Docker image and can carry a reasonable performance hit. since it's an automated build it won't matter too much but it's common to join these into one `RUN` statement. put differently, we've created these on separate `RUN` lines for organizational purposes but that creates an unintended side-effect in the Docker image we didn't want.

line-continuation marks and indentation are our friends, and now we can alphabetize the dependencies and make future diffs clearer

```Dockerfile
RUN dpkg --add-architecture i386 \
    && apt update \
    && apt -y install \
         g++-multilib \
         gcc-multilib \
         libx11-dev libx11-dev:i386 \
         libxkbfile-dev libxkbfile-dev:i386 \
         libxss-dev \
         make \
         python \
         rpm
```

dmsnell:
Asking out of ignorance: will this impact #2042? I notice that we have to create a Docker environment and add the `i386` architecture, presumably to install build deps for it? Would we need to, would we be able to, still cross-compile for ARM?

belcherj:
Closing as much has changed and I think we need to modify the build in ways that would make this PR obsolete. We shouldn't build the desktop app to ship the web app.

PR #2004, opened 2020-04-16
Refactor: Move isElectron and isMacApp checks to a util

belcherj:
### Fix
Moves isElectron and isMacApp checks to a util which allows us to remove a bunch of prop drilling.

### Test
1. Do you see settings in tag panel at the bottom on web?
1. Do you not see settings in tag panel at the bottom on electron?
1. In the import dialog do not you see Evernote on web?
1. In the import dialog do you see Evernote on electron?

### Release
Moved isElectron and isMacApp checks to a util

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -16,20 +16,12 @@ import DevBadge from './components/dev-badge';
 import DialogRenderer from './dialog-renderer';
 import { getIpcRenderer } from './utils/electron';
 import exportZipArchive from './utils/export';
+import { isElectron, isMac } from './utils/platform';
 import { activityHooks, getUnsyncedNoteIds, nudgeUnsynced } from './utils/sync';
 import { setLastSyncedTime } from './utils/sync/last-synced-time';
 import analytics from './analytics';
 import classNames from 'classnames';
-import {
-  debounce,
-  get,
-  has,
-  isObject,
-  matchesProperty,
-  overEvery,
-  pick,
-  values,
-} from 'lodash';
+import { debounce, get, has, isObject, overEvery, pick, values } from 'lodash';
 import {
   createNote,
   closeNote,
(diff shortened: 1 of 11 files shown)


lib/utils/is-electron.ts
```diff
@@ -0,0 +1,9 @@
+// https://github.com/atom/electron/issues/22
+
+export const isElectron: () => boolean = () => {
+  debugger;
+  return !!window?.process?.type;
+};
+
+export const isMacElectron: () => boolean = () =>
+  window?.process?.platform === 'darwin';
```
dmsnell, line 9: what if we called this `lib/utils/platform.ts` since it carries more information than just Electron?

lib/utils/is-electron.ts
```diff
@@ -0,0 +1,9 @@
+// https://github.com/atom/electron/issues/22
+
+export const isElectron: () => boolean = () => {
+  debugger;
+  return !!window?.process?.type;
+};
+
+export const isMacElectron: () => boolean = () =>
```
dmsnell, line 8: the name `isMacApp` or `isMac` seems like a closer match to what this function returns and to what we want to know. what are your thoughts on renaming it?

lib/utils/is-electron.ts
```diff
@@ -0,0 +1,9 @@
+// https://github.com/atom/electron/issues/22
+
+export const isElectron: () => boolean = () => {
+  debugger;
+  return !!window?.process?.type;
+};
```
dmsnell, line 6: both of these exports are functions and they shouldn't need to be. let's simplify now. there was a reason they were IIFEs before, but I don't remember. maybe to lazy-evaluate the code? that wouldn't be a strong reason…

```ts
export const isElectron = !! window?.process?.type;

export const isMac = window?.process?.platofrm === 'darwin';
```

PR #2006, opened 2020-04-17
Fix: Remove tags from state on logout

belcherj:
### Fix
After moving the tags state to redux we were not clearing out tags on log out. This PR sets tags to an empty array

### Test
1. Be logged in
2. Have tags
3. Log out
4. Look at redux chrome dev tools plugin
5. Observe no tags in state

### Release
Fix: Remove tags from state on logout

### DIFF
--- a/lib/state/tags/reducer.ts
+++ b/lib/state/tags/reducer.ts
@@ -21,6 +21,11 @@ const tags: A.Reducer<T.TagEntity[]> = (state = [], action) => {
       }
       return sortedTags;
     }
+    case 'AUTH_SET':
+      if (action.status === 'not-authorized') {
+        return [];
+      }
+      return state;
     default:
       return state;
   }


dmsnell (changes requested):
Apparently I lost a comment somewhere.

We have similar issues all around state and it might be advantageous for us to consider completely clearing app state during auth changes. For example, the search filter persists if you logout/login without restarting the app.

```ts
// state/index.ts
const resetOnUnauthorized = ( reducer ) => ( state, action ) =>
	'AUTH_SET' === action.type && 'not-authorized' === action.status
		? reducer( undefined, action )
		: reducer( state, action );

export const store = createStore(
	resetOnUnauthorized( reducers ),
	…
)
```

We may have to make other adjustments as necessary to account for conditions that need to persist but this gives us a framework to handle a full reset without doing it 100 times.

belcherj:
CLosing as auth refactor took care of this.

PR #2008, opened 2020-04-17
Types: Add types to ProgressBar component

belcherj:
### Fix
Adds Types to the ProgressBar component.

### Test
Just types

### Release
Types: Add types to ProgressBar component

### DIFF
--- a/lib/components/progress-bar/index.tsx
+++ b/lib/components/progress-bar/index.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
-import { LinearProgress, LinearProgressProps } from '@material-ui/core';
+import { LinearProgress } from '@material-ui/core';
 
-const ProgressBar: React.FunctionComponent<LinearProgressProps> = (props) => {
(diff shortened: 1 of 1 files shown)


lib/components/progress-bar/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
-import { LinearProgress } from '@material-ui/core';
+import { LinearProgress, LinearProgressProps } from '@material-ui/core';
 
-const ProgressBar = props => {
+const ProgressBar: React.FunctionComponent<LinearProgressProps> = props => {
```
dmsnell, line 4: Did you try this?

```ts
const ProcessBar: LinearProgress = props => …
```

belcherj: Screen Shot 2020-04-17 at 4 51 10 PM

dmsnell: this works for me

```ts
const ProgressBar: typeof LinearProgress = props => {
	…
}
```

PR #2010, opened 2020-04-17
Use actions import over direct import of actions

belcherj:
### Fix
Import actions using `import actions from '../state/actions';` rather than directly importing actions.

### Test
This affects the whole app. I think the checklist might be necessary.

### Release
Use actions import over direct import of actions


belcherj:
Closing to recreate later

PR #2013, opened 2020-04-18
Fix: print stylesheets for Firefox

codebykat:
### Fix
I happened to notice that Firefox only prints the first page of longer notes. Turns out this is a Firefox bug with `display:flex;`. In order to fix it, we need to set every `div` up the tree that is `display:flex` to `display:block`.

See https://bugzilla.mozilla.org/show_bug.cgi?id=939897

### Todo
There's still something wrong with printing in Markdown preview mode

### Test
1. Open a long note in Firefox
2. Print the note
3. Verify that all pages print and no extraneous elements show up
4. Verify that printing long notes still looks good in other browsers

### Release
Not updated: Fix a bug that caused long notes to be truncated to the first page when printing in Firefox.

### DIFF
--- a/scss/print.scss
+++ b/scss/print.scss
@@ -7,9 +7,25 @@
     border: 0;
   }
 
+  .dev-badge {
+    display: none;
+  }
+
   .tag-field,
   .note-toolbar-wrapper {
     display: none;
+  }
+
+  /* Firefox only prints the first page if any divs are display: flex */
+  .app,
+  .simplenote-app,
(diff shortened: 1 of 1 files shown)


codebykat:
Tested further and realized this PR is actually making the print stylesheet worse for Markdown preview, so please stand by.

codebykat:
I take it back, it seems fine after a rebase. I just had styles cached I think. Tested in Safari, Chrome, Firefox.

PR #2015, opened 2020-04-19
Feature: Import from Google Keep

mgunyho:
### Fix
This PR adds the option to import notes from Google Keep. Notes can be either imported from a zip file exported from Google Takeout, or as individual JSON files. [Attached](https://github.com/Automattic/simplenote-electron/files/4499355/takeout-20200412T094422Z-001-test.zip) is a small test file that contains notes with various features (check boxes, tags, etc.).

A couple of things to note:
- Images and other attachments are not imported. I played around a bit with embedding the images as base64 in the markdown, but that is maybe not a very good idea with large images. It's also a bit of a hassle with the zipped files.
- Archived notes get imported as just regular notes. I think the best way to handle this would be to add an option to the modal (next to 'enable markdown on all notes'), where the user could choose to move archived notes to the trash, add an user-specified tag or to import them regularly. I'm not familiar enough with react/redux to implement this myself.
- It would be cool to parse links in notes and enable markdown on them (using e.g. [linkify](https://soapbox.github.io/linkifyjs/)), not sure if it's worth it to add another dependency. The notes also have the `annotations` attribute that could be used for this.

I'm fairly new to typescript, so let me know if something looks wrong in the code!

### Test
1. Go to File-> Import Notes -> Google Keep
2. Drag and drop e.g. [this](https://github.com/Automattic/simplenote-electron/files/4499355/takeout-20200412T094422Z-001-test.zip) test zip file
3. Observe that the notes have been imported

### Release
- Added possibility to import notes from Google Keep

### DIFF
--- a/lib/utils/import/googlekeep/index.ts
+++ b/lib/utils/import/googlekeep/index.ts
@@ -38,26 +38,28 @@ class GoogleKeepImporter extends EventEmitter {
       let textContent = title ? title + '\n\n' : '';
       textContent += get(importedNote, 'textContent', '');
 
-      const note = {
-        content: textContent,
-        // note: the exported files don't tell us the creation date...
-        modificationDate: importedNote.userEditedTimestampUsec / 1e6,
-        pinned: importedNote.isPinned,
-        tags: get(importedNote, 'labels', []).map(item => item.name),
-      };
-
       if (importedNote.listContent) {
         // Note has checkboxes
-        note.content += importedNote.listContent
+        textContent += importedNote.listContent
           .map(item => `- [${item.isChecked ? 'x' : ' '}] ${item.text}`)
           .join('\n');
       }
 
-      const opts = { ...this.options, isTrashed: importedNote.isTrashed };
-      return coreImporter.importNote(note, opts).then(() => {
-        importedNoteCount++;
-        this.emit('status', 'progress', importedNoteCount);
-      });
+      return coreImporter
+        .importNote(
+          {
+            content: textContent,
+            // note: the exported files don't tell us the creation date...
+            modificationDate: importedNote.userEditedTimestampUsec / 1e6,
+            pinned: importedNote.isPinned,
+            tags: get(importedNote, 'labels', []).map(item => item.name),
+          },
+          { ...this.options, isTrashed: importedNote.isTrashed }
+        )
+        .then(() => {
+          importedNoteCount++;
+          this.emit('status', 'progress', importedNoteCount);
+        });
     };
 
     const importZipFile = fileData =>
--- a/lib/utils/import/googlekeep/index.ts
+++ b/lib/utils/import/googlekeep/index.ts
@@ -47,15 +47,16 @@ class GoogleKeepImporter extends EventEmitter {
       }
 
       const title = importedNote.title;
-      let textContent = title ? title + '\n\n' : '';
-      textContent += get(importedNote, 'textContent', '');
 
-      if (importedNote.listContent) {
-        // Note has checkboxes
-        textContent += importedNote.listContent
-          .map(item => `- [${item.isChecked ? 'x' : ' '}] ${item.text}`)
-          .join('\n');
-      }
+      const importedContent = importedNote.listContent
+        ? importedNote.listContent // Note has checkboxes, no text content
+            .map(item => `- [${item.isChecked ? 'x' : ' '}] ${item.text}`)
+            .join('\n')
+        : importedNote.textContent;
+
+      const textContent = title
+        ? `${title}\n\n${importedContent}`
+        : importedContent;
 
       return coreImporter
         .importNote(
--- a/lib/utils/import/googlekeep/index.ts
+++ b/lib/utils/import/googlekeep/index.ts
@@ -48,7 +48,7 @@ class GoogleKeepImporter extends EventEmitter {
       const note = {
         content: textContent,
         // note: the exported files don't tell us the creation date...
-        modificationDate: importedNote.userEditedTimestampUsec / 1000000,
+        modificationDate: importedNote.userEditedTimestampUsec / 1e6,
         pinned: importedNote.isPinned,
         tags: get(importedNote, 'labels', []).map(item => item.name),
       };
--- a/lib/utils/import/googlekeep/index.ts
+++ b/lib/utils/import/googlekeep/index.ts
@@ -38,40 +38,34 @@ class GoogleKeepImporter extends EventEmitter {
 
     let importedNoteCount = 0;
 
-    const extractAndImport = zipObj => {
-      return zipObj
-        .async('string')
-        .then(content => {
-          const importedNote = JSON.parse(content);
+    const importJsonString = jsonString => {
+      const importedNote = JSON.parse(jsonString);
 
-          const title = importedNote.title;
-          let textContent = title ? title + '\n' : '';
-          textContent += get(importedNote, 'textContent', '');
-          const note = {
-            content: textContent,
-            // note: the exported files don't tell us the creation date...
-            modificationDate: importedNote.userEditedTimestampUsec / 1000000,
-            pinned: importedNote.isPinned,
-            tags: get(importedNote, 'labels', []).map(item => item.name),
-          };
+      const title = importedNote.title;
+      let textContent = title ? title + '\n' : '';
+      textContent += get(importedNote, 'textContent', '');
 
-          if (importedNote.listContent) {
-            note.content += importedNote.listContent
-              .map(item => {
-                return '- [' + (item.isChecked ? 'x' : ' ') + '] ' + item.text;
-              })
-              .join('\n');
-          }
+      const note = {
+        content: textContent,
+        // note: the exported files don't tell us the creation date...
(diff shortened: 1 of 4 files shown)


dmsnell:
@mgunyho this looks great and it's so cool that you added it. we're a bit busy preparing the 1.16 release but I'd love to prioritize your PR. Please ping us again here if we haven't done more review by Monday.

mgunyho:
Hi, any update on this?

dmsnell:
> Hi, any update on this?

Very nice @mgunyho you held me accountable 😄 

Sadly no, we got a bit more distracted by some issues we uncovered while testing the 1.16 release. Those are mostly done. I will make sure to get you some review by the end of this weekend - setting an alert on my calendar to remind me.

Thanks for your patience! We appreciate your contribution so much and are excited to move it forward!

lib/utils/import/googlekeep/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const coreImporter = new CoreImporter({
+      noteBucket: this.noteBucket,
+      tagBucket: this.tagBucket,
+    });
+
+    let importedNoteCount = 0;
+
+    const importJsonString = jsonString => {
+      const importedNote = JSON.parse(jsonString);
```
dmsnell, line 35: this could fail if the export file is corrupted or simply not what we expect.
could we emit an error here or should we skip any files we can't parse?

if all we wanted to do was skip them then we don't need to change the code since the `Exception` will be swallowed by the promise, but if we do that it would be ideal to indicate in a comment that we expect this to happen and that that's our expected resolution.

lib/utils/import/googlekeep/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      const title = importedNote.title;
+      let textContent = title ? title + '\n\n' : '';
+      textContent += get(importedNote, 'textContent', '');
```
dmsnell, line 39: this is a minor style issue, but we don't need to mutate this variable when we can instead set it in one go

```ts
const title = importedNote.title;
const importedContent = importedNote.textContent ?? '';
const textContent = title ? `${ title }\n\n${ importedContent }`	: importedContent;
```

lib/utils/import/googlekeep/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      return coreImporter
+        .importNote(
+          {
+            content: textContent,
+            // note: the exported files don't tell us the creation date...
+            modificationDate: importedNote.userEditedTimestampUsec / 1e6,
+            pinned: importedNote.isPinned,
+            tags: get(importedNote, 'labels', []).map(item => item.name),
+          },
+          { ...this.options, isTrashed: importedNote.isTrashed }
```
dmsnell, line 57: if it were me I might suggest sending archived notes to the trash _or_ add a tag to them - `archived` and leave them in the inbox.

trash is generally safe but someone could accidentally "empty the trash" and wipe out their archive

if, on the other hand, they import their archive into the "All Notes" section then they might get more notes in the list than they expect.

maybe we just need a setting…

```
Archived Notes: [ ] Import into trash [ ] Import with tag _________
```

lib/utils/import/googlekeep/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    const coreImporter = new CoreImporter({
+      noteBucket: this.noteBucket,
+      tagBucket: this.tagBucket,
+    });
+
+    let importedNoteCount = 0;
+
+    const importJsonString = jsonString => {
```
dmsnell, line 34: it might help to do some basic checks here to make sure we don't import something that looks like a note but whose fields are all empty or `undefined`

```ts
const importedNote = JSON.parse(jsonString);

if ( ! importedNote.title && ! importedNote.textContent && ! importedNote.listContent ) {
	// empty note, skip
	return;
}
```

lib/utils/import/googlekeep/index.ts
```diff
@@ -0,0 +1,101 @@
...
+        textContent += importedNote.listContent
+          .map(item => `- [${item.isChecked ? 'x' : ' '}] ${item.text}`)
+          .join('\n');
+      }
+
+      return coreImporter
+        .importNote(
+          {
+            content: textContent,
+            // note: the exported files don't tell us the creation date...
```
dmsnell, line 52: in the absence of a creation date the modification date seems like a good substitute - what do you think?

```ts
creationDate: importedNote.userEditedTimestampUsec / 1e6, // Keep doesn't store a creation date
```

lib/utils/import/googlekeep/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    const importZipFile = fileData =>
+      JSZip.loadAsync(fileData).then(zip => {
+        const promises = zip
+          .file(/.*\/Keep\/.*\.json/)
+          .map(zipObj => zipObj.async('string').then(importJsonString));
+        return Promise.all(promises);
+      });
+
+    const promises = filesArray.map(file =>
+      fs.promises
```
dmsnell, line 74: is this API stable? I thought I might remember seeing something like it would go away or was marked as experimental. would it be safer if we created our own wrapper?

```ts
const readFile = path => new Promise( resolve => setTimeout( () => resolve( fs.readFileSync( path ) ), 0 ) );

const promises = filesArray.map( file => readFile( file.path ) ).then( … ).catch( … );
```

dmsnell (commented):
This is looking great. I left some feedback on things that might be good to update. I'm sure our existing importers aren't the best examples of this, but it looks like we pass over a number of different errors/exceptions/failures in this PR and it would be nice to either handle those explicitly or add a comment in the code explaining why we're not and what we expect to happen - skip note vs. emit an error etc…

Thanks for your patience with this and thanks so much for your efforts! ❤️

lib/utils/import/googlekeep/index.ts
mgunyho: I think just skipping failing notes is the best solution, emitting a `warning` status will display the error message only until the next update so it isn't much use for the user anyway. I'll add a comment mentioning that.

Ideally a warning for each failing note could show up in addition to the "complete" message but I imagine this would require more work.

lib/utils/import/googlekeep/index.ts
mgunyho: Yes, this is how I would do it as well. I don't know how to add a (importer-specific) setting to the dialog.

lib/utils/import/googlekeep/index.ts
mgunyho: Looks like it's stable, at least as of node v12, see [here](https://github.com/nodejs/node/commit/5583d4d73ef750b27c9068ea0f05fae0de9f7766) and the related PR.

lib/utils/import/googlekeep/index.ts
mgunyho: I removed the mutable variable in `bd9425b`, do you think the code is readable?

Side note: is the `??` operator widely enough supported? With a quick glance, I couldn't find it used elsewhere in this repo.

lib/utils/import/googlekeep/index.ts
mgunyho: Fair enough, added in c6c1bf7. Although I think `CoreImporter` already does this if the creation date is missing.

lib/utils/import/googlekeep/index.ts
dmsnell: yeah skipping is fine for now.

lib/utils/import/googlekeep/index.ts
dmsnell: sorry for missing that automatic behavior. probably not bad to be explicit in any case. leaving it empty I would guess from reading here that it would use `now()` as the creation date 🤷

lib/utils/import/googlekeep/index.ts
```diff
@@ -0,0 +1,115 @@
...
+        return;
+      }
+
+      const title = importedNote.title;
+
+      const importedContent = importedNote.listContent
+        ? importedNote.listContent // Note has checkboxes, no text content
+            .map(item => `- [${item.isChecked ? 'x' : ' '}] ${item.text}`)
+            .join('\n')
+        : importedNote.textContent;
```
dmsnell, line 55: interesting. a note can only _either_ be a list _or_ some text?

lib/utils/import/googlekeep/index.ts
dmsnell: I will try to play around with it this week and see if it wouldn't be too hard. if not, maybe adding `archive` to the tags could be a good interim solution?

lib/utils/import/googlekeep/index.ts
mgunyho: Yeah, in Keep, you can "show checkboxes", which turns the whole note into a checklist, and as far as I can tell there's no way to have regular text then.

ghost (approved):

KnifeFed:
Any updates?

dmsnell:
@KnifeFed if there were updates you'd see them here, so no, unfortunately there are none.

PR #2019, opened 2020-04-21
Types: TransitionFadeInOut

belcherj:
### Fix
Adds types to TransitionFadeInOut

### Test
Just Types no need to test

### Release
Added types to TransitionFadeInOut

### DIFF
--- a/lib/components/transition-fade-in-out/index.tsx
+++ b/lib/components/transition-fade-in-out/index.tsx
@@ -1,9 +1,17 @@
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 import PropTypes from 'prop-types';
 import { CSSTransition } from 'react-transition-group';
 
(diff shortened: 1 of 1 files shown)


lib/components/transition-fade-in-out/index.tsx
```diff
@@ -1,9 +1,17 @@
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 import PropTypes from 'prop-types';
 import { CSSTransition } from 'react-transition-group';
 
-const TransitionFadeInOut = props => {
-  const { children, wrapperClassName, shouldMount } = props;
+type OwnProps = {
```
dmsnell, line 5: `children` should be `ReactElement` or `Element` or `ReactNode` or something like that - I think `ReactNode`

PR #2020, opened 2020-04-21
Types: Spinner

belcherj:
### Fix
Adds types to the spinner component

### Test
Just types no functional change

### Release
Added types to the Spinner component

### DIFF
--- a/lib/components/spinner/index.tsx
+++ b/lib/components/spinner/index.tsx
@@ -1,9 +1,13 @@
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 import { CircularProgress } from '@material-ui/core';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-const Spinner = ({ isWhite, ...props }) => {
+type OwnProps = typeof CircularProgress & {
+  isWhite: boolean;
+};
+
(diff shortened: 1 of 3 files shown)


lib/components/spinner/index.tsx
```diff
@@ -1,9 +1,12 @@
...
 import { CircularProgress } from '@material-ui/core';
-import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-const Spinner = ({ isWhite, ...props }) => {
+type OwnProps = typeof CircularProgress & {
+  isWhite: boolean;
+};
+
+const Spinner: FunctionComponent<OwnProps> = ({ isWhite, ...props }) => {
```
dmsnell, line 9: we might as well get rid of the `...props` ambiguity at the same time…

```ts
const Spinner = ({
  isWhite,
  ...props
}: CircularProgressProps & { isWhite: boolean }) => {
```

belcherj: Doesn't what I have do that? I like what you have better but want to know what doesn't work with what I had.

dmsnell: maybe because I missed what you had before? I didn't notice the `CircularPropgressProps` in the first line of the type definition 😉

PR #2021, opened 2020-04-21
Types: TagChip

belcherj:
### Fix
Adds types to the TagChip component

### Test
1. Does the tag chip work correctly?

### Release
Added types to the TagChip component

### DIFF
--- a/lib/components/tag-chip/index.tsx
+++ b/lib/components/tag-chip/index.tsx
@@ -1,24 +1,20 @@
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 import classNames from 'classnames';
-import PropTypes from 'prop-types';
-import { noop } from 'lodash';
 
-export const TagChip = ({ onSelect = noop, selected, tag: tagName }) => (
+type OwnProps = {
+  onSelect: () => {};
+  selected: boolean;
+  tag: string;
+};
+
+const TagChip: FunctionComponent<OwnProps> = ({ onSelect, selected, tag }) => (
   <div
     className={classNames('tag-chip', { selected })}
-    data-tag-name={tagName}
+    data-tag-name={tag}
     onClick={onSelect}
   >
-    {tagName}
+    {tag}
   </div>
 );
-
-TagChip.displayName = 'TagChip';
-
-TagChip.propTypes = {
-  onSelect: PropTypes.func,
-  selected: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


lib/components/tag-chip/index.tsx
```diff
@@ -1,24 +1,20 @@
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 import classNames from 'classnames';
-import PropTypes from 'prop-types';
-import { noop } from 'lodash';
 
-export const TagChip = ({ onSelect = noop, selected, tag: tagName }) => (
+type OwnProps = {
+  onSelect: () => {};
```
dmsnell, line 5: it's probably helpful to follow suit with our other prop types by using `() => any` here

dmsnell (approved):

lib/components/tag-chip/index.tsx
```diff
@@ -1,24 +1,20 @@
...
-import { noop } from 'lodash';
 
-export const TagChip = ({ onSelect = noop, selected, tag: tagName }) => (
+type OwnProps = {
+  onSelect: () => any;
+  selected: boolean;
+  tag: string;
+};
+
+const TagChip: FunctionComponent<OwnProps> = ({ onSelect, selected, tag }) => (
```
dmsnell, line 10: out of curiosity, why did you rename the `tagName` here while adding types?

if my memory serves me right we renamed the prop inside the component because it read more clearly as "tag name" than "tag" did, which conflates with one of several types of "tag" (`TagEntity`, `Tag`, `TagName`) but in part due to lack of typing around the app we didn't update all the call-sites

belcherj: I'm all for renaming but we should rename the property given to the TagChip

dmsnell: > I'm all for renaming 

I might have miscommunicated. This PR does more than add types - it renames an internal variable. That stood out to me, even if I can agree to it. In this case, it seems like we went from something that in the past we made more specific to decrease confusion to something more generic and I didn't see an explanation of why we changed it.

belcherj: I understood what you are saying, I think the wrong place to change the name is in the component. It should be changed everywhere that the component is called.

I agree that changing the variable name was out of scope for this PR.

PR #2023, opened 2020-04-21
Types: Checkbox

belcherj:
### Fix
Adds types to the Checkbox component

### Test
Just types no functional change

### Release
Added types to the Checkbox component

### DIFF
--- a/lib/controls/checkbox/index.tsx
+++ b/lib/controls/checkbox/index.tsx
@@ -1,8 +1,17 @@
 import React from 'react';
 import classNames from 'classnames';
-import PropTypes from 'prop-types';
 
-function CheckboxControl({ className, ...props }) {
+type OwnProps = {
+  className?: string;
+  checked: boolean;
+  id: string;
+  name: string;
+  onChange: () => {};
+  type: 'radio';
+  value: string;
+};
+
+function CheckboxControl({ className, ...props }: OwnProps) {
   return (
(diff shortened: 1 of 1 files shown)


lib/controls/checkbox/index.tsx
```diff
@@ -1,8 +1,17 @@
...
-function CheckboxControl({ className, ...props }) {
+type OwnProps = {
+  className?: string;
+  checked: boolean;
+  id: string;
+  name: string;
+  onChange: () => {};
+  type: 'radio';
+  value: string;
+};
```
dmsnell, line 7: we could apply the same trick here…

```ts
type OwnProps = typeof HTMLInputElement & {
  className: string;
};
```

lib/controls/checkbox/index.tsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
 import classNames from 'classnames';
-import PropTypes from 'prop-types';
 
-function CheckboxControl({ className, ...props }) {
+type OwnProps = React.HTMLProps<HTMLInputElement> & {
+  className?: string;
+  onChange: () => {};
```
dmsnell: can we make this `() => any` for consistency?

dmsnell (approved):
Correction: Resistance is the voltage across a component divided by the current moving through it 😉 

Looks good

PR #2037, opened 2020-04-24
Fix note list toggle

### LINKED ISSUE #2029: Can't access note list with narrow app window
### Expected

When my app window is narrow, I expect to be able to go back to the note list using the back arrow in the top left or the keyboard shortcut `CTRL/CMD + SHIFT + L`.

### Observed

When my app window is narrow, the note editor takes the full frame and the back arrow appears in the top left. However, clicking on that back arrow or using the keyboard shortcut doesn't do anything. I can't get back to the note list unless I make the app window wider.

### Reproduced

1. With a wide app window, notice you can see both the note list and note editor.
2. Drag to make the app window narrow until the note list disappears.
3. Click the back arrow in the top left and notice that nothing happens.
4. Type `CTRL/CMD + SHIFT + L` (CTRL or CMD depending on your OS) and notice that nothing happens.

-|-|-|-

codebykat:
### Fix
When we refactored the previousIndex (#1989) and note list, we didn't account for behavior on small screens. This PR disentangles that by clarifying when we want `CLOSE_NOTE` (unselect the selected note) and when we want `TOGGLE_NOTE_LIST` (keep a note selected, but switch to list view).

This has the interesting side effect of making it possible to maintain a selected note on the small screen. I removed the check that disables the selected note highlight on small views. You can now use the note list keyboard command to toggle between list and note view without losing the selected note, and use next/previous note from either view to select and open a new note.

This will fix #2029 

### Test
1. Load the app on a narrow window and ensure the note list shows up (not a blank screen)
2. Select a note. The note should show
3. Ensure the back button on the note takes you back to the note list
4. Test keyboard commands to show/hide note list and move between next/previous notes

### Release
Not updated: Fix a bug that made it impossible to access the note list on small screens

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -38,16 +38,12 @@ const isElectron = (() => {
   return () => foundElectron;
 })();
 
-type OwnProps = {
-  isSmallScreen: boolean;
-};
-
 type StateProps = {
   editingEnabled: boolean;
   searchQuery: string;
 };
 
-type Props = OwnProps & StateProps;
+type Props = StateProps;
 
 class NoteContentEditor extends Component<Props> {
   static propTypes = {
@@ -366,7 +362,6 @@ class NoteContentEditor extends Component<Props> {
         <Editor
           key={this.editorKey}
           ref={this.saveEditorRef}
-          readOnly={!this.props.editingEnabled}
           spellCheck={this.props.spellCheckEnabled}
           stripPastedStyles
           onChange={this.handleEditorStateChange}
@@ -379,11 +374,9 @@ class NoteContentEditor extends Component<Props> {
   }
 }
 
-const mapStateToProps: S.MapState<StateProps, OwnProps> = (
-  { ui: { searchQuery, showNoteList } },
-  { isSmallScreen }
-) => ({
-  editingEnabled: !(isSmallScreen && showNoteList),
+const mapStateToProps: S.MapState<StateProps> = ({
+  ui: { searchQuery, showNoteList },
+}) => ({
   searchQuery,
 });
 
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -38,11 +38,16 @@ const isElectron = (() => {
   return () => foundElectron;
 })();
 
+type OwnProps = {
+  isSmallScreen: boolean;
+};
+
 type StateProps = {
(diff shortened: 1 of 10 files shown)


lib/note-content-editor.tsx
```diff
@@ -352,6 +357,7 @@ class NoteContentEditor extends Component<Props> {
   };
 
   render() {
+    console.log(this.props.editingEnabled);
```
belcherj, line 360: Remove console.log

lib/note-detail/index.tsx
```diff
@@ -249,6 +249,7 @@ export class NoteDetail extends Component<Props> {
     const {
       note,
       fontSize,
+      isSmallScreen,
```
belcherj, line 252: I don't think this is needed due to https://github.com/Automattic/simplenote-electron/pull/2037/files#diff-52e457c4f342fbeecf796739f2492ccaR385

lib/note-editor/index.tsx
```diff
@@ -139,14 +140,15 @@ export class NoteEditor extends Component<Props> {
...
-    const { editMode, note, noteBucket, fontSize } = this.props;
+    const { editMode, isSmallScreen, note, noteBucket, fontSize } = this.props;
     const revision = this.props.revision || note;
     const tags = (revision && revision.data && revision.data.tags) || [];
     const isTrashed = !!(note && note.data.deleted);
 
     return (
       <div className="note-editor theme-color-bg theme-color-fg">
         <NoteDetail
+          isSmallScreen={isSmallScreen}
```
belcherj, line 151: I don't think this is needed due to https://github.com/Automattic/simplenote-electron/pull/2037/files#diff-52e457c4f342fbeecf796739f2492ccaR385

belcherj (changes requested):

codebykat:
Addressed the review comments and I think this is looking great! Maintaining the currently selected note and using `Enter` to open into note view feels really slick. I think this a lot more usable than it was before honestly.

I have one really minor edge case, which is that on first load in wide screen, a note is seemingly selected (the first note is in the editor window), but if you resize the window to narrow screen, you get the note list. If you instead select a note first and then resize, it sticks with the note view. I think this is probably fine since the user hasn't actively selected a note yet, but it's a little inconsistent.

belcherj (approved):
Ship It!

PR #2039, opened 2020-04-24
Update `node-simperium` to fix some duplicate-write issues

dmsnell:
In certain situations we can find more than one WebSocket connection to
Simperium open and messages come back from the server over both. This
tends to happen when the network connection is spotty: `node-simperium`
thinks the WebSocket is closed but the browser still flushes out the
last messages on it as soon as network connectivity returns.

In the update to `node-simperium` it guards against certain kinds of
"ghost writing" in these situations by making sure that duplicate or
out-of-order messages from the server don't get applied more than once.


dmsnell:
Not merging now because we noticed a change in behavior that was unexpected. Though it seems like this did stop ghost-writing in some cases, it stopped syncing as well.

PR #2041, opened 2020-04-25
Hotkeys: Always allow using the Ctrl-T shortcut to jump between note/…

### LINKED ISSUE #2028: Shortcut to toggle between note/tag editing breaks after removing tag
### Expected

I expect to always be able to toggle between editing the note content and editing tags using the documented shortcut for "Toggle editing content/tags" (e.g. `CTRL + T` on Windows).

### Observed

The shortcut works as expected in these cases:

* Toggle without making any changes
* Toggle to tag editor after changing note content
* Toggle to note editor after adding a tag

The shortcut doesn't work as expected in this case:

* Toggle to note editor after removing a tag

### Reproduced

Note: The steps below use the Windows keyboard shortcut but the same behavior was observed on macOS as well, using the macOS shortcut.

1. Open a note with a tag.
2. With focus on the tag editor, backspace twice to remove a tag.
3. Type `CTRL + T` to toggle focus to the note editor.
4. Start typing in the note editor and notice that the cursor disappears and the content is not added.

This screencast shows the working scenarios and then the non-working case:

electron_tag-toggle-shortcut

-|-|-|-

dmsnell:
Resolves #2028

Previously we were only allowing a switch _from_ the note editor _to_ the
tag editor, or _from_ the tag editor _to_ the note editor, but we really
want to allow jumping between these fields any time, even if none is
selected.

With the work to improve keyboard shortcut usability this is more important
since we're emphasizing keyboard navigation.

In this patch we change the conditional from one that gates the behavior
to one that uses the current focus to determine which to jump to.

See #2028 for testing instructions.

### DIFF
--- a/lib/note-editor/index.tsx
+++ b/lib/note-editor/index.tsx
@@ -106,7 +106,7 @@ export class NoteEditor extends Component<Props> {
         event.stopPropagation();
         event.preventDefault();
         return false;
-      } else if (!this.tagFieldHasFocus()) {
+      } else {
         this.focusTagField && this.focusTagField();
 
         event.stopPropagation();
--- a/lib/note-editor/index.tsx
+++ b/lib/note-editor/index.tsx
@@ -101,13 +101,13 @@ export class NoteEditor extends Component<Props> {
     ) {
       // prefer focusing the edit field first
       if (!this.editFieldHasFocus()) {
-        this.focusNoteEditor && this.focusNoteEditor();
+        this.focusNoteEditor?.();
 
         event.stopPropagation();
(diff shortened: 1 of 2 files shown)


dmsnell:
@codebykat this is ready to try again. I think that it might be enough to listen for that hotkey and unselect the current tag when it occurs.

in the end it's not the deletion of a tag that triggers this behavior but the selecting of a tag, so you can recreate it by backspacing once over the tag and then hitting <kbd>Ctrl</kbd> + <kbd>t</kbd>. There's a race condition between `setState()` and `focus()` and the key presses.

Issue #1986, opened 2020-04-04
ARM v8 binaries


Feakster:
Will you be releasing ARM v8 Linux binaries in the future?

dmsnell:
Thanks for the question @Feakster - I don't anticipate us releasing any ARM binaries in the near future. If Electron and `electron-builder` support it then we will too, as we have no native code in the Simplenote part of the app, but that support is still largely missing.

Electron has apparently started supporting Windows ARM64 binaries but I think even that is still in its early stages. We'll announce if it we do start though.

Just out of curiosity, what is this for? A Raspberry Pi?

Feakster:
A Pinebook Pro.

dmsnell:
Very good. I wouldn't hold my breath for this. We mostly focus on the app code and rely on `electron-builder` for building the binaries. You can keep track of what's happening in Electron or in other Electron apps. It looks like there are ways to make it work but I'm pretty confident that, for example, you won't find Simplenote on ARM before VSCode or Slack are officially available on ARM.

Closing the issue for now. Thanks for using Simplenote and sharing!

Feakster:
Thanks for getting back to me. Will use the website interface for now.

rschouwenburg:
Electron and electron-builder support both armv7l and arm64.

I just compiled and ran simplenote-electron on a Pinebook Pro (arm64). No modificatons to the source was necessary except for adding arm64 to the electron-builder.js

Would it be possiblee to see if a release can be made for these two platforms?

Currently Pinebook Pro, Pinephone, Rock Pro, Raspberry Pi and Purism Librem 5 benefit from this and we all love our beloved note app.

dmsnell:
@rschouwenburg would you be willing to submit a patch to our build process? that would definitely help us get to Arm faster.

rschouwenburg:
Sure no problem. Or would you like a pull request?

dmsnell:
@rschouwenburg I meant a pull request, thanks!

rschouwenburg:
Here you go. Let me know if I did not do this correctly.

Happy to beta test on production!

Feakster:
I also wouldn't mind beta-testing this if an arm64 version becomes available.

Feakster:
Thanks for creating the aarch64 and armv7h binaries. just tried the aarch64 one on Manjaro ARM. Works like a charm.

dmsnell:
[...]

PR #2042, opened 2020-04-26
Added armv7l and arm64 build targets to Linux build

rschouwenburg:
### Fix
Add Linux armv7l and arm64 build target support for Linux build. Standard Linux build environment  worked for me. I build on Ubuntu LTS 18.0.4.3 x64 and tested build packages on arm64 (Pinebook Pro running Manjaro 20.04) and armv7l (Raspberry Pi running Debian 10.3).

PR asked for by @dmsnell on https://github.com/Automattic/simplenote-electron/issues/1986

### Test

Linux build now creates armv7 and arm64 packages:

`make package-linux`

### Release

Simplenote now supports armv7l (aka armhf on Debian) and arm64 platforms

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -60,19 +60,19 @@
     "target": [
       {
         "target": "AppImage",
-        "arch": ["x64", "ia32"]
+        "arch": ["x64", "ia32", "armv7l", "arm64"]
       },
       {
         "target": "deb",
-        "arch": ["x64", "ia32"]
+        "arch": ["x64", "ia32", "armv7l", "arm64"]
       },
       {
         "target": "rpm",
-        "arch": ["x64", "ia32"]
+        "arch": ["x64", "ia32", "armv7l", "arm64"]
       },
       {
         "target": "tar.gz",
-        "arch": ["x64", "ia32"]
+        "arch": ["x64", "ia32", "armv7l", "arm64"]
       }
     ],
     "synopsis": "The simplest way to keep notes",


dmsnell:
@nsakaimbo would you be willing to give this a review?

belcherj (approved):
I am happy with these changes.  By providing these builds we will not necessarily be testing releases on arm based systems.

belcherj:
We will need to see if this builds before merging.

nsakaimbo (approved):
The change looks straightforward enough - couldn't find `arm7l` in the Electron docs, but it is in the electron-builder code `https://github.com/electron-userland/electron-builder/blob/cd28eb093f0bf5969ad87c64773cc0714dc8d8ea/packages/builder-util/src/arch.ts`.

To echo what @belcherj said, as long as there are clear expectations around debugging/troubleshooting issues that come up on these specific platforms then should be fine. (From time-to-time there are platform-specific issues that arise when trying to build or run any specific variant, so if we're willing to accept the maintenance overhead as well then LGTM.)

👍

belcherj:
The build succeeded.

valiant66:
I'm just an end user, not a coder.

I've tried installing Simplenote onto my Raspberry Pi Four using Snap and gdebi. Both throw "not available on stable for this architecture" type errors.

I know very little about git.

a. Is there an unstable version of Simplenote that runs on Raspberry Pi Four?
b. where would I turn to learn how to install it.

THanks

belcherj:
The next release will have arm versions available. The releases are available here: https://github.com/Automattic/simplenote-electron/releases

The betas currently there are pretty stable.

We do not have access to arm devices for testing or have knowledge of installing on Raspberry Pi Four

rschouwenburg:
Excellent news! Thank you!

I did some testing on both Raspberry Pi (deb, armv7l) and Pinebook Pro (AppImage, arm64) and both versions work. I tried most common activities and everything works as expected. I did encounter a couple of bugs but I do not know if they are related to the 1.7 beta or the arm platform builds. None of these bugs were present in 1.15.1 I compiled on myself on these platforms.

Let me know how you would like to go about this? I am happy to try to fix any arm platform related bugs but I rather wait until I can get my hands on a stable version if that makes sense.

belcherj:
@rschouwenburg I think opening a new issue is the best bet! Let us know what the bugs are and we can figure out the cause.

PR #2046, opened 2020-04-28
Fix keyboard shortcut issues identified in release testing

### LINKED ISSUE #2025: Search: Keyboard shortcut for "search within note" doesn't work as expected
In this case there's a mismatch between the expected and observed behavior, but the problem mainly stems from how the keyboard shortcut is documented (i.e. calling it "search within note" is misleading).

### Expected

I see a keyboard shortcut `CTRL + G` listed as "Search within note." I expect to be able to start with the focus in the note editor and use that shortcut to start searching within a note.

### Observed

When the focus is in the note editor and I type `CTRL + G` nothing happens.

That keyboard shortcut only works if I already did a regular search (e.g. starting with `CTRL + SHIFT + F`) and then use `CTRL + G` to cycle through the keyword matches in the selected note.

### Reproduced

1. Select a note.
2. Click in the note editor to give it focus.
3. Type `CTRL + G` and notice that nothing happens.
4. Type `CTRL + SHIFT + F` and notice the search bar gets focus.
5. Type a search with results in a note.
6. Type `CTRL + G` and notice that there is additional highlighting in the selected note, and each time you type the keyboard shortcut the next match within the note is highlighted.

-|-|-|-

### LINKED ISSUE #2026: Focus mode keyboard shortcut broken
### Expected

I expect to toggle Focus Mode with the keyboard shortcut `CTRL + SHIFT + F`/`CMD + SHIFT + F` (as documented in the View menu).

### Observed

As of https://github.com/Automattic/simplenote-electron/pull/1983 the shortcut `CTRL + SHIFT + F`/`CMD + SHIFT + F` activates search instead of focus mode.

### Reproduced

1. Open View menu and note the keyboard shortcut listed for Focus Mode (`CTRL + SHIFT + F`/`CMD + SHIFT + F`).
2. Open Help > Keyboard Shortcuts and note the same keyboard shortcut is listed as "Focus search field" there.
3. Type `CTRL + SHIFT + F`/`CMD + SHIFT + F` and notice that the search field gets focus.

-|-|-|-

dmsnell:
Resolves #2025
Resolves #2026

Changes:
 - restores "focus mode" hotkey to Ctrl + Shift + F
 - resets "focus search field" hotkey to Ctrl + F
 - rewords description of Ctrl + G
 - rewords "select previous/next" to "above/below"
 - disables browser default behavior on Ctrl + G
 - removes hotkey Ctrl + Backspace to prevent accidentally trashing note
 - inlines keyboard shortcut explainer to one column only for visual styling purposes

### DIFF
--- a/lib/dialogs/keybindings/index.tsx
+++ b/lib/dialogs/keybindings/index.tsx
@@ -46,8 +46,8 @@ const Keys = ({
 export class AboutDialog extends Component<Props> {
   render() {
     const { closeDialog, isElectron, isMacApp } = this.props;
-
-    const CmdOrCtrl = isMacApp ? 'Cmd' : 'Ctrl';
+    const isMac = isMacApp || navigator.userAgent.indexOf('Mac OS X');
+    const CmdOrCtrl = isMac ? 'Cmd' : 'Ctrl';
 
     return (
       <div className="keybindings">
@@ -145,7 +145,7 @@ export class AboutDialog extends Component<Props> {
                 <li>
                   <Keys
                     keys={
-                      isMacApp
+                      isMac
                         ? ['fn', CmdOrCtrl, 'Delete']
                         : [CmdOrCtrl, 'Delete']
                     }
--- a/lib/dialogs/keybindings/index.tsx
+++ b/lib/dialogs/keybindings/index.tsx
@@ -61,11 +61,21 @@ export class AboutDialog extends Component<Props> {
                 </li>
                 <li>
                   <Keys keys={[CmdOrCtrl, 'Shift', 'F']}>
-                    Focus search field
+                    Toggle focus mode
                   </Keys>
                 </li>
                 <li>
-                  <Keys keys={[CmdOrCtrl, 'G']}>Search within note</Keys>
+                  <Keys keys={[CmdOrCtrl, 'F']}>Focus search field</Keys>
+                </li>
+                <li>
+                  <Keys keys={[CmdOrCtrl, 'G']}>
+                    Jump to next match in note
+                  </Keys>
+                </li>
+                <li>
+                  <Keys keys={[CmdOrCtrl, 'Shift', 'G']}>
+                    Jump to previous match in note
+                  </Keys>
                 </li>
                 {isElectron && (
                   <li>
@@ -105,11 +115,13 @@ export class AboutDialog extends Component<Props> {
                 </li>
                 <li>
                   <Keys keys={[CmdOrCtrl, 'Shift', 'K']}>
-                    Select previous note
+                    Open note above current one
                   </Keys>
                 </li>
                 <li>
-                  <Keys keys={[CmdOrCtrl, 'Shift', 'J']}>Select next note</Keys>
+                  <Keys keys={[CmdOrCtrl, 'Shift', 'J']}>
+                    Open note below current one
+                  </Keys>
                 </li>
                 <li>
                   <Keys keys={[CmdOrCtrl, 'T']}>
@@ -118,9 +130,7 @@ export class AboutDialog extends Component<Props> {
                 </li>
                 <li>
                   <Keys keys={[CmdOrCtrl, 'Shift', 'L']}>
-                    Toggle note list
-                    <br />
-                    (on narrow screens)
+                    Toggle note list (on narrow screens)
                   </Keys>
                 </li>
               </ul>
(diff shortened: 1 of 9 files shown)


codebykat:
This is in good shape aside from `T H E`  `B E E P I N G` which I have confirmed to be present on other websites with a text field, as well as in Chrome. [For those not in the Slack: `Ctrl+/` with a cursor in a text field causes a mysterious error beep on the Mac.] But I haven't found another site for which it's a valid key combo that also happens to beep.

Unfortunately I think the right answer here might be to display `Cmd+/` for this on the Mac (whether or not we still respond to both). Just to note, everything in View now works without conflicts, since we're overriding the browser search completely.

Speaking of which -- the search is much nicer now! No more inconsistencies where you can have both browser and Simplenote search competing with each other.

The wider view was sooooooooort of nicer when there was space to display it all at once, but that's something we can always revisit.

belcherj:
The note list does not open and close.

belcherj:
CMD + Delete is not deleting notes for me in macOS Electron

belcherj:
ctrl + / does not open keyboard shortcuts on Edge in Windows 10

belcherj:
ctrl + t does not work on fedora firefox, it opens a new tab

belcherj:
ctrl + shift + p does not work on Fedora Firefox. It opens a new private browsing tab.

belcherj:
ctrl + t does not work in Windows 10 Firefox, it opens a new browser tab.

belcherj:
The Edge issues occurred on version  44.18362.449.0

belcherj:
After updating to Chromium Edge things appear to work better.  We may want to just add a note to say that Edge < 80 is not supported.

dmsnell:
> CMD + Delete is not deleting notes for me in macOS Electron

The problem here is more that the macOS keyboards lie. The big button marked "delete" is sending "Backspace" and truly it has to be that way because certain OS-level text shortcuts use it. We had it in here but then I noticed that I was accidentally trashing notes all the time because <kbd>⌘</kbd>+<kbd>⌫</kbd> deletes all the text in the current text field from the before the cursor to the start of the line.

macOS distinguishes a _forward delete_ <kbd>⌦</kbd> from a backspace or _backward delete_ <kbd>⌫</kbd>. Considering that our macOS builds are mostly incidental vs PCs where the delete key works as expected I'm leaning towards leaving this as-is (you should still be able to access the shortcut with <kbd>fn</kbd> + <kbd>delete</kbd>). Maybe we can verify that and add it to the keyboard explainer somehow. I'd like a better way to explain this than a paragraph 🙃

dmsnell:
@belcherj for issues with existing keybindings that are browser keybindings what do you propose we do? for <kbd>ctrl</kbd> + <kbd>t</kbd> and <kbd>ctrl</kbd> + <kbd>shift</kbd> + <kbd>p</kbd> these are not new hotkeys or issues.

codebykat:
> The note list does not open and close.

https://github.com/Automattic/simplenote-electron/pull/2037

belcherj:
Something needs to change because:

macOS Electron:

Screen Shot 2020-04-28 at 2 42 25 PM

macOS Web:

Screen Shot 2020-04-28 at 2 42 57 PM

Neither work as expected.  

The keyboards say delete so seeing cmd + delete or ctrl + delete should work or it should be changed.

belcherj:
ctrl + t and ctrl + shift + p We can leave this as is but I'd like to add an additional hotkey that does work change it in the menu and then deprecate the old ones a couple of version later.

belcherj:
I think we can also ignore the old edge issues as they are existing.

lib/dialogs/keybindings/index.tsx
```diff
@@ -143,7 +143,15 @@ export class AboutDialog extends Component<Props> {
                   <Keys keys={[CmdOrCtrl, 'Shift', 'N']}>Create new note</Keys>
                 </li>
                 <li>
-                  <Keys keys={[CmdOrCtrl, 'Delete']}>Trash note</Keys>
+                  <Keys
+                    keys={
+                      isMacApp
```
dmsnell, line 148: @belcherj we want to detect macOS from the browser as well. I don't think we care too much about people who are masking their user agent (they probably know what's going on here anyway), though we might lose our user-agent sniffing ability with privacy controls in browser updates

PR #2048, opened 2020-04-29
Feature/support markdown inline code highlight

### LINKED ISSUE #1891: Code elements are hard to distinguish
\`code markdown\` is hard to distinguish from the rest of the text. The font is definitely different from regular text but it can be hard to pick out. Giving it a little `background color` would make it easy to locate. Alternatively the iOS app uses a gray font color for code, which works too.

happybridge14-totoro:
### Fix
Support inline code highlight mode in markdown preview
[Inline code in markdown preview](https://github.com/Automattic/simplenote-electron/issues/1891)
* A quick development for inline codes feature.
* Apply `highlight.js` to all code tags, set `display: inline` CSS style attribute to those who are not under `pre` tag

### Test
1. Click "INFO" and enable Markdown
2. Add inline codes with signs or code tag
3. Add code blocks
4. Click "Preview"
5. Successfully display both inline codes and code blocks

### Release
RELEASE-NOTES.txt was updated with:
- Highlight inline code in markdown preview

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,9 @@
 
 ## [v1.15.1]
 
+### Enhancements
+- Highlight inline code in markdown preview
+
 ### Other Changes
 
 - Fix application signing to generate proper appx for Windows Store [#1960](https://github.com/Automattic/simplenote-electron/pull/1960)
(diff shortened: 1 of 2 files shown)


dcalhoun:
@happybridge14-totoro thank you for taking the time to contribute improvements to Simplenote. After a discussion in https://github.com/Automattic/simplenote-electron/issues/1891#issuecomment-794159658, we have decided to apply a background to inline code elements rather than enabling syntax highlighting on them. Therefore I am going to close this PR as it is now superseded by #2739.

Thanks again!

PR #2051, opened 2020-04-29
Keyboard: Fix reopen-last-closed-tab

codebykat:
### Fix
Cmd+shift+T is supposed to reopen the most recently closed tab but we were overriding it.

With this change, the browser behavior still works, but the tag list is toggled as well, which is kind of weird. It would probably be better if we could only toggle the tag list on Ctrl in the Mac browser. I'm not sure about Windows shortcut keys.

### Test
1. Go to Simplenote
2. Open a new tab
3. Close it so that the Simplenote tab is now focused
4. Type Cmd+shift+T and note that it reopens the closed tab.

### Release
Not updated: Fix an issue where Cmd+Shift+T could not be used to reopen a recently closed tab in the browser.

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -27,7 +27,7 @@ import {
   closeNote,
   setUnsyncedNoteIds,
   toggleNavigation,
-  toggleSimperiumConnectionStatus
+  toggleSimperiumConnectionStatus,
 } from './state/ui/actions';
 
 import * as settingsActions from './state/settings/actions';
@@ -58,10 +58,10 @@ export type DispatchProps = {
 
 export type Props = OwnProps & StateProps & DispatchProps;
 
-const mapStateToProps: S.MapState<StateProps> = state => ({
+const mapStateToProps: S.MapState<StateProps> = (state) => ({
   ...state,
   authIsPending: selectors.auth.authIsPending(state),
-  isAuthorized: selectors.auth.isAuthorized(state)
+  isAuthorized: selectors.auth.isAuthorized(state),
 });
 
 const mapDispatchToProps: S.MapDispatch<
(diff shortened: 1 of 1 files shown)


codebykat:
This is failing prettier even though I rebased to develop and it's a one-line change.  Not sure how to fix.

belcherj:
Did you mean to make this change against develop?

lib/app.tsx
```diff
@@ -234,7 +234,6 @@ export const App = connect(
         this.props.openTagList();
 
         event.stopPropagation();
-        event.preventDefault();
```
dmsnell, line 237: if we do this, won't we _always_ re-open the last-closed tab if we can? I don't think we want that either.

it seems like it could be a good idea to consider a different keybinding. I spoke with @belcherj about this yesterday and we didn't have any good ideas, what do you think about changing it?

it's the "tag list" or the "group list" or the "side sidebar"

<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>o</kbd>

`o` for "organization" ?
<del>`c` for "categorization" ?</del> (dev-tools select element)
`e` because it's available?
`m` because it's available?

codebykat: Yeah, it's kind of annoying. On the Mac you can choose to use Ctrl in the browser and not Cmd (and I wish we could just exclude Cmd from being valid). On Windows I guess you're stuck. I would be open to changing it.

codebykat:
> Did you mean to make this change against develop?

Yeah, it's not specifically a patch for the release (i.e. not a release blocker as this problem has existed for some time -- unless I'm wrong about that)

codebykat:
2965582-zen:

>Just wanted to say new keyboard shortcuts in Simplenote are really inconvenient, as they interfere with established shortcuts in Safari.
> ⌘ + F ➙ should start a Safari search, not Simplenote search. While Simplenote search is ok for finding the note that contains a keyword, it wouldn’t show the exact location of the keyword within the note. I have notes with thousands of words, so finding anything with Simplenote search is impossible. Safari search lets you scroll through every instance of the keyword, and now I have to manually pick that option from the menu bar.
> ⌘ + Shift + N ➙ should start a private window and instead starts a new note.

codebykat:
Abandoning in favor of https://github.com/Automattic/simplenote-electron/pull/2079

PR #2057, opened 2020-05-04
Merge release 1.16.0 into develop

mokagio:


lib/dialogs/keybindings/index.tsx
```diff
@@ -46,7 +46,7 @@ const Keys = ({
 export class AboutDialog extends Component<Props> {
   render() {
     const { closeDialog, isElectron, isMacApp } = this.props;
-
+    const isMac = isMacApp || navigator.userAgent.indexOf('Mac OS X');
```
dmsnell, line 49: looks like this change was an improper merge-conflict resolution.
when I ran test-merges I found a few conflicts that required careful resolution so we don't wipe-out the changes made in `develop` since we forked `release/1.16`

it's probably most helpful to review the PRs in `develop` while resolving the conflicts to see what changed and how it will need to be resolved. in this case there was a utility function introduced in #2004 that should be taken into consideration here.

off the top of my head I don't remember the others but we need to be careful with this merge or we'll erase the work that @belcherj did

dmsnell (changes requested):

lib/dialogs/keybindings/index.tsx
mokagio: Hey @dmsnell, thanks for picking this up, phew... 😅 

> looks like this change was an improper merge-conflict resolution.

I'm not sure I follow, sorry. _This is my first run at opening this kind of PRs for the release process, bare with me._

I didn't attempt any conflict resolution, just opened the PR. In fact, GitHub is showing a few conflicting files, including this one.

Screen Shot 2020-05-05 at 11 35 34 am

I can _try_ to resolve the conflicts, but I think it's better if one of you core devs does it, right?

dmsnell: > bare with me

No worries. As long as we don't rush we'll be fine.

> I can try to resolve the conflicts

We should absolutely avoid Github's tools to resolve the conflicts since they have a tendency to force-push and destroy the repo.

Typically someone adds new commits to the release branch to resolve the conflicts. Technically that violates our official release process and practically it can make for some confusing state in the repo, leading to weird divergence between `develop` and `master`.

> I think it's better if one of you core devs does it, right?

we can try. maybe we can pull a cherry-pick from the release into `develop`. this PR would ideally need to be recreated after that.

mokagio:
Closing in favor of #2062.

@dmsnell and I had a call and agreed that that PR is the way to go. I cleanly resolves the conflict we see here in a merge commit, so we don't have extra commits in the `release/1.16` branch. Because the conflicts are solved already, it's also mergeable into `develop` immediately. 

The other option we looked at was to just push the merge commit of `release/1.16` into `develop` from one local to the Git server here. That would have required disabling the branch security and wouldn't have left a paper trail that we can look at on GitHub.

Overall, creating a new branch and PR seemed a clearer approach.

dmsnell:
> The other option we looked at was to just push the merge commit of release/1.16 into develop from one local to the Git server here. That would have required disabling the branch security and wouldn't have left a paper trail that we can look at on GitHub.

For for the sake of history, I want to point out that this was never an option we would have considered. We may have acknowledge that it was possible, but we did not deliberate whether or not to do it.

PR #2058, opened 2020-05-04
Merge release 1.16.0 into master

mokagio:


mokagio:
Closing in favor of #2061.

@dmsnell and I had a call and agreed that that PR is the way to go. I cleanly resolves the conflict we see here in a merge commit, so we don't have extra commits in the `release/1.16` branch. Because the conflicts are solved already, it's also mergeable into `master` immediately. 

The other option we looked at was to just push the merge commit of `release/1.16` into `master` from one local to the Git server here. That would have required disabling the branch security and wouldn't have left a paper trail that we can look at on GitHub.

Overall, creating a new branch and PR seemed a clearer approach.

dmsnell:
> The other option we looked at was to just push the merge commit of release/1.16 into master from one local to the Git server here. That would have required disabling the branch security and wouldn't have left a paper trail that we can look at on GitHub.

For for the sake of history, I want to point out that this was never an option we would have considered. We may have acknowledge that it was possible, but we did not deliberate whether or not to do it.

PR #2066, opened 2020-05-07
Refactor: Extract authentication and login screen from main app

dmsnell:
Explored in 5f1f65d
Explored in e3c660e

This work exists in preparation for further and deeper work to decouple Simperium in the app data flows and to finish the internal state refactors.

The goal here is to initialize Simperium only after having proper authentication in order to allow us to move Simperium into Redux middleware. The further goal is to remove the race condition that exists in many places between making an edit or clicking a button, making a network call to Simperium (or not), updating `indexedDB`, and rerendering the app.

App boot is now handled on its own and centralizes the token-loading process. Upon logout it force-reloads the browser window to clear out app state. This is necessary due to the ways that things like `client` and `app-state` initialize their variables in the module global scope. We can't reload the app state once the module has been imported the first time. Reloading the page completely resets this. There is a flash of a white screen when logging out.

The `auth` component of app state has been correspondingly removed because the app will not load without an authorization. A new action `LOGOUT` has been created in order to trigger a logout in the app, driven from the Simperium middleware.

## Status

Ready for merge, deployed to https://app.simplenote.com already

## Todo

 - [x] Add back analytics events
 - [x] Create welcome note when creating an account
 - [x] Fix issue where you can get stuck logging out


codebykat (commented):
Code looks great and I didn't see any issues so far in testing. I hope to spend some more time testing this before we merge, but it's not a blocker if you feel confident.

lib/boot-without-auth.tsx
```diff
@@ -0,0 +1,117 @@
...
+        });
+    });
+  };
+
+  tokenLogin = (username: string, token: string) => {
+    this.props.onAuth(token, username, false);
+  };
+
+  render() {
+    const systemTheme = window.matchMedia('(prefers-color-scheme: dark)')
```
belcherj, line 90: I think this needs to be smarter pulling from localstorage before checking system theme.

belcherj: I'm happy to leave this for another PR, I'm not sure how it currently works.

lib/boot.ts
```diff
@@ -1,246 +1,81 @@
...
-        }
+const run = (
+  token: string | null,
+  username: string | null,
+  createWelcomeNote: boolean
+) => {
+  if (token) {
+    import('./boot-with-auth').then(({ bootWithToken }) => {
+      bootWithToken(
+        () => {
```
belcherj, line 96: boot feels like the wrong place to have all this logic. Not now but in future PR's how do we move all this out of boot.

lib/state/simperium/middleware.ts
```diff
@@ -9,48 +11,88 @@ import * as T from '../../types';
...
-      }
+    const note = state.ui.note;
 
-      const thisState = store.getState();
-      if (!(thisState.ui.note && note.id === thisState.ui.note.id)) {
-        return;
-      }
+    noteBucket.getRevisions(
+      note.id,
+      (error: unknown, revisions: T.NoteEntity[]) => {
```
belcherj, line 74: is this truly unknown? You later use string interpolation. Can we assume it is not a string?

lib/state/simperium/middleware.ts
```diff
@@ -9,48 +11,88 @@ import * as T from '../../types';
...
+            `Failed to load revisions for note ${note.id}: ${error}`
+          );
+        }
 
-      store.dispatch(actions.ui.storeRevisions(note.id, revisions));
-    }
-  );
-};
+        const thisState = store.getState();
+        if (!(thisState.ui.note && note.id === thisState.ui.note.id)) {
```
belcherj, line 82: why use the parens with the bang?

`thisState.ui.note && note.id !== thisState.ui.note.id`

lib/boot-without-auth.tsx
dmsnell: Note that at this point we aren't logged in. In the existing code we don't clear settings and so that stays in `localStorage`, which while convenient for places like this also seems wrong. If we really want to log out then I think we should forget the settings that person may have saved.

Do you feel differently?

lib/boot.ts
dmsnell: What feels wrong about it?

lib/state/simperium/middleware.ts
dmsnell: it's unknown to me right now 😄 

the string interpolation relies on `toString()` which almost every type will have but that isn't itself meaningful enough for us do anything with. at some point I'd like an error audit, also a type definition for the Simperium library which would resolve this.

in the meantime `unknown` is serving like `any` but backwards - it prevents us from making assumptions about `error` in any way, which might keep us from raising type errors in runtime if we were to do something like `error.message` (which was causing type errors on sign-out and is something I fixed in this PR in the settings dialog).

lib/state/simperium/middleware.ts
dmsnell: hm. I think you are asking why I didn't write different logic? as written, the parentheses is important because the `!` takes precedence in associativity over the `&&`

the logic you wrote in the comment _is_ an alternative to the `!` but it fails to grab the case when `thisState.ui.note` is `null` - the condition would fail and we would continue to dispatch the revisions for a note that's not longer selected.

whenever we want to invert logic we need to rely on deMorgan's laws for the full flip, otherwise we'd miss cases like this. `&&` and `||` swap and every term has to be negated.

```ts
if (!(thisState.ui.note && note.id === thisState.ui.note.id)) {}
if (!thisState.ui.note || note.id !== thisState.ui.note.id) {}
```

at this point I often prefer to write the first case since it reads easier for me. the difference is "if we don't have: a note and one which matches the id" vs. "if we don't have a note or we have a note but the id doesn't match." this tends to be more notable as the list gets larger with more terms but in general I do find the symmetry of this construction to be helpful. "look for all the requirements and _then_ if we don't have all of them, take this other route"

lib/boot-without-auth.tsx
belcherj: If you delete the settings that person set then every time they log in they will have to go change their settings. Until we have a way to preserve settings then I think it cannot be cleared.

lib/boot.ts
belcherj: The code to sign a user out exists in the boot file. Analytics should be moved to middleware. Logging out functionality should be moved to an action.

dmsnell: will discuss in Slack, but for the sake of responding here…

> the code to sign a user out exists in the boot file

yeah I did this intentionally, specifically with the `cookie` handling. I wanted to move all the persistence and clearing of the authentication into one place and at the top so we can control how to load the app (with a side benefit of slimming up the logged-out page load)

if we move it out of `boot` where would you want it to move?

> Analytics should be moved to middleware

this is something I discovered while doing this work - we're sending out analytics before we boot the app, even in `develop`. the `analytics` module appears to queue up these events until we load and until we have a reading on `analytics_enabled`, which is good, but it definitely _looks_ wrong to be calling for trackers before being logged in (it looks like we're not waiting for consent even though we are)

if we want these edge calls to be in middleware then we're going to need to initialize the Redux store before login (which is what currently happens in `develop`). If we do that, I'd still really like to replace the store during the login, which we can do with `replaceReducer()`

I actually thought about doing this when creating this PR but I wanted to avoid the complexity of creating that replacement mechanism. if we move this in a future PR, I guess that would be a reasonable way to go about it.

> Logging out functionality should be moved to an action.

It currently is handled by an action within the app but the Simperium middleware calls this function in response to that action or in response to an authentication expiration.

belcherj (approved):
Smerge

PR #2067, opened 2020-05-07
Add 32 bit version for Windows Store

belcherj:
### Fix

This will build both a 32 and 64-bit version for the app store. I don't this is entirely needed but since we are moving to 64-bit we should have a fallback for 32 bit.

### Test

1. Create a tag, push it
2. Observe builds for win-store being created.


dmsnell:
Related #1986 #2042

nsakaimbo (approved):
Seems straightforward - LGTM. 👍

PR #2068, opened 2020-05-07
Update tag chip to make it a bit cleaner

belcherj:
### Fix

- changes the property name to tagName from tag to make it clearer
- make onClick type accurate
- make onSelect an optional type

### Test

1. Test the TagChip

### DIFF
--- a/lib/components/tag-chip/index.tsx
+++ b/lib/components/tag-chip/index.tsx
@@ -2,9 +2,7 @@ import React, { FunctionComponent } from 'react';
 import classNames from 'classnames';
 
 type OwnProps = {
-  onSelect?: (
-    event: React.MouseEvent<HTMLDivElement, globalThis.MouseEvent>
-  ) => any;
+  onSelect?: (event: React.MouseEvent<HTMLDivElement>) => any;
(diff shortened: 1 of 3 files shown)


lib/components/tag-chip/index.tsx
```diff
@@ -2,18 +2,24 @@ import React, { FunctionComponent } from 'react';
 import classNames from 'classnames';
 
 type OwnProps = {
-  onSelect: () => any;
+  onSelect?: (
+    event: React.MouseEvent<HTMLDivElement, globalThis.MouseEvent>
```
dmsnell: what's `globalThis.MouseEvent` here? I haven't come across it in the React types yet

dmsnell: it looks like it might be the default parameter so do we need it or could we also use `React.MouseEvent<HTMLDivElement>`?

dmsnell (approved):
Looks like our type might be needlessly specific on the `onSelect` prop but otherwise this looks fine to merge.

belcherj:
You are right, I grabbed the type from itellisense.

PR #2075, opened 2020-05-12
Add option to disable hotkeys

belcherj:
### Fix

Adds setting to disable hotkeys

### Test

1. Load page
2. Try hotkeys, they should work
3. Disable in settings
4. They should not work

Screen Shot 2020-05-13 at 10 10 49 PM

### DIFF
--- a/lib/state/settings/actions.ts
+++ b/lib/state/settings/actions.ts
@@ -49,6 +49,10 @@ export const setLineLength: A.ActionCreator<A.SetLineLength> = (
   lineLength,
 });
 
+export const toggleKeyboardShortcuts: A.ActionCreator<A.ToggleKeyboardShortcuts> = () => ({
+  type: 'KEYBOARD_SHORTCUTS_TOGGLE',
+});
+
 export const toggleSortOrder = () => (dispatch, getState) => {
   dispatch({
     type: 'setSortReversed',
--- a/lib/tag-field/index.tsx
+++ b/lib/tag-field/index.tsx
@@ -45,7 +45,11 @@ type DispatchProps = {
   updateNoteTags: (args: { note: T.NoteEntity; tags: T.TagEntity[] }) => any;
 };
 
-type Props = OwnProps & DispatchProps;
+type StateProps = {
+  keyboardShortcuts: boolean;
+};
+
+type Props = OwnProps & DispatchProps & StateProps;
 
 const KEY_BACKSPACE = 8;
 const KEY_TAB = 9;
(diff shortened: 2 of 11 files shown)


lib/state/settings/actions.ts
```diff
@@ -49,6 +49,15 @@ export const setLineLength: A.ActionCreator<A.SetLineLength> = (
   lineLength,
 });
 
+export const toggleDisableHotkeys = () => (dispatch, getState) => {
+  const newValue = !getState().settings.disableHotkeys;
```
codebykat: For simplicity and consistency with the rest of the actions, I don't think we need to set a variable here, we can just pass the value below like `sortReversed` etc. is doing.

codebykat (changes requested):
Okay I have 2.5 comments about this that I feel pretty strongly about:

Screen Shot 2020-05-13 at 3 18 29 AM

1. Can we not call them "hotkeys"? They're already "Keyboard shortcuts" in the popup dialog and that sounds more friendly / obvious.

2. Right now this is an "Enable to disable" toggle which personally I find annoying at best, and kind of malicious at worst. Can we make this setting "Enable keyboard shortcuts" and have it default to on? Same functionality, much less confusing.

For comparison, Fastmail does this, very classy:

Screen Shot 2020-05-13 at 3 20 32 AM

And Gmail does this:

Screen Shot 2020-05-13 at 3 21 01 AM

The last half-a-comment is, should this really be under Display? There's hella room on Tools (whereas Display has a ton of stuff in it already) and IMO it's marginally more logical there.

All of that aside, the code and functionality are okay.

SylvesterWilmott:
Agree with @codebykat 

1. Let's call them "Keyboard Shortcuts" everywhere
2. Default to on and have the toggle just be labelled "Keyboard Shortcuts"

SylvesterWilmott:
>The last half-a-comment is, should this really be under Display? There's hella room on Tools (whereas Display has a ton of stuff in it already) and IMO it's marginally more logical there.

Tools makes a lot more sense to me

lib/tag-field/index.tsx
```diff
@@ -282,6 +293,12 @@ export class TagField extends Component<Props, OwnState> {
...
->)(TagField);
+const mapStateToProps: S.MapState<StateProps> = ({
+  settings: { keyboardShortcuts },
+}) => ({
+  keyboardShortcuts,
+});
+
+export default connect(mapStateToProps, {
+  updateNoteTags,
+} as S.MapDispatch<DispatchProps>)(TagField);
```
codebykat, line 304: Fun fact: this is the only place in the codebase we're doing this (as opposed to defining a separate `MapDispatchToProps` variable) and I don't think I like it. I do realize it was already like that, though.

codebykat (approved):
One minor code comment, but this looks beautiful and works like a dream now. :shipit:

SylvesterWilmott:
Design looks good! Thanks @belcherj

lib/tag-field/index.tsx
belcherj: I'll do a follow up

PR #2076, opened 2020-05-12
Remove delete note shortcut

belcherj:
### Fix

Removes the delete note shortcut

### Test

1. Use the delete note shortcut: ctrl+fn+delete
2. It should not work


codebykat (approved):
Yes. 🚮 🗑️ 🔥 🚮

PR #2078, opened 2020-05-12
Change search hotkey so it does not replace the native find

belcherj:
### Fix

Change the search hotkey to ctrl + shift + s so that it does not conflict with the native browser find.

### Test

1. Goto app
2. Use cmd/ctrl + f, does it open browser find?
3. Use cmd/ctrl + shift + s, does it focus search field?

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -210,7 +210,7 @@ export const App = connect(
         return false;
       }
 
-      if (cmdOrCtrl && !shiftKey && 'KeyF' === code) {
+      if (cmdOrCtrl && shiftKey && 'KeyS' === code) {
         this.props.focusSearchField();
 
         event.stopPropagation();
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -210,7 +210,10 @@ export const App = connect(
         return false;
       }
 
-      if (cmdOrCtrl && shiftKey && 'KeyS' === code) {
+      if (
+        (cmdOrCtrl && shiftKey && 'KeyS' === code) ||
+        (isElectron && cmdOrCtrl && !shiftKey && 'KeyF' === code)
+      ) {
         this.props.focusSearchField();
 
         event.stopPropagation();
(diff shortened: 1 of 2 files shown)


codebykat (changes requested):
I think we need to remap Ctrl-G too. Right now you can get into a state where you have two different searches in the app search and browser find dialog, and it's confusing. Like from here if I click in the app and use Ctrl-G, it cycles through "tags", but using Ctrl-F and then Ctrl-G makes it cycle through "trash". It's bad.

Screen Shot 2020-05-13 at 3 49 59 AM

Our Ctrl-G only exists because we were trying to take over the browser-based search entirely, but in the browser, it makes things worse. There's no reason to have it when it does the same thing as the browser's Find.

In Electron I really think having Ctrl-F and Ctrl-G is really nice. There's no built-in Find functionality so it makes sense there. Can we consider dropping these shortcuts only for the browser? We could ensure a nice cross-platform experience if we support Ctrl+Shift+S in Electron as well as Ctrl+F. 

Overall I'm concerned about this entire search thing, it feels like we made a pretty major change on a whim when we should've looped more people into the discussion, and now we're reacting very quickly to feedback in a kind of piecemeal way (without a larger vision of how all this is supposed to fit together and ALSO without taking the time to think through all the flows).

codebykat (approved):
Works great on OSX Firefox and Chrome. However, we have a shortcut conflict with Firefox on Windows, sigh... Ctrl+shift+S takes a screenshot 🙄.

That being said, I **think** this is what the shortcut was set to, pre-1.16? Maybe @dmsnell can confirm. If so, I would say let's run with it (AKA revert to what it was originally) and we can rethink from there.

Interestingly, I just tested in Parallels and Ctrl+F brings up the browser Find dialog anyway on Edge. o_O (Firefox behaves as expected.)

belcherj:
Grrrr, thought i tested all the browsers. There was no search before.  

Screenshotting still works so I guess Firefox just wont get a hotkey. There are basically no hotketys left to use.

PR #2083, opened 2020-05-13
Refactors settings reducer to be seperate

belcherj:
### Fix

Due to the way the reducer was structured adding a setting that must initialize is difficult since in most cases if the user is logged in the reducer is already initialized.

So adding a default value of True for Keyboard Shortcuts is darn near impossible without adding code to initialize users for that value until we can be sure redux state is reinitialized.

### Test

1. Load app before checking out branch
2. Check out branch
3. Refresh
4. Does app work as intended?
5. Clear site data
6. Refresh
7. login
8. Play with settings
9. Do settings work?

### DIFF
--- a/lib/state/settings/reducer.ts
+++ b/lib/state/settings/reducer.ts
@@ -1,61 +1,139 @@
 import { clamp } from 'lodash';
 
+import { combineReducers } from 'redux';
+
 import * as A from '../action-types';
 import * as T from '../../types';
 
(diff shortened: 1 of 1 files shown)


codebykat (approved):
Okay, I have played around with this quite a bit and it seems solid! The code is cleaner now, even if we don't have the benefit of having the initial settings all in one place.

PR #2088, opened 2020-05-14
Add protection against insecure passwords

belcherj:
### Fix

If the user has a password that does not meet the strength requirements we are now displaying an error message that they need to reset their password.

### Test

1. Have insecure password (less than 8 characters)
2. In the electron app, try to login
3. Observe you are redirected to the reset form with email pre-filled

Screen Shot 2020-05-15 at 2 27 41 PM


codebykat:
https://github.com/Automattic/Simplenote-United/issues/60

PR #2095, opened 2020-05-18
Simplify login/signup form validation

codebykat:
### Fix

Simplifies and cleans up form validation for the signup/login form. Also takes advantage of the HTML form validation API.

These changes echo the updates made on the GAE side in https://github.com/Simperium/simplenote-gae/pull/280 and https://github.com/Simperium/simplenote-gae/pull/294

### Test

Try to sign up and log in on Electron. See if form validation behaves in a reasonable way (correct error messages, blocking form submission, etc).
Some specific things to try:
- Insecure password
- Password containing email address
- Invalid email
- Existing email
- Sign up should succeed with valid email and password

### Release

Not updated: Simplified login/signup form validation

### DIFF
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -124,6 +124,8 @@ export class Auth extends Component<Props> {
             id="login__field-username"
             onInput={this.onInput}
             onInvalid={this.onInput}
+            onFocus={this.onFocus}
+            onBlur={this.onBlur}
             placeholder="Email"
             ref={(ref) => (this.usernameInput = ref)}
             spellCheck={false}
@@ -141,6 +143,8 @@ export class Auth extends Component<Props> {
             id="login__field-password"
             onInput={this.onInput}
             onInvalid={this.onInput}
+            onFocus={this.onFocus}
+            onBlur={this.onBlur}
             placeholder="Password"
             ref={(ref) => (this.passwordInput = ref)}
             spellCheck={false}
@@ -210,7 +214,49 @@ export class Auth extends Component<Props> {
     );
   }
 
+  onBlur = (event) => {
+    // nothing has been entered yet, don't validate
+    if (
+      event.currentTarget.value === '' &&
+      !event.currentTarget.getAttribute('data-touched')
+    ) {
+      return;
+    }
+
+    // a value has been entered, validate from now on
+    event.currentTarget.className = 'validate';
+  };
+
+  onFocus = (event) => {
+    // nothing has been entered yet, don't validate
+    if (!event.currentTarget.getAttribute('data-touched')) {
+      return;
+    }
+
+    // if the input is valid when focused, don't validate on each keystroke
+    if (event.currentTarget.validity.valid) {
+      event.currentTarget.className = '';
+    } else {
+      event.currentTarget.className = 'validate';
+    }
+  };
+
   onInput = (event) => {
+    // don't be annoying about red borders during input:
+    // - don't start checking validity until field has been changed
+    // - after invalid, check at every input to see if it becomes valid
+    // see https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-ui-invalid
+
+    // monitor when the field has been changed by the user
+    if (event.currentTarget.value !== '') {
+      event.currentTarget.setAttribute('data-touched', true);
+    }
+
+    // we're using a className for validation so we can style in Electron
+    if (event.currentTarget.className !== 'validate') {
+      return;
+    }
+
     const passwordError =
       'Sorry, that password is not strong enough. Passwords must be at least 8 characters long and may not include your email address.';
 
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -31,10 +31,6 @@ export class Auth extends Component<Props> {
   };
 
   componentDidMount() {
-    if (this.usernameInput) {
-      this.usernameInput.focus();
-    }
-
     window.addEventListener('online', this.setConnectivity, false);
     window.addEventListener('offline', this.setConnectivity, false);
   }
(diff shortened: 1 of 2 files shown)


lib/auth/index.tsx
```diff
@@ -69,11 +65,15 @@ export class Auth extends Component<Props> {
       ? 'Could not create account. Please try again.'
       : 'Could not log in with the provided email address and password.';
 
+    const mainClasses = classNames('login', {
```
belcherj, line 68: I'm not sure this is needed since it will only ever be electron.

lib/auth/index.tsx
```diff
@@ -212,37 +214,126 @@ export class Auth extends Component<Props> {
     );
   }
 
+  onBlur = (event) => {
+    // nothing has been entered yet, don't validate
+    if (
+      event.currentTarget.value === '' &&
```
belcherj, line 219: In a future PR we should be using local state or useRef to manipulate the class names. Interacting directly with the DOM is not reccomended.

belcherj (changes requested):
When I give an invalid email then go to password it turns red as expected. Then when I enter an invalid password and go back to email field, the password field does not turn red.

I was unable to submit the form with the following 64 char password:

NF{ejLDxvLvgPiCGrecpufoPFsFTz8irtigJVBibgT)HBwRLaupvPwWfGzb6gyAu

lib/auth/index.tsx
codebykat: It avoids applying the class for local development 🤷

codebykat:
> When I give an invalid email then go to password it turns red as expected. Then when I enter an invalid password and go back to email field, the password field does not turn red.

I can't reproduce this with the steps provided, on either the login or signup form. Can you post the exact steps? Is it possible you were on the login form and it was only requiring a minimum of 4 characters (so the password was actually valid)?

> I was unable to submit the form with the following 64 char password:

Whoops, the regex lost its curly braces somewhere. Fixed in this PR. Do we want to patch the release branch to allow those characters?

lib/utils/validate-password/index.ts
```diff
@@ -5,6 +5,6 @@ export const validatePassword = function (password: string, email: string) {
   }
   // minimum of 8 characters
   // allow symbols ~!@#$%^&*_-+=`|(){}[]:;"',.?/
-  const re = /^[a-zA-Z0-9~ !@#$%^&*_\-+=`|()[\]:;"',.?/]{8,}$/;
+  const re = /^[a-zA-Z0-9~ !@#$%^&*_\-+=`|(){}[\]:;"',.?/]{8,}$/;
```
dmsnell: just for kicks and giggles, if our policy is "no newlines or tabs" then we can also encode that, which is more permissive for the full range of valid inputs

```ts
const re = /^[^\t\n\r]{8,}$/;
```

this can also simplify the validation if we also want to hard-code the rules in the clients (which I don't think we want to but…)

```ts
const digits = /\d/;
const symbols = /[~ !@$%^&*_\-+=`|(){}\[\]:;"',.?]/;

const isLong = re.test(password);
const hasDigits = digits.test(password);
const hasSymbols = symbols.test(password);

return isLong && hasDigits && hasSymbols
	? { isValid: true }
	: { isValid: false, needs: { digits: ! hasDigits, symbols: ! hasSymbols } }
```

anyhoo… :whistles: carry on 🥃

codebykat: https://github.com/Automattic/simplenote-electron/pull/2114

codebykat:
Fixed regex separately in https://github.com/Automattic/simplenote-electron/pull/2114 . This will need a rebase

codebykat:
Rebased 👍

lib/auth/index.tsx
```diff
@@ -145,18 +141,24 @@ export class Auth extends Component<Props> {
...
+            onInvalid={this.onInput}
+            onFocus={this.onFocus}
+            onBlur={this.onBlur}
             placeholder="Password"
             ref={(ref) => (this.passwordInput = ref)}
             spellCheck={false}
             type="password"
+            required
+            minLength="4"
+            maxLength="64"
```
belcherj: Do we still have a max length?

lib/auth/index.tsx
```diff
@@ -212,37 +214,127 @@ export class Auth extends Component<Props> {
...
+        return;
+      }
+
+      this.passwordInput.setCustomValidity('');
+      this.props.signup(username, password, true);
+    }
+
+    // login - slightly more relaxed password rules
+    else {
+      if (!this.passwordInput.validity.valid) {
```
belcherj, line 326: why are we allowing relaxed passwords?

belcherj (changes requested):

lib/auth/index.tsx
codebykat: Ah we don't good catch!

lib/auth/index.tsx
codebykat: We haven't implemented the check for forcing-reset-on-login yet... but even so, we first need to submit the auth, which means we need to allow existing passwords to go through

belcherj: Ahh, yes. I thought we do have the check: https://github.com/Automattic/simplenote-electron/pull/2088/

belcherj (approved):
Ship it!

lib/auth/index.tsx
codebykat: Oh yeah I got confused about what was merged back into develop. Let's make sure to test this flow

codebykat: Yeah okay I still stand by this code. #2088 uses `authorize(username, password).then(...` which means the form needs to accept the insecure password in order to submit it for authorization, at which point we prompt them to reset. Confusing but I think we do need to do it, otherwise we don't know if it's the correct password to begin with.

codebykat: https://github.com/Automattic/simplenote-electron/pull/2088/files#diff-967127579e021461234f5de838baf700R47

PR #2101, opened 2020-05-18
Update dependencies outside of Electron

belcherj:
### Fix

Updated dependencies to the latest.

### Test

1. There are a couple major point releases in here
2. Test eslint, wait-on, highlight.js
3. Test everything in the app


codebykat (approved):
Smoke-tested and ran the linter, looks good to me

PR #2102, opened 2020-05-18
Updates electron and associated dependencies to the latest

### LINKED ISSUE #2031: Print only works once per session
### Expected

I expect to be able to print notes as many times as I want.

### Observed

I can only print once each time I open the app. After that, the print dialog seems to work but nothing happens when I confirm I want to print. (This is the same for both printing and printing to a PDF.) I have to close/reopen the app for print to work again.

This only seems to happen in the packaged apps; the web app prints every time as expected.

### Reproduced

1. Open the app and select a note.
2. Select File > Print or use the keyboard shortcut.
3. When the print dialog comes up, confirm you want to print the note.
4. Notice that the note is printed (or a PDF is created if you print to PDF).
5. Repeat steps 2-3 and notice that nothing happens.

-|-|-|-

belcherj:
### Fix

- This updates all the Electron dependencies
- Removes electron-spellchecker because Electron now has one built-in.

Should fix: https://github.com/Automattic/simplenote-electron/issues/2031

### Test

- Run the checklist on the artifacts on this PR
- Test spellcheck.
- Does it turn off and on correctly?
- Test all the menu items in electron.


nsakaimbo:
@belcherj This is a good start! Simplenote currently doesn't have spellchecking as a preference (or does it?).

In order to enable spellchecking, you'll need to toggle `spellcheck` in the window's web preferences to `true`. In wp-desktop, that value is loaded from the user's "spellcheck enabled" setting at app start. (Also, it's not overridable at runtime, so requires an application restart).

desktop/preload.js
```diff
@@ -1 +0,0 @@
-require('./context-menu');
```
belcherj, line 1: I didn't delete the file as I'll need to add it back in a future PR.

belcherj:
Screen Shot 2020-06-03 at 4 27 14 PM
Screen Shot 2020-06-03 at 4 30 01 PM

belcherj:
Screen Shot 2020-06-03 at 4 37 02 PM

belcherj:
Fixed the above issues

desktop/app.js
```diff
@@ -46,11 +47,6 @@ module.exports = function main() {
       defaultHeight: 768,
     });
 
-    // Create the browser window.
-    const iconPath = path.join(
-      __dirname,
-      '../lib/icons/app-icon/icon_256x256.png'
```
codebykat, line 52: Does this mean this icon is now unused, or is it being loaded somewhere else? I would like to remove it if possible (see #2158)

codebykat (changes requested):
When I try to logout, it hangs on "Logging out...", console shows this crash:

```
boot.ts?1649:24 Uncaught (in promise) ReferenceError: require is not defined
    at eval (boot.ts?1649:24)
    at new Promise (<anonymous>)
    at clearStorage (boot.ts?1649:13)
    at eval (boot.ts?1649:99)
    at eval (middleware.ts?2ac7:97)
```

Otherwise this tests okay and the code looks good. Small question about the icon change.

codebykat:
Still getting a hang on logout, but without any console errors now :/

Makefile
```diff
@@ -52,7 +52,7 @@ PUBLISH ?= onTag
 # Main targets
 .PHONY: start
 start:
-	@NODE_ENV=$(NODE_ENV) DEV_SERVER=$(DEV_SERVER) npx electron .
+	@NODE_ENV=$(NODE_ENV) DEV_SERVER=$(DEV_SERVER) npx electron . --inspect
```
codebykat, line 55: Did you intend to leave this debugging in for `start` and not just `dev`? Not sure if I am reading it right.

codebykat (approved):
Logging out works now! 🙆‍♀️ 

One tiny question about the `--inspect` line but merge at will.

Makefile
belcherj: I left it there on purpose.

dmsnell: why?

also why did we remove `rebuild-deps`?

Issue #2011, opened 2018-03-06
Firefox populates search field with username


kristarella:
## Issue

When password management is on in Firefox, it fills the web app search field with the username.

## Reported

Over the course of several years:
- https://simplenotep2.wordpress.com/2015/10/26/email-address-appearing-in-search-box-firefox/
- https://help.automattic.com/staff/index.php?_m=tickets&_a=viewticket&ticketid=2616730
- https://woothemes.zendesk.com/agent/tickets/989626

User has said:
> In any case, I've found that changing the search field's "type" attribute to "search" instead of "text" fixes the issue. Setting the field's "autocomplete" attribute to "off" ought to fix it as well I suppose, but I don't have an easy way to test that.

Changing the type might be a suitable fix?

codebykat:
Closed by #2104

PR #2104, opened 2020-05-19
Fixes autocomplete filling search field with username

belcherj:
### Fix

User reported in https://github.com/Automattic/simplenote-electron/issues/2011 that the search field was getting autopopulated with their username in Firefox.

I was not able to replicate but I do not see why this would have a negative impact

### Test

1. Cannot replicate bug

### DIFF
--- a/lib/search-field/index.tsx
+++ b/lib/search-field/index.tsx
@@ -79,6 +79,7 @@ export class SearchField extends Component<Props> {
       <div className="search-field">
         <input
           aria-label={screenReaderLabel}
+          autoComplete="off"
           ref={this.inputField}
           type="text"
           placeholder={placeholder}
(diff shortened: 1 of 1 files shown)


codebykat (changes requested):
Ahh nice! I had also tried to reproduce this and couldn't.

Unless there's some reason you ruled this out, I actually think the better answer here is to change it to `type=search` as that is also an a11y win:

> A role attribute of value search on the &lt;form&gt; element will cause screenreaders to announce that the form is a search form.

`autocorrect` is marked as "A Safari extension" which has not been standardized, though it looks like it has wide browser support in practice.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

belcherj:
Works for me. Change made.

codebykat (approved):
Ship it 🚢

PR #2109, opened 2020-05-19
Remove loadAndSelectNote from app-state

belcherj:
### Fix

Removes unused app-state action. It is not referenced anywhere in the code.

### Test

1. Does app run as expected?
2. Does app build as expected?


codebykat (approved):
Noice 👍

PR #2111, opened 2020-05-20
Note List: Refactor selection, stop selecting notes

dmsnell:
# Fix

This changes the selection logic to stop notes being unselected unintentionally.

Some discussion here: https://github.com/Automattic/simplenote-electron/pull/2120

### Test

1. Cycle through the app searching, trashing, un-trashing, delete forever, selecting trash, selecting tags
2. Is the selected note what you expect?
3. Is the height of items in the note list correct?
4. Delete a note from another window
5. Search by tag, select a note, untag from another window

### Release

Fixed bug where notes could be unintentionally unselected.

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -96,7 +96,7 @@ const renderNote = (
     noteDisplay: T.ListDisplayMode;
     highlightedIndex: number;
     onPinNote: DispatchProps['onPinNote'];
-    openNote: DispatchProps['openNote'];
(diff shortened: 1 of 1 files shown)


codebykat (approved):
This is actually testing out really well for me now! I did a bunch of stuff quickly with both web and Electron open and the selection was behaving correctly. Didn't see noteflip (of course I almost never see it anyway so something about "absence of evidence"). Still a lag after creating a new note and before it gets selected but that's not new or a blocker.

PR #2112, opened 2020-05-21
Update: Passwords can contain but not equal email, remove max character limit

codebykat:
### Fix

Passwords can contain the email but cannot equal it (this avoids copy-paste errors). No max length.

### Test

1. Run `make test`
2. Try to sign up with a password the same as email, should fail
3. Try to sign up with a password that includes email but is otherwise valid, should succeed
4. Try to sign up with a password over 64 characters long, should succeed

### Release

Release notes not required, since this is a minor update to a change already listed in this release.

### DIFF
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -358,7 +358,7 @@ export class Auth extends Component<Props> {
     if (!validatePassword(password, username)) {
       this.setState({
         passwordErrorMessage:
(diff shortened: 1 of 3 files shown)


belcherj (approved):
LGTM, is this fully tested?

codebykat:
I have tested it but I always prefer a second person to test!

belcherj:
Tested and works

PR #2114, opened 2020-05-22
Allow all special characters in passwords

codebykat:
### Fix

Change regex from inclusive (listing allowed characters) to exclusive (only disallowing tabs and newlines). Also adds a test for symbols.

### Test

1. `make test`
2. Try to sign up with a variety of passwords

### Release

No release notes required


belcherj (approved):
Ship it.

PR #2120, opened 2020-05-28
Fix note selection

belcherj:
This work was mostly done by @dmsnell in https://github.com/Automattic/simplenote-electron/pull/2111

### Fix

This changes the selection logic to stop notes being unselected unintentionally.

### Test

1. Cycle through the app searching, trashing, un-trashing, delete forever, selecting trash, selecting tags
2. Is the selected note what you expect?
3. Is the height of items in the note list correct?

### Release

Fixed bug where notes could be unintentionally unselected.

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -228,28 +228,15 @@ export class NoteList extends Component<Props> {
       heightCache.clearAll();
     }
 
-    // reselect when a note is removed
-    if (
-      !this.props.selectedNote &&
-      this.state.selectedIndex !== null &&
-      this.props.notes[this.state.selectedIndex]
-    ) {
-      this.props.onSelectNote(this.props.notes[this.state.selectedIndex]);
-    }
-
     // reselect when a note should be removed
     if (
-      this.props.selectedNote &&
-      this.props.selectedNote.data.deleted !== this.props.showTrash &&
+      prevProps.selectedNote &&
+      prevProps.selectedNote.data.deleted &&
+      prevProps.selectedNote.data.deleted !== this.props.showTrash &&
       this.state.selectedIndex &&
       this.props.notes[this.state.selectedIndex]
     ) {
       this.props.onSelectNote(this.props.notes[this.state.selectedIndex]);
-    } else if (
-      this.props.selectedNote &&
-      this.props.selectedNote.data.deleted !== this.props.showTrash
-    ) {
-      this.props.onSelectNote(this.props.notes[0]);
     }
   }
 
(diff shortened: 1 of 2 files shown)


belcherj:
Not sure why package-lock changed.  I'll revert out that change.

lib/note-list/index.tsx
```diff
@@ -215,86 +215,53 @@ export class NoteList extends Component<Props> {
...
-      showTrash !== this.props.showTrash
+      !this.props.selectedNote &&
+      this.state.selectedIndex !== null &&
+      this.props.notes[this.state.selectedIndex]
     ) {
-      const hasNotes = notes.length > 0;
-      this.setState({ selectedIndex: hasNotes ? 0 : null });
-      this.props.onSelectNote(hasNotes ? notes[0] : null);
-      return;
+      this.props.onSelectNote(this.props.notes[this.state.selectedIndex]);
```
dmsnell: does this still belong? I thought we got into a quick loop at times when the note would disappear from the sort selection and then reappear.

codebykat (commented):
Man, I dunno. I feel uneasy about this. The code seems reasonable enough, but I'm seeing so many bugs in trying to test it, and I can't tell if it's worse than the release branch.

Of course, I was never able to reproduce the note selection bug, so I have no idea whether this fixes that.

Spinning this up to test it on Electron, I had almost constant bugs. I witnessed a bunch of terrible behavior:
* Add new note, title characters were repeated (I typed "ello" and got "elloelloello")
* After typing a longer note title and then adding a new note, changes to the longer note were deleted
* Noticeable lag between adding new note and it being selected, causing drop of the first character(s) typed -- this for sure also exists on release branch and `develop`

Reverting to the release branch and trying to do pretty much the same series of actions, I'm seeing some of the same bugs, but maybe less frequently? On develop I haven't been able to reproduce the ghost typing / deletions yet.

codebykat:
One specific thing, when you delete a note, the list jumps to a different scroll position. I think it's scrolling to move the newly-selected note to the top of the list?

I saw some weirdly jumpy scrolling behavior too, but I can't reproduce it consistently. It's like when you're scrolling and you are about to scroll the selected note off the list, it jumps it back to keep the selected note in view.

codebykat:
Note list jumping on scroll; I am fairly confident this was introduced in this PR, as I haven't seen it on `release/1.17` or `develop`.

Screen-Recording-2020-05-28-at-1

codebykat:
If I comment out the line suggested by @dmsnell, deleting a note leads to a new note being highlighted in the note list but not loaded into the editor.

(Though it may improve some of the other issues I saw; still testing.)

dmsnell:
maybe for now we just stop selecting notes altogether unless someone navigates to the prev/next note?

it would be a small annoyance I guess but would stop frustratingly switching notes. based on the work in https://github.com/Automattic/simplenote-electron/commit/164b54a076d16c47daee72af2c17679d1ab56bbe I think we could undo this, remove the highlighted index altogether, and get the "real fix" into `develop` before the big refactor is done

I used middleware to annotate the action with the next note to select if we need to, and then used the presence of that new note meta to switch the opened note. the _only_ thing in `note-list` that matters is `state.ui.note.id === thisNote.id`

just a thought

belcherj:
Closing in favor of https://github.com/Automattic/simplenote-electron/pull/2111

PR #2122, opened 2020-05-28
Move clipboard types to dev dependencies and remove hat

belcherj:
### Fix

This PR:
- Moves @types/clipboard to dev dependencies.
- Pins its version by removing the hat `^`.


codebykat (approved):
👍

PR #2126, opened 2020-06-02
Merge 1.17 into master

mokagio:
I opened this because I saw `master` and `develop` are both still in the repo. I thought we were meaning to drop the `master`/`develop` parallelism, so this might be unnecessary. If so, feel free to close it.


belcherj:
We just haven't removed the branch

PR #2140, opened 2020-06-05
Update coloured elements to match new design guidelines

codebykat:
### Fix

After noticing that icons and text were slightly different colors under the old guidelines, @SylvesterWilmott simplified to using the same color for both colored elements and text. They should all be `simplenote-blue-50` in light mode and `simplenote-blue-30` in dark mode.

### Screenshots

Screen Shot 2020-06-05 at 3 18 19 PM

Screen Shot 2020-06-05 at 3 18 06 PM

### DIFF
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -88,7 +88,7 @@
 
   .login__signup a,
   .terms a {
-    color: $studio-simplenote-blue-60;
+    color: $studio-simplenote-blue-50;
     margin-left: 5px;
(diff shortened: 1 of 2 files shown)


SylvesterWilmott (approved):

SylvesterWilmott:
>Should we change the background color of the cookie banner? We now have three different shades of blue on the login page

Absolutely let's do that.

I would also remove the white fill from the logo svg.

PR #2146, opened 2020-06-09
Use allow list not whitelist

belcherj:
### Fix

Changes a variable name to stop using whitelist.

### Test

No testing needed


codebykat (approved):
Thank you! ❤️

PR #2147, opened 2020-06-10
User browser for authing a user via WP Login

belcherj:
See p2XJRt-1Y0-p2

### Fix

In order to pass the Windows store certification, we cannot do auth in the app. Instead, it needs to be pushed to a browser.

There is still uncertainty that this will resolve the issue with certification however this is the only solution that I can see as a path forward.

### Test

1. Run app
2. In Electron click the WordPress.com login button
3. Log in
4. Were you able to log in?

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -22,18 +22,22 @@ const spellcheck = require('./spellchecker');
 require('module').globalPaths.push(path.resolve(path.join(__dirname)));
 
 module.exports = function main() {
+  // Keep a global reference of the window object, if you don't, the window will
+  // be closed automatically when the JavaScript object is GCed.
+  let mainWindow = null;
+
   app.on('will-finish-launching', function () {
     setTimeout(updater.ping.bind(updater), config.updater.delay);
+    app.on('open-url', function (event, url) {
+      event.preventDefault();
+      mainWindow.webContents.send('wpLogin', url);
+    });
   });
 
   const url =
     isDev && process.env.DEV_SERVER
       ? 'http://localhost:4000' // TODO: find a solution to use host and port based on make config.
       : 'file://' + path.join(__dirname, '..', 'dist', 'index.html');
-
-  // Keep a global reference of the window object, if you don't, the window will
-  // be closed automatically when the JavaScript object is GCed.
-  let mainWindow = null;
 
   const activateWindow = function () {
     // Only allow a single window
@@ -118,6 +122,10 @@ module.exports = function main() {
       mainWindow.setMenuBarVisibility(!autoHideMenuBar);
     });
 
+    ipcMain.on('wpLogin', function (event, url) {
+      shell.openExternal(url);
+    });
+
     mainWindowState.manage(mainWindow);
 
     mainWindow.webContents.on('new-window', function (event, linkUrl) {
@@ -165,6 +173,11 @@ module.exports = function main() {
 
   if (!gotTheLock) {
     return app.quit();
+  }
+
+  if (!app.isDefaultProtocolClient('simplenote')) {
+    // Define custom protocol handler. This allows for deeplinking into the app from simplenote://
+    app.setAsDefaultProtocolClient('simplenote');
   }
 
   // Quit when all windows are closed.
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -6,7 +6,6 @@ import getConfig from '../../get-config';
 import SimplenoteLogo from '../icons/simplenote';
 import Spinner from '../components/spinner';
 import { validatePassword } from '../utils/validate-password';
-
 import { isElectron, isMac } from '../utils/platform';
 import { viewExternalUrl } from '../utils/url-utils';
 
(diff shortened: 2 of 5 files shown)


desktop/app.js
```diff
@@ -168,6 +172,11 @@ module.exports = function main() {
     return app.quit();
   }
 
+  if (!app.isDefaultProtocolClient('simplenote')) {
+    // Define custom protocol handler. Deep linking works on packaged versions of the application!
+    app.setAsDefaultProtocolClient('simplenote');
+  }
```
dmsnell, line 181: is this part of the same fix or is this a secondary fix?

electron-builder.json
```diff
@@ -6,6 +6,10 @@
     "buildResources": "resources"
   },
   "files": ["desktop", "dist", "shared"],
+  "protocols": {
+    "name": "Simplenote",
+    "schemes": ["simplnote"]
```
dmsnell: typo: `simplnote` → `simplenote`

lib/auth/index.tsx
```diff
@@ -338,80 +338,49 @@ export class Auth extends Component<Props> {
...
-
     const redirectUrl = encodeURIComponent(config.wpcc_redirect_url);
     this.authState = `app-${cryptoRandomString(20)}`;
     const authUrl = `https://public-api.wordpress.com/oauth2/authorize?client_id=${config.wpcc_client_id}&redirect_uri=${redirectUrl}&response_type=code&scope=global&state=${this.authState}`;
+    const shell = __non_webpack_require__('electron').shell;
+    shell.openExternal(authUrl);
+
+    const ipc = getIpcRenderer();
+    ipc.on('wpLogin', (event, url) => {
+      debugger;
```
dmsnell: noting to remove later

desktop/app.js
belcherj: This is part of the fix

dmsnell:
Can you explain more what the problem is and how this resolves it?

Is the problem that we create a second window to handle auth whereas replacing the app itself with loading the WP login solves the issue?

electron-builder.json
belcherj: Ohhhh jeez

belcherj:
The issue is that we cannot use a window inside the app to do WordPress.com login because it violates the Microsoft store policy for certification.  We need to do it in the browser.

desktop/app.js
dmsnell: can you maybe explain this more and explain why we have to set this here? (in a comment)

belcherj: We have to set a custom simplenote:// protocol so when the browser gets the callback URL from WordPress.com it deep links into the app.

lib/auth/index.tsx
```diff
@@ -338,80 +338,49 @@ export class Auth extends Component<Props> {
 
   onWPLogin = () => {
     const config = getConfig();
-    this.setupAuthWindow();
-
     const redirectUrl = encodeURIComponent(config.wpcc_redirect_url);
     this.authState = `app-${cryptoRandomString(20)}`;
     const authUrl = `https://public-api.wordpress.com/oauth2/authorize?client_id=${config.wpcc_client_id}&redirect_uri=${redirectUrl}&response_type=code&scope=global&state=${this.authState}`;
+    const shell = __non_webpack_require__('electron').shell;
+    shell.openExternal(authUrl);
```
dmsnell: won't this cause issues for in-browser views or does the Electron code work in the browser context too?

desktop/app.js
belcherj: Comment added

dmsnell:
> we cannot use a window inside the app to do WordPress.com login

thanks. did we confirm that this change resolves their issue? I know App Store rules can seem bureaucratic but it definitely feels like we're doing the same thing here - did they provide a link to their rules documentation?

The interface inside of `lib/auth/index` also feels a bit ad-hoc or convenient. I know we're wanting to hurry to get it in the store but I wonder if we shouldn't isolate that more in a separate module or at least provide clear documentation in the comments what's happening and where/how the signals are coming from the external window.

Also, what is supposed to happen if we don't complete the auth process in the external window? When I read the code it looks like we should always be sending `wpLogin` based on `desktop/app.js` and in `auth/index` it looks like we should only receive the `wpLogin` message after the WPCC process is done?

Mainly I'm saying that I'm confused reading the code and the login flow is hard for me to figure out.

lib/auth/index.tsx
belcherj: This code never gets rendered in the browser, except for in dev. Will add back code from L181

belcherj:
This is the 'get it working' stage of this PR. I need to do some cleanup work for sure.

We are doing the same thing but not in app, we are doing it in the browser. If this does not fix the issue for certification then we will likely have to remove WordPress.com login from the app that gets submitted to the MS app store.

dmsnell:
> If this does not fix the issue for certification then we will likely have to remove WordPress.com login from the app that gets submitted to the MS app store.

Do they provide a way for us to know this before we change the code? Do they link to the guidelines that they argue we violate?

belcherj:
I asked via email.

https://docs.microsoft.com/en-us/windows/uwp/publish/store-policies

These are the two non-compliance issues: 

10.1.5 Software Distribution
10.8.2 In-App Purchase: Physical Goods & Services

There is no way to tell if this will comply.

dmsnell:
> 10.1.5 Software Distribution
> 10.8.2 In-App Purchase: Physical Goods & Services

It sounds like they don't understand the role of the WordPress.com signin since those rules apply to purchases or other software. How does opening in an external window address these?

> There is no way to tell if this will comply.

Do they not respond to the email or just say "submit again and see if it works." Seems like we wouldn't be the only ones having this issue and I wonder if there's something we could do to get their attention to give an adequate review.

It doesn't seem like Simplenote login and WP login are any different based on their qualifications. I guess they should be blocking us if we were trying to allow people to login with their Live accounts too?

belcherj:
> It sounds like they don't understand the role of the WordPress.com signin since those rules apply to purchases or other software. How does opening in an external window address these?

Once you pop open the auth window you can navigate on Apple and Google website which allow for purchases within the app.

By sending them to a browser window they would not be able to make purchases within the app. 

Other apps I am seeing in the Windows store you a browser window to do 3rd party auth.

codebykat (changes requested):
I don't know what's up but I'm getting a white screen on Electron with `npm run dev`. I deleted my `node_modules` and did a full `npm install` several times. No errors in the console.

codebykat (approved):
Okay the build thing was unrelated (due to having been on the rewrite branch and not clearing site data).

This is not ideal but it's fine for now so we can get the release process unblocked.

Clicking "Log in with WordPress.com" runs validation on the form, causing this field to pop up:

Screen Shot 2020-07-07 at 11 54 58 AM

Got this when returning from the browser, in the standalone build does it say Simplenote rather than Electron?

Screen Shot 2020-07-07 at 11 52 56 AM

belcherj:
Closing in favour of https://github.com/Automattic/simplenote-electron/pull/2191

PR #2167, opened 2020-06-25
Preview: Block internal links

codebykat:
### Fix

Right now if you have a note with intra-note links (such as anchor tags) it will open a broken `localhost` link. This disallows those links from opening.

### Test

1. Create a Markdown note containing some relative links, e.g.`* [first section](#first)`
2. Preview
3. Try to click on the links. They should not do anything

### Release

Not updated: Prevent internal links from opening in a new browser window


dmsnell (approved):
Looks like a good chance. I wish we could prevent creating relative links entirely but this seems like it should do.

PR #2168, opened 2020-06-25
Fix: App bootup fails in browser due to missing isElectron checks

codebykat:
### Fix

Introduced in #2102. Right now loading in browser yields the following error: `TypeError: window.electron is undefined`

fyi @belcherj

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -127,12 +127,14 @@ export const App = connect(
     }
 
     componentDidMount() {
-      window.electron.receive('appCommand', this.onAppCommand);
-      window.electron.send(
-        'setAutoHideMenuBar',
-        this.props.settings.autoHideMenuBar
-      );
-      window.electron.send('settingsUpdate', this.props.settings);
+      if (isElectron) {
+        window.electron.receive('appCommand', this.onAppCommand);
+        window.electron.send(
+          'setAutoHideMenuBar',
+          this.props.settings.autoHideMenuBar
+        );
+        window.electron.send('settingsUpdate', this.props.settings);
+      }
 
       this.props.noteBucket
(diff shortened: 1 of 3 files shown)


lib/app.tsx
```diff
@@ -127,12 +127,14 @@ export const App = connect(
...
-        this.props.settings.autoHideMenuBar
-      );
-      window.electron.send('settingsUpdate', this.props.settings);
+      if (isElectron) {
+        window.electron.receive('appCommand', this.onAppCommand);
+        window.electron.send(
+          'setAutoHideMenuBar',
+          this.props.settings.autoHideMenuBar
+        );
+        window.electron.send('settingsUpdate', this.props.settings);
```
dmsnell: also acceptable would be to use optional chaining, which I slightly prefer at the moment due to its brevity… (doesn't require `isElectron`)

```ts
window.electron?.receive('appCommand', this.onAppCommand);
window.electron?.send( 'setAutoHideMenuBar', this.props.settings.autoHideMenuBar );
window.electron?.send('settingsUpdate', this.props.settings);
```

the tradeoff could be argued as losing the context that this is intended to run only in Electron but I feel like that's obvious given the project and also given the naming

lib/app.tsx
```diff
@@ -178,7 +180,7 @@ export const App = connect(
     componentDidUpdate(prevProps) {
       const { settings } = this.props;
 
-      if (settings !== prevProps.settings) {
+      if (settings !== prevProps.settings && isElectron) {
```
dmsnell: same here… with optional chaining we don't need the additional check

dmsnell (approved):
Do as you see fit. I like the optional chaining but don't feel strongly about it.

PR #2170, opened 2020-06-26
Fix: Render checkboxes on initial note load

### LINKED ISSUE #2141: Checklist first displays as bracket - requires switching note and returning
### Expected
checklist to be displayed with bullet points instead of `-[]` in display mode

### Observed
checklist is displayed with `- []` at startup, after moving to another note and coming back to the previous note the checklist is properly displayed.
This is true for the Arch Linux 1.17.0 Simplenote and accessing Simplenote on Firefox as well.

Arch Linux
1.17.0

codebykat:
### Fix

This fixes https://github.com/Automattic/simplenote-electron/issues/2141, an issue causing checkboxes to be rendered as plain-text when the note is first loaded.

However, I am pretty sure we deferred decorating the note for performance reasons so this may leave us in a bind. I'm not sure if `componentDidMount` is the best place to put this call.

### Test

1. Create a note with checkboxes (doesn't matter if it's a Markdown or plain-text note)
1. Close and re-open app
1. Open the note and verify that the checkboxes are rendered correctly

### Release

Not updated: Fixed an issue causing checkboxes to be rendered incorrectly when a note was first opened

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -123,6 +123,9 @@ class NoteContentEditor extends Component<Props> {
     }
 
     window.addEventListener('keydown', this.handleKeydown, false);
+
+    this.queueDecoratorUpdate();
+    this.queueDecoratorUpdate.flush();
   }
 
   handleEditorStateChange = (editorState) => {


dmsnell (approved):
Looks good. I just dealt with this in my branch too, though with entirely different code and editor components. 🎈

PR #2178, opened 2020-06-29
Add notices to logged out screen

belcherj:
### Fix
Not entirely sure that this is needed:

Screen Shot 2020-06-29 at 5 11 16 PM

### Test

Open app to login screen, do you see footer.


codebykat:
As per Slack, not needed

PR #2179, opened 2020-06-29
Add privacy policy

belcherj:
### Fix

Add privacy policy

### Test

Do you see it in the readme?

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,5 +54,5 @@ For other platforms, see:
 
 ## Privacy Policy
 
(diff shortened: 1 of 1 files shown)


codebykat (changes requested):
I think this should be https://automattic.com/privacy/#california-consumer-privacy-act-ccpa and not wp.me. I would also change the wording to match what we're using everywhere else, "Privacy notice for California users"

belcherj:
Closing as not needed

PR #2180, opened 2020-06-30
Fix Export

belcherj:
### Fix

Fix export on rewrite/no-races

### Test

1. Export from web local build
2. Export from electron local build

### DIFF
--- a/lib/utils/export/index.ts
+++ b/lib/utils/export/index.ts
@@ -4,10 +4,12 @@ import { get } from 'lodash';
 import exportNotes from './export-notes';
 import exportToZip from './to-zip';
 
+import * as T from '../../types';
+
 const filename = 'notes.zip';
 
-const exportZipArchive = () => {
-  return exportNotes()
+const exportZipArchive = (notes: T.Note[]) => {
+  return exportNotes(notes)
     .then(exportToZip)
     .then((zip) =>
       zip.generateAsync({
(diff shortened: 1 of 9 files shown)


lib/utils/export/types.ts
```diff
@@ -0,0 +1,18 @@
...
+  id: T.EntityId;
+  markdown?: boolean;
+  modificationDate: T.SecondsEpoch;
+  pinned?: boolean;
+  publicURL?: string;
+  tags: T.TagName[];
+};
+
+export type GroupedExportNotes = {
+  activeNotes: ExportNote[] | [];
```
dmsnell: no need for the `| []` here as an empty array is still `ExportNote[]`

lib/utils/export/index.ts
```diff
@@ -4,10 +4,12 @@ import { get } from 'lodash';
 import exportNotes from './export-notes';
 import exportToZip from './to-zip';
 
+import * as T from '../../types';
+
 const filename = 'notes.zip';
 
-const exportZipArchive = () => {
-  return exportNotes()
+const exportZipArchive = (notes: T.Note[]) => {
```
dmsnell, line 11: adding `notes` here is great because I plan on adding this as an option when logging out if there are unconfirmed changes in the account:

```
There are changes which may not be synchronized with the server.

 - Lose changes and logout?
 - Export affected notes and logout?
 - Don't logout yet?
```

lib/utils/export/types.ts
belcherj: I was wondering about that. Thx!

dmsnell (approved):

PR #2181, opened 2020-07-02
[Rewrite Branch] Fixes undo redo selectall

belcherj:
### Fix

This fixes the undo, redo, and selectall in the Electron menu

### Test

1. Run PR
2. Do undo, redo, and selectall function from the menu


belcherj:
Closing as this only applies to Monaco

belcherj:
@dmsnell What is the best way to add the types for window.editor?

dmsnell (changes requested):
Despite things I've said in the past which probably sound like the opposite I think it's appropriate for us to listen to these inside the editor component itself to keep them from sprawling around the app. I don't want to have to tightly couple the specific editor implementation with the Electron middleware

While apparently it isn't there now in many of my editor replacements I've attached a `window.electron?.receive()` inside of `note-content-editor` to interact with the editor. It's particularly helpful there because we can start listening inside of `editorReady` where we know the editor is available.

While `insertChecklist` is left in there you may notice it's a dud: no reducer listens anymore for `INSERT_TASK` and that action could potentially disappear - I left it in not knowing whether we'd stick with it.

The editor is a gray area of high interactivity. I'd like to leave as much of it out of the rest of the app as possible but this has been a central challenge with the editor component.

belcherj:
Moved as requested.

dmsnell (approved):
Thanks for rearranging this

belcherj:
Actually, I think there is an issue, I need to remove the listeners when switching notes.

dmsnell:
> I need to remove the listeners when switching notes

Good catch. The use of `key={ openedNote }` means that the editor gets recreated from scratch when switching notes

belcherj:
This is broken till we rebase in develop. Need the code that removes events from the eletron bridge.

belcherj:
This is ready for review

desktop/context-menu/index.js
```diff
@@ -0,0 +1,45 @@
...
+        label: 'Paste',
+        role: 'paste',
+        enabled: editFlags.canPaste,
+      },
+      {
+        type: 'separator',
+      },
+    ];
+    const menu = Menu.buildFromTemplate(template);
+    menu.popup({});
```
dmsnell: I assume these are split up as part of the development process? if we find it useful we could inline the function. how could it be useful? if we found more value in the brevity than in the verbosity

```js
mainWindow.webContents.on('context-menu', (event, { editFlags }) =>
	Menu.buildFromTemplate([
		{
			…
		},
		…
	]).popup({});
```

I'm assuming that argument-destructuring works since the object destructuring is working.
if you don't think this is a positive change just ignore it.

desktop/menus/utils.js
```diff
@@ -14,10 +14,10 @@ const buildRadioGroup = ({ action, propName, settings }) => {
   };
 };
 
-const appCommandSender = (arg) => {
+const appCommandSender = (commandName = 'appCommand', arg) => {
```
dmsnell: can we avoid adding optional arguments before required arguments? this can be pretty confusing especially since JS doesn't require a fixed arity when calling functions

dmsnell (changes requested):
looks like we have lots of `appCommnad` that should be `appCommand`
also I think that making the first argument to `appCommandSender` is more of an immediate convenience and a long-term hassle. I'd feel more comfortable if we either spent the time to update all the calls to `appCommandSender()` or created a new `editCommandSender()` for the edit operations. 

if we create `editCommandSender()` (which would mirror the way the we are using the commands and names here) then the size of this PR would dramatically shrink I think. one change to `appCommandSender()`, the new `editCommandSender()`, the new editing actions, and the listener

desktop/menus/utils.js
belcherj: I think we can make it mandatory.

dmsnell (approved):
Thanks again for changing things around. How do you like using `editCommandSender`?

belcherj:
I like it however I think we will have to do a refactor that makes it more general use. I'd like to have named commands for every action. That way we can verify the data being passed. Also makes typing better since we will be able to determine types  of the payload based on the command.

PR #2148, opened 2020-06-10
Rewrite data flow in app to remove races and sync bugs

dmsnell:
## Description

We've been unable to resolve the worst sync bugs in the application and feature development has been artificially limited by the fact that we have to fight a number of data-flow issues tightly coupling different levels of the application's architecture together.

In this branch, we're ripping apart the entire app state and Simperium data flow to rebuild it in order to remove a number of those couplings and races.

## Blockers

 - [ ] ~Enter key not working in Firefox, programmatic edits to note content not working - execCommand~
 - [x] Cannot quit app if unsync'd notes exist (for example, emptied trash, notes in bucket, missing from ghost, looping to sync notes) [`beforeunload` works differently in Electron]
 - [ ] Deleted-forever notes may not actually delete (refresh the app and they might come back). This may be related to the items below on adding a *delete-queue*
 - [x] Add cut/copy/paste to context menu when editing #2280 
 - [x] Make sure notifications do not spam users on login
 - [x] Make sure notifications does not spam users when requesting permission
 - [x] Make sure notifications off by default
 - [ ] Add Electron confirm close dialog
 - [x] Test all keyboard shortcuts [#](https://github.com/Automattic/simplenote-electron/pull/2148#issuecomment-675743700)
  - [x] Fix broken shortcuts (Done except for "next match?") #2284 
 - [x] Debounce `editor_note_edited` analytics to follow Simperium's edits, not the editor's (@dmsnell, @codebykat
 - [x] Printing in edit mode results in blank page, preview works @belcherj #2271
 - [x] Search is missing #2313 

## Not-really blockers

 - [x] Audit tag operations to make sure affected notes are synchronized (@dmsnell)
 - [x] Only select the next note during `DELETE_NOTE_FOREVER` if the opened note is the one that was deleted
 - [x] Create checkbox, mark as completed, enter down, checkbox created is marked as complete.
     - When continuing a task list the new checkboxes will now all be unchecked if automatically created
 - [x] Disable "matching parenthesis and brackets" from the editor. they are visually wrong and not "Simple" #2283 

### Safari
 - [x] Buttons in revision history and trash view do not have a proper line-height
 - [x] Line height for search highlights is incorrect.
 - [x] Text on staging is #d4d4d4 not #fff in dark mode (should be fixed by #2272)
 - [x] Tag names in edit mode have bottom chopped off, edit tag a `qqq`, observe
 - [x] In search field descenders are cut off, type `qqq`
 - [x] When coming back online sever connection does not go fully green

[...]

PR #2182, opened 2020-07-03
Fix/search case sensitivity

codebykat:
### Fix

1. Fix: a search string containing uppercase letters would return no matches (e.g. "Markdown").
2. Fix: Some search results were hidden; react-virtualized was pushing them far to the right of the screen because of a negative column offset in a previous header.

See https://github.com/Automattic/simplenote-electron/pull/2148

### Test

Search for notes and ensure that all notes get matched and appear in the notes list properly.

### Follow-up tasks

There is still something funky with the note-list cache/renderer, I can't reproduce it consistently but occasionally I was able to get glitches like this:

Screen Shot 2020-07-03 at 12 31 55 AM

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -91,7 +91,7 @@ const renderNote = (
     ) : (
       <CellMeasurer
         cache={heightCache}
-        columnIndex={-3}
+        columnIndex={0}
         key="notes-header"
         parent={parent}
         rowIndex={index}
--- a/lib/search/index.ts
+++ b/lib/search/index.ts
@@ -203,7 +203,7 @@ export const middleware: S.Middleware = (store) => {
 
       if (
         searchTerms.length > 0 &&
-        !searchTerms.every((term) => note.content.includes(term))
+        !searchTerms.every((term) => note.content.match(RegExp(term, 'gi')))
       ) {
         continue;
       }
--- a/lib/search/index.ts
+++ b/lib/search/index.ts
@@ -203,7 +203,9 @@ export const middleware: S.Middleware = (store) => {
 
       if (
         searchTerms.length > 0 &&
-        !searchTerms.every((term) => note.content.match(RegExp(term, 'gi')))
+        !searchTerms.every((term) =>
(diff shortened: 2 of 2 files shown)


lib/search/index.ts
```diff
@@ -203,7 +203,7 @@ export const middleware: S.Middleware = (store) => {
 
       if (
         searchTerms.length > 0 &&
-        !searchTerms.every((term) => note.content.includes(term))
+        !searchTerms.every((term) => note.content.match(RegExp(term, 'gi')))
```
dmsnell: instead of creating a new RegExp on every search term, which has proven to be quite slow in the past, let's transform the term itself. the content in the search index is [already transformed](https://github.com/Automattic/simplenote-electron/blob/1baaf0aa01609b4e55a05df956f7b206b93d9af3/lib/search/index.ts#L33).

```ts
!searchTerms.every((term) => note.content.includes(term.toLocaleLowerCase()))
```

lib/note-list/index.tsx
```diff
@@ -91,7 +91,7 @@ const renderNote = (
     ) : (
       <CellMeasurer
         cache={heightCache}
-        columnIndex={-3}
+        columnIndex={0}
```
dmsnell, line 94: my only suggestion here is that we make sure this doesn't interfere with the tag suggestions which also uses `columnIndex={0}`. when reading on `CellMeasurer` I had a hard time truly understanding the role in play here; there's some association with the caches we clear in `deriveStateFromProps`

dmsnell (commented):
thanks @codebykat! I had been wondering why search seemed to be hiding so many results

lib/note-list/index.tsx
codebykat: As per Slack discussion, the -3 index was an attempt to set it out of bounds so it wouldn't conflict with the notes in the list, but doesn't seem to work that way and causes other issues. Since this tests out okay and the height cache glitch isn't new, gonna merge this and we can keep arguing with CellMeasurer another day.

PR #2183, opened 2020-07-03
Fix: restore tag icons and scrolling

codebykat:
### Fix

Trash and reorder icons were invisible on Firefox and improperly-sized on Chrome. Restored commented-out styles to fix this.

n.b. If a width isn't set, the flex-box will grow out-of-bounds greedily, and the svgs will be resized to 0 width. This is fixable by telling flex not to shrink them (`flex-shrink: 0`). Setting an explicit width on the container removes the need for this by telling flex not to expand it beyond the sidebar.

I can't get overflow scrolling to work on just the `<ul>` for the tag list, I put it on the parent div for now. Someone who understands CSS should look at this. This at least makes it usable again.

See #2148 

### Test

1. Open tag sidebar
2. Verify you can scroll through tags
3. Click edit
4. Verify that icons are displayed to trash and reorder tags, and that they work

### DIFF
--- a/lib/tag-list/style.scss
+++ b/lib/tag-list/style.scss
@@ -10,6 +10,7 @@
   flex-direction: column;
   flex: 1 1 auto;
   overflow: hidden;
+  overflow-y: scroll;
 
   .tag-list-title {
     display: flex;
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
> n.b. If a width isn't set, the flex-box will grow out-of-bounds greedily, and the svgs will be resized to 0 width. 

interesting. any idea why the SVGs shrink to zero-width?

looks good - I knew something had to be off. the icons were also hidden in Safari

codebykat:
> interesting. any idea why the SVGs shrink to zero-width?

I only sort-of comprehend flex, but as I understand it, the parent container was too wide, so the browser shrunk any child elements it could in an attempt to narrow the width of the parent. Since the svgs didn't have a min-width, the browser was happy to reduce them to a width of 0.

PR #2184, opened 2020-07-03
Fix: Copy all html if nothing selected

codebykat:
### Fix

Copy all HTML from preview if nothing is selected.

See https://github.com/Automattic/simplenote-electron/blob/776002ead9548132de01142cfd888d25b1fa146d/lib/note-detail/index.tsx#L111

#2148

In porting this over I realized that the existing functionality doesn't work in Firefox and needed to use the Clipboard API instead. See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard

### Test

1. Preview a Markdown note
2. Hit Ctrl+C to copy
3. Paste and verify it is in your clipboard
4. Test that you can still copy and paste a smaller selection
5. Test that it does not copy all content if not in preview mode
6. It should not copy all content if the note info or any dialogs are showing on top of the preview

### DIFF
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -50,21 +50,13 @@ export const NotePreview: FunctionComponent<Props> = ({
 
       const div = document.createElement('div');
       renderToNode(div, note.content, searchQuery).then(() => {
-        // try {
-        /* this works in Chrome but not Firefox */
-        // event.clipboardData.setData('text/plain', div.innerHTML);
-        // } catch (DOMException) {
-        // console.log("that didn't work, trying something else");
-        /* try it the Firefox way - this works in Firefox and Chrome */
-        navigator.clipboard.writeText(div.innerHTML).then(
-          () => {
-            /* success */
-          },
-          () => {
-            console.log("it didn't work :(");
-          }
-        );
-        // }
+        try {
+          // this works in Chrome and Safari but not Firefox
+          event.clipboardData.setData('text/plain', div.innerHTML);
+        } catch (DOMException) {
+          // try it the Firefox way - this works in Firefox and Chrome
+          navigator.clipboard.writeText(div.innerHTML);
+        }
       });
 
       event.preventDefault();
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -48,22 +48,31 @@ export const NotePreview: FunctionComponent<Props> = ({
         return true;
       }
 
-      // const node = document.createDocumentFragment();
       const div = document.createElement('div');
-      renderToNode(div, note!.content, searchQuery);
-      // node.appendChild(div);
+      renderToNode(div, note.content, searchQuery).then(() => {
+        // try {
+        /* this works in Chrome but not Firefox */
+        // event.clipboardData.setData('text/plain', div.innerHTML);
+        // } catch (DOMException) {
+        // console.log("that didn't work, trying something else");
+        /* try it the Firefox way - this works in Firefox and Chrome */
+        navigator.clipboard.writeText(div.innerHTML).then(
+          () => {
+            /* success */
+          },
+          () => {
+            console.log("it didn't work :(");
+          }
+        );
+        // }
+      });
 
-      // TODO this is a bug
-      console.log(div); // innerHTML shows up here in console
-      console.log(div.innerHTML); // innerHTML is now an empty string???????
-
-      event.clipboardData.setData('text/plain', div.innerHTML);
       event.preventDefault();
     };
 
     document.addEventListener('copy', copyRenderedNote, false);
     return () => document.removeEventListener('copy', copyRenderedNote, false);
-  }, [isFocused]);
+  }, [isFocused, searchQuery]);
 
   useEffect(() => {
     const handleClick = (event: MouseEvent) => {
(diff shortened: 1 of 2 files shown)


lib/components/note-preview/index.tsx
```diff
@@ -29,12 +31,46 @@ type Props = OwnProps & StateProps & DispatchProps;
...
   noteId,
   searchQuery,
 }) => {
   const previewNode = useRef<HTMLDivElement>();
 
+  useEffect(() => {
+    const copyRenderedNote = (event: ClipboardEvent) => {
+      console.log(isNoteInfoOpen);
+      // TODO this isn't working - these are false when I log them, but in console they are true
+      // e.g. state.ui.showNoteInfo in console will print "true"
```
dmsnell: the reason for this oddity is that we're enclosing the copies of the variables inside this closure when we create it. that is, when grabbing them from the outer props they have been bound to `isNoteInfoOpen` and `isDialogOpen` with the values they had when `useEffect()` ran.

with `this` referring to a component's backing instance we can still access the props at the time of invocation because we follow the reference through the objects to the values of `props` (vs reading a copied value).

one way around this is to add those values as dependencies to the `useEffect()` and have it recreate the closure when they change.

```ts
useEffect(() => {
	…
}, [isDialogOpen, isNoteInfoOpen])
```

That should recreate the effect whenever those props change and update the values bound inside of it.

lib/components/note-preview/index.tsx
```diff
@@ -29,12 +31,46 @@ type Props = OwnProps & StateProps & DispatchProps;
...
+      }
+
+      // Only copy the rendered content if nothing is selected
+      if (!document.getSelection().isCollapsed) {
+        return true;
+      }
+
+      const node = document.createDocumentFragment();
+      const div = document.createElement('div');
+      renderToNode(div, note!.content, searchQuery);
```
dmsnell: nitpick and personal style: I'm more in favor of handling these possible-null errors in one of two other ways than to use the `!`. why? `!` tells TypeScript to stop checking and leaves it a fully-manual operation meaning that (a) we can get it wrong and (b) if things change we may not realize it.

If all we need is a safe fallback I have so-far preferred using optional-chaining and the null-coalescing operator. It's explicit what we're falling back to…

```ts
renderToNode(div, note?.content ?? '', searchQuery );
```

…or if we "know" that it's there even though the types and TypeScript show that there's a possibility that it might not be, then I actually prefer leaving the error in the code since it can make it easy for us to find later. there's a decent amount of future reworking we could do in this app to eliminate these case. basically anywhere we have components rendering that depend on the opened or selected note we're going to have these errors.

as far as I can tell right now this may never happen because it would be a hefty change to do that for relatively little gain. we could pass a copy of the `note` down from a parent component instead of sourcing it from state: this would eliminate the possibility of a null/undefined value but introduce duplication of data and reintroduce prop-drilling.

lib/components/note-preview/index.tsx
```diff
@@ -106,6 +142,8 @@ export const NotePreview: FunctionComponent<Props> = ({
 
 const mapStateToProps: S.MapState<StateProps, OwnProps> = (state, props) => ({
   fontSize: state.settings.fontSize,
+  isDialogOpen: state.ui.dialogs.length > 0,
+  isNoteInfoOpen: state.ui.showNoteInfo,
```
dmsnell: might also consider reframing this from the perspective of the inner component and the API we want to expose from it…

```ts
copyFullRenderedContent: state.ui.dialogs.length = 0 && !state.ui.showNoteInfo
```

…this way it's clearer why we're using them and easier to push into a known testable state. this is not a blocking issue.

lib/components/note-preview/index.tsx
codebykat: I copied this from https://github.com/Automattic/simplenote-electron/blob/rewrite/no-races/lib/components/note-preview/index.tsx#L87 fwiw. I'm happy to change it if you think we ought.

dmsnell: okay lol sorry. that's me being lazy and having a different "writing code" vs. "reviewing code" mindset

if I were writing it again I'd probably leave out the `!` for the reasons mentioned above. it's not a big deal but as I go I get more persuaded of the value of not overriding TypeScript's inference.

can do it later if you want outside of this PR

lib/components/note-preview/index.tsx
```diff
@@ -29,12 +30,41 @@ type Props = OwnProps & StateProps & DispatchProps;
...
+      }
+
+      // const node = document.createDocumentFragment();
+      const div = document.createElement('div');
+      renderToNode(div, note!.content, searchQuery);
+      // node.appendChild(div);
+
+      // TODO this is a bug
+      console.log(div); // innerHTML shows up here in console
+      console.log(div.innerHTML); // innerHTML is now an empty string???????
```
dmsnell: looks like `renderToNode()` runs inside a Promise, meaning that `div.innerHTML` shouldn't be populated at least until this function returns and the Promise `.then()` executes, setting the node's `innerHTML`. this is why it works in the console - because the Promise `.then()` has already run (it probably takes microseconds after this outer function finishes).

we can hook into the `Promise` chain by modifying `renderToNode()` to return the `Promise` instead of just running it (and returning `undefined`)

```diff
  export const renderToNode = (
      …
- ) => {
-   renderNoteToHtml(content)
+ ) => renderNoteToHtml(content)
  …
-   });
+ });
- }
```

lib/components/note-preview/index.tsx
codebykat: oh I removed it and TS didn't even warn me it could be null so I guess we're good

dmsnell (approved):

PR #2189, opened 2020-07-06
show pointer cursor on checkboxes in the editor view

### LINKED ISSUE #2185: Mouse cursor & checklist items
While hovering checked / un-checked checklist item, mouse cursor appears as text cursor.

I expected pointer cursor, like the one which appears when hovering or selecting notes.

It also applies for Preview mode.

### Expected
Screenshot from 2020-07-03 17-43-41(1)

### Observed
Screenshot from 2020-07-03 17-43-41

|OS Version|App Version|
|-|-|
|Ubuntu 18.04 LTS|Simplenote 1.19.0|

arpitbatra123:
### Fix

This PR adds a fix to show the `pointer` type cursor for checkboxes rendered in the editor view.

### Test

> 1. Create a new note with content that contains markdown style checkboxes. You can use the below as an example
```md
### Shopping List
- [ ] Teas
- [x] Brownie
- [x] Cookies
- [x] Notepad
```
> 2. The editor view should show a pointer icon over checkboxes. See below for an example of how this should look
 
simplenote-cursor-fix

### DIFF
--- a/lib/note-detail/style.scss
+++ b/lib/note-detail/style.scss
@@ -81,6 +81,10 @@
     cursor: inherit;
   }
 
+  .checkbox {
+    cursor: pointer;
+  }
+
   &::first-line {
     font-size: 120%;
   }


lib/note-detail/style.scss
```diff
@@ -81,6 +81,11 @@
     cursor: inherit;
   }
 
+  // Show pointer cursor on checkboxes
```
dmsnell: we probably don't need this comment as the CSS is pretty clear

dmsnell (commented):
Nice change! And thanks for the contribution. This seems like a straightforward fix.

lib/note-detail/style.scss
arpitbatra123: You're right. I'll remove it!

arpitbatra123:
@dmsnell I've addressed the comment! 😄

PR #2191, opened 2020-07-07
Use browser for authing a user via WP Login

belcherj:
Cherry-picked this [PR](https://github.com/Automattic/simplenote-electron/pull/2147) on top of the release. Approval on the PR was there.

See p2XJRt-1Y0-p2

### Fix

In order to pass the Windows store certification, we cannot do auth in the app. Instead, it needs to be pushed to a browser.

There is still uncertainty that this will resolve the issue with certification however this is the only solution that I can see as a path forward.

### Test

1. Run app
2. In Electron click the WordPress.com login button
3. Log in
4. Were you able to log in?

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -22,18 +22,22 @@ const spellcheck = require('./spellchecker');
 require('module').globalPaths.push(path.resolve(path.join(__dirname)));
 
 module.exports = function main() {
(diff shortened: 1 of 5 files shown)


codebykat (approved):
🚢 it!

PR #2192, opened 2020-07-07
Fix lazing loading

belcherj:
### Fix

I removed lazy loading as an oversight in: https://github.com/Automattic/simplenote-electron/pull/2191 This adds it back.

### Test

1. Can you sign into the app


dmsnell (approved):
@belcherj thanks for restoring this. did it break anything though? I'm not seeing anything obviously wrong with either version lazy or eager.

PR #2194, opened 2020-07-08
Adds logging and ability to download logs

belcherj:
### Fix

Stole liberally from https://github.com/Automattic/wp-desktop/pull/814/files.

This PR adds logging to the app. It also adds a menu item to download those logs. This will be helpful in diagnosing issues.

### Test

1. Download installers from Circle
2. Install
3. Download logs

### Release

Adds logging and ability to download logs

### DIFF
--- a/desktop/logger/archiver.js
+++ b/desktop/logger/archiver.js
@@ -4,8 +4,7 @@
  * External dependencies
  */
 const fs = require('fs');
-const path = require('path');
-const archiver = require('archiver');
+var JSZip = require('jszip');
 
 /**
  * Internal dependencies
@@ -20,43 +19,42 @@ module.exports = {
    * @param {String} dst Path to destination archive
    * @param {function():void} onZipped Callback invoked when archive is complete
    */
-  zipContents: (contents, dst, onZipped) => {
-    let output = fs.createWriteStream(dst);
-    let archive = archiver('zip', {
-      zlib: { level: 9 },
+  zipContents: (logPath, dst, onZipped) => {
+    const zip = new JSZip();
+    const contentPromise = new JSZip.external.Promise(function (
+      resolve,
+      reject
+    ) {
+      fs.readFile(logPath, function (err, data) {
+        if (err) {
+          log.warn('Unexpected error: ', err);
+          reject(err);
+        } else {
+          resolve(data);
+        }
+      });
     });
+    zip.file('simplenote.log', contentPromise);
 
-    output.on('close', function () {
-      log.debug('Archive finalized: %s bytes written', archive.pointer());
-      if (typeof onZipped === 'function') {
-        onZipped();
-      }
-    });
+    let output = fs.createWriteStream(dst);
 
     // Catch warnings (e.g. stat failures and other non-blocking errors)
-    archive.on('warning', function (err) {
+    output.on('warning', function (err) {
       log.warn('Unexpected error: ', err);
     });
 
-    archive.on('error', function (err) {
+    output.on('error', function (err) {
       throw err;
     });
 
-    archive.pipe(output);
-
-    for (let i = 0; i < contents.length; i++) {
-      const src = contents[i];
-      const zipSubdir = path.basename(dst).replace('.zip', '');
-      const dstInZipSubdir = path.join(zipSubdir, path.basename(src));
-
-      const stat = fs.lstatSync(src);
-      if (stat.isDirectory()) {
-        archive.directory(path.join(src, path.sep), dstInZipSubdir);
-        continue;
-      }
-      archive.file(src, { name: dstInZipSubdir });
-    }
-
-    archive.finalize();
+    zip
+      .generateNodeStream({ type: 'nodebuffer', streamFiles: true })
+      .pipe(output)
+      .on('finish', function () {
+        log.debug('Archive finalized');
+        if (typeof onZipped === 'function') {
+          onZipped();
+        }
+      });
   },
 };
--- a/desktop/logger/archiver.js
+++ b/desktop/logger/archiver.js
@@ -1,0 +1,62 @@
+'use strict';
+
+/**
+ * External dependencies
+ */
+const fs = require('fs');
+const path = require('path');
+const archiver = require('archiver');
+
+/**
+ * Internal dependencies
+ */
+const log = require('./')('desktop:lib:archiver');
+
+module.exports = {
+  /**
+   * Compresses `contents` to the archive at `dst`.
+   *
+   * @param {String[]} contents Paths to be zipped
+   * @param {String} dst Path to destination archive
+   * @param {function():void} onZipped Callback invoked when archive is complete
+   */
+  zipContents: (contents, dst, onZipped) => {
+    let output = fs.createWriteStream(dst);
+    let archive = archiver('zip', {
+      zlib: { level: 9 },
+    });
+
+    output.on('close', function () {
+      log.debug('Archive finalized: %s bytes written', archive.pointer());
+      if (typeof onZipped === 'function') {
+        onZipped();
+      }
+    });
+
+    // Catch warnings (e.g. stat failures and other non-blocking errors)
+    archive.on('warning', function (err) {
+      log.warn('Unexpected error: ', err);
+    });
+
+    archive.on('error', function (err) {
+      throw err;
+    });
+
+    archive.pipe(output);
+
+    for (let i = 0; i < contents.length; i++) {
+      const src = contents[i];
+      const zipSubdir = path.basename(dst).replace('.zip', '');
+      const dstInZipSubdir = path.join(zipSubdir, path.basename(src));
+
+      const stat = fs.lstatSync(src);
+      if (stat.isDirectory()) {
+        archive.directory(path.join(src, path.sep), dstInZipSubdir);
+        continue;
+      }
+      archive.file(src, { name: dstInZipSubdir });
+    }
+
+    archive.finalize();
+  },
+};
--- a/desktop/logger/archiver.js
+++ b/desktop/logger/archiver.js
@@ -21,22 +21,21 @@ module.exports = {
    */
   zipContents: (logPath, dst, onZipped) => {
     const zip = new JSZip();
-    const contentPromise = new JSZip.external.Promise(function (
-      resolve,
-      reject
-    ) {
-      fs.readFile(logPath, function (err, data) {
-        if (err) {
-          log.warn('Unexpected error: ', err);
-          reject(err);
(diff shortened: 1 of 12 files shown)


desktop/logger/namespaces/index.js
```diff
@@ -0,0 +1,45 @@
...
+   */
+  check: function (namespace) {
+    if (!this.populated) this.populate();
+    if (this.namespaces.indexOf('*') !== -1) return true;
+    if (this.namespaces.indexOf(namespace) !== -1) return true;
+    /* If it is as 'server:api:controller', it could have a wildcard as 'server:*' */
+    if (namespace.indexOf(':') !== -1) {
+      /* Different levels of the namespace. Using the example of above: 'server' is level 0, 'api' is level 1 and
+       * 'controller' is level 2. */
+      let levels = namespace.split(':');
```
dmsnell: looks like this could be `const` here

desktop/logger/namespaces/index.js
```diff
@@ -0,0 +1,45 @@
...
+   * Checks if the namespace is available to debug. The namespace could be contained in wildcards.
+   * Ex: 'server:api:controller' would pass the check (return true) if the 'server:api:controller' is in the
+   * environment variable or if 'server:api:*' or 'server:*' is in the environment variable.
+   * @param namespace {String} - Namespace to check.
+   * @returns {boolean} Whether or not the namespace is available.
+   */
+  check: function (namespace) {
+    if (!this.populated) this.populate();
+    if (this.namespaces.indexOf('*') !== -1) return true;
+    if (this.namespaces.indexOf(namespace) !== -1) return true;
```
dmsnell: can we include `{}`s in these checks? I understand the value of the cleanliness of one-liners but I'm worried about the kinds of bugs that creep in due to oversight about missing brackets

desktop/logger/archiver.js
```diff
@@ -0,0 +1,62 @@
...
+
+    archive.on('error', function (err) {
+      throw err;
+    });
+
+    archive.pipe(output);
+
+    for (let i = 0; i < contents.length; i++) {
+      const src = contents[i];
+      const zipSubdir = path.basename(dst).replace('.zip', '');
```
dmsnell: if we want to strip the trailing `.zip` then we should make that explicit. we might cause unexpected behavior if someone passes a path like `files.ziparchives.zip`

```js
.replace(/\.zip$/, '')
```

desktop/logger/index.js
```diff
@@ -0,0 +1,100 @@
...
+ * Module variables
+ */
+const maxFiles = 3;
+const maxsize = 15000000;
+
+module.exports = (namespace, options) => {
+  if (!options || typeof options !== 'object') options = {};
+
+  const formatMessageWithMeta = (info) => {
+    const args = info[Symbol.for('splat')];
```
dmsnell, line 39: what's the reason for a `Symbol` here and what's the reason for `splat`?

desktop/logger/index.js
```diff
@@ -0,0 +1,100 @@
...
+  return {
+    error: (message, meta) => logger.error(message, meta),
+    warn: (message, meta) => logger.warn(message, meta),
+    info: (message, meta) => logger.info(message, meta),
+    debug: (message, meta) => {
+      if (enabled) logger.debug(message, meta);
+    }, // eslint-disable-line brace-style
+    silly: (message, meta) => {
+      if (enabled) logger.silly(message, meta);
+    }, // eslint-disable-line brace-style
```
dmsnell: commented above: can we note exclude these and enforce the lint rule to avoid the risk of thinking we're in the `if` body when in fact we're not?

desktop/logger/archiver.js
belcherj: deleted that code.

desktop/logger/index.js
belcherj: done

desktop/logger/namespaces/index.js
belcherj: done

desktop/logger/namespaces/index.js
belcherj: done

desktop/logger/index.js
nsakaimbo: It's been a while so my recollection is a little fuzzy - wish I added a comment when I initially wrote this. 😅  Per the Winston docs, `Symbol.for('splat')` is how splat arguments are extracted. I believe this piece of logic handles the case where splat arguments could be populated with an array like `[undefined]` or some such, leading to some funky output.

I _think_ this is moot if splatting is entirely not used (e.g. `log.info("message: %s", x)`), although you may encounter it even with `log.info("message: ", x)`, but not 100 percent sure -  I may have been working around some default behavior. You may need to verify output with/without this piece of logic.

desktop/logger/archiver.js
```diff
@@ -0,0 +1,60 @@
...
+    ) {
+      fs.readFile(logPath, function (err, data) {
+        if (err) {
+          log.warn('Unexpected error: ', err);
+          reject(err);
+        } else {
+          resolve(data);
+        }
+      });
+    });
```
dmsnell: is this Promise necessary since our platform support already assumes a native `Promise`? (and arrow functions)

```js
zip.file( 'simplenote.log', new Promise( ( resolve, reject ) =>
	fs.readFile( logPath, ( error, data ) => {
		if (error) {
			log.warn('Unexpected error: ', error);
			reject(error);
		} else {
			resolve(data);
		}
} ) ) );
```

I suppose there could be some special interface that `JSZip.external.Promise` provides but I'm guessing they offer that as a compatibility layer for older Node/browser environments

desktop/logger/archiver.js
```diff
@@ -0,0 +1,60 @@
...
+    zip.file('simplenote.log', contentPromise);
+
+    let output = fs.createWriteStream(dst);
```
dmsnell: small nitpick, this can be a `const`

desktop/logger/index.js
```diff
@@ -0,0 +1,106 @@
...
+  );
+
+  const logPath = path.join(
+    (app && app.getPath('appData')) || '',
+    'logs',
+    'simplenote-main.log'
+  );
+
+  const baseOptions = {
+    level: process.env.LOG_LEVEL || 'silly',
```
dmsnell: this is just silly 😉

desktop/logger/index.js
```diff
@@ -0,0 +1,106 @@
...
+        filename: path.basename(logPath),
+        maxFiles,
+        maxsize,
+        format: baseformat,
+      }),
+    ],
+  };
+
+  const enabled = namespaces.check(namespace);
+  let logger = createLogger({ ...baseOptions, ...options });
```
dmsnell: `const`

desktop/logger/index.js
```diff
@@ -0,0 +1,106 @@
...
+  if (process.env.DEBUG) {
```
dmsnell: did we intend on using this `ENV`?
if so maybe we should add a mention in the `README` about it, or in the `CONTRIBUTING` document

otherwise we could base it on something more in line with what we already use, like `process.env.NODE_ENV !== 'production'`

desktop/logger/namespaces/index.js
```diff
@@ -0,0 +1,51 @@
...
+   * Boolean indicating if the object is populated with the environment data.
+   */
+  populated: false,
+  /**
+   * Populates the private data 'namespaces' as an array with the different namespaces from the DEBUG
+   * environment variable. It splits the data with ',' as separator.
+   * @private
+   */
+  populate: function () {
+    let envString = process.env.DEBUG;
```
dmsnell: same note on `ENV` check, also this could be `const`

desktop/logger/namespaces/index.js
```diff
@@ -0,0 +1,51 @@
...
+      const levels = namespace.split(':');
+      let level;
+      for (let i = 1; i < levels.length; i++) {
+        level = levels.slice(0, i).join(':') + ':*';
+        if (this.namespaces.indexOf(level) !== -1) {
+          return true;
+        }
+      }
+    }
+    return false;
```
dmsnell: this last bit _looks_ like it's trying to find out if we have _some_ level which is in the namespace…

```js
return levels.split(':').some(
	( _, level, levels) => this.namespaces.indexOf( levels.slice(0, level).join(':'). ':*' ) !== -1
)
```

…I also think we're safe to use `String.prototype.includes` since it's been available since Node 4.9.1 and `Array.prototype.includes` which came in with Node 6.4.0

```js
check: function(namespace) {
	if (!this.populated) {
		this.populate();
	}

	if (this.namespaces.includes('*')) {
		return true;
	}

	if (this.namespaces.includes(namespace)) {
		return true;
	}

	return namespace.split(':').some((_, level, levels) =>
		this.namespaces.includes(levels.slice(0,level).join(':') + ':*')
	);
}
```

desktop/logger/index.js
```diff
@@ -0,0 +1,106 @@
...
+  );
+
+  const logPath = path.join(
+    (app && app.getPath('appData')) || '',
+    'logs',
+    'simplenote-main.log'
+  );
+
+  const baseOptions = {
+    level: process.env.LOG_LEVEL || 'silly',
```
dmsnell: on second thought, by default to `silly` we're saying that customers will by default have `silly` as their log level and any attempt to change that will require that they first quit the application and figure out how to start it again with an `ENV` variable set.

would it do us well to default to something like `warn` instead?

desktop/logger/namespaces/index.js
```diff
@@ -0,0 +1,46 @@
...
+   * Boolean indicating if the object is populated with the environment data.
+   */
+  populated: false,
+  /**
+   * Populates the private data 'namespaces' as an array with the environment from the NODE_ENV
+   * environment variable. It splits the data with ',' as separator.
+   * @private
+   */
+  populate: function () {
+    this.namespaces = [process.env.NODE_ENV];
```
dmsnell: mmm…

namespaces is supposed to refer to log-namespacing. I think we meant to use another `ENV` here or at least map the `NODE_ENV` to something…

```js
this.namespaces = [process.env.NODE_ENV === 'production' ? '' : '*'];
```

I know I suggested using `NODE_ENV` earlier but I meant to suggest that we use that as a guide to establish the namespace but not use it directly _as_ the namespace, if that makes sense

desktop/logger/index.js
belcherj: Makes sense.

dmsnell (approved):

PR #2201, opened 2020-07-13
Fixes Evernote import

### LINKED ISSUE #2200: Simplenote 1.20.0 unusable after trying to import notes
Simplenote unusable after trying to import notes. As you can see in the animation below the whole window goes white.

import-notes-broken

### Reproduced
This happens every time when trying to import notes.
> 1. Go to "File"
> 2. Click on "Import Notes"
> 3. Select one of the import options.
> 4. See error...

OS: Solus Budgie 4.1
Make: Re-packaged the .deb file to .eopkg using Solus build tools.
App Version: Simplenote 1.20.0

belcherj:
### Fix

Evernote import required fs which is no longer available in the web app due to security changes. This moves the Evernote import to the Electron side so that it can use fs.

This removes the progress on the import and removes a differentiation between no notes and an import error. I can add this but the plumbing required is significant. With more time my hope is to move this code back into the react side and make it available in the web app.

Fixes: https://github.com/Automattic/simplenote-electron/issues/2200

### Test

1. Get a .enex file from Evernote
2. Import into the app.
3. Does it import correctly? 

### Release
- Fix Evernote Import


desktop/preload.js
```diff
@@ -1,26 +1,32 @@
...
+      ipcRenderer.on(channel, newCallback);
+    }
+  },
+  once: (channel, callback) => {
     if (validChannels.includes(channel)) {
-      // Deliberately strip event as it includes `sender`
-      ipcRenderer.on(channel, (event, ...args) => {
-        return func(...args);
-      });
+      const newCallback = (_, data) => callback(data);
```
dmsnell: why strip out the event value? it seems odd to intentionally hide it.

in any case, looking at the type signature I see `callback` taking two arguments, the `event` and `notes` but this `newCallback` specifically drops anything past the second argument, so it's almost like we're preventing the app from behaving the way we're encoding it. I'm sure I might be missing something though.

if the issue were simply that we didn't want to need to worry about `event` in `addNotesToApp` I'd definitely be more in favor of managing that specific detail when passing `addNotesToApp` in as the callback so that we don't cause its particular need to ripple into other cases of `once`

```ts
window.electron?.once('notesImported', (event, notes) => addNotesToApp(notes));
```

lib/utils/import/evernote/index.ts
```diff
@@ -37,84 +28,28 @@ class EvernoteImporter extends EventEmitter {
...
-    // Read the file via stream
-    fs.createReadStream(file.path).pipe(saxStream);
+      notes.forEach((note) => {
+        coreImporter.importNote(note, this.options);
+      });
+      this.emit('status', 'complete', notes.length);
+    };
+
+    window.electron?.once('notesImported', addNotesToApp);
+    window.electron?.send('importNotes', file.path);
```
dmsnell, line 64: unlike other parts of the app where integrations may or may not be present there is no way to make the Evernote importer work without Electron, right? 

if this is true then I think we need to hide it when not in an Electron environment and abort if `window.electron` isn't defined. that means we shouldn't need to use the optional chaining for safety and can rely on `window.electron.once()`.

that will probably leave a type error here but if we let that error stay then it will be a reminder for us to either fix it deeper in our types and how we structure the code or to remember that if something fails it could be because we're assuming `window.electron` is there (vs. hiding that fact in a graceful fail for a process that directly depends on it).

codebykat (approved):
This tested well for me with a HUGE Evernote import (900+ notes). I don't have any opinions on the code.

codebykat (changes requested):
Unfortunately, I retract my approval. 😭 

After importing a large import (900+ notes) I am unable to logout on Electron.

I couldn't reproduce this with a smaller import.

This was a dump of an extremely old Evernote account of mine which contains all sorts of random stuff (images etc) -- if it looks like a problem with the content rather than the size I can try to narrow it down more to find the problem note(s).

Here's the error I get when I click Log Out. Looks like it might be related to a large import. I don't know if this is a regression - when is the last time Evernote import worked? I can test that with the same export.

```
index.tsx?5e84:79 Uncaught (in promise) ReferenceError: require is not defined
    at SettingsDialog.eval [as showElectronWarningDialog] (index.tsx?5e84:79)
    at SettingsDialog.eval [as showUnsyncedWarning] (index.tsx?5e84:75)
    at eval (index.tsx?5e84:51)
    at eval (bucket.js?d231:50)
```

This was in the console too, but I'm not sure when it happened.

`arget: IDBRequest
error: DOMException: The serialized keys and/or value are too large (size=178044082 bytes, max=133169152 bytes).`

Screen Shot 2020-07-14 at 11 19 43 PM

codebykat:
I redid it. This error comes in when the import is reported as complete:

```
Uncaught (in promise) 
Event {isTrusted: true, type: "error", target: IDBRequest, currentTarget: null, eventPhase: 0, …}
bubbles: true
cancelBubble: false
cancelable: true
composed: false
currentTarget: null
defaultPrevented: false
eventPhase: 0
isTrusted: true
path: []
returnValue: true
srcElement: IDBRequest {result: undefined, error: DOMException: The serialized keys and/or value are too large (size=178044082 bytes, max=133169152 b…, source: IDBObjectStore, transaction: IDBTransaction, readyState: "done", …}
target: IDBRequest
error: DOMException: The serialized keys and/or value are too large (size=178044082 bytes, max=133169152 bytes).
onerror: ƒ (e)
onsuccess: ƒ ()
readyState: "done"
result: undefined
source: IDBObjectStore {name: "note", keyPath: "id", indexNames: DOMStringList, transaction: IDBTransaction, autoIncrement: false}
transaction: IDBTransaction {objectStoreNames: DOMStringList, mode: "readwrite", db: IDBDatabase, error: DOMException: The serialized keys and/or value are too large (size=178044082 bytes, max=133169152 b…, onabort: null, …}
__proto__: IDBRequest
timeStamp: 1005092.1850000159
type: "error"
__proto__: Event
```

desktop/preload.js
belcherj: This is intentional to not expose the event to the non-electron side of things. The event includes the `sender`. Here are the details: https://www.electronjs.org/docs/api/ipc-main#sending-messages

lib/utils/import/evernote/index.ts
belcherj: I can add a check to the top of the file to return if it is not Electron. This code should never be executed outside of Electron: https://github.com/Automattic/simplenote-electron/blob/develop/lib/dialogs/import/source-selector.tsx#L17

belcherj:
@codebykat Could you try that Evernote file on an old version of Electron? I'd like to see what happens there.  We limit the size of text files imports to 1mb and Simplenote imports to 5mb.

codebykat:
Comparing to 1.17:

* In both cases, the import fails with the above error (`The serialized keys and/or value are too large`) and the error is not caught. It erroneously reports that the import is complete. We should handle that better but that can be a follow-up task.
* After importing, I **can** successfully log out on 1.17. (though it complains about unsynced notes and there are some console errors). I think that makes this **a blocker for this PR** as the too-large import leaves the app in a broken state.
* On this branch, it does not report the number of notes processed as it is processing them. That is really nice to have for larger imports.

My full note export is 130.1 MB and it DOES appear to contain 905 notes (the number reported as successful by the import), so I don't know whether it actually imported fewer than 905. There's obviously a lot of image content which doesn't get handled well.

1.17:

Screen Shot 2020-07-15 at 12 49 07 PM

Here:

Screen Shot 2020-07-15 at 12 53 08 PM

codebykat (approved):
Confirmed this is working now! We're still not catching the promise error, which maybe we should at some point, but it correctly counts and imports the notes and I am able to log out afterward (after rebasing to include #2230)

PR #2202, opened 2020-07-14
Update/offline badge

belcherj:
### Fix

Moves the offline badge from the bottom right to the note toolbar.

On mobile and smaller screens the badge will look quite odd and will not be visible when the note list is open. I think we should consider leaving it where it is.

Design:
Screen Shot 2020-07-14 at 4 05 58 PM

This PR:
Screen Shot 2020-07-14 at 3 55 36 PM
Screen Shot 2020-07-14 at 3 55 53 PM
Screen Shot 2020-07-14 at 4 03 10 PM

### Test

1. Load app
2. Turn off wifi

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -24,7 +24,6 @@ type OwnProps = {
 type StateProps = {
   autoHideMenuBar: boolean;
   hotkeysEnabled: boolean;
-  isOffline: boolean;
   isSmallScreen: boolean;
   lineLength: T.LineLength;
   showNavigation: boolean;
@@ -137,7 +136,6 @@ class AppComponent extends Component<Props> {
   render() {
     const {
       isDevConfig,
-      isOffline,
       lineLength,
       showNavigation,
       showNoteInfo,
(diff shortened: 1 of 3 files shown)


dmsnell (approved):
The narrow screenshot is a bit confusing but if there are any problems we can fix them in follow-up work.

It looks like the offline indicator might be uncomfortably close to the back-arrow, or on the wrong side of it, but it's hard to see in the context of the picture and I don't see it blocking this work. We can follow-up quickly after getting it in the branch.

belcherj:
I have concerns with this approach in general. I think it should be left as is. The PR was to show the shortcomings with this approach.

SylvesterWilmott:
>I have concerns with this approach in general. I think it should be left as is. The PR was to show the shortcomings with this approach.

Yeah it doesn't look great with the current small screen UI.

I want to avoid using a fixed position, non-dismissable element. It's not a warning or alert so I think it should integrate into the UI.

Currently, on the notes list on small devices, the navigational icons are positioned some distance from the edge of the viewport, however, in the editor, they are equally spaced within the bar. This already creates alignment issues when going between those views.

Instead, we can keep the same icon alignment as on desktop with sufficient left and right margins on the offline indicator to be able to offset it when on very narrow mobile devices.

Screenshot 2020-07-15 at 00 43 51

And on a smaller screen:

Screenshot 2020-07-15 at 00 47 22

One other thing to note is that the planned work with note options means that we will have less icons to display in the navigation (spring cleaning) so that offset won't be quite as dramatic.

dmsnell (approved):
Looks like an improvement! We can continue to refine in-branch if we think it needs more adjustments.

PR #2203, opened 2020-07-14
Style the last synced indicator

belcherj:
### Fix

Design:
Screen Shot 2020-07-14 at 4 20 11 PM

This PR:
Screen Shot 2020-07-14 at 4 17 53 PM
Screen Shot 2020-07-14 at 4 18 15 PM

Note: The indicator is currently broken when you first open app since it doesn't have the last updated.

Another Note: How important is it that is say "Today". This is more difficult since it is not localized. The code here will put the date/time in a localized format.

### Test

1. Open a note
2. Observe the last synced indicator

### DIFF
--- a/lib/note-editor/index.tsx
+++ b/lib/note-editor/index.tsx
@@ -126,9 +126,8 @@ export class NoteEditor extends Component<Props> {
 
     return (
       <div className="note-editor theme-color-bg theme-color-fg">
-        <div>
-          Note last updated: {Math.floor((Date.now() - lastUpdated) / 1000)}s
-          ago ({new Date(lastUpdated).toLocaleTimeString()})
+        <div className="last-sync">
+          Last synced: {new Date(lastUpdated).toLocaleString()}
         </div>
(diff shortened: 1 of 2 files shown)


dmsnell:
> The indicator is currently broken when you first open app since it doesn't have the last updated.

This PR should be good to merge by why not hide the indicator when it's unavailable? Or is it more complex knowing what to do when it's not available?

```ts
Last synced: {lastUpdated ? new Date(lastUpdated).toLocaleString() : 'unknown'}
```

We can get it out either way and see how it feels.

:shipit:

dmsnell (approved):

SylvesterWilmott:
>why not hide the indicator when it's unavailable

Agreed, if we can do this, let's do it.

>How important is it that is say "Today". This is more difficult since it is not localized. The code here will put the date/time in a localized format.

The format in your screenshots looks good to me @belcherj 

Would you mind adding full screenshots?

belcherj:
As requested: https://github.com/Automattic/simplenote-electron/pull/2236

PR #2204, opened 2020-07-15
Update note list note status icons

belcherj:
### Fix

Notes:

1. When in condensed list mode or a note that does not have the content the icons took up too much space. I increased the minimum height to 64px; Another option would be to decrease the spacing between icons so that they fit.

Design:
Screen Shot 2020-07-15 at 1 16 14 PM

This PR:

NoteStatusIcons
NoteStatusIconsDarkMode

### Test

- Open app
- Make a change, observe attention icon in the note list
- Hover over where the pin icon would observe it appears
- Click it, does it pin note?
- Do pinned notes show pin icon?

### DIFF
--- a/lib/icons/attention.tsx
+++ b/lib/icons/attention.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+
+export default function AlertIcon() {
+  return (
+    <svg
+      className="icon-attention"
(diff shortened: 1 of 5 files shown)


dmsnell (approved):
This looks so good! Thanks @belcherj and @SylvesterWilmott

PR #2206, opened 2020-07-15
Updates unsynced notes log out dialog

belcherj:
### Fix

Notes:

- Font is off but I don't think we should change that here
- Much of this dialog is hardcoded css that should apply to all dialogs. I think this should wait till spring cleaning
- I did change "Close" to the "X" icon on all dialogs.
- Some of the line heights are off from a bug that stems from the editor font used to render checkboxes.

Design:

Screen Shot 2020-07-15 at 2 54 59 PM

This PR:

Screen Shot 2020-07-15 at 4 03 51 PM
Screen Shot 2020-07-15 at 4 05 07 PM
Screen Shot 2020-07-15 at 4 05 34 PM

### Test

1. Turn off wifi
2. Make changes to a note
3. Try logging out.

### DIFF
--- a/lib/dialogs/logout-confirmation/index.tsx
+++ b/lib/dialogs/logout-confirmation/index.tsx
@@ -1,5 +1,7 @@
 import React, { Component, Fragment } from 'react';
 import { connect } from 'react-redux';
+
+import AttentionIcon from '../../icons/attention';
 import Dialog from '../../dialog';
 import { closeDialog } from '../../state/ui/actions';
 import { getUnconfirmedChanges } from '../../state/simperium/functions/unconfirmed-changes';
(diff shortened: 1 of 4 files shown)


lib/dialogs/logout-confirmation/index.tsx
```diff
@@ -32,56 +34,45 @@ export class LogoutConfirmation extends Component<Props> {
...
       <div className="logoutConfirmation">
-        <Dialog onDone={closeDialog} title="Unsynchronized Notes Found">
-          <section>
-            <h1>Are you sure you want to logout?</h1>
-          </section>
-
+        <Dialog onDone={closeDialog} title="Unsynced Notes">
+          <p className="explanation">
+            {notes.size > 0
+              ? 'Logging out will delete any unsynced notes.'
```
dmsnell: Not sure if anyone else has feelings about this but I find the use of the word form `unsynced` a bit too casual for official use. Any thoughts about "unsynchronized" as a replacement? Am I over-thinking it?

lib/dialogs/logout-confirmation/index.tsx
```diff
@@ -32,56 +34,45 @@ export class LogoutConfirmation extends Component<Props> {
...
+              : 'All notes have syncronized!'}
```
dmsnell: typo: "synchronized" (also one of the reasons I don't like `unsynced` and `synced` is because of the missing `h` - this is why I prefer the other short-hand version `sync'd` which opaquely shows we're removing letters vs. `synced` which makes it look like we aren't when in fact we are)

lib/dialogs/logout-confirmation/index.tsx
```diff
@@ -32,56 +34,45 @@ export class LogoutConfirmation extends Component<Props> {
...
+          </p>
+
+          <p className="explanation-secondary">
+            {notes.size > 0 && 'Possibly unsynchronized notes'}
+          </p>
```
dmsnell: Do we need an empty paragraph if there are no changes?

lib/dialogs/logout-confirmation/index.tsx
```diff
@@ -32,56 +34,45 @@ export class LogoutConfirmation extends Component<Props> {
...
-              safe.
-            </p>
+            <section className="change-list">
+              <ul>
+                {[...notes.entries()].map(([noteId, note]) => {
+                  const { title, preview } = noteTitleAndPreview(note);
+
+                  return (
+                    <li key={noteId}>
+                      <AttentionIcon />
```
dmsnell: when I saw the screenshot with the new icon I wondered if it might not be fun to keep the notes in the list but instead to replace the icon with one confirming the sync status.

that is, instead of removing the indicator that it could be unsynchronized…_add_ an indicator that it _has_ been synchronized

belcherj: The designs did not have a resolved state. I do like this idea!!

dmsnell: I just realized that this will increase the complexity in the code since we grab this data in a stateless fashion.

Should just need some initial state when the component mounts to store the initial set of note ids and then on render we can compare that set against the set passed in to know if they are confirmed or not.

lib/dialogs/logout-confirmation/index.tsx
belcherj: I agree.

lib/dialogs/logout-confirmation/index.tsx
belcherj: Im thinking we leave that for another PR.

lib/dialogs/logout-confirmation/index.tsx
belcherj: done

lib/dialogs/logout-confirmation/index.tsx
belcherj: fixed it

dmsnell (approved):
Code looks good, visuals look nice. Thanks!

SylvesterWilmott:
Looks good!

I know this is merged but just some small feedback. The blue link in dark mode should be blue30.

PR #2208, opened 2020-07-16
Add beta invitation banner

codebykat:
### Fix

This adds a banner to the bottom of the page that links to the beta on staging.simplenote.com.

### Screenshots

Screen Shot 2020-07-16 at 12 17 35 AM

### Release

Not updated: Add beta invitation banner

### DIFF
--- a/lib/components/beta-bar/index.tsx
+++ b/lib/components/beta-bar/index.tsx
@@ -1,0 +1,17 @@
+import React, { useState } from 'react';
+import SmallCrossIcon from '../../icons/cross-small';
+
+const BetaBar = () => {
+  const [isVisible, setIsVisible] = useState(true);
+  return isVisible ? (
+    <div className="beta-bar">
+      You&lsquo;re invited to try Simplenote Beta.
+      <a href="https://staging.simplenote.com">Try it now.</a>
+      <a className="icon" onClick={() => setIsVisible(false)}>
+        <SmallCrossIcon />
+      </a>
+    </div>
+  ) : null;
+};
+
+export default BetaBar;
--- a/lib/components/beta-bar/index.tsx
+++ b/lib/components/beta-bar/index.tsx
@@ -2,35 +2,34 @@ import React, { useState } from 'react';
 import SmallCrossIcon from '../../icons/cross-small';
 import { isElectron } from '../../utils/platform';
 
+const loadOptOut = (): boolean => {
+  try {
+    const stored = localStorage.getItem('betaBannerDismissedAt');
+    const dismissedAt = stored ? parseInt(stored, 10) : -Infinity;
+
+    return (
+      isNaN(dismissedAt) || Date.now() - dismissedAt > 7 * 24 * 60 * 60 * 1000
+    );
+  } catch (e) {
+    return true;
+  }
+};
+
+const storeOptOut = () => {
+  try {
+    localStorage.setItem('betaBannerDismissedAt', Date.now().toString());
+  } catch (e) {
+    // pass
+  }
+};
+
 const BetaBar = () => {
-  const [isVisible, setIsVisible] = useState(true);
-
-  const shouldShowBanner = () => {
-    if (isElectron) {
-      return false;
-    }
-    try {
-      const stored = localStorage.getItem('betaBannerDismissedAt');
-      const dismissedAt = stored ? parseInt(stored, 10) : -Infinity;
-
-      return (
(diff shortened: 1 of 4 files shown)


SylvesterWilmott:
Looks good!

belcherj:
Do we want to persist this to localstorage so it doesn't appear on every refresh?

We should also gate this to Electron

dmsnell:
> Do we want to persist this to localstorage so it doesn't appear on every refresh?

That seems reasonable but maybe if we do then expire it after a week or two. Remember that `localStorage` will throw an exception if unavailable…

```ts
const shouldShowBanner = () => {
	try {
		const stored = localStorage.getItem('betaBannerDismissedAt');
		const dismissedAt = stored ? parseInt( stored, 10 ) : -Infinity;

		return isNaN( dismissedAt ) || ( Date.now() - dismissedAt ) > 7 * 24 * 60 * 60 * 1000;
	} catch ( e ) {
		return true;
	}
}

const dismissBanner = () => {
	try {
		localStorage.setItem( 'betaBannerDismissedAt', Date.now() );
	} catch ( e ) {
		// pass
	}
}
```

codebykat:
Okay, not shown in Electron and the opt-out is saved in localStorage for a week. @dmsnell I'm curious if there's a better way to organize the code here -- it feels like the state variable is now vestigial and redundant, but without it, it won't trigger a re-render when the banner is dismissed.

belcherj (approved):
I think the useState is fine in this case. Since interacting with local storage is a side effect it could be put in a useEffect but since this is a short term banner I think this implementation is sound.

lib/components/beta-bar/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  };
+
+  const dismissBanner = () => {
+    setIsVisible(false);
+    try {
+      localStorage.setItem('betaBannerDismissedAt', Date.now().toString());
+    } catch (e) {
+      // pass
+    }
+  };
```
dmsnell, line 31: @belcherj is right about not wanting to call this on every render. we can put it in `useEffect()` or more ideally use it to set the default value for `useState()` since we never want to make the banner appear once we've loaded if it shouldn't have already been showing at the start of loading

```ts
const loadDismissal = (): boolean => {
	try {
		const stored = localStorage.getItem('betaBannerDismissedAt');
		const dismissedAt = stored ? parseInt( stored, 10 ) : -Infinity;

		return isNaN( dismissedAt ) || ( Date.now() - dismissedAt ) > 7 * 24 * 60 * 60 * 1000;
	} catch ( e ) {
		return true;
	}
}

const storeDismissal = () => {
	try {
		localStorage.setItem( 'betaBannerDismissedAt', Date.now() );
	} catch ( e ) {
		// pass
	}
}

const BetaBar = () => {
	const [isVisible, setIsVisible] = useState(loadDismissal());
	const dismissBanner = () => { storeDismissal(); setIsVisible(false); };

	…
}
```

codebykat: this is very clever thank you!

codebykat:
updated to rework per @dmsnell's suggestion, I think this is good to go

lib/components/beta-bar/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+    localStorage.setItem('betaBannerDismissedAt', Date.now().toString());
+  } catch (e) {
+    // pass
+  }
+};
+
+const BetaBar = () => {
+  const [isVisible, setIsVisible] = useState(loadOptOut());
+  const dismissBanner = () => {
+    storeOptOut(), setIsVisible(false);
```
dmsnell: you've probably already spotted this, and while I love the comma operator and originally doodled out a solution using it, it's probably a typo here

we _could_ do `() => (storeOptOut(), setIsVisible(false))` which would be more normative for comma operator expressions, but within the `{}`s it's awkward; probably best separated by semi-colons `() => { storeOptOut(); setIsVisible(false); }`

dmsnell: I say "normative" because comma-operator expressions return the last value in their chain but when put in `{}`s without a return statement nothing is returned, meaning we're using a construct built to sequence operations and return the last value but intentionally ignoring that value then not returning anything

codebykat: noooope that was a typo and yet somehow worked

dmsnell (approved):
beta beta beta

PR #2209, opened 2020-07-16
Empty trash

codebykat:
### Fix

Re-implements Empty Trash functionality

### Test

1. Go to Trash
2. Empty trash
3. Do all the deleted notes (and none of the non-deleted notes) get deleted?


dmsnell (approved):
Weee trash pickup 😄

PR #2210, opened 2020-07-16
Fix tag list CSS

codebykat:
### Fix
* The tag list should scroll while still keeping the header and edit button visible. Follow-up to https://github.com/Automattic/simplenote-electron/pull/2183
* Add some padding on the right so long tags don't go right up to the edge and the grab icons aren't under the scrollbar
* Remove top/bottom margins from the tag list since we have a bit less space now
* Make trash icon turn red on hover
* Pointer cursors over trash and grab icons
* Added a shadow to the tag title (possibly controversial!! but I think it looks better when you're scrolling tags)

### Screenshots

Screen Shot 2020-07-16 at 3 18 26 PM
Screen Shot 2020-07-16 at 3 18 29 PM
Screen Shot 2020-07-16 at 3 18 30 PM

### DIFF
--- a/lib/tag-list/style.scss
+++ b/lib/tag-list/style.scss
@@ -48,6 +48,7 @@
   list-style: none;
   padding-inline-start: 20px;
   overflow-y: scroll;
+  padding-right: 20px;
 }
 
 .tag-list-input {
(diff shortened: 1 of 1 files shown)


dmsnell (approved):

dmsnell (approved):
bleep bleep bloop looks good to me

PR #2211, opened 2020-07-17
Fix scroll bars on narrow line length

belcherj:
### Fix

When the narrow line length is set the scroll bar was not on the edge of the screen. This moves it back to the edge.

### Test

1. Have narrow line length set
2. Resize screen to multiple widths
3. Have wide line length set
4. Resize screen to multiple widths

### DIFF
--- a/scss/theme.scss
+++ b/scss/theme.scss
@@ -23,11 +23,23 @@
     height: 100%;
     width: 100%;
     margin: 0;
-    padding: 0;
     border: none;
     box-shadow: none;
     outline: none;
+    padding: 0 calc((100% - 768px) / 2);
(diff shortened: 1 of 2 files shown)


dmsnell (approved):

scss/theme.scss
```diff
@@ -23,12 +23,24 @@
     height: 100%;
     width: 100%;
     margin: 0;
-    padding: 0;
     border: none;
     box-shadow: none;
     outline: none;
+    padding: 0 calc((100% - 768px) / 2);
```
dmsnell, line 29: nice use of `calc()` here

PR #2212, opened 2020-07-17
Display unknown in last updated when it is unknown

belcherj:
### Fix

Display unknown in the last synced indicator when we do not know when it was last synced.

Screen Shot 2020-07-17 at 12 28 33 PM

### Test

1. Load app
2. Observe that the Last Synced displays unknown until you make an edit

### DIFF
--- a/lib/note-editor/index.tsx
+++ b/lib/note-editor/index.tsx
@@ -124,11 +124,12 @@ export class NoteEditor extends Component<Props> {
 
     const isTrashed = !!note.deleted;
 
+    const lastUpdatedDate =
+      (lastUpdated && new Date(lastUpdated).toLocaleString()) || 'Unknown';
+
     return (
       <div className="note-editor theme-color-bg theme-color-fg">
-        <div className="last-sync">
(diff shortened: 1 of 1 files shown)


lib/note-editor/index.tsx
```diff
@@ -124,11 +124,12 @@ export class NoteEditor extends Component<Props> {
 
     const isTrashed = !!note.deleted;
 
+    const lastUpdatedDate =
+      (lastUpdated && new Date(lastUpdated).toLocaleString()) || 'Unknown';
```
dmsnell: I'm sure you'll see this in hindsight, but this is structurally the ternary operator…

```ts
const lastUpdatedDate = lastUpdated
	? new Date(lastUpdated).toLocaleString()
	: 'Unknown';
```

belcherj: hahahahahahahahah, brain not working.

belcherj: Fixed

dmsnell (approved):

PR #2215, opened 2020-07-17
Update version to 2.0.0

belcherj:
### Fix

2.0.0!  This branch changes enough that it deserves a full point release.

### Test

None required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "email": "support@simplenote.com"
   },
   "productName": "Simplenote",
-  "version": "1.20.0",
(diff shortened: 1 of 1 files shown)


dmsnell (commented):
This is certainly warranted, but don't we typically update the version number in a release branch, partly to prevent merge conflicts with `develop` if we release another update to it before this merges?

belcherj:
Was thinking that we should get it updated so when we ask what version people are on it is easy to differentiate.

belcherj:
Let's do this later after we merge to develop.

PR #2219, opened 2020-07-20
Add loading state for analytics toggle

belcherj:
### Fix

This adds the ability to have a null state for a toggle in the settings dialog. This enables us to display `Loading` if we do not have the analytics enabled value.

I don't love this as it makes the settings code murkier but we have a project to clean up settings in the backlog so I think it is appropriate to address how settings work at that time.

### Test

This is hard to test until Analytics is loaded back into the app. It will always display as loading.

1. Open Settings dialog.
2. Observe that the analytics toggle does not display and that you see 'Loading'

### DIFF
--- a/lib/dialogs/settings-group.tsx
+++ b/lib/dialogs/settings-group.tsx
@@ -10,6 +10,7 @@ export const SettingsGroup = ({
   slug: groupSlug,
   activeSlug,
   description,
+  isEnabled,
   onChange,
   learnMoreURL,
   renderer,
@@ -47,7 +48,8 @@ export const SettingsGroup = ({
                 groupSlug,
                 slug,
                 title,
-                isEnabled: slug === activeSlug,
+                isEnabled:
+                  isEnabled === undefined ? slug === activeSlug : isEnabled,
                 onChange,
               })}
             </div>
--- a/lib/dialogs/toggle-settings-group.tsx
+++ b/lib/dialogs/toggle-settings-group.tsx
@@ -1,11 +1,11 @@
-import React, { FunctionComponent } from 'react';
+import React, { FunctionComponent, Fragment } from 'react';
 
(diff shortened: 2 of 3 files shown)


lib/dialogs/settings-group.tsx
```diff
@@ -47,7 +48,8 @@ export const SettingsGroup = ({
                 groupSlug,
                 slug,
                 title,
-                isEnabled: slug === activeSlug,
+                isEnabled:
+                  isEnabled === undefined ? slug === activeSlug : isEnabled,
```
dmsnell, line 52: we can use the `??` operator to pick an alternative value if `isEnabled` is `null` or `undefined`

```ts
isEnabled: isEnabled ?? slug === activeSlug,
```

lib/dialogs/toggle-settings-group.tsx
```diff
@@ -17,13 +17,19 @@ const ToggleGroup: FunctionComponent<Props> = ({
...
+      <span>Loading</span>
+    ) : (
+      <ToggleControl
+        name={groupSlug}
+        value={slug}
+        id={`settings-field-${groupSlug}-${slug}`}
+        checked={isEnabled}
+        onChange={onChange}
+      />
+    )}
```
dmsnell, line 31: interesting approach; I didn't think of this.

if we haven't loaded the setting then I think we have to presume that sharing analytics is disabled and not send them. do you have an idea how to communicate that?

alternatively we might not have a good network connection but someone should theoretically be able to set this and let Simperium sync it. if they do that we should be able to use the value they set even though it hasn't synchronized with the server.

do you think it's worth the effort to let them toggle the setting in the loading state? flipped around, do you think it's worth preventing them from toggling the setting before it loads?

dmsnell:
Closing because we need to rethink our needs for this and how we approach it.

PR #2220, opened 2020-07-20
Fixes missing ids in note export

belcherj:
### Fix

Exported notes were missing ids. This occurred because we were passing the values from the note Map. We need to pass the notes Map and iterate using entries to retain the ids.  The manipulation of the notes Map should occur as close to where it is being used. This reduces complexity upstream.

### Test

1. Have notes
2. Export them
3. Look in the source folder
4. In the json file
5. Do notes have ids?


dmsnell (approved):
Thanks!

PR #2222, opened 2020-07-21
[rewrite branch] Make condensed style actually condensed

codebykat:
### Fix

The note list condensed style was just removing the preview text but not actually condensing anything. Now it does.

I couldn't easily figure out how to make it less than 49px because `CellMeasurer` hates me, but this is an improvement anyway.

### Before

Screen Shot 2020-07-20 at 11 36 12 PM

### After

Screen Shot 2020-07-20 at 11 32 44 PM

Screen Shot 2020-07-20 at 11 32 06 PM

### DIFF
--- a/lib/note-list/style.scss
+++ b/lib/note-list/style.scss
@@ -21,6 +21,13 @@
     white-space: nowrap;
   }
 
+  .condensed .note-list-item {
+    min-height: 16px;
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
Awesome! Thanks

PR #2223, opened 2020-07-22
[rewrite branch] Add analytics as middleware

codebykat:
### Fix

This is an updated version of #2001 for the rewrite branch!

### TODO

* Figure out why analytics needs to be toggled twice to enable it SEE #2219 

### Test

Add a console.log in `recordEvent` and make sure the correct events are firing in response to actions taken throughout the app.

### DIFF
--- a/lib/dialogs/share/index.tsx
+++ b/lib/dialogs/share/index.tsx
@@ -3,7 +3,6 @@ import { connect } from 'react-redux';
 import { includes, isEmpty } from 'lodash';
 import MD5 from 'md5.js';
 
-import analytics from '../../analytics';
 import ClipboardButton from '../../components/clipboard-button';
 import isEmailTag from '../../utils/is-email-tag';
 import Dialog from '../../dialog';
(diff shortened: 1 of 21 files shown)


lib/dialogs/share/index.tsx
```diff
@@ -51,7 +52,7 @@ export class ShareDialog extends Component<Props> {
 
     if (collaborator !== '' && tags.indexOf(collaborator) === -1 && !isSelf) {
       this.props.editNote(noteId, { tags: [...tags, collaborator] });
-      analytics.tracks.recordEvent('editor_note_collaborator_added');
+      this.props.recordEvent('editor_note_collaborator_added');
```
dmsnell: this would be an ideal place to wrap the `editNote` action with analytics middleware so that we can avoid a double-dispatch. by adding it here we make two quick state updates in succession but if we had `{ meta: { analytics: [ [ 'editor_note_collaborator_added' ] ] } }` then we could keep it to one dispatch (which mirrors the reality of all that's changing in the app) and fire the analytics in the middleware layer

```ts
const analyticsMiddleware = store => next => action => {
	if (action.meta?.analytics?.length) {
		action.meta.analytics.forEach( ( [ name, properties ] ) => analytics.tracks.recordEvent( name, properties ) )
	}

	return next(action);
}

const withEvent = (name, properties) => action => ( {
	...action,
	meta: {
		...action.meta,
		analytics: [
			...action.meta.analytics,
			[name, properties]
		]
	}
} )

const recordEvent = (name, properties) => 
	withEvent(name, properties)( {
		type: 'RECORD_EVENT'
	} )
```

no reducer need listen for `RECORD_EVENT` - it can be a hollow event purely to carry the analytics meta for those analytics events which do not jump onto other actions

```ts
const mapDispatchToProps = {
	recordEvent,
	addCollaborator: (noteId, tags, collaborator) =>
		withEvent( 'editor_note_collaborator_added' )( editNote( noteId, { tags: [ ...tags, collaborator ] } ) )
}
```

makes me think we could do well to add a `ADD_COLLABORATOR` action type

PR #2224, opened 2020-07-22
Update the editor width css to allow narrow/wide views

belcherj:
### Fix

Added css to properly display the editor in wide and narrow views.

### Test

1. Observe editor, does it look good?
2. Change your browser width, does it look good?
3. Switch the line length setting, does it look good?


dmsnell (approved):
Didn't test. Looks harmless.

PR #2225, opened 2020-07-22
Checks for Notification on window

belcherj:
### Fix

iOS and IE break when Notification is not available.

### Test

1. Load the app on iOS
2. Does it work?

### DIFF
--- a/lib/state/settings/reducer.ts
+++ b/lib/state/settings/reducer.ts
@@ -90,17 +90,17 @@ const noteDisplay: A.Reducer<T.ListDisplayMode> = (state = 'comfy', action) => {
 };
 
 const sendNotifications: A.Reducer<boolean> = (
-  state = window?.Notification?.permission === 'granted',
+  state = window.Notification?.permission === 'granted',
   action
 ) => {
   switch (action.type) {
     case 'REQUEST_NOTIFICATIONS':
       return action.sendNotifications
-        ? window?.Notification?.permission === 'granted'
+        ? window.Notification?.permission === 'granted'
         : false;
 
     default:
-      return state && window?.Notification?.permission === 'granted';
+      return state && window.Notification?.permission === 'granted';
   }
 };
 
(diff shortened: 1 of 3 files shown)


dmsnell (commented):
we want to remove the toggle from the settings when they aren't supported as well.

lib/state/settings/reducer.ts
```diff
@@ -90,17 +90,21 @@ const noteDisplay: A.Reducer<T.ListDisplayMode> = (state = 'comfy', action) => {
 };
 
 const sendNotifications: A.Reducer<boolean> = (
-  state = Notification.permission === 'granted',
+  state = 'Notification' in window && Notification.permission === 'granted',
```
dmsnell: should be possible to use optional chaining here, though maybe it has trouble given the top-level naming?

```ts
state = Notification?.permission === 'granted'
```

if it does complain then it should be a quick resolution, add `window`

```ts
state = window?.Notification?.permission === 'granted'
```

dmsnell (commented):
the approach I had planned on taking was more holistic than adding some exceptions to the logic. we shouldn't have much trouble (apart from the settings dialog) in removing the `sendNotifications` reducer/middleware/etc… entirely if indeed it's not supported

in fact while maybe the reducer is fine/safe to leave in we can at least only conditionally-load the `functions/…` file that does anything with them so that unsupported platforms won't even download the code to send them

lib/state/settings/reducer.ts
belcherj: I tried this `state = Notification?.permission === 'granted'` didn't work. Can't believe that I didn't try `window?.Notifications` my brain is not working today.

dmsnell: that's an easy one. top-level bindings can be awkward. I think the difference is that property access is different than binding access. this is part of what makes JS ugly, because you can access things that don't exist or don't exist in the current scope.

PR #2226, opened 2020-07-23
Fix: Ability to log in with mixed case email address

### LINKED ISSUE #2193: Auto logging out at first start. Solution: use only lower case on email field
### Expected
After first start and log in normal work of application with viewing,editing tasks,etc...

### Observed
I'm download latest AppImage version and starting application.Them I'm log in but after 1-2 seconds this app automated logging out and wish to log in again

***(Optional)***
Before log in
изображение
Apps is logging in
изображение
Apps log out
изображение

***(Optional)*** 
-|-|-|-
AppImage|Dell Inspiron 5000 Series|Linux Mint 20 Cinnamon|1.19.0

codebykat:
### Fix

This issue only arises in the Electron apps because App Engine authentication lowercases these arguments before creating or logging in to an account.

I also added a trim to match the App Engine logic.

Fixes #2193 

### Test

1. On the standalone Electron build (**not on web**)
2. Log in with any capital letters in your email address
3. You should still be able to log in
4. Ditto for sign up

### Release

Not updated: Fixed a bug that prevented login if capital letters were typed in the email address

### DIFF
--- a/lib/boot-without-auth.tsx
+++ b/lib/boot-without-auth.tsx
@@ -35,12 +35,12 @@ class AppWithoutAuth extends Component<Props, State> {
     authStatus: 'unsubmitted',
   };
 
-  authenticate = (username: string, password: string) => {
+  authenticate = (usernameArg: string, password: string) => {
+    const username = usernameArg.trim().toLowerCase();
+
     if (!(username && password)) {
       return;
     }
-
-    username = username.trim().toLowerCase();
 
     this.setState({ authStatus: 'submitting' }, () => {
       auth
@@ -68,12 +68,11 @@ class AppWithoutAuth extends Component<Props, State> {
     });
   };
 
-  createUser = (username: string, password: string) => {
+  createUser = (usernameArg: string, password: string) => {
+    const username = usernameArg.trim().toLowerCase();
     if (!(username && password)) {
       return;
     }
-
-    username = username.trim().toLowerCase();
 
     this.setState({ authStatus: 'submitting' }, () => {
       auth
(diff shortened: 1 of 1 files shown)


lib/boot-without-auth.tsx
```diff
@@ -40,6 +40,8 @@ class AppWithoutAuth extends Component<Props, State> {
       return;
     }
 
+    username = username.trim().toLowerCase();
```
dmsnell: this is probably going to sound really nit-picky but shouldn't have any cases any more of this kind of mutation. would you mind splitting this into an _immutable_ change so we don't have "the value of `username` above this arbitrary line" and also "the value of `username` below this arbitrary line?"

```ts
authenticate = (givenUsername: string, password: string) => {
	…
	const username = givenUsername.trim().toLowerCase();
```

…or skip the update and only transform the username where it matters, during validation…

```ts
.authorize(username.trim().toLowerCase(), password)
.then(…this.props.onAuth(user.access_token, username, false))
```

lib/boot-without-auth.tsx
```diff
@@ -71,6 +73,8 @@ class AppWithoutAuth extends Component<Props, State> {
...
```
dmsnell: same as above comment…

dmsnell (commented):
Really like this change but it's too bad there was the discrepancy. Too bad we lower-case the email in the backend but I don't suppose we can change that.

dmsnell (approved):
Thanks @codebykat - at some point we should be able to do as you suggest and move this to the backend, I think/hope

Issue #2137, opened 2020-06-05
When logged out most menu items don't work (About, theme, etc)


rachelmcr:
### Expected
I expect that when the Simplenote > About Simplenote menu item is enabled, selecting it will open an about window like this (the about window that opens while logged in):

![Screenshot 2020-06-05 14 28 19](https://user-images.githubusercontent.com/8658164/83881946-50469980-a739-11ea-98fc-246377f7636b.png)

### Observed
Even though that menu item is enabled when I'm logged out, selecting it does nothing. No information about the app is available. (This is a little annoying when I am logged out and want to check what version of the app I'm using, since I have to log in first.)

### Reproduced

1. Log out of the app.
2. Go to the Simplenote system menu.
3. Select "About Simplenote."

Result: Nothing happens.

Make|Model|OS Version|App Version
-|-|-|-
Mac|Macbook Pro|macOS 10.15.4|1.18.0-beta1

codebykat:
Oh yeesh. Yeah actually none of the menu items work while logged out (including changing the theme to dark mode). Stuff like "Insert Checklist" is still there. Sort options. They also seem to get cleared once you log in.

belcherj:
I think they should all be greyed out while logged out. This would require a large refactor to allow for dark mode switches while logged out.

I do think the about page should be accessible while logged out.

codebykat:
Closed in #2228, #2231, #2232

PR #2228, opened 2020-07-23
Disable most menu items if user isn't logged in

codebykat:
### Fix

Most menu items aren't applicable when logged out, and don't work anyhow, so let's hide them.

We could also gray them out instead of remove them entirely; anyone have a preference?

### Test

1. Open the app, signed out
2. Note that irrelevant menu options do not appear (Format menu, Preferences, etc)
3. Log in
4. Note that all menu items are now visible and work as intended
5. Log out again
6. Note that menu items are once again disabled

### Release

Not updated: Disable inapplicable menu items when logged out

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -91,7 +91,7 @@ module.exports = function main() {
 
     ipcMain.on('settingsUpdate', function (event, settings) {
       Menu.setApplicationMenu(
-        Menu.buildFromTemplate(createMenuTemplate(settings))
+        Menu.buildFromTemplate(createMenuTemplate(settings), mainWindow)
       );
     });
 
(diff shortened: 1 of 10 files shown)


belcherj:
I think we should also remove:
- File -> Print
- View:
  - Zooms
  - Theme

I am going to approve as this works great and code looks good.

belcherj (approved):
Ship

codebykat:
I removed Print (I suppose there's no valid reason to want to print the login screen).

I also removed Zoom and Theme; I feel like these should work on the login screen, but at the moment they don't, so let's remove them for now. Ditto for the spellcheck toggle; maybe it should work, but at the moment, it doesn't, so out it goes.

The other big elephant in the room is About, which I'm going to try to get working in a follow-up PR.

codebykat:
Just to note if we'd rather gray these out instead of remove them it's easily enough done by using `enabled` instead of `visible`.

https://github.com/Automattic/simplenote-electron/issues/2137#issuecomment-643301865

PR #2229, opened 2020-07-23
Fix: Table alignment and strikethrough in Markdown preview

codebykat:
### Fix

We were sanitizing the `text-aligns` into oblivion, so table alignment wasn't working properly. These changes mirror the ones we needed to make for published notes on App Engine.

~I also added the emoji module here, but in a comment, because I want to remind myself that it exists if we ever want to support those!~

### Test

Markdown example:

```| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| **col 3 is**  | right-aligned | $1600 |
| col 2 is      | *centered*    |   $12 |
| zebra stripes | ~~are neat~~  |    $1 |
```

Should appear like:

Screen Shot 2020-07-23 at 4 49 09 PM

### Release

Not updated: Support table alignment and strikethrough in Markdown preview

### DIFF
--- a/lib/utils/render-note-to-html.ts
+++ b/lib/utils/render-note-to-html.ts
@@ -17,6 +17,9 @@ export const renderNoteToHtml = (content: string) => {
       markdownConverter.setOption('simpleLineBreaks', false); // override GFM
       markdownConverter.setOption('ghMentions', false);
       markdownConverter.setOption('smoothLivePreview', true);
+      markdownConverter.setOption('tables', true); // table syntax
+      markdownConverter.setOption('strikethrough', true); // ~~strikethrough~~
+      // markdownConverter.setOption('emoji', true); // emoji support like :smile:
 
       const withNormalizedBullets = content.replace(
         /([ \t\u2000-\u200a]*)\u2022(\s)/gm,
--- a/lib/utils/render-note-to-html.ts
+++ b/lib/utils/render-note-to-html.ts
@@ -19,7 +19,6 @@ export const renderNoteToHtml = (content: string) => {
       markdownConverter.setOption('smoothLivePreview', true);
       markdownConverter.setOption('tables', true); // table syntax
       markdownConverter.setOption('strikethrough', true); // ~~strikethrough~~
-      // markdownConverter.setOption('emoji', true); // emoji support like :smile:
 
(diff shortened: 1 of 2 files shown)


lib/utils/sanitize-html.ts
```diff
@@ -138,6 +138,22 @@ const isAllowedAttr = (tagName: string, attrName: string, value: string) => {
...
+    case 'td':
+      switch (attrName) {
+        case 'style':
+          switch (value) {
+            case 'text-align:center;':
+            case 'text-align:left;':
+            case 'text-align:right;':
+              return true;
+            default:
+              return false;
```
dmsnell, line 151: nice special-casing here. it's much better than allowing _any_ `style` value

lib/utils/render-note-to-html.ts
```diff
@@ -17,6 +17,9 @@ export const renderNoteToHtml = (content: string) => {
       markdownConverter.setOption('simpleLineBreaks', false); // override GFM
       markdownConverter.setOption('ghMentions', false);
       markdownConverter.setOption('smoothLivePreview', true);
+      markdownConverter.setOption('tables', true); // table syntax
+      markdownConverter.setOption('strikethrough', true); // ~~strikethrough~~
+      // markdownConverter.setOption('emoji', true); // emoji support like :smile:
```
dmsnell: I'm assuming this isn't supposed to be here?

it's not something I would want to support just because of how inconsistent and undocumented it would seem (at least at this time).

dmsnell (approved):

lib/utils/render-note-to-html.ts
codebykat: > I also added the emoji module here, but in a comment, because I want to remind myself that it exists if we ever want to support those!

But okay OKAY fine I'll take it out and make this note to myself somewhere else

dmsnell: oh I missed that in the description, sorry!

it's fine to leave it in but maybe instead explain that _in the comment_

```ts
// markdownConverter.setOption('emoji', true); // earmarked to remind us that this exists in case we decide to support it in the future
```

probably we don't need to earmark it because if we want to add emoji it won't be hard to find, but I don't mind commented out code if there's an explained reason for it

PR #2230, opened 2020-07-24
Fixes unsync'd notes dialog on Electron log out

belcherj:
### Fix

Fixes the unsync'd notes dialog to use the same thing as web version.

### Test

1. Have electron open
2. Turn off proxy or wifi
3. Modify a note
4. Try to log out
5. Get dialog
6. Cancel
7. Try to log out again
8. Get dialog and confirm
9. Do you log out?


codebykat (approved):
Works great 👍

PR #2231, opened 2020-07-25
Fix: About dialog while logged out

codebykat:
### Fix

The About dialog should be accessible whether logged in or logged out.

Penultimate fix for #2137 (complementary to #2228 ); still need to fix the keybindings one way or the other (either show the dialog now that I've figured out how to do it, or hide the menu item when logged out)

### Release

Not updated: Restore ability to view About while logged out

### DIFF
--- a/lib/boot-without-auth.tsx
+++ b/lib/boot-without-auth.tsx
@@ -46,6 +46,10 @@ class AppWithoutAuth extends Component<Props, State> {
     window.electron?.receive('appCommand', this.onAppCommand);
   }
 
+  componentWillUnmount() {
+    window.electron?.removeListener('appCommand');
+  }
+
   onAppCommand = (event) => {
     if ('showDialog' === event.action && 'ABOUT' === event.dialog) {
       this.setState({ showAbout: true });
--- a/lib/boot-without-auth.tsx
+++ b/lib/boot-without-auth.tsx
@@ -1,9 +1,14 @@
 import React, { Component } from 'react';
+import { createStore } from 'redux';
+import { Provider } from 'react-redux';
 import { render } from 'react-dom';
 import { Auth as AuthApp } from './auth';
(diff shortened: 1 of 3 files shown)


codebykat:
~I look forward to @dmsnell telling me what crimes I have committed against React~

Update: Committing fewer crimes now, still kind of sad that it's back to prop-drilling

lib/boot-without-auth.tsx
```diff
@@ -33,6 +37,25 @@ const auth = new SimperiumAuth(config.app_id, config.app_key);
...
     authStatus: 'unsubmitted',
+    showAbout: false,
+  };
+
+  componentDidMount() {
+    window.electron?.receive('appCommand', this.onAppCommand);
+  }
+
+  componentWillUnmount() {
+    window.electron?.removeListener('appCommand');
```
belcherj, line 48: We need to be careful with this one since it removes all 'appCommand' listeners. In this case, it is correct.

belcherj (approved):
This is great, TY!

lib/dialogs/about/index.tsx
```diff
@@ -134,8 +131,4 @@ export class AboutDialog extends Component<Props> {
   }
 }
 
-const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
-  closeDialog,
-};
-
-export default connect(null, mapDispatchToProps)(AboutDialog);
+export default AboutDialog;
```
dmsnell, line 134: we can also pass in an optional `closeDialog` through `OwnProps` and prefer that to the connected props, leaving the normal connected interface

```ts
type OwnProps = {
	closeDialog?: () => any;
}

const mapDispatchToProps: S.MapDispatch<DispatchProps, OwnProps> = (dispatch, props) => ({
	closeDialog: () => dispatch( props.closeDialog?.() ?? closeDialog() ),
})
```

PR #2232, opened 2020-07-25
Fix: disable keyboard shortcuts menu when logged out

codebykat:
### Fix

Missed this one in #2228. Keyboard shortcuts dialog should only be visible if user is logged in.

Last fix for https://github.com/Automattic/simplenote-electron/issues/2137


belcherj (approved):
Ship it

PR #2234, opened 2020-07-27
Fixes styling of offline badge

belcherj:
### Fix

The offline badge line height was not right and it displayed incorrectly in Electron.

Screen Shot 2020-07-27 at 10 57 37 AM

### Test

1. Go offline
2. View on web
3. View on Electron
4. Does it look good?


dmsnell (approved):
I blindly trust all CSS patches; we can adjust if we catch a problem

PR #2236, opened 2020-07-27
Hides last syncd message when it is unknown

belcherj:
### Fix

Hides the last syncd when it is unknown

Screen Shot 2020-07-27 at 11 08 37 AM
Screen Shot 2020-07-27 at 11 08 44 AM

### Test

1. Open app
2. Make change observe last sync'd
3. Open another not that did not have changes
4. Should not see last sync'd
5. Make change
6. Last sync'd should appear after sync

@SylvesterWilmott should we keep the padding so when last sync'd appears the note doesn't shift down?

### DIFF
--- a/lib/note-editor/index.tsx
+++ b/lib/note-editor/index.tsx
@@ -124,13 +124,13 @@ export class NoteEditor extends Component<Props> {
 
     const isTrashed = !!note.deleted;
 
-    const lastUpdatedDate = lastUpdated
-      ? new Date(lastUpdated).toLocaleString()
-      : 'Unknown';
-
     return (
       <div className="note-editor theme-color-bg theme-color-fg">
-        <div className="last-sync">Last synced: {lastUpdatedDate}</div>
+        <div className="last-sync">
+          {lastUpdated && (
+            <span>Last synced: {new Date(lastUpdated).toLocaleString()}</span>
+          )}
+        </div>
         {editMode || !note.systemTags.includes('markdown') ? (
           <NoteDetail
             storeFocusEditor={this.storeFocusEditor}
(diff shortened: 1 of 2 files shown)


lib/note-editor/index.tsx
```diff
@@ -124,13 +124,13 @@ export class NoteEditor extends Component<Props> {
...
     const isTrashed = !!note.deleted;
 
-    const lastUpdatedDate = lastUpdated
-      ? new Date(lastUpdated).toLocaleString()
-      : 'Unknown';
-
     return (
       <div className="note-editor theme-color-bg theme-color-fg">
-        <div className="last-sync">Last synced: {lastUpdatedDate}</div>
+        <div className="last-sync">
```
belcherj, line 129: I left the div so we would keep some of the padding at the top of the note.

dmsnell: Does the note text bounce when we go from unknown to known?

Also, I feel like we can eventually make sure the status is always known, so this should be moot. Leaving it in could be a decent reminder that our work is unfinished.

The effect I'm after with the bounce is the difference between not rendering this vs. adding `opacity: 0` or something like that so it takes the same vertical space but isn't visible.

I did not test to find out - trying to get through the PR queue…

SylvesterWilmott: The "bounce" would be the decider for me. Ideally, we would hide the div entirely when it's unknown but more importantly, we don't want notes jumping around.

belcherj:
lastupdated

belcherj:
Jul-30-2020 10-58-02

SylvesterWilmott:
Yeah, it's pretty clear now. Setting opacity to 0 is better than hiding the div entirely. Thanks @belcherj

PR #2237, opened 2020-07-27
Fixes editor width when narrow line length is selected

belcherj:
### Fix

My previous fix for the editor width only worked on the left margin. This fixes it so the editor width is constrained on both sides. It does however cause the scroll bar to be just inside the narrow view and not on the edge of the screen, see screenshot. I am not sure how to fix it to push it to the side.

This PR also fixes the jerkiness of the note when it opens. The padding applied to the content before the editor is ready is now the same as it is once Monaco renders.

Screen Shot 2020-07-27 at 1 03 48 PM

Known issue, when you switch line length the editor does not update and requires you to resize the browser. To be handled in another PR.

### Test

1. Have a note with a long line length
2. Switch to a narrow line length
3. Resize screen
4. Observe note

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -294,7 +294,7 @@ class NoteContentEditor extends Component<Props> {
         }`}
       >
         {editor === 'fast' ? (
-          <div style={{ padding: '0.7em', whiteSpace: 'pre-wrap' }}>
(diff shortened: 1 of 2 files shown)


dmsnell (approved):
> This PR also fixes the jerkiness of the note when it opens.

❤️

PR #2239, opened 2020-07-27
Allow scroll bar to be any height fixing it only scrolling partially to the bottom

belcherj:
### Fix

We had a fixed height on the scroll bar. This caused it to look like it hadn't scrolled all the way to the bottom. This removes that max height so that the scroll bar displays correctly.

Screen Shot 2020-07-27 at 1 00 08 PM
Screen Shot 2020-07-27 at 12 59 57 PM

### Test

1. Open long note and scroll


dmsnell (approved):
:shipit:

PR #2247, opened 2020-07-28
Make monaco rerender after linelength changes

belcherj:
### Fix

After changing the line length in settings Monaco does not resize.

This is probably not the right way to do it but it works. Any thoughts @dmsnell?

### Test

1. Change lineLength in settings
2. Observe Monaco editor resizes to new width


lib/note-content-editor.tsx
```diff
@@ -132,6 +133,11 @@ class NoteContentEditor extends Component<Props> {
         }
       }, SPEED_DELAY);
     }
+    if (prevProps.lineLength !== this.props.lineLength) {
+      setTimeout(() => {
+        window?.editor.layout();
+      }, 400);
+    }
```
dmsnell, line 140: why the timeout?

belcherj: Doesn't work without the timeout. I think it happens too soon for some reason. Not sure what is going on there but I figured I'd get what I had working up before going deeper.

lib/note-content-editor.tsx
```diff
@@ -132,6 +133,11 @@ class NoteContentEditor extends Component<Props> {
         }
       }, SPEED_DELAY);
     }
+    if (prevProps.lineLength !== this.props.lineLength) {
+      setTimeout(() => {
+        window?.editor.layout();
```
dmsnell: should be `window.editor?.layout()`

lib/note-content-editor.tsx
dmsnell: did you try `window.editor?.updateOptions({ lineHeight: this.props.lineLength })`

lib/note-content-editor.tsx
belcherj: If the editor exists on window layout will exist. This checks to make sure editor is there.

lib/note-content-editor.tsx
belcherj: That does not work either without the timeout.

lib/note-content-editor.tsx
belcherj: Updated to: window?.editor?.layout();

dmsnell: you don't need `window?.` because `window` will always exist, hence `window.editor?.` not `window?.editor`

`window?.editor.layout()` says "if indeed `window` exists then I assert that `editor` will always exist on it and `layout()` will be a method on `editor`"

`window.editor?.layout()` says "if indeed `editor` exists on `window` then I assert that `layout()` will be a method on it"

lib/note-content-editor.tsx
dmsnell: strange. I'll try and examine this too. something smells fishy about needing that timeout and I want to make sure we don't overlook something important

lib/note-content-editor.tsx
belcherj: 🤦 yeah...

dmsnell:
Hm. while trying to test this I wasn't able to properly understand what's broken. I can run `dispatch({type: 'setLineLength', lineLength: 'narrow'})` (or also with `full`) and watch the editor resize. this is confirmed in the `rewrite/no-races` branch

adjustingWidth mov

Am I missing something?

belcherj:
You are not starting from full line length. Notice the width in your gif does not change.


Aug-04-2020 13-36-44

dmsnell:
>You are not starting from full line length

I'll re-test but I was moving between the narrow and full-width variants. The note I have in there I think had hard word wraps which is why it probably appeared to not change.

belcherj:
@dmsnell Notice in your gif how the line length does not change. Make the line length full then resize the browser or reload.

dmsnell:
Somehow I botched a rebase or this got out of sync. I've recreated it in https://github.com/Automattic/simplenote-electron/pull/2148/commits/ffed150d548b0017e7b3d6936d1f9e6972f1a539 so I'm closing this PR.

PR #2248, opened 2020-07-28
Clean up disable restore note button when is newset revision

belcherj:
### Fix

The disable on the restore note button when you are currently on the most recent revision is CSS and was not actually disabling the functionality. Additionally, the logic failed when you first opened the revisions panel. This gets everything into tip top.

### Test

1. Open the revisions panel for a note.
2. Is button disabled?
3. Move revisions slider around
4. Can you restore a revision?
5. When you slide the slider all the way to the right does it become disabled again?

### DIFF
--- a/lib/revision-selector/index.tsx
+++ b/lib/revision-selector/index.tsx
@@ -36,7 +36,6 @@ export class RevisionSelector extends Component<Props> {
 
   onAcceptRevision = () => {
     const { noteId, openedRevision, restoreRevision } = this.props;
-
     restoreRevision(noteId, openedRevision);
   };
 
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
There are some deeper issues here if we have unsynchronized changes in the buffer. I have let them slide for now but if we have a bucket value different than the ghost we really need an additional option in the slider which is "Latest local copy" upon which selecting it cancels/closes the revision selector.

A while ago I tried to fix it but stopped because it was distracting me from other priorities.

Suppose we have 13 revisions:
 - if the note is fully synchronized we should have positions at 1 - 13
 - if not sync'd we should have positions 1 - 14 where 14 says "Latest" and is tantamount to not having the revision slider open

belcherj:
Yikes, that does sound like fun. Will have to revisit this.

PR #2250, opened 2020-07-29
Fix logged-out Format menu on Windows

codebykat:
### Fix

Apparently Windows doesn't respect the `visible` option on a menu, so let's just not return it unless we're logged in. Follow-up to #2228 

### Test
1. Open app logged-out
2. Verify Format menu is not shown
3. Log in
4. Verify Format menu is shown


belcherj (approved):
Ship

PR #2256, opened 2020-08-03
Disable remote module in electron

belcherj:
### Fix

The remote module in Electron carries security issues. We don't use it so why not disable it. https://github.com/doyensec/electronegativity/wiki/REMOTE_MODULE_JS_CHECK

### Test
1. Open app
2. Click around
3. Try an Evernote import

### Release

Security: Disabled Electron remote module


codebykat (approved):
Took this for a spin, seems to work fine 👍

PR #2260, opened 2020-08-05
Fixes the polyfill for Promise to make IE work

belcherj:
### Fix

The polyfils need to be moved up to boot so that they exist everywhere for Internet Explorer. We should look later at removing them when not necessary in browsers.

### Test
1. Open this branch in IE.
2. Login
3. Do you see the app?

### DIFF
--- a/lib/boot-with-auth.tsx
+++ b/lib/boot-with-auth.tsx
@@ -4,9 +4,6 @@ if (__TEST__) {
   window.testEvents = [];
 }
 
-import 'core-js/stable';
-import 'regenerator-runtime/runtime';
-
 import React from 'react';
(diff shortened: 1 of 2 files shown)


dmsnell (approved):
> We should look later at removing them when not necessary in browsers.

Agreed. That's on my lower-priority list after this merges and is safe. We should be able to do this through the Babel config and look into creating two builds as Calypso does with it's "evergreen" build. When the time comes we can poke @sgomes who I think made that possible.

sgomes:
Thanks for looping me in, @dmsnell! I'm more than happy to help pick out and implement a differential build serving approach once you feel the time is right 👍

PR #2262, opened 2020-08-05
Add/sentry

belcherj:
### Fix

Adds Sentry to app so that we can track and fix errors

### Test
1. Build this branch
2. Throw a JS error
3. Ping JB to make sure it got logged

### DIFF
--- a/lib/boot.ts
+++ b/lib/boot.ts
@@ -6,12 +6,6 @@ import analytics from './analytics';
 import getConfig from '../get-config';
 import { boot as bootWithoutAuth } from './boot-without-auth';
 import { boot as bootLoggingOut } from './logging-out';
-import * as Sentry from '@sentry/react';
-
-Sentry.init({
-  dsn:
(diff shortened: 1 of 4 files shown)


dmsnell:
Not sure how big the `Sentry` dependency is. We could measure it which would be helpful. If it's hefty then given the fact that it's toggled on after boot and not going to be needed in many cases I think it might be a trivial optimization to lazy-load it…

```ts
if (action.allowAnalytics) {
	import(/* webpackChunkName: 'sentry' */ '@sentry/react').then( Sentry => Sentry.init( … ) )
```

that being said while the code seems fine it'd be nice to discuss the broader implications of this, which is that we're making a big change to how we handle privacy because this change might start sending very specific private information to a third-party service during app-crash scenarios. I guess the mobile apps already use this service so that's a fair counterpoint.

currently I don't believe any information in this app goes to any third party even with WordPress.com login (since that's a first-party system in our control).

belcherj:
Conditionally loads the module and only runs in production.

Issue #1410, opened 2019-06-26
Log Out button is hard to find


rachelmcr:
### What

The Log Out button is hard to find, under Simplenote > Preferences.

![image](https://user-images.githubusercontent.com/8658164/60176795-3f926180-980f-11e9-8149-c6c22d775ea9.png)

### How

There are two points of confusion that I see:

* It isn't necessarily clear that the app Preferences contain Account information and logout.
* The logout button is placed at the bottom of the screen, separate from the information about what account you're currently logged in to.

h/t @msilbers for the original report!

rachelmcr:
This came up again in internal beta testing (version 1.7.0-beta1, tested on Ubuntu 19.04, internal ref: p5T066-Xr-p2). h/t @aicee for the feedback!

thehenrybyrd:
I was poking around Simplenote for some troubleshooting and getting familiar with it, and realized I had _no idea_ how to log out. 10 minutes later someone pointed out this button - I hadn't seen it at all in my clicking through the Settings tabs. 😮

thehenrybyrd:
User report in #2273353-zen

> After the latest update on web, I can't find the logout button!!

I'm assuming the new webapp has the logout button in a different place than the old one.

thehenrybyrd:
User report in #2273353-zen

> After the latest update on web, I can't find the logout button!!

I'm assuming the new webapp has the logout button in a different place than the old one.

thehenrybyrd:
Here's another in #2273429-zen

> How do I sign out from the web app? I don't see my profile icon on the top right corner which used to have the sign out option, nor I found no other option to click on sign out.

PR #2263, opened 2020-08-05
Harmonize menu bar and settings in sidebar

dmsnell:
The web app has settings/keyboard shortcuts/about items in the navigation bar but those are absent from Electron.
Electron already provides access to the settings through the OS menu but it's confusing to jump between the web and
the electron app to discover those items are missing in one place.

In this PR we stop hiding the settings and other items in the navigation sidebar when running in Electron.

The Electron OS menu has also been missing "Empty Trash" and "Sign Out" as the native apps have. "Sign Out" is particularly
convenient to have there and its absence makes logging-out particularly inconvenient.

As a result of the overall state refactor in the app we're adding those menu items to make it more convenient again.

**Before**

Screen Shot 2020-08-05 at 6 11 56 PM
Screen Shot 2020-08-05 at 6 12 00 PM

**After**

Screen Shot 2020-08-05 at 6 11 20 PM
Screen Shot 2020-08-05 at 6 11 29 PM

### DIFF
--- a/desktop/menus/mac-app-menu.js
+++ b/desktop/menus/mac-app-menu.js
@@ -21,6 +21,13 @@ const buildMacAppMenu = (isAuthenticated) => {
     { role: 'hide' },
     { role: 'hideothers' },
     { role: 'unhide' },
+    ...(isAuthenticated
+      ? [
+          { type: 'separator' },
+          menuItems.emptyTrash(isAuthenticated),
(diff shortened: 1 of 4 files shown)


codebykat (approved):
This is great! The logout button particularly is a longstanding request #1410

dmsnell:
@belcherj I'm going to merge this. we can revert it no problem if you have reservations

PR #2271, opened 2020-08-11
[Rewrite Branch] Fix printing of notes

belcherj:
### Fix

This PR adds a display none div that contains the content of a note. When the user prints it displays that div and hides other content on the page.

Adding a display none div with the content allows us to do this purely with CSS rather than watching for print events and potentially running into race conditions with React rendering.

### Test
1. Have a note
2. Print it
3. Save to pdf
4. Do this in Safari, Chrome, and Firefox

### DIFF
--- a/scss/print.scss
+++ b/scss/print.scss
@@ -41,7 +41,7 @@
     }
   }
 
-  .note-content-print {
+  .note-content-plaintext {
     left: 0px;
     top: 0px;
     position: absolute;
--- a/scss/print.scss
+++ b/scss/print.scss
@@ -1,51 +1,51 @@
-.theme-dark .theme-color-bg {
-  background-color: #ffffff;
-  padding-top: 20px;
-}
+@media print {
+  .theme-dark .theme-color-bg {
+    background-color: #ffffff;
+  }
 
-.app-layout__source-column {
-  display: none;
-}
+  .app-layout__note-column {
+    border: 0;
+  }
 
-.app-layout__note-column {
-  border: 0;
-}
+  .dev-badge,
+  .app-layout__source-column,
+  .tag-field,
+  .note-toolbar-wrapper,
+  .last-sync,
+  .react-monaco-editor-container {
+    display: none;
+  }
 
-.dev-badge {
-  display: none;
-}
+  /* Firefox only prints the first page if any divs are display: flex */
+  .app,
+  .simplenote-app,
+  .app-layout,
+  .app-layout__note-column,
+  .note-editor,
+  .note-detail-wrapper,
+  .note-detail,
+  .note-detail-textarea,
+  .note-content-editor-shell,
+  .note-detail-wrapper {
+    overflow: visible;
+  }
 
-.tag-field,
-.note-toolbar-wrapper,
-.last-sync {
-  display: none;
-}
-
-/* Firefox only prints the first page if any divs are display: flex */
-.app,
-.simplenote-app,
-.app-layout,
-.app-layout__note-column,
-.note-editor,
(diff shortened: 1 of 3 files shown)


codebykat:
This is working well on OSX/Firefox in light and dark mode. Works okay on Electron, but there's an odd visual glitch about half the time where it looks like the note briefly becomes wider? Horizontal scrollbar shows up briefly and the contents resize. Anyone else seeing it?

scss/theme.scss
```diff
@@ -41,6 +41,16 @@
...
 
+.note-content-print {
+  display: none;
+}
+
+.note-content-fast {
+  display: inherit;
+  padding: 0 10px;
+  white-space: pre-wrap;
+}
```
dmsnell, line 52: did you try the inverse of this situation? hide the `div` by default, add visibility when in fast mode, and on the print stylesheet?

I think we should be able to eliminate this `.note-content-print` class altogether.
that is, something like this…

```ts
 <div className={`plaintext-note-content ${
		editor === 'fast' ? 'visible' : 'hidden'
	}`}
>
```

```css
.plaintext-note-content {
	display: none;

	&.visible {
		display: inherit;
	}
}

@media print {
	.plaintext-note-content.hidden {
		display: inherit; // or "block" or whatever is required
	}
}
```

the normal caveat applies: I have no idea what I'm doing with styling. just thought this model might feel simpler if it works and in the rare chance that you didn't try it.

belcherj: This is hiding by default: https://github.com/Automattic/simplenote-electron/pull/2271/files#diff-b4aef3f86a137f3edaf5a4e2fbb2a23fR44

Then I add .note-content-fast to make it visible. I can certainly shorten to .visible though.

dmsnell: >This is hiding by default

yeah sorry for being confusing in my wording. I understand that the behavior is to be hidden by default but I was referring to the nomenclature and usage of the CSS classes. we have `note-content-print` as the CSS `class` even when we're not printing and then the only CSS properties for that are `display: none` which again reads a bit surprisingly.

my suggestion was more around the idea of naming the CSS classes according to what that component is doing - showing the untransformed note contents - and then using visibility overrides to show it. does that make any more sense than what I wrote the first time?

scss/print.scss
```diff
@@ -23,26 +23,29 @@
...
 
     &.theme-color-fg {
       color: $studio-black;
+      overflow-y: auto !important;
     }
   }
+
+  .note-content-print {
+    left: 0px;
+    top: 0px;
```
dmsnell: if `display: none` wasn't taking this out of the layout flow maybe we need to add `visibility: hidden;`?

I don't think we should need to position it at the top left of the screen and am a bit worried this could run into unexpected consequences: maybe even the flickering that @codebykat noticed

belcherj: Removed the absolute positioning.

The flickering is due to adding the display none to and then removing it. I am not sure how we fix the flicker without removing the animations. You can open the animations panel in chrome dev tools to see what it is doing.

scss/print.scss
```diff
@@ -23,26 +23,26 @@
...
 
     &.theme-color-fg {
       color: $studio-black;
+      overflow-y: auto !important;
     }
   }
+
+  .note-content-plaintext {
+    display: inherit;
+    white-space: pre-wrap;
```
dmsnell: why do we have `white-space` added here and on `&.visible` in `theme.scss` instead of on the root `.note-content-plaintext` rule? are we taking it out of the hidden view for some side-effect?

dmsnell (approved):

PR #2272, opened 2020-08-11
[rewrite branch] Use custom styles for Monaco; fixes colors of various editor components

### LINKED ISSUE #2265: Staging: Dark mode text colors don't match
On the staging beta, the text color within notes not only differs from on app.simplenote.com (which I am not sure is expected), but it also differs from the text color in the note list.

### Expected
That the note text and the note list text colors match.

### Observed
Note text color: #d4d4d4
Note list text color: #fff 

Chrome Version 84.0.4147.105
Mac OS 10.14.6

codebykat:
### Fix

Monaco can accept custom theming; instead of using the default `vs` and `vs-dark` themes, we can create Simplenote light and dark themes to pass in.

Options are documented over here: https://microsoft.github.io/monaco-editor/playground.html#customizing-the-appearence-exposed-colors and scrollbars here: https://microsoft.github.io/monaco-editor/playground.html#customizing-the-appearence-scrollbars

This fixes #2265 as well as other oddities such as the highlight colors not matching.

### Test

Observe the colors of various things within the editor component (text color, highlight, scrollbar). They should match the rest of the app (and, hopefully, production).

### Release

Not updated: Use Monaco styles to pass Simplenote theme to editor component

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -1,6 +1,10 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import Monaco, { ChangeHandler, EditorDidMount } from 'react-monaco-editor';
+import Monaco, {
+  ChangeHandler,
+  EditorDidMount,
+  EditorWillMount,
+} from 'react-monaco-editor';
 import { editor as Editor, Selection, SelectionDirection } from 'monaco-editor';
 
 import actions from './state/actions';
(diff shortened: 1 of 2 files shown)


belcherj (commented):
Should we take a look at how things works underneath: https://reactjs.org/docs/react-component.html#unsafe_componentwillmount

dmsnell:
>Should we take a look at how things works underneath

@belcherj what are you commenting on or suggesting? I'm not following so I assume I might be missing something you're pointing toward but otherwise this seems like a solid update. It mimics some of my [initial work in #692](https://github.com/Automattic/simplenote-electron/pull/692/files#diff-6646bac18aa042ee2a7fb3cc0a5f669cR225) but in a way that seems more consistent overall.

codebykat:
I'm not following @belcherj's comment either. `EditorWillMount` is a function provided by the `react-monaco-editor` package: https://github.com/react-monaco-editor/react-monaco-editor#properties

Aside from the name, I don't think it has anything to do with React's `ComponentWillMount`.

belcherj:
https://github.com/react-monaco-editor/react-monaco-editor/blob/94af8886155034c1f20ad82538e6c99e05b4d294/README.md#properties

> editorWillMount(monaco) an event emitted before the editor mounted (similar to componentWillMount of React).

I was commenting that Monaco may be using componentWillMount internally. I am happy to approve the PR.

I guess it doesn't matter since it is in a dependency. Also, React 17 doesn't remove componentWillMount. So if React Monaco changes the method we will just have to change our code.

dmsnell:
> I was commenting that Monaco may be using componentWillMount internally

that's kind of what I inferred might be the case. I don't think there's any problem using `UNSAFE_componentWillMount` or any liability from code using it. the React team seems to believe that they found a better API through hooks for most of what people use this lifecycle method for (largely due to some mistakes that were easy to make or common inconveniences with `willMount` and `didMount`) but even still there are some good cases where that deprecated API might still be right.

so that is to say that even if it were using it internally it wouldn't bother me or give me any reason to be suspicious - unless that is, it were demonstrably broken in some way as a result of it, which I don't believe it is.

dmsnell (approved):
:shipit:

PR #2276, opened 2020-08-14
[rewrite branch] Add back search highlighting

belcherj:
### Fix

In the note-content -editor, when nextProps for searchQuery do not match existing props, open the Monaco search widget and populate with the text. After doing that refocus the search field. When there are no ranges in Monaco that match the search query it closes the search widget.

This PR also updated the search highlight and scrollbar highlights style so that they are Simplenote blue 50. It also hides the native Monaco search widget with css.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -115,14 +115,20 @@ class NoteContentEditor extends Component<Props> {
 
     if (this.editor && prevProps.searchQuery !== this.props.searchQuery) {
       const model = this.editor.getModel();
+      // Look to see if the current note has matches for the search term and get range of first match
       const range = model?.findMatches(this.props.searchQuery)[0]?.range;
+      // get the currently active element, to be restored later
       const { activeElement } = document;
       if (range) {
+        // if range exists select it
         this.editor.setSelection(range);
+        // open Monaco search widget, will populate with selected range
         this.editor.getAction('actions.find').run();
       } else {
+        // if no range close widget
         this.editor.trigger('keyboard', 'closeFindWidget');
       }
+      // restore focus to previously active element
       activeElement?.focus();
     }
 
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -116,14 +116,14 @@ class NoteContentEditor extends Component<Props> {
     if (this.editor && prevProps.searchQuery !== this.props.searchQuery) {
       const model = this.editor.getModel();
       const range = model?.findMatches(this.props.searchQuery)[0]?.range;
-      const searchArea = document.activeElement;
+      const { activeElement } = document;
       if (range) {
         this.editor.setSelection(range);
         this.editor.getAction('actions.find').run();
       } else {
         this.editor.trigger('keyboard', 'closeFindWidget');
       }
-      searchArea?.focus();
+      activeElement?.focus();
     }
 
     if (
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -32,7 +32,6 @@ type StateProps = {
 
 type DispatchProps = {
   editNote: (noteId: T.EntityId, changes: Partial<T.Note>) => any;
-  focusSearch: () => any;
   insertTask: () => any;
   storeEditorSelection: (
     noteId: T.EntityId,
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -183,8 +183,8 @@ class NoteContentEditor extends Component<Props> {
         'editor.foreground': '#2c3338', // $studio-gray-80 AKA theme-color-fg
         'editor.background': '#ffffff',
         'editor.selectionBackground': '#ced9f2', // $studio-simplenote-blue-5
-        'editor.findMatchHighlightBackground': '#3361cc', // $studio-simplenote-blue-50
-        'editorOverviewRuler.findMatchForeground': '#3361cc', // Editor marker navigation widget error color.
+        'editor.findMatchHighlightBackground': '#3361cc', // $studio-simplenote-blue-50 Search highlights
+        'editorOverviewRuler.findMatchForeground': '#3361cc', // $studio-simplenote-blue-50 Search highlights in scrollbar
         'scrollbarSlider.activeBackground': '#8c8f94', // $studio-gray-30
         'scrollbarSlider.background': '#c3c4c7', // $studio-gray-10
         'scrollbarSlider.hoverBackground': '#a7aaad', // $studio-gray-20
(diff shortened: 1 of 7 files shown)


lib/note-content-editor.tsx
```diff
@@ -152,6 +152,7 @@ class NoteContentEditor extends Component<Props> {
 
   editorReady: EditorDidMount = (editor, monaco) => {
     this.editor = editor;
+    window.editor = editor;
```
dmsnell: this was here before for debugging. mind adding a `@TODO` comment to make it easy to find before final merge? (I don't think we want to leave it hanging on the `window` object even though it may not be the worst thing)

lib/state/ui/search-field-middleware.ts
```diff
@@ -10,10 +10,19 @@ export const middleware: S.Middleware = () => {
...
+        if (range) {
+          window.editor.setSelection(range);
+          window.editor.getAction('actions.find').run();
+        } else {
+          window.editor.setSelection();
+          window.editor.getAction('actions.find').run();
+        }
         break;
-
+      }
```
dmsnell: this sits a big uneasily with me as another change did before where we split Monaco-specific functionality between the editor and the Electron middleware.

I've considered pulling the Redux store into the editor in order to listen to these events even as a tradeoff, as bad as that would normally seem, to keep all the specifics about Monaco in one place.

`SEARCH` fires on every keystroke in the search box? if so, do we even need to listen here for changes and imperatively update the editor or is it already convenient enough that we can watch for the pure state prop `searchQuery` in  `note-content-editor::getDerivedStateFromProps` and update the editor when it changes?

dmsnell (commented):
this is awesome @belcherj - thanks for finding this. I suspect that somewhere in this functionality is a key to highlighting spell-check suggestions as well

lib/state/action-types.ts
```diff
@@ -68,7 +68,10 @@ export type FilterNotes = Action<
   'FILTER_NOTES',
   { noteIds: T.EntityId[]; tagHashes: T.TagHash[] }
 >;
-export type FocusSearchField = Action<'FOCUS_SEARCH_FIELD'>;
+export type FocusSearchField = Action<
+  'FOCUS_SEARCH_FIELD',
+  { select?: boolean }
```
dmsnell: I'm nervous about adding an unlabeled boolean argument here and wondering if it's even necessary. it seems like the role this is fulfilling is to be able to give us a chance to restore focus to the search field when we "steal" it in Monaco to perform searches

could we trap `document.activeElement` or somehow find the focused element inside of the Monaco code and then restore that element there, eliminating the need for changing this action?

belcherj: Interesting, I will play with document.activeElement. If that won't work I will make it a string.

lib/note-content-editor.tsx
```diff
@@ -113,6 +114,18 @@ class NoteContentEditor extends Component<Props> {
...
+    if (this.editor && prevProps.searchQuery !== this.props.searchQuery) {
+      const model = this.editor.getModel();
+      const range = model?.findMatches(this.props.searchQuery)[0]?.range;
+      if (range) {
+        this.editor.setSelection(range);
+        this.editor.getAction('actions.find').run();
+      } else {
+        this.editor.trigger('keyboard', 'closeFindWidget');
+      }
+      this.props.focusSearch();
```
dmsnell: `findMatches` accepts a [`searchScope`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.itextmodel.html#findmatches) we could use to force the full document range.

```ts
const model = this.editor.getModel();
if (!model) {
	break;
}

model.findMatches(this.props.searchQuery, model.getFullModelRange())
```

maybe this would prevent us from stealing focus from the search bar?

belcherj: findMatches will give us all the matches however we would then have to do some custom highlighting.

lib/state/action-types.ts
dmsnell: thanks. reasonable string values could be something like `'select-text'` and `'do-not-select-text'`

lib/note-content-editor.tsx
```diff
@@ -113,6 +113,19 @@ class NoteContentEditor extends Component<Props> {
       editorSelection: [thisStart, thisEnd, thisDirection],
     } = this.props;
 
+    if (this.editor && prevProps.searchQuery !== this.props.searchQuery) {
+      const model = this.editor.getModel();
+      const range = model?.findMatches(this.props.searchQuery)[0]?.range;
+      const searchArea = document.activeElement;
```
dmsnell: part of my suggestion was based on the fact that it might not be the search box that's focused when this update occurs. we might be clearer if we said something like `const activeElement = document.activeElement` or just `const { activeElement } = document`

lib/note-content-editor.tsx
belcherj: `this.editor.getAction('actions.find').run()` will start a search and do all the highlighting with whatever is currently selected in the editor.  

I will see if there is a way to run the search programmatically rather than relying on the text selection. As I remember I could not find a way.

lib/note-content-editor.tsx
belcherj: Fair enough!

lib/note-content-editor.tsx
```diff
@@ -113,6 +113,19 @@ class NoteContentEditor extends Component<Props> {
...
+      if (range) {
+        this.editor.setSelection(range);
+        this.editor.getAction('actions.find').run();
```
dmsnell, line 126: if I understand your last comment right, then it's this line that highlights the search results?

what's the relationship between `findMatches()` and `getAction('actions.find').run()`?
do we define `actions.find` somewhere?

dmsnell: …answered while I was asking - https://github.com/Automattic/simplenote-electron/pull/2276#discussion_r474888632

belcherj: `actions.find` is a Monaco action. It opens the search widget with whatever the currently selected text is. 

We have to findMatches in the note. If there is a match, highlight it. Then open the search widget.

If there are no matches close the search widget

belcherj: I added comments as it seems what is happening here is unclear

lib/note-content-editor.tsx
dmsnell: Might try to set `seedSearchStringFromSelection` to `false` [IEditorFindOptions](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.ieditorfindoptions.html) in [setOption](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.html.languageservicedefaults.html#setoptions)

It's not clear to me if that'll do what we want.

belcherj:
Closing for now as this approach does not seem viable. #2292 PR adds this functionality.

PR #2277, opened 2020-08-18
Add Electron close dialog

belcherj:
### Fix

When we close the app we want to ensure that the user knows if there are unsynchronized notes. Closing the app will cause those changes to sit unsynchronized and could lead to data loss.

This PR utilizes the work done on the Log out unsynchronized notes dialog. When the user clicks "Quit Simplenote" in the system menu or uses the close button on the window we send an event to the app to check if there are unsynchronized notes. If there are no unsynchronized notes we send an event back to electron to close the application. If there are unsynchronized notes we pop the confirmation dialog. If the click the continue button we will send an event to electron to close the app.

### Test
1. Close the app is a variety of ways with fully sync'd notes.
2. Close the app in a variety of ways with unsync'd notes.
3. Interact with all the features of the unsync'd notes dialog


lib/dialogs/logout-confirmation/index.tsx
```diff
@@ -29,15 +33,28 @@ export class LogoutConfirmation extends Component<Props> {
...
+  };
+
   render() {
-    const { closeDialog, notes, reallyLogout } = this.props;
+    const { action, closeDialog, notes } = this.props;
+
+    const actionDescription =
+      action === 'REALLY_LOGOUT'
+        ? 'Logging out will delete any unsynchronized notes.'
+        : 'Closing the app with unsynchronized notes could cause data loss.';
```
dmsnell: the way we're passing in the action name and then making decisions based on that inside of the component feels like what I call _slanted abstractions_ where we have one concern split across separate layers of the app. in this case I feel like the confirmation dialog shouldn't have to know about the Redux actions and we shouldn't have to create a hollow function to reconstruct a _real_ Redux action based on the _partial_ action type we pass in.

Can we think of ways to reuse this component without diving in and doing surgery on it? I think I see two points of change based on the role for which we're using the dialog: labeling the description at the top; and providing a _continuation_ action for when we're done.

In the first case I wonder if we even need to have different descriptions? We _could_ pass them into the component but also would something like this work for both cases?

> If you continue and leave while there are unsynchronized changes there's a risk those changes will be lost.

Seems like something of that nature could cover both cases and the user will know the context from their interactions that they are closing vs. logging-out.

Concerning the button I'd rather see us pass in an actual action that this component doesn't know about than to pass in an action name, coupling this component to the details of our Redux system and the expectations of the parent rendering this component.

```ts
type OwnProps {
	finisher: () => any;
}

…

<button … onCLick={this.props.finisher}>
```

To that point we still have labeling on the button that's specific. Again, can we return control back up the parent? I don't know if this is the right solution, but we can have the parent supply the button itself and even avoid the question of the action.

```ts
<LogoutConfirmation key="logout-confirmation">
	({hasPendingChanges, className}) => (
		<button className={className} onClick={this.props.reallyLogOut}>
			{hasPendingChanges ? "Log out" : "Safely log out"}
		</button>
	)
</LogoutConfirmation>

<LogoutConfirmation key="close-confirmation">
	({hasPendingChanges, className}) => (
		<button className={className} onClick={this.props.reallyLogOut}>
			{hasPendingChanges ? "Close window" : "Safely close window"}
		</button>
	)
</LogoutConfirmation>
```

belcherj: I agree with everything here except that we should supply the button to the dialog. I think it is cleaner to pass the redux action to the dialog and let the dialog control the UI.

lib/dialogs/unsynchronized/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+            <button
+              className="export-unsynchronized"
+              onClick={this.exportUnsyncedNotes}
+            >
+              Export unsynchronized notes
+            </button>
+          )}
+
+          <section className="action-button">
+            <button className="log-out" onClick={this.triggerContinueAction}>
```
dmsnell: there's no reason to create `triggerContinueAction` if we're already passing in `continueAction`. all it's doing it adding a function as a middleman that literally does nothing but call the other one and add a layer of indirection

```ts
<button className="log-out" onClick={this.props.continueAction}>
```

belcherj: Ohh geeez, fixed

lib/dialog-renderer/index.tsx
```diff
@@ -52,6 +53,8 @@ export class DialogRenderer extends Component<Props> {
               <AboutDialog key="about" closeDialog={closeDialog} />
             ) : 'BETA-WARNING' === dialog ? (
               <BetaWarning key="beta-warning" />
+            ) : 'CLOSE-WINDOW-CONFIRMATION' === dialog ? (
+              <CloseWindowConfirmation key="close-window-confirmation" />
```
dmsnell, line 57: While I love that the specifics are separated from the dialog I feel like we don't really need to have separate components for them. In this case it almost masquerades the intrinsics of what we're opening. That is, is `<CloseWindowConfirmation />` all that better than this?

```ts
<UnsynchronizedConfirmation
     actionDescription="Closing the app with unsynchronized notes could cause data loss."
     actionName="Close window"
     actionSafeName="Safely close window"
     continueAction={this.props.reallyCloseWindow}
/>
```

This second example appears to break the pattern in this dialog-renderer list but it also means we don't have to go hunting to realize that the close and logout dialogs are basically the same. It also means we don't have to create a seemingly unnecessary `continueAction` prop just for the `<UnsynchronizedConfirmation />`.

belcherj: I disagree. I think adding the complexities of the redux actions and the copy to DialogRenderer is a mistake.

I think if we were to do this it would make more sense to push it back to the component.

lib/dialogs/unsynchronized/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { noteTitleAndPreview } from '../../utils/note-utils';
+
+import type * as S from '../../state';
+import type * as T from '../../types';
+
+type OwnProps = {
+  actionDescription: string;
+  actionName: string;
+  actionSafeName: string;
+  continueAction: () => any;
```
dmsnell: Given the usage above and specificity of these the repeated `action` prefixes and suffix may not be necessary.

```ts
<UnsynchronizedDialog
	description="You may lose your changes if you logout!"
	unsafeAction="Logout"
	safeAction="Safely Logout"
	continue={this.props.reallyLogout}
/>
```

`continueAction` - `action` - `finisher` - `onFinish` - `confirmAction` - `doAnyway`

the name of this one isn't as important but it seems like something else out there fits it better. if you don't like any of the others `continueAction` is fine or maybe just `continue` works.

lib/state/electron/middleware.ts
```diff
@@ -100,6 +104,12 @@ export const middleware: S.Middleware = ({ dispatch, getState }) => {
     const result = next(action);
     const nextState = getState();
 
+    switch (action.type) {
+      case 'REALLY_CLOSE_WINDOW':
+        console.log('REALLY_CLOSE_WINDOW');
+        window.electron.send('reallyCloseWindow');
+        return result;
+    }
```
dmsnell, line 111: probably want a newline here. I'm surprised `prettier` didn't add it.

also let's get rid of that console statement

lib/state/simperium/actions.ts
```diff
@@ -9,3 +9,7 @@ export const remoteNoteUpdate: A.ActionCreator<A.RemoteNoteUpdate> = (
   noteId,
   note,
 });
+
+export const reallyLogOut: A.ActionCreator<A.ReallyLogOut> = () => ({
+  type: 'REALLY_LOG_OUT',
+});
```
dmsnell: as an action this feels more like a UI action than a Simperium one. is there a reason to have it in this file?

lib/dialogs/logout-confirmation/index.tsx
```diff
@@ -1,104 +1,28 @@
...
+    actionName="Log out"
+    actionSafeName="Safely log out"
+    continueAction={continueAction}
+  />
+);
 
 const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
-  closeDialog,
-  reallyLogout: () => ({ type: 'REALLY_LOGOUT' }),
+  continueAction: actions.simperium.reallyLogOut,
```
dmsnell: if we leave in these wrapper components then changing `reallyLogout` to `continueAction` feels like it's a step backwards in indicating what the intention is for the component. it's exposing the internals of the component to the caller of the component because we know that as a _logout_ thing the action it wants is some function that _really logs out_. with `continueAction` I'm not sure what that implies.

lib/dialogs/unsynchronized/index.tsx
belcherj: Resolved in https://github.com/Automattic/simplenote-electron/pull/2277/commits/bb069aed17218e442be9646f4dd46c85201bf610

lib/dialog-renderer/index.tsx
dmsnell: >I think if we were to do this it would make more sense to push it back to the component.

Can you identify the demonstratives here?

If we were to do what?
Push what back to the component?

lib/state/simperium/actions.ts
belcherj: Resolved in https://github.com/Automattic/simplenote-electron/pull/2277/commits/973f4d9f0dd18b1bf5467297f617df1bac2ccbf2

lib/dialogs/close-window-confirmation/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  <UnsynchronizedConfirmation
+    description="Closing the app with unsynchronized notes could cause data loss."
+    unsafeAction="Close window"
+    safeAction="Safely close window"
+    action={action}
+  />
+);
+
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
+  action: actions.electron.reallyCloseWindow,
```
dmsnell: sorry for the duplicate comment - adding here since the last one might have been lost during the other code update.

this component, the `CloseWindowConfirmation`, isn't asking for some abstract `action` - it specifically wants to close the window and so why not maintain that honesty in the prop names. dialogs are already so confusing in this app and I don't want to see us keep the abstraction bleeding by exposing internals up the React ladder.

as a default-wrapped component nobody else should ever know but I don't think we named this `action` because that's what we felt was the component API of `CloseWindowConfirmation` - that would ring closer to what the `LogoutConfirmation` dialog previously used - in this case, `reallyCloseWindow`.

lib/state/electron/middleware.ts
belcherj: Console resolved in: https://github.com/Automattic/simplenote-electron/pull/2277/commits/bb069aed17218e442be9646f4dd46c85201bf610

New line resolved in: https://github.com/Automattic/simplenote-electron/pull/2277/commits/6dfd9843c1cdd0bb802a4fdd174202fa0bdbdc07

lib/dialogs/logout-confirmation/index.tsx
```diff
@@ -1,104 +1,28 @@
...
+    unsafeAction="Log out"
+    safeAction="Safely log out"
+    action={action}
+  />
+);
 
 const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
-  closeDialog,
-  reallyLogout: () => ({ type: 'REALLY_LOGOUT' }),
+  action: actions.ui.reallyLogOut,
```
dmsnell: same comment as above: `reallyLogout` is closer to what this component is looking for in reality. it's not asking for _any_ action but specifically some function that will _really logout_ of the app.

lib/dialog-renderer/index.tsx
belcherj: If we were to remove the separate components and put the props in `lib/dialog-renderer/index.tsx` as you described. This component should not be concerned with the specifics of the logout and close window dialogs.

lib/dialogs/logout-confirmation/index.tsx
belcherj: resolved in https://github.com/Automattic/simplenote-electron/pull/2277/commits/094570fc5044611800546774d72d58d716113af6

lib/dialogs/close-window-confirmation/index.tsx
belcherj: resolved in https://github.com/Automattic/simplenote-electron/pull/2277/commits/094570fc5044611800546774d72d58d716113af6

lib/dialogs/logout-confirmation/index.tsx
belcherj: resolved in https://github.com/Automattic/simplenote-electron/pull/2277/commits/094570fc5044611800546774d72d58d716113af6

belcherj:
Merging as all comments have been addressed.

PR #2280, opened 2020-08-18
[Rewrite Branch] Use Monaco context menu for editor actions

belcherj:
### Fix

Moves to the Monaco context menu. This enables undo/redo/select all/copy/cut from the context menu.

Screen Shot 2020-08-25 at 10 01 13 AM

### Test
1. Test all the items in the context menu on multiple browsers
2. Test all the items in the edit menu on multiple browsers

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -327,9 +327,9 @@ class NoteContentEditor extends Component<Props> {
 
   editorReady: EditorDidMount = (editor, monaco) => {
     this.editor = editor;
-    this.monaco = monaco;
-
     window.editor = editor;
+
+    this.monaco = monaco;
     window.monaco = monaco;
 
     // remove keybindings; see https://github.com/microsoft/monaco-editor/issues/287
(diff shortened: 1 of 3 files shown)


codebykat:
I spent awhile trying to find a better way to disable the Command Palette and Change All Occurrences aside from hiding the CSS, and that might actually be best, even though I hate it. Would be good to comment that CSS block.

~#2284 shouldn't conflict with this one; over there I put some editor config and `addAction`s in `editorReady`, which I think is where they belong here too.~ (done)

~Select all needs a keybinding (`keybindings: [monaco.KeyMod.CtrlCmd | monaco.KeyCode.KEY_A],`).~ (done)

We should figure out how to make Cut and Copy show their keybindings as well. Again see #2284.

Given that we might well end up suppressing everything and adding everything we want to the context menu, I'm seriously wondering whether we want to build our own from scratch rather than fight with Monaco's.

belcherj:
> Given that we might well end up suppressing everything and adding everything we want to the context menu, I'm seriously wondering whether we want to build our own from scratch rather than fight with Monaco's.

Was thinking the same thing.

codebykat:
I made a couple minor changes, mostly just moving things around as I noted in the previous comment. Disabling the keybindings for Electron (so it's handled by the native keybindings when possible).

Annoyingly, Firefox has Undo/Redo disabled, I may dig into that more and see if there's a way to enable it. Might be an a11y issue.

Screen Shot 2020-08-24 at 1 00 41 AM

Still needs work but it's better..

codebykat:
Ready for review I think. We are now removing every native thing from this menu and adding our own 🙃 

Screen Shot 2020-08-24 at 5 13 09 PM

I'm not sure if we want to worry about things being contextual, like should we disable cut/copy if there's no selection?

codebykat:
We might need to remove Paste, I don't think `execCommand('paste')` works everywhere. Having the action added when the paste action is not supported is breaking the keyboard shortcut too.

I also tried `editor.trigger( ...,  'editor.action.clipboardCopyAction', ...);` which does not appear to work anywhere.

In my testing it does not work on Firefox/Windows.

See https://github.com/Microsoft/monaco-editor/issues/999

codebykat:
I removed Paste from the browser unless/until we find a workaround (kept in for Electron, where it works)

belcherj:
I cannot approve since this is my PR. I have tested this in a bunch of different browsers and have not found any major issues. 

Issues to resolve later:
- ctrl + v is not working on pre-chromium edge.
- select all is not working on pre-chromium edge.

I am going to merge. If you (@dmsnell) have reservations about this change I will revert.

codebykat:
Just noting I circled back and tested the shortcuts on pre-chromium Edge (in Parallels) and they work for me.

PR #2281, opened 2020-08-18
[rewrite branch] Fix plain-text revision preview

codebykat:
### Fix

We're overloading the preview component to preview revisions as well as rendered Markdown, so it needs to be aware of whether we're in `editMode`. This allows viewing revisions of Markdown notes either rendered or plain-text, depending on whether preview was selected when opening the history viewer.

### Test

1. Open a Markdown note with some history
2. Open revision history and scrub through
3. Verify that the revisions are shown in plain-text
4. Toggle Markdown preview and open revision history again
5. Verify that revisions are previewed in Markdown
6. Verify that revision history works as expected for a non-Markdown note

### DIFF
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -21,6 +21,7 @@ type StateProps = {
   note: T.Note | null;
   noteId: T.EntityId | null;
   searchQuery: string;
+  showRevisions: boolean;
 };
 
 type DispatchProps = {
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
Love it.

If there were any change I'd consider passing in some boolean like `showRendered` which is provided by `mapStateToProps` and moves the logic of checking the markdown tag and the `editMode` outside of the actual UI component.

the benefit of course would be removing some coupling to the app state from the component but the downside (while insignificant for us) is that it does slightly more work in `mapStateToProps`.

so this seems like a decent tradeoff. I'm not sure which approach I would prefer but the other would change `editMode` as a prop to `showRenderedView` or something and the check in line 135/137 would be `if (note?.content && showRenderedView)`

codebykat:
I keep forgetting I can move logic into `mapStateToProps`! Okay I like this better @dmsnell

PR #2285, opened 2020-08-19
Fix deep-linking for Windows app store

belcherj:
### Fix

Fixes wpcom login on the Windows store.

The appx deeplinking was broken due to the way that appx registers itself. You have to specify the protocol in the app registry which means it needs to be moved to the electron builder config.

I moved the `will-finish-launching` below the instantiation of mainWindow as I managed to catch an error that mainWindow was null. This ensures that we won't get into that error state because the watcher isn't set up until mainWindow is no longer null.

Additionally, I added some code to the ''second-instance' to correctly send the URL that we get back from wpcom to windows.

### Test
1. Download the appx from circleCI for this PR
2. Login via wpcom login
3. Login via email and password

### Release

Fixes WordPress.com login

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -27,6 +27,14 @@ module.exports = function main() {
   // be closed automatically when the JavaScript object is GCed.
   let mainWindow = null;
 
+  app.on('will-finish-launching', function () {
+    setTimeout(updater.ping.bind(updater), config.updater.delay);
+    app.on('open-url', function (event, url) {
+      event.preventDefault();
+      mainWindow.webContents.send('wpLogin', url);
+    });
+  });
+
   const url =
     isDev && process.env.DEV_SERVER
       ? 'http://localhost:4000' // TODO: find a solution to use host and port based on make config.
@@ -155,14 +163,6 @@ module.exports = function main() {
     // wait until window is presentable
     mainWindow.once('ready-to-show', mainWindow.show);
   };
-
-  app.on('will-finish-launching', function () {
-    setTimeout(updater.ping.bind(updater), config.updater.delay);
-    app.on('open-url', function (event, url) {
-      event.preventDefault();
-      mainWindow.webContents.send('wpLogin', url);
-    });
-  });
 
   const gotTheLock = app.requestSingleInstanceLock();
   if (gotTheLock) {
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -170,6 +170,14 @@ module.exports = function main() {
   }
 
   app.on('second-instance', (e, argv) => {
+    // Someone tried to run a second instance, we should focus our window.
+    if (mainWindow) {
+      if (mainWindow.isMinimized()) {
+        mainWindow.restore();
+      }
+      mainWindow.focus();
+    }
+
     if (process.platform === 'darwin') {
       // macOS communicates deep-linking via the `open-url` event (see above)
       // but we might still end up with this message so ignore it if we do
@@ -179,14 +187,6 @@ module.exports = function main() {
     // argv: An array of the second instance’s (command line / deep linked) arguments
     // The last index of argv is the full deeplink url (simplenote://SOME_URL)
     mainWindow.webContents.send('wpLogin', argv[argv.length - 1]);
-
-    // Someone tried to run a second instance, we should focus our window.
-    if (mainWindow) {
-      if (mainWindow.isMinimized()) {
-        mainWindow.restore();
-      }
-      mainWindow.focus();
-    }
   });
 
   if (!app.isDefaultProtocolClient('simplenote')) {
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -167,14 +167,15 @@ module.exports = function main() {
   const gotTheLock = app.requestSingleInstanceLock();
   if (gotTheLock) {
     app.on('second-instance', (e, argv) => {
+      if (process.platform === 'darwin') {
+        // macOS communicates deep-linking via the `open-url` event (see above)
+        // but we might still end up with this message so ignore it if we do
+        return;
+      }
       // Protocol handler for platforms other than macOS
       // argv: An array of the second instance’s (command line / deep linked) arguments
       // The last index of argv is the full deeplink url (simplenote://SOME_URL)
-      // second-instance should not be called by macOS however we should check
-      if (process.platform !== 'darwin') {
-        const url = argv[argv.length - 1] || '';
-        mainWindow.webContents.send('wpLogin', url);
-      }
+      mainWindow.webContents.send('wpLogin', argv[argv.length - 1]);
 
       // Someone tried to run a second instance, we should focus our window.
       if (mainWindow) {
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -165,9 +165,6 @@ module.exports = function main() {
   };
 
   const gotTheLock = app.requestSingleInstanceLock();
-  if (!gotTheLock) {
-    return app.quit();
-  }
 
   app.on('second-instance', (e, argv) => {
     // Someone tried to run a second instance, we should focus our window.
@@ -188,6 +185,10 @@ module.exports = function main() {
     // The last index of argv is the full deeplink url (simplenote://SOME_URL)
     mainWindow.webContents.send('wpLogin', argv[argv.length - 1]);
   });
+
+  if (!gotTheLock) {
+    return app.quit();
+  }
 
   if (!app.isDefaultProtocolClient('simplenote')) {
     // Define custom protocol handler. This allows for deeplinking into the app from simplenote://
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -27,14 +27,6 @@ module.exports = function main() {
   // be closed automatically when the JavaScript object is GCed.
   let mainWindow = null;
 
-  app.on('will-finish-launching', function () {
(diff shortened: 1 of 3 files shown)


desktop/app.js
```diff
@@ -30,8 +30,13 @@ module.exports = function main() {
   app.on('will-finish-launching', function () {
     setTimeout(updater.ping.bind(updater), config.updater.delay);
     app.on('open-url', function (event, url) {
+      let deeplinkingUrl = url;
       event.preventDefault();
-      mainWindow.webContents.send('wpLogin', url);
+      if (process.platform === 'win32') {
+        // Keep only command line / deep linked arguments
+        deeplinkingUrl = process.argv.slice(1);
+      }
```
dmsnell: no need for the mutation: can easily assign this as a `const`

```ts
const deeplinkingUrl = process.platform === 'win32'
	? process.argv.slice(1) // keep only command-line / deep-linked arguments
	: url;
```

also, just a check - we want _either_ the given URL _or_ an array of strings for this value?

dmsnell (commented):
it's such a relief that you found a solution for this. I was afraid it was going to be another nightmare

bravo, you are my hero

dmsnell:
> I moved the will-finish-launching below the instantiation of mainWindow as I managed to catch an error that mainWindow was null.

This is a great catch @belcherj but I don't think moving the code is going to address it. The assignment to `mainWindow` is in `activateWindow()` which is called on `app.on('ready', …)` and `app.on('activate', …)`, meaning that even when `will-finish-launching` is registered after `activateWindow()` is defined it won't change when that code runs. If it started working after the move it's likely it was coincidence. I might be missing something but if we want to eliminate the race condition we would need to add a sequencing structure to ensure the order of execution.

We could move that entire block inside the `activateWindow()` callback and get rid of `will-finish-launching`
```js
const activateWindow = function() {
	…

	mainWindow = new BrowserWindow(…);

	app.on('open-url', function(event, url) {
		event.preventDefault();
		mainWindow.webContents.send('wpLogin', url);
	} );

	…
}
```

If we needed to make sure we are listening for the `will-finish-launching`  or `open-url` events and they might come before `mainWindow` is around then we can still grab it there and consider adding a Promise for sequencing…

```js
let mainWindow = null;
let resolveMainWindow;
const waitForMainWindow = new Promise( resolve => resolveMainWindow = resolve );

…

app.on('will-finish-launching', function() {
	setTimetout(…);
	app.on('open-url', function(event, url) {
		event.preventDefault();
		waitForMainWindow.then( () => mainWindow.webContents.send('wpLogin', url) );
	});
});
```

This raises another issue to where things start creeping into this PR so maybe take this with a grain of salt, but the file has a number of these races related to `mainWindow` because we assume all the events come after initialization. A single Promise-interface might clear them all up whereby we get `mainWindow` through the `Promise` instead of through the function-level "global" variable. This approach would be more type-safe as well since we would eliminate the possibility of using a `null`-valued `mainWindow`

```js
const { activateWindow, waitForMainWindow } = (() => {
	let mainWindow = null;
	let mainWindowResolver;
	const waitForMainWindow = new Promise( resolve => mainWindowResolver = resolve )
		.then( mainWindow => mainWindow ? Promise.resolve(mainWindow) : Promise.reject() );

	const activateWindow = () => {
		
		…

		mainWindow = new BrowserWindow({…})
		mainWindowResolver( mainWindow );

		…
	}

	return { activateWindow, waitForMainWindow };
})();

app.on('will-finish-launching', … {
	…
		waitForMainWindow.then( mainWindow => mainWindow.webContents.send('wpLogin', url));
	…
});

app.on('second-instance', () => waitForMainWindow.then(mainWindow => {
	if (mainWindow.isMinimized()) {
		mainWindow.restore();
	}
	mainWindow.focus();
}));
```

that approach removes some possible errors but also I realize that only `second-instance` and `will-finish-launching` add the possibility to use `mainWindow` before it's available. maybe we just add those inside `activateWindow` and call it a day…all depends on whether we have to trap the events before `mainWindow` appears or not.

these sequencing chains could be done with our own new events using an `EventEmitter` or a `MessageChannel` but I thought the `Promise` might be simpler.

desktop/app.js
```diff
@@ -164,20 +156,37 @@ module.exports = function main() {
...
+  const gotTheLock = app.requestSingleInstanceLock();
+  if (gotTheLock) {
+    app.on('second-instance', (e, argv) => {
+      // Someone tried to run a second instance, we should focus our window.
+      // Protocol handler for win32
+      // argv: An array of the second instance’s (command line / deep linked) arguments
+      if (process.platform !== 'darwin') {
+        // Keep only command line / deep linked arguments
+        const url =
+          (argv.slice(-1) && argv.slice(-1).length && argv.slice(-1)[0]) || '';
```
dmsnell: can we show in a comment what these might be that we expect? it now looks like we're trying to get the last-passed arg from the CLI but it's not clear what that argument would or should be

(The comment I'm leaving in this snippet might be complete garbage but that's just because I'm trying to understand what's happening but don't know).
```js
app.on( 'second-instance', (event, argv) => {
	// when we register the `simplenote://` protocol in Windows and Linux they handle those
	// links by spawning a new invocation of Electron and pass the URL as the last parameter
	// in this we case we expect an instance to be already open and so we'll intercept that event
	// here and translate the URL into an action for the app to handle
	//
	// e.g. argv: [ '/some/path/to/simplenote.appx', 'http://localhost:4000' ]
	//
	// in macOS it doesn't send the URL so we'll assume this is for WPCC login but eventually
	// we'll have to differentiate to support deep-linking to individual notes if we add that

	const isMac = process.platform === 'darwin';
	const loginUrl = isMac ? url : argv[ argv.length - 1]

	mainWindow.webContents.send('wpLogin', loginUrl);
} );
```

after writing this comment I believe I understand the problem better but it begs a few questions:
 - am I correct in what I described in the comment?
 - what is the role of the URL? is it passed back from the WPCC auth server managed in the OAuth client application? (the `redirect_url`)?
 - why is macOS different?
 - should we verify that the URL is well-formed according to our expectations? should we differentiate what we do based on the URL?

desktop/app.js
```diff
@@ -164,20 +156,37 @@ module.exports = function main() {
...
-    return app.quit();
+      if (mainWindow) {
+        if (mainWindow.isMinimized()) {
+          mainWindow.restore();
+        }
+        mainWindow.focus();
+      }
+    });
+  } else {
+    app.quit();
```
dmsnell: is this here because we think of this process…

 1. Click "Sign-in with WordPress.com"
 2. Close the Electron app
 3. Sign-in with the browser
 4. Browser/OS invokes new Electron app _after_ closing the window but _before_ the app process quits

?

desktop/app.js
belcherj: macOS uses the open-url event. macOS does allow for multiple instances. At least it is not supposed to. There were some bugs saying that this event was fired on macOS, thus the check.

desktop/app.js
belcherj: This doesn't have specifically to do with wpcom sign in. When we get a second instance we need to quit it so we only have one instance open. This was already in the code as:

```
if (!gotTheLock) {
    return app.quit();
```

desktop/app.js
belcherj: We only want to run the code if it is not macOS

dmsnell: > macOS uses the open-url event

this would be good to include in the comment. I can easily see someone spending time trying to understand why something is different on Windows/macOS where that little sentence could save them (or us) time.

> We only want to run the code if it is not macOS

ah. my code in PR comments is usually wrong at least in some way.

```js
if (process.platform === 'darwin') {
	// macOS communicates deep-linking via the `open-url` event (see above)
	// but we might still end up with this message so ignore it if we do
	return;
}

const loginUrl = argv[argv.length - 1]; // still not sure what we expect of this URL
mainWindow.webContents.send('wpLogin', loginUrl);
```

desktop/app.js
dmsnell: Looks like I was confused by the diff. What were two separate things before now are joined into one.

Would it makes sense to early-abort to keep the window management separate from the login-URL handling?

```js
const gotTheLock = app.requestSingleInstanceLock();
if (!gotTheLock) {
	return app.quit();
}

// dispatch `wpLogin` based on `argv`
app.on('second-instance', …);
```

belcherj: Done. 

second-instance does more than "// dispatch `wpLogin` based on `argv`", it also focuses the window.

desktop/app.js
```diff
@@ -166,14 +166,24 @@ module.exports = function main() {
...
+      // but we might still end up with this message so ignore it if we do
+      return;
+    }
+    // Protocol handler for platforms other than macOS
+    // argv: An array of the second instance’s (command line / deep linked) arguments
+    // The last index of argv is the full deeplink url (simplenote://SOME_URL)
+    mainWindow.webContents.send('wpLogin', argv[argv.length - 1]);
   });
 
   if (!gotTheLock) {
```
dmsnell, line 189: do we need to move this above `app.on('second-instance', …)` or is it fine since a second invocation wouldn't get the _second_ instance signal?

dmsnell (approved):
Left a comment asking about `!gotTheLock` but I think it's sound since any additional invocations probably won't get the `second-instance` event (a guess on my part)

thanks for putting in so much work on this. it's clearer to me at least what's happening and how we're connecting the login flows

PR #2286, opened 2020-08-21
Add/inter note link searching

dmsnell:
This PR adds an auto-complete handler to use the search index for the app to generate a list of inter-note links.

It triggers automatically after typing a `[`, or if you dismiss the popup via `ESC` or by typing until there are no matches, you can trigger it with <kbd>Ctrl</kbd> + <kbd>Space</kbd>.

~n.b. even if the key combo is used, it will only open if there is an open bracket!~ Changed this to allow inserting a note link with Ctrl+Space at any time.

## Notes

 - ~Can use the full note list search query syntax. E.g. to search for notes with tag `classes` you type `[tag:classes` and then hit the autocomplete trigger~
- Only searches the note title
 - Autocomplete results will match the sort settings for the note list
 - Limits matches so to find specific note you need to type more search to narrow the list
 - The icon for pinned notes is different. It can look weird in the sort order if they aren't distinguished from each other, but it needs a better icon.
 - ~We should do a better job adding the singleton autocompleter. It's easy but hacky here.~

**(Graphic somewhat outdated but gives a general idea)**
interLinkAutoComplete mov

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -602,6 +602,38 @@ class NoteContentEditor extends Component<Props> {
       window.electron ? false : true
     );
 
+    editor.addCommand(
+      monaco.KeyMod.WinCtrl | monaco.KeyCode.Space,
+      function () {
+        // only show suggestions if we have an open bracket
+        const model = editor.getModel();
+        if (!model) {
+          return;
+        }
+        // @todo DRY - code duplicated from the completion handler
+        const position = editor.getPosition();
+        const line = model.getLineContent(position.lineNumber);
+        if (!line.includes('[')) {
+          return;
+        }
+        const precedingOpener = line.lastIndexOf('[', position.column);
+        const precedingCloser = line.lastIndexOf(']', position.column);
+        const precedingBracket =
+          precedingOpener >= 0 && precedingCloser < precedingOpener
+            ? precedingOpener
+            : -1;
+
+        const soFar =
+          precedingBracket >= 0
+            ? line.slice(precedingBracket + 1, position.column)
+            : '';
+
+        if (soFar) {
+          editor.trigger('suggestions', 'editor.action.triggerSuggest', null);
+        }
+      }
+    );
+
     editor.addAction({
       id: 'context_undo',
       label: 'Undo',
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -164,8 +164,6 @@ class NoteContentEditor extends Component<Props> {
     window.electron?.removeListener('editorCommand');
     window.removeEventListener('input', this.handleUndoRedo, true);
     this.toggleShortcuts(false);
-
-    this.completionProviderHandle?.dispose();
   }
 
   toggleShortcuts = (doEnable: boolean) => {
@@ -530,11 +528,6 @@ class NoteContentEditor extends Component<Props> {
         'textLink.foreground': '#ced9f2', // studio-simplenote-blue-5
       },
     });
-
-    this.completionProviderHandle = monaco.languages.registerCompletionItemProvider(
-      'plaintext',
-      this.completionProvider()
-    );
   };
 
   editorReady: EditorDidMount = (editor, monaco) => {
@@ -753,6 +746,13 @@ class NoteContentEditor extends Component<Props> {
     // make component rerender after the decorators are set.
     this.setState({});
     editor.onDidChangeModelContent(() => this.setDecorators());
+
+    // register completion provider for internal links
+    this.completionProviderHandle = monaco.languages.registerCompletionItemProvider(
+      'plaintext',
+      this.completionProvider()
+    );
+    editor.onDidDispose(() => this.completionProviderHandle?.dispose());
 
     document.oncopy = (event) => {
       // @TODO: This is selecting everything in the app but we should only
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -201,10 +201,6 @@ class NoteContentEditor extends Component<Props> {
       provideCompletionItems(model, position, context, token) {
         const line = model.getLineContent(position.lineNumber);
         const precedingOpener = line.lastIndexOf('[', position.column);
-        if (-1 === precedingOpener) {
-          editor.trigger('completions', 'hideSuggestWidget', null);
-          return null;
-        }
         const precedingCloser = line.lastIndexOf(']', position.column);
         const precedingBracket =
           precedingOpener >= 0 && precedingCloser < precedingOpener
@@ -231,12 +227,6 @@ class NoteContentEditor extends Component<Props> {
           isPinned: note.systemTags.includes('pinned'),
           ...noteTitleAndPreview(note),
         }));
-
-        if (0 === notes.length) {
-          // dismiss the widget if we don't have anything to suggest
-          editor.trigger('completions', 'hideSuggestWidget', null);
-          return null;
-        }
 
         const additionalTextEdits =
           precedingBracket >= 0
@@ -776,6 +766,11 @@ class NoteContentEditor extends Component<Props> {
       this.completionProvider(this.state.noteId, editor)
     );
     editor.onDidDispose(() => completionProviderHandle?.dispose());
+    monaco.languages.setLanguageConfiguration('plaintext', {
+      // Allow any non-whitespace character to be part of a "word"
+      // This prevents the dictionary suggestions from taking over our autosuggest
+      wordPattern: /[^\s]+/g,
+    });
 
     document.oncopy = (event) => {
       // @TODO: This is selecting everything in the app but we should only
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -174,8 +174,9 @@ class NoteContentEditor extends Component<Props> {
     }
   };
 
-  completionProvider: () => languages.CompletionItemProvider = () => {
-    const selectedNoteId = this.state.noteId;
+  completionProvider: (
+    selectedNoteId: T.EntityId
+  ) => languages.CompletionItemProvider = (selectedNoteId) => {
     return {
       triggerCharacters: ['['],
 
@@ -750,7 +751,7 @@ class NoteContentEditor extends Component<Props> {
     // register completion provider for internal links
     this.completionProviderHandle = monaco.languages.registerCompletionItemProvider(
       'plaintext',
-      this.completionProvider()
+      this.completionProvider(this.state.noteId)
     );
     editor.onDidDispose(() => this.completionProviderHandle?.dispose());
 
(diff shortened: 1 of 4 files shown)


codebykat:
I tinkered with this a bit. After re-enabling the suggestion shortcut, it was still not working for me, but I added bracket as a trigger and it comes up.

The search string is very weird:

Screen Shot 2020-10-22 at 7 19 19 PM

And I think it should only match on titles instead of anywhere in the note?

I moved the registration to `editorWillMount`, but I'm not sure I understand your comment about it needing to be a singleton.

codebykat:
I fixed the issue where ESC didn't dismiss the popup, but still seeing very odd search behavior. Will have to debug more tomorrow.

codebykat:
It's weird because it seems to start out working okay, but if you do a few searches, it seems to kind of lose the plot... the search results get worse and worse or seem to only search for disconnected characters, as if rather than putting in "mark" you put in "ma", "r", "k".

dmsnell:
@codebykat the weird behavior is that once we return the results to Monaco it continues to narrow them down by its own autocomplete algorithm, which searches inside the previews we supply. that is, we don't have full control over the searching or _narrowing_ behavior.

one thing I thought might be possible is to specify that our results are `incomplete` and maybe if we do that we can continue to provide more results but I also think that might just add more options rather than filter them.

for some reason I can't remember I had to make sure to add the autocompleter as a singleton in the module and prevent it from being recreated when switching notes or when changing to-and-from the preview view.

honestly after my experiments I remember thinking that it might be easier for us to create a new custom dialog and control the autocomplete process ourselves rather than try to make the built-in behavior work, as nice as it is.

codebykat:
I added the incomplete flag and that honestly seems to solve a lot of the strange behavior I was seeing. I also did verify the need for the singleton (in testing and from an upstream developer comment over on the Monaco repo) and added that back.

Humorously, it does do a full text search of the note, meaning that the current note will by definition always be offered as a match.

I think we should limit search to the titles only and exclude the current note and this might actually work out.

dmsnell:
> I think we should limit search to the titles only and exclude the current note and this might actually work out.

Since the other platforms do that I agree, though I wish everything were text search, or full-text but with title priority. There are a couple things we could explore without changing a substantial amount of code:

 - wrap whatever is typed inside the autocomplete area in quotes, which will treat it as a phrase and naturally surface things that might be identically written in a title. this is a trick but could end up being a neat one
 - duplicate search functionality temporarily in this component but only search the title - skip sorting and other features of the search indexer
 - supply a very large number of max results from the search indexer and then narrow down the results looking only at the title. in most cases this would get everything we want and do it quickly, I would expect, except when the search query is too general to be of good use otherwise.

long term I guess adding a title-search to the search indexer is probably a sound bet anyway.

codebykat:
Found this again, just want to leave it here for posterity. There's no API to check if the completions are bound already, and since they are bound to the model, not the editor instance, they can persist, which is why we need to ensure we only add them once. https://github.com/microsoft/monaco-editor/issues/2084#issuecomment-692632522

codebykat:
This is ready for testing! I hacked it into submission. I probably owe @dmsnell a few cat pictures after he tells me what crimes I have committed.

I added extra parameters to the search function to allow searching titles only and excluding an array of IDs. This gives us:
* Search titles only
* Exclude current note

In the process I found the completion provider needed to access the current note ID, so I had to rewire things a little. I ended up removing the singleton tracker and using a state prop and the editor's `onDispose` method.

I also found that the bracket trigger is not sufficient to get the suggestions to pop up when you would expect, so I added a command that binds to Ctrl+Space and only calls `triggerSuggest` if we're inside a bracket. (More or less what was suggested [here](https://github.com/microsoft/monaco-editor/issues/1764#issuecomment-574554086).) Needs refactoring because I copied the code from the completion provider. Also it may need a different keyboard combo for Windows, I need to test that. But I'm going to mark this as ready for review so folks can test it as-is and I can refine it further.

codebykat:
Oh I also tested passing in the typed text as a quoted phrase rather than splitting it up, but I like the current behavior (split by word) better personally.

lib/note-content-editor.tsx
```diff
@@ -167,6 +173,86 @@ class NoteContentEditor extends Component<Props> {
...
+    selectedNoteId: T.EntityId | null
+  ) => languages.CompletionItemProvider = (selectedNoteId) => {
+    return {
+      triggerCharacters: ['['],
+
+      provideCompletionItems(model, position, context, token) {
+        const line = model.getLineContent(position.lineNumber);
+        if (!line.includes('[')) {
+          return null;
+        }
```
dmsnell: we're getting the `precedingOpener` right below. we can probably avoid this check and avoid scanning the line twice.

```ts
const precedingOpener = line.lastIndexOf(…);
if (-1 === precedeingOpener) {
	return null;
}
```

lib/note-content-editor.tsx
```diff
@@ -167,6 +173,86 @@ class NoteContentEditor extends Component<Props> {
...
+              ]
+            : [];
+
+        return {
+          incomplete: true,
+          suggestions: notes.map((note, index) => ({
+            additionalTextEdits,
+            kind: note.isPinned
+              ? languages.CompletionItemKind.Snippet
+              : languages.CompletionItemKind.File,
```
dmsnell, line 252: do we all agree that this is right here? I put it in originally as I was exploring the API but I never had strong feelings about it, or if we should try and do anything better with the icons

lib/note-content-editor.tsx
```diff
@@ -167,6 +173,86 @@ class NoteContentEditor extends Component<Props> {
...
+            label: note.title,
+            // detail: note.preview,
+            documentation: note.content,
+            insertText: `[${note.title}](simplenote://note/${note.noteId})`,
+            sortText: index.toString(),
```
dmsnell, line 257: I believe that this property is used to sort the results. sorting by index should then reflect the sort order in the notes list, as that's how we proceed through notes in the search.

if we want any kind of alphabetical sort we could use `note.title` and if we wanted some relevance search we could perform a quick search distance calculation on the note title against the search query. that would put closer matches at the top

lib/note-content-editor.tsx
```diff
@@ -515,6 +601,38 @@ class NoteContentEditor extends Component<Props> {
...
+        const soFar =
+          precedingBracket >= 0
+            ? line.slice(precedingBracket + 1, position.column)
+            : '';
+
+        if (soFar) {
+          editor.trigger('suggestions', 'editor.action.triggerSuggest', null);
+        }
+      }
+    );
```
dmsnell: nitpick: even though we don't use `this` here I do like using function shorthand syntax since it ensures that if we run into `this` issues we will know that `this` is lexically-scoped and not dynamically-scoped as is done when using `function(){}` syntax

not a big deal - just noted the syntax

lib/note-content-editor.tsx
```diff
@@ -660,6 +779,13 @@ class NoteContentEditor extends Component<Props> {
     this.setState({});
     editor.onDidChangeModelContent(() => this.setDecorators());
 
+    // register completion provider for internal links
+    const completionProviderHandle = monaco.languages.registerCompletionItemProvider(
+      'plaintext',
+      this.completionProvider(this.state.noteId)
+    );
+    editor.onDidDispose(() => completionProviderHandle?.dispose());
```
dmsnell, line 768: good call here. this is exactly how we ought to handle this

lib/note-content-editor.tsx
```diff
@@ -167,6 +173,86 @@ class NoteContentEditor extends Component<Props> {
     }
   };
 
+  completionProvider: (
+    selectedNoteId: T.EntityId | null
+  ) => languages.CompletionItemProvider = (selectedNoteId) => {
+    return {
+      triggerCharacters: ['['],
```
dmsnell, line 180: it doesn't look like we're using this - is that just lost because the diff is hiding lines that aren't changing?

lib/search/index.ts
```diff
@@ -211,6 +232,13 @@ export const middleware: S.Middleware = (store) => {
       }
 
       if (
+        titleOnly &&
+        !searchTerms.every((term) =>
+          note.content.split('\n', 1)[0].includes(term.toLocaleLowerCase())
```
dmsnell: I don't think that this is going to give us what we want.
we might prefer to trap the searchable content into a separate variable and then keep the rest of the search flow the same

```ts

const searchText = titleOnly ? getNoteTitle(note) : note.content;
…
```

I'm pretty sure I don't have that right, but I mean to point out that we have a function that extracts the title exactly how it's done in the note list and in the note editor and that apart from that the searching is performed identically whether the text is truncated or not.

lib/search/index.ts
```diff
@@ -229,6 +257,11 @@ export const middleware: S.Middleware = (store) => {
     return [...pinnedMatches.values(), ...matches.values()];
   };
 
+  searchNotes = (args, maxResults) =>
+    runSearch(args, maxResults)
+      .map((noteId) => [noteId, store.getState().data.notes.get(noteId)])
+      .filter(([, a]) => 'undefined' !== typeof a) as [T.EntityId, T.Note][];
```
dmsnell: I probably wrote this filtering line but we don't need it because we can have `runSearch` perform the check. this is faster and also avoids the case where we request a max number of results and get those, but then prune the list to return fewer than the maximum

```ts
const storeNotes = store.getState().data.notes;
…
const note = notes.get(noteId);
if (!note || !storeNotes.has(noteId)) {
	continue;
}
```

dmsnell:
looking good!

lib/note-content-editor.tsx
codebykat: I think we can do something better with it in a follow-up 🤷

lib/search/index.ts
codebykat: ~Hmm I did this and it seems to work but I've made TypeScript sad.~ nvm got it

belcherj (approved):
Looks good!!

codebykat:
This could use some refinement and I'm not sure if the trigger key interaction is going to annoy people (pops up automatically when you type a `[`), but I have addressed the review comments and fixed the bugs I know about, so let's merge this and we can build on it!

lib/search/index.ts
```diff
@@ -52,8 +55,14 @@ const tagsFromSearch = (query: string) => {
   return searchTags;
 };
 
+export let searchNotes: (
+  args: Partial<SearchState>,
+  maxResults: number
+) => [T.EntityId, T.Note | undefined][] = () => [];
```
dmsnell, line 61: this `undefined` jumps off the screen - why would we return results for notes that don't exist?

should we not be filtering out those notes before returning them? what is a caller supposed to do if they get a note id without a note?

codebykat: @belcherj and I were of consensus that it would never return `undefined`, but TypeScript was unhappy, so we added this. Please do fix in a follow-up PR if you know the fix for it :)

PR #2289, opened 2020-08-21
[Rewrite Branch] Add debounce of 2 seconds to analytics for note edit

belcherj:
### Fix

Current production has a debounce of 2 seconds on `editor_note_edited`. Additionally, it could have a performance impact sending analytics on every key press. I was not able to see any performance impact on staging.

Note we are not sending what was written only that an edit was made.

### Test
1. Analytics does not run on dev so you will have to comment out: https://github.com/Automattic/simplenote-electron/blob/develop/lib/analytics/index.ts#L58 to test locally
2. Turn on analytics
3. Open Chrome dev tools to the network panel
4. Type something in a note
5. Notice there was a 2-second delay before network call gets sent

### DIFF
--- a/lib/state/analytics/middleware.ts
+++ b/lib/state/analytics/middleware.ts
@@ -1,8 +1,8 @@
+import { debounce } from 'lodash';
+
 import analytics from '../../analytics';
 import getConfig from '../../../get-config';
 import isDevConfig from '../../utils/is-dev-config';
-
-import { debounce } from 'lodash';
 
 const config = getConfig();
 
@@ -36,6 +36,7 @@ export const middleware: S.Middleware = (store) => {
         return;
     }
   };
+
   const recordNoteEdit = debounce(() => record('editor_note_edited'), 2000);
 
   return (next) => (action: A.ActionType) => {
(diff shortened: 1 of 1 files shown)


codebykat (approved):
Thank you!!

lib/state/analytics/middleware.ts
```diff
@@ -2,6 +2,8 @@ import analytics from '../../analytics';
 import getConfig from '../../../get-config';
 import isDevConfig from '../../utils/is-dev-config';
 
+import { debounce } from 'lodash';
```
dmsnell: we don't have any hard rules in this repo but in most Automattic JS code we put external exports at the top. would be nice if we could move this before merging but it's not a blocker.

lib/state/analytics/middleware.ts
```diff
@@ -34,6 +36,7 @@ export const middleware: S.Middleware = (store) => {
         return;
     }
   };
+  const recordNoteEdit = debounce(() => record('editor_note_edited'), 2000);
```
dmsnell, line 40: nitpick: we need a blank line above this.

dmsnell (approved):
There are a couple things I think it would be worth updating here but it's good to merge after that/otherwise. Thanks for adding this in and preventing our stats from surging artificially.

codebykat:
Nits picked

dmsnell (approved):
Ship it!

PR #2290, opened 2020-08-23
Add document highlight provider to highlight search results

dmsnell:
In this patch we're adding a new document highlight provider to highlight the
search results in a note. This may be a useful approach because it relies
entirely on Monaco's API to provide the functionality. Unfortunately in my
exploration it wasn't clear to me to how get this to highlight when the cursor
isn't on any word. The API appears to be designed to highlight related words to
the word under the cursor. If the cursor is on an empty line then nothing else
gets highlighted and the provider never gets called.

If we can force-trigger the highlighting to run this might be workable. We need
to do that and also be able to select an active highlight if we want to use
this for search.

We might have `editor.action.wordHighlight.trigger` as an action to call but I
gave up trying to figure out how to properly call it.

highlightProvider mov

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -9,9 +9,10 @@ import { editor as Editor, Selection, SelectionDirection } from 'monaco-editor';
 
 import actions from './state/actions';
 import * as selectors from './state/selectors';
+import { getTerms } from './utils/filter-notes';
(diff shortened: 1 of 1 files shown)


belcherj:
Closing for now as this approach does not seem viable. #2292 adds this functionality.

PR #2292, opened 2020-08-24
Adds highlighting in note for search query

belcherj:
### Fix

Alternative to: https://github.com/Automattic/simplenote-electron/pull/2290 and https://github.com/Automattic/simplenote-electron/pull/2276

In this approach to search highlighting we are using Monaco decorations. We have added a search function that will find all of the terms we want to decorate. We create decorations from their ranges and combine that with the note title decoration.

The decoration function is run when editor is ready, on editor.onDidChangeModelContent, and in componentDidUpdate when the search query does not equal the previous search query.

```
{
  options: {
    inlineClassName: 'searchDecoration',
    overviewRuler: { color: '#3361cc' },
  },
  range: {
    startLineNumber: start.lineNumber,
    startColumn: start.column,
    endLineNumber: end.lineNumber,
    endColumn: end.column,
  },
}
```

### Test
1. Enter a search term
2. Search using a tag
3. Cancel search
4. Edit a result in the note so it is no longer a match, does highlight disappear?
5. Add a match, does it get highlighted?

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -267,8 +267,6 @@ class NoteContentEditor extends Component<Props> {
         'editor.foreground': '#2c3338', // $studio-gray-80 AKA theme-color-fg
         'editor.background': '#ffffff',
         'editor.selectionBackground': '#ced9f2', // $studio-simplenote-blue-5
-        'editor.selectionHighlightBackground': '#3361cc', // $studio-simplenote-blue-50
-        'editorOverviewRuler.selectionHighlightForeground': '#3361cc', // $studio-simplenote-blue-50
         'scrollbarSlider.activeBackground': '#8c8f94', // $studio-gray-30
         'scrollbarSlider.background': '#c3c4c7', // $studio-gray-10
         'scrollbarSlider.hoverBackground': '#a7aaad', // $studio-gray-20
@@ -283,8 +281,6 @@ class NoteContentEditor extends Component<Props> {
         'editor.foreground': '#ffffff',
         'editor.background': '#1d2327', // $studio-gray-90
         'editor.selectionBackground': '#50575e', // $studio-gray-60
-        'editor.selectionHighlightBackground': '#3361cc', // $studio-simplenote-blue-50
-        'editorOverviewRuler.selectionHighlightForeground': '#3361cc', // $studio-simplenote-blue-50
         'scrollbarSlider.activeBackground': '#646970', // $studio-gray-50
         'scrollbarSlider.background': '#2c3338', // $studio-gray-80
         'scrollbarSlider.hoverBackground': '#1d2327', // $studio-gray-90
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -19,7 +19,7 @@ import * as S from './state';
 import * as T from './types';
 
 const SPEED_DELAY = 120;
-let decorations = [];
+let decorations: string[] = [];
 
 type OwnProps = {
   storeFocusEditor: (focusSetter: () => any) => any;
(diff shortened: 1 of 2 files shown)


dmsnell:
Liking how this is shaping up. Can we keep this open until after we've started the staged rollout?

A couple quick thoughts:
 - this model gives me higher confidence that we'll highlight the right ranges
 - we should evaluate performance when editing large notes
 - if performance is an issue we can look at moving this computation into a web worker. if it's bad enough to need that now we probably want to defer merging until that's done well. moving into a worker not only implies asynchronous messaging but also it means that once the messages returns we need to verify that we're still working against the same raw note content as when we sent the message. it could be that computation in a worker is too slow and the editor updates in between the time we request the computation and when it returns.

belcherj:
This is a blocker for staged rollout. We cannot start rolling out without search highlighting.

belcherj:
No performance issues in a note with 1.5 million characters.

codebykat (approved):
This is looking good to me!

In a follow-up I'd like to see if we can make the scrollbar highlights look a little more polished. It seems kind of odd to me that they aren't the same width as the scrollbar.

We also need to enable the keyboard shortcut for "jump to next match". I'm okay with that being a follow-up too -- I'll add it to the big list.

lib/note-content-editor.tsx
```diff
@@ -9,15 +9,17 @@ import { editor as Editor, Selection, SelectionDirection } from 'monaco-editor';
...
-  withCheckboxSyntax,
   withCheckboxCharacters,
+  withCheckboxSyntax,
 } from './utils/task-transform';
 
 import * as S from './state';
 import * as T from './types';
 
 const SPEED_DELAY = 120;
+let decorations: string[] = [];
```
dmsnell: it's convenient for us to use a module-global variable but only because we never have more than one editor displayed at a time. putting this inside the class as a class member would be more reliable, for example, even when writing tests.

lib/note-content-editor.tsx
```diff
@@ -150,8 +152,96 @@ class NoteContentEditor extends Component<Props> {
...
+  setDecorators = () => {
+    const searchMatches = this.searchMatches();
+    const titleDecoration = this.getTitleDecoration();
+    if (!searchMatches.length && !titleDecoration) {
+      return;
+    }
+    if (titleDecoration) {
+      searchMatches.push(titleDecoration);
+    }
+    decorations = this.editor.deltaDecorations(decorations, [...searchMatches]);
```
dmsnell: we have some syntax confusion. on one hand we're _spreading_ something that's already a list and on the other we're mutating a list that's later spread.

we can resolve this mismatch in one of two ways: make everything immutable, or make everything mutable. you know my preference would be towards the immutable…

```ts
setDecorators = () => {
     const searchMatches = this.searchMatches() ?? [];
     const titleDecoration = this.getTitleDecoration() ?? [];

     this.decorations = this.editor.deltaDecorations(
          this.decorations,
          [...searchMatches, ...titleDecoration]
     );
}
```

in this snippet I've made three changes:
 - I've used `this.decorations` assuming that we've [made `decorations` a class instance member](https://github.com/Automattic/simplenote-electron/pull/2292/files#r475878496) of the `NoteContentEditor` vs. a module-global shared by all instances
 - in the case that one of the components has no matches we're defaulting to `[]` so we can safely spread them
 - we're _always_ updating the decorations. if we early-abort when there are no new decorations then we'll continue to highlight the old ranges when we are supposed to stop highlighting altogether

lib/note-content-editor.tsx
```diff
@@ -150,8 +152,96 @@ class NoteContentEditor extends Component<Props> {
...
+      return;
+    }
+    const titleDecoration = (line: number) => ({
+      range: new monaco.Range(line, 1, line, 1),
+      options: {
+        isWholeLine: true,
+        inlineClassName: 'note-title',
+        stickiness: Editor.TrackedRangeStickiness.NeverGrowsWhenTypingAtEdges,
+      },
+    });
```
dmsnell: as this is a static function we can move it into the module global scope above `class NoteContentEditor`

won't make a meaningful impact but it will do less than defining it here does in the VM and it will keep this function smaller. maybe we should consider a rename: `titleDecorationForLine(line: number)`

dmsnell (commented):
When I was testing #2290 I found some problem with literal phrases. that is, if I search for `cat dog` it will search for either `cat` or `dog`, when I searched for `"cat dog"` it will only search for the sequence `cat dog`, but if I search for `"cat dog" bird` it would only search for `cat dog` when I want it to search for either `cat dog` or `bird`

not sure why or if that has to do with `getTerms()` - just something we might want to look over.

not a blocker because the search is pretty good already and this could be related to something already existing. needs an audit.

lib/note-content-editor.tsx
```diff
@@ -9,16 +9,26 @@ import { editor as Editor, Selection, SelectionDirection } from 'monaco-editor';
...
+  withCheckboxSyntax,
 } from './utils/task-transform';
 
 import * as S from './state';
 import * as T from './types';
 
 const SPEED_DELAY = 120;
 
+const titleDecorationForLine = (line: number) => ({
+  range: new monaco.Range(line, 1, line, 1),
```
dmsnell, line 24: I need to update code - I've been using `0`

oops

lib/note-content-editor.tsx
```diff
@@ -150,8 +161,86 @@ class NoteContentEditor extends Component<Props> {
         }
       }, 400);
     }
+
+    if (this.editor && prevProps.searchQuery !== this.props.searchQuery) {
```
dmsnell: we also need to gate this on `this.state.editor === 'full'` otherwise we'll be trying to update the decorators for a plain `<div>`

belcherj: Done

PR #2294, opened 2020-08-24
Add missing style to fix react-virtualized autoSizing

belcherj:
### Fix

This PR adds the styles to the tags and note list header so that they resize properly. Before, as the number of tags appearing in the note list changed the vertical layout was broken.

I have no idea why this fixes it but react-virtualized was throwing a warning so it seemed like a good place to start.

### Test
1. Have notes with the word "said"
2. Add tags for "s", "sa", "sai", and "said"
3. Type and delete the letters: s, a, i, and d into the search
4. Do the note list heights resize properly?

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -87,7 +87,7 @@ const renderNote = (
         parent={parent}
         rowIndex={index}
       >
-        <TagSuggestions />
+        <TagSuggestions style={{ ...style }} />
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
Sweet. In hindsight it seems obvious. They're adding height properties and we were preventing them. I've wondered about this one ever since I added React Virtualized. Never could figure it out.

PR #2297, opened 2020-08-25
Fixes the width of content in the editor

belcherj:
### Fix

Notes have a maximum width which is incorrect. This makes the max column width 400. Using the bounded strategy it will fill viewport till it hits 400 columns. By using advanced it fills in available space better.

Before:

Screen Shot 2020-08-24 at 10 13 43 PM

After:

Screen Shot 2020-08-24 at 10 13 06 PM

### Test
1. Have note with long line length
2. Make your screen really wide
3. Observe that note grows.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -10,6 +10,7 @@ import { editor as Editor, Selection, SelectionDirection } from 'monaco-editor';
 import actions from './state/actions';
 import * as selectors from './state/selectors';
 import { getTerms } from './utils/filter-notes';
+import { isSafari } from './utils/platform';
 import {
   withCheckboxCharacters,
   withCheckboxSyntax,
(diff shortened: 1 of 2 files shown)


dmsnell (commented):
We started out with advanced but I think it crashed Safari. Will want to verify that this works on all browsers. The advanced wrap was much better. I figured that at some point we would resort to a browser check so we could provide it where it works.

codebykat (approved):
This doesn't fix my random scrollbar issues sadly, but it IS a huge improvement for the layout. I tested in Safari, Firefox and Electron and didn't see any issues!

~I do wish we could fix the weirdly large right gutter. Do we know what's up with that?~ #2298 

Update: I think there might be something up with the performance on this branch. I saw slow loading times while testing something else -- occasional long delays on page reload before the note was displayed. Not positive it's worse here but it might be.

belcherj:
Put in a fix for Safari as it is still breaking there. Going to merge, this is an easy one to undo.

PR #2298, opened 2020-08-25
[rewrite branch] Align the scrollbar to the right

codebykat:
### Fix
The huge gutter to the right of the scrollbar was really bothering me. I tried a million things but there's seriously no good way to do this with padding of the parent containers...

So, this is maybe terrible, but we can trick Monaco by claiming the scrollbar is as wide as we want our padding, then use CSS to resize it back down. (Props to https://github.com/microsoft/monaco-editor/issues/1368 for this idea.)

I originally tried to (ab)use the `lineDecorationsWidth` for left padding, but noticed that checkbox lines are clickable anywhere in that margin, which is kind of annoying. So I removed it.

We'll need to test and see if this messes up the search highlights or anything else.

### Test
1. Set line width of both narrow and full
2. Look at some long notes with long lines
3. Make sure the right scrollbar is always flush to the right
4. Make sure Markdown preview is scrollable and has scrollbars that make sense

### Screenshots

Before:
Screen Shot 2020-08-25 at 12 00 45 AM

After (this will also look better after #2297):

Screen Shot 2020-08-24 at 11 59 46 PM

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -617,6 +617,11 @@ class NoteContentEditor extends Component<Props> {
         return;
       }
 
+      // a range spanning more than one column means we're over the gutter
+      if (range.endColumn - range.startColumn > 1) {
+        return;
+      }
+
       const model = editor.getModel();
       if (!model) {
         return;
@@ -640,6 +645,11 @@ class NoteContentEditor extends Component<Props> {
       } = event;
 
       if (!range) {
+        return;
+      }
+
+      // a range spanning more than one column means we're over the gutter
+      if (range.endColumn - range.startColumn > 1) {
         return;
       }
 
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { Component, createRef, RefObject } from 'react';
 import { connect } from 'react-redux';
 import Monaco, {
   ChangeHandler,
@@ -75,6 +75,7 @@ class NoteContentEditor extends Component<Props> {
   bootTimer: ReturnType<typeof setTimeout> | null = null;
   editor: Editor.IStandaloneCodeEditor | null = null;
   monaco: Monaco | null = null;
+  contentDiv: RefObject<HTMLDivElement> = createRef();
   decorations: string[] = [];
   matchesInNote: [] = [];
 
@@ -811,8 +812,23 @@ class NoteContentEditor extends Component<Props> {
     const { fontSize, noteId, searchQuery, theme } = this.props;
     const { content, editor, overTodo, selectedSearchMatchIndex } = this.state;
     const searchMatches = searchQuery ? this.searchMatches() : [];
+
+    let editorPadding = 25;
+
+    if (lineLength === 'narrow' && this.contentDiv?.current) {
+      const width = this.contentDiv.current?.offsetWidth;
+      if (width <= 1400) {
+        // should be 10% up to 1400px wide
+        editorPadding = width * 0.1;
+      } else {
+        // after 1400, calc((100% - 768px) / 2);
+        editorPadding = (width - 768) / 2;
+      }
+    }
+
     return (
       <div
+        ref={this.contentDiv}
         className={`note-content-editor-shell${
           overTodo ? ' cursor-pointer' : ''
         }`}
@@ -844,6 +860,7 @@ class NoteContentEditor extends Component<Props> {
                 '"Simplenote Tasks", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell", "Helvetica Neue", sans-serif',
               fontSize,
               hideCursorInOverviewRuler: true,
+              lineDecorationsWidth: 0,
               lineHeight: fontSize > 20 ? 42 : 24,
               lineNumbers: 'off',
               links: true,
@@ -854,7 +871,11 @@ class NoteContentEditor extends Component<Props> {
               quickSuggestions: false,
               renderIndentGuides: false,
               renderLineHighlight: 'none',
-              scrollbar: { horizontal: 'hidden', useShadows: false },
+              scrollbar: {
+                horizontal: 'hidden',
+                useShadows: false,
+                verticalScrollbarSize: editorPadding,
+              },
               scrollBeyondLastLine: false,
               selectionHighlight: false,
               wordWrap: 'bounded',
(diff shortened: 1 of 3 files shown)


codebykat:
~Back to draft for now. Need to figure out padding in Markdown preview mode.~ OK fixed.

lib/note-content-editor.tsx
```diff
@@ -74,6 +75,7 @@ class NoteContentEditor extends Component<Props> {
 
   state: OwnState = {
     content: '',
+    contentDiv: createRef(),
     editor: 'fast',
```
dmsnell: This is a rare situation but we might want to set the `contentDiv` directly on the class instance as `this.contentDiv` instead of through `this.state.contentDiv`

While it works here the refs operate differently than normal values. It matches the directly-mutable model we usually avoid in React more than it does the one-way data flow we adhere to.

belcherj:
Screen Shot 2020-08-25 at 9 42 26 AM

This blows up search. Let's wait on this one for more testing.

codebykat:
> This blows up search. Let's wait on this one for more testing.

Dang it. I sort of thought it might. I'll see if I can CSS it into submission after rebasing the new changes.

codebykat:
Search highlights were at least straightforward to fix:

Screen Shot 2020-08-25 at 10 15 38 AM

@belcherj can you test this some more? 😅

lib/note-content-editor.tsx
codebykat: @dmsnell should it look like this? https://github.com/Automattic/simplenote-electron/pull/2298/commits/535393d9dc11ffe844c4be2d465cfc2eec7ee4fc

TypeScript makes a sadface about how `Property 'contentDiv' does not exist on type 'NoteContentEditor'`.

belcherj:
Staging:
Screen Shot 2020-08-27 at 11 00 37 AM

This branch:
Screen Shot 2020-08-27 at 11 00 21 AM

The really long title is an existing issue but the width seems off and the editor is not centered.

lib/note-content-editor.tsx
```diff
@@ -113,6 +113,7 @@ class NoteContentEditor extends Component<Props> {
     }, SPEED_DELAY);
     this.props.storeFocusEditor(this.focusEditor);
     this.props.storeHasFocus(this.hasFocus);
+    this.contentDiv = createRef();
```
dmsnell: ooh ooh! this can go up above right next to `decorations: string[] = []`

this is part of the fun of the class syntax. we'll additionally get class-level typing by putting it there and since it's there always we won't have to convince TypeScript that it truly exists - it will know.

lib/note-content-editor.tsx
```diff
@@ -731,11 +732,25 @@ class NoteContentEditor extends Component<Props> {
...
 
   render() {
-    const { fontSize, noteId, theme } = this.props;
+    const { fontSize, lineLength, noteId, theme } = this.props;
     const { content, editor, overTodo } = this.state;
 
+    let editorPadding = 25;
+
+    if (lineLength === 'narrow' && this.contentDiv?.current) {
+      const width = this.contentDiv?.current?.offsetWidth;
```
dmsnell: this is the line that will lose a `?` once `contentDiv` is defined at the top of the class

```ts
const width = this.contentDiv.current?.offsetWidth;
```

scss/theme.scss
```diff
@@ -27,18 +27,43 @@
   unicode-range: U+E000-E001;
 }
 
-.note-detail {
+.note-detail-textarea {
+  padding-left: calc((100% - 768px) / 2);
```
dmsnell: every time I see `calc()` in CSS I'm amazed.

lib/note-content-editor.tsx
codebykat: Ooh nice, this made TS happy

codebykat:
I think the long line lengths issue came in via the upstream fix to wrapping, I will investigate.

codebykat:
A few updates to this. The 120% `font-size` on the first line is causing a lot of our problems, I took it out for now.

The last thing I need to solve here is the fact that you can click anywhere in the left margin to check/uncheck a checkbox o_O

codebykat:
Checkbox clicking fixed.

Safari is leaving extra space on the right, don't know why yet :/

Update: [I have my suspicions](https://github.com/microsoft/vscode/blob/f74e473238aca7b79c08be761d99a0232838ca4c/src/vs/editor/browser/viewParts/overviewRuler/decorationsOverviewRuler.ts#L84)

dmsnell:
@codebykat might as well conditionally update state to avoid needless updates…

```ts
const overTodo = …;
if (this.state.overTodo !== overTodo) {
  this.setState({overTodo});
}
```

belcherj:
The note area shifts when switching notes even when a note is not using the fast trick for long notes. Can you rebase? I think I had some tweaks to fix that in rewrite/no-races

codebykat:
@belcherj rebased and I think that did fix it!

@dmsnell addressed this comment too 👍 

I think the Safari issues are pre-existing since the line wrap is shorter on Safari, so this is ready for more review. It tests well for me on OSX/Firefox, Chrome, Safari (though alignment looks off on Safari), Windows/Edge and Firefox.

I don't see any way to re-implement the title size decoration without throwing off Monaco's built-in measurement of its container, unfortunately.

lib/note-content-editor.tsx
dmsnell: we should create the `createRef()` before `componentDidMount()`. it's going to be easiest to declare it on the class instance even outside the constructor function. we can remember here that React is creating a class (which initializes the JavaScript class) and _then_ calling `componentDidMount()` only _after_ the actual virtual DOM nodes have been returned from `render()`. since we are passing that ref into `render()` it should be there already

```ts
class NoteContentEditor extends Component<Props> {
	bootTimer… = null;
	editor: Editor.IStandaloneCodeEditor | null = null;
	…
	contentDiv = createRef<HTMLDivElement>();

	state: OwnState = {
		…
	}

	…
}
```

note that TypeScript will infer the type of `contentDiv` for us and we won't have to import `RefObject`

scss/theme.scss
```diff
@@ -54,7 +67,7 @@
   font-family: 'Simplenote Tasks', sans-serif;
 
   .note-title {
-    font-size: 120%;
+    // font-size: 120%;
```
dmsnell: is this commented-out on purpose? if so we should explain why in a comment

codebykat:
Here's a before/after of Safari on a wide monitor.

Narrow line width:

Screen Shot 2020-09-09 at 3 09 36 PM

Screen Shot 2020-09-09 at 3 10 07 PM

Full width:

Screen Shot 2020-09-09 at 3 10 26 PM

  
Screen Shot 2020-09-09 at 3 10 31 PM

scss/theme.scss
codebykat: Haha it's removed on purpose but I should remove the comment, yes 😂

dmsnell (approved):
Trusting your testing of this. Thanks for working through a few code things with me. People are going to be happy to have this in.

PR #2299, opened 2020-08-25
[rewrite branch] Too many scrollbars

codebykat:
### Fix

I was testing on Windows/Edge and noticed we've got some scrollbars on note-detail and friends that are cluttering up the place.

The Monaco container is inside of .note-detail, and handles its own overflows and scrolling, so we don't need these.

Special-cased the Markdown preview; make sure you can scroll in preview if necessary!

### Screenshots

Before:

Screen Shot 2020-08-24 at 9 17 14 PM

After:

Screen Shot 2020-08-25 at 12 07 24 AM

### DIFF
--- a/scss/print.scss
+++ b/scss/print.scss
@@ -22,7 +22,6 @@
   .app-layout,
   .app-layout__note-column,
   .note-editor,
-  .note-detail-wrapper,
   .note-detail,
   .note-detail-textarea,
   .note-content-editor-shell,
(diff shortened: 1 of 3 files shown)


scss/print.scss
```diff
@@ -22,7 +22,6 @@
   .app-layout,
   .app-layout__note-column,
   .note-editor,
-  .note-detail-wrapper,
```
codebykat, line 25: Preview is cutting this off but it's a duplicate rule.

belcherj (approved):
Tested in Safari, Edge, Chrome, Firefox

PR #2302, opened 2020-08-26
Update the tag panel UI

belcherj:
### Fix

This PR updates the tag panel UI to match the new designs. It also fixes a bug where the dragged tag does not know that it should style dark mode.

Designs:

Screen Shot 2020-09-21 at 4 39 11 PM
Screen Shot 2020-09-21 at 4 39 27 PM

App:

image
image

### Test
1. Open up tag panel
2. Have a few tags, can you edit, reorder, delete
3. View the panel in light and dark mode
4. View it in several browsers
5. If you have a log of tags does it scroll?

### DIFF
--- a/lib/tag-list/index.tsx
+++ b/lib/tag-list/index.tsx
@@ -10,9 +10,11 @@ import {
 import isEmailTag from '../utils/is-email-tag';
 import PanelTitle from '../components/panel-title';
 import ReorderIcon from '../icons/reorder';
-import TrashIcon from '../icons/trash';
+import CrossOutlineIcon from '../icons/cross-outline';
 import TagListInput from './input';
 import { openTag, toggleTagEditing } from '../state/ui/actions';
+
+import * as selectors from './../state/selectors';
 
 import type * as S from '../state';
 import type * as T from '../types';
@@ -21,6 +23,7 @@ type StateProps = {
   editingTags: boolean;
   openedTag: T.TagHash | null;
   sortTagsAlpha: boolean;
+  theme: 'light' | 'dark';
   tags: Map<T.TagHash, T.Tag>;
 };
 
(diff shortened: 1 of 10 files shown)


belcherj:
This is ready for review

lib/tag-list/index.tsx
```diff
@@ -116,7 +128,7 @@ export class TagList extends Component<Props> {
   reorderTag: SortEndHandler = ({ newIndex, nodes, oldIndex }) => {
     const tagName = nodes[oldIndex].node.dataset.tagName;
 
-    this.props.reorderTag(tagName, newIndex);
+    this.props.reorderTag(tagName, newIndex + 1);
```
dmsnell: What was broken before that `+1` fixes? That is, what was causing the wrong behavior and why does `newIndex` gives us the wrong value?

belcherj: It was giving us the correct value. Our reorder logic was flawed. I fixed out logic.

belcherj: https://github.com/Automattic/simplenote-electron/pull/2302/files#diff-159d5dbfee67d319843d07f814f1e2115598b5b40e076aef780b73e09c1cadf3R417

codebykat (changes requested):
This is looking good, but there seems to be an off-by-one bug when dragging-and-dropping. The tag seems to always be moved one tag below where it should be:

tag-move-test

belcherj:
Odd, my code should fix that issue. Can you do a fresh pull and try again?


Oct-29-2020 18-33-39

codebykat:
Still happening 🧐

I did a `git pull` on `develop`, deleted the entire branch and checked it out again. Seeing the behavior on Firefox, Safari, Electron.

codebykat (approved):
This is working now 👍 Can you update the description about the long-standing index bug to avoid future confusion?

belcherj:
This PR updates the tag panel UI to match the new designs. It also fixes a bug where the dragged tag does not know that it should style dark mode.

PR #2305, opened 2020-08-27
Only request revisions when the revisions panel is opened

belcherj:
### Fix

Only request revisions when the revisions panel is opened. We are making a lot of Simperium requests when they are unneeded.

### Test
1. Test revisions panel

### DIFF
--- a/lib/state/simperium/middleware.ts
+++ b/lib/state/simperium/middleware.ts
@@ -274,9 +274,8 @@ export const initSimperium = (
       }
 
       case 'REVISIONS_TOGGLE': {
-        const state = nextState;
-        const showRevisions = state.ui.showRevisions;
-        const noteId = state.ui.openedNote;
+        const showRevisions = nextState.ui.showRevisions;
+        const noteId = nextState.ui.openedNote;
 
         if (noteId && showRevisions) {
           noteBucket.getRevisions(noteId).then((revisions) => {
--- a/lib/state/simperium/middleware.ts
+++ b/lib/state/simperium/middleware.ts
@@ -274,7 +274,7 @@ export const initSimperium = (
       }
 
       case 'REVISIONS_TOGGLE': {
-        const state = getState();
+        const state = nextState;
         const showRevisions = state.ui.showRevisions;
         const noteId = state.ui.openedNote;
 
--- a/lib/state/simperium/middleware.ts
+++ b/lib/state/simperium/middleware.ts
@@ -255,15 +255,20 @@ export const initSimperium = (
         //  Preload the revisions when opening a note but only do it if no revisions are in memory
         if (noteId && !state.data.noteRevisions.get(noteId)?.size) {
           setTimeout(() => {
-            noteBucket.getRevisions(noteId).then((revisions) => {
-              dispatch({
-                type: 'LOAD_REVISIONS',
-                noteId: noteId,
-                revisions: revisions
-                  .map(({ data, version }): [number, T.Note] => [version, data])
-                  .sort((a, b) => a[0] - b[0]),
+            if (getState().ui.openedNote === noteId) {
+              noteBucket.getRevisions(noteId).then((revisions) => {
+                dispatch({
+                  type: 'LOAD_REVISIONS',
+                  noteId: noteId,
+                  revisions: revisions
+                    .map(({ data, version }): [number, T.Note] => [
+                      version,
+                      data,
+                    ])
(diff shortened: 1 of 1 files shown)


codebykat:
Code looks good, this tests fine. If we can't preload revisions we ultimately will need to add a loading state for the revisions slider.

codebykat (approved):

lib/state/simperium/middleware.ts
```diff
@@ -246,27 +246,45 @@ export const initSimperium = (
       case 'FILTER_NOTES':
       case 'OPEN_NOTE':
       case 'SELECT_NOTE': {
+        const state = getState();
```
dmsnell: this is already available from above as `prevState` and `nextState` depending on if you want the state before or after the reducers updated. in this case it shouldn't matter because these actions won't affect revisions. might as well use `nextState`

lib/state/simperium/middleware.ts
```diff
@@ -246,27 +246,45 @@ export const initSimperium = (
...
         const noteId =
           action.noteId ??
           action.meta?.nextNoteToOpen ??
           getState().ui.openedNote;
 
-        if (noteId) {
+        //  Preload the revisions when opening a note but only do it if no revisions are in memory
+        if (noteId && !state.data.noteRevisions.get(noteId)?.size) {
```
dmsnell: let's leave `getState().ui.openedNote === noteId` in there. that is actually a guard against this same kind of proliferation. what's happening is that when jumping through the list of notes we end up opening every one for about 50ms and we don't want to issue requests for all those all at once.

lib/state/simperium/middleware.ts
```diff
@@ -246,12 +246,14 @@ export const initSimperium = (
...
           setTimeout(() => {
             if (getState().ui.openedNote === noteId) {
```
dmsnell, line 264: note that here we're using `getState()` because this code is run inside the callback and `prevState` and `nextState` will already be expired. this code runs 250ms in the future and we have to call `getState()` again to the get the up-to-date value

lib/state/simperium/middleware.ts
```diff
@@ -269,6 +271,27 @@ export const initSimperium = (
...
+        return result;
+      }
+
+      case 'REVISIONS_TOGGLE': {
+        const state = getState();
+        const showRevisions = state.ui.showRevisions;
+        const noteId = state.ui.openedNote;
+
+        if (noteId && showRevisions) {
+          setTimeout(() => {
```
dmsnell: we may not need this delay here since we know we're opening the revisions slider. it's probably not bad to keep in in case someone opens and immediately closes it - we wouldn't need to issue the request. if that's the case though we'd want to check `showRevisions` from inside the `setTimeout` callback

probably the most straightforward thing here would be to remove the timeout and issue the fetch every time we open up the slider.

lib/state/simperium/middleware.ts
```diff
@@ -269,6 +270,27 @@ export const initSimperium = (
             }
           }, 250);
         }
+        return result;
+      }
+
+      case 'REVISIONS_TOGGLE': {
+        const state = getState();
```
dmsnell: can use `nextState` here - it's important because if we use `prevState` then `showRevisions` will be opposite of what the toggle flips it to

belcherj (commented):
Tests well merge it.

lib/state/simperium/middleware.ts
```diff
@@ -251,7 +251,8 @@ export const initSimperium = (
           action.meta?.nextNoteToOpen ??
           getState().ui.openedNote;
 
-        if (noteId) {
+        //  Preload the revisions when opening a note but only do it if no revisions are in memory
+        if (noteId && !nextState.data.noteRevisions.get(noteId)?.size) {
```
dmsnell: we might need to add a new flag saying "we're getting them" otherwise we might issue a bunch of requests in the time before the first one comes back.

something like a set in this middleware could do.

```ts
const noteQueue = new BucketQueue(noteBucket);
const queueNoteUpdate = …

const hasRequestedRevisions = new Set<T.EntityId>();
```

then when checking here…

```ts
if (noteId && !nextState.data.noteRevisions.get(noteId)?.size && !hasRequestedRevisions.has(noteId)) {
	hasRequestedRevisions.add(noteId);
	setTimeout( … );
	…
}
```

dmsnell (approved):
These look like they should stop frantically re-requesting the revisions while still maintaining the pre-fetch.

PR #2308, opened 2020-08-28
[rewrite branch] Only dispatch SET_ANALYTICS if it has changed, take 2

codebykat:
### Fix

#2306 take two

### DIFF
--- a/lib/state/simperium/middleware.ts
+++ b/lib/state/simperium/middleware.ts
@@ -19,7 +19,6 @@ import { stopSyncing } from '../persistence';
 import type * as A from '../action-types';
 import type * as S from '../';
 import type * as T from '../../types';
-import { update } from 'lodash';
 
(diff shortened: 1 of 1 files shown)


lib/state/simperium/middleware.ts
```diff
@@ -155,10 +155,12 @@ export const initSimperium = (
       return;
     }
 
-    dispatch({
-      type: 'SET_ANALYTICS',
-      allowAnalytics: !!updatedEntity.analytics_enabled,
-    });
+    if (updatedEntity.analytics_enabled !== getState().data.analyticsAllowed) {
```
dmsnell: Because Simperium may use non-`true` truthy values we need to coerce to Boolean here with `!!updatedEntity.analytics_enabled`

dmsnell (approved):

PR #2312, opened 2020-09-01
[rewrite branch] Enable Monaco find widget

codebykat:
### Fix

This PR uses Monaco's Find widget to provide search within note (find in page) functionality.

### Todo // Known Issues & Questions
* The title of the note gets bumped down when the widget is shown
* Scrollbar highlight color is wrong
* If text focus is within the find widget, the combo to toggle between tag/editor focus erroneously toggles to the editor text, and leaves the find widget visible
* In the browser only, if the editor is not focused, Find and Find Next get eaten. We disable the browser behavior but it would be ideal if we could focus the editor and trigger the editor action. (This works in Electron because we don't override the default behavior, since it doesn't have one.)
* Search prefills with whatever word you're next to, or if you're next to punctuation, were previously closest to, which is kind of confusing. Probably this isn't customizable. Are we okay with it?
* Not sure about the word highlight colors (different for current match / all matches in page) -- check w/Sly
* Should this interact better with our existing / global search functionality? Right now they are totally separate which is somewhat odd. IF we do want that maybe we should use different highlight colors?

### DIFF
--- a/desktop/menus/edit-menu.js
+++ b/desktop/menus/edit-menu.js
@@ -56,6 +56,12 @@ const buildEditMenu = (settings, isAuthenticated) => {
         click: editorCommandSender({ action: 'find' }),
         accelerator: 'CommandOrControl+F',
       },
+      {
+        label: 'Find Again',
+        visible: isAuthenticated,
+        click: editorCommandSender({ action: 'findAgain' }),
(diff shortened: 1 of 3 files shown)


codebykat:
Closing in favor of https://github.com/Automattic/simplenote-electron/pull/2313

PR #2313, opened 2020-09-01
[Rewrite Branch] Adds the new search interface

belcherj:
### Fix

Implements the new search interface. This PR captures the browser "finds" (cmd+f) and displays a search UI that allows users to navigate through results. This also enables the user to use the (cmd+g) and (cdm+shift+g)

Screen Shot 2020-09-02 at 7 49 26 PM

### Test
1. Use the cmd+F, cmd+shift+g, and cmd+g to navigate search results
2. Use the search on a note that doesn't have the result
3. Use the search bottom bar to navigate results

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -8,7 +8,12 @@ import DevBadge from './components/dev-badge';
 import DialogRenderer from './dialog-renderer';
 import { isElectron, isMac } from './utils/platform';
 import classNames from 'classnames';
-import { createNote, closeNote, toggleNavigation } from './state/ui/actions';
+import {
+  createNote,
+  closeNote,
+  search,
+  toggleNavigation,
+} from './state/ui/actions';
 import { recordEvent } from './state/analytics/middleware';
 
 import * as settingsActions from './state/settings/actions';
@@ -27,12 +32,14 @@ type StateProps = {
   hotkeysEnabled: boolean;
   isSmallScreen: boolean;
   lineLength: T.LineLength;
+  isSearchActive: boolean;
   showNavigation: boolean;
   showNoteInfo: boolean;
   theme: 'light' | 'dark';
 };
 
 type DispatchProps = {
+  clearSearch: () => any;
   closeNote: () => any;
   createNote: () => any;
   focusSearchField: () => any;
@@ -86,13 +93,17 @@ class AppComponent extends Component<Props> {
 
     if (
       (cmdOrCtrl && shiftKey && 'KeyS' === code) ||
-      (isElectron && cmdOrCtrl && !shiftKey && 'KeyF' === code)
+      (cmdOrCtrl && !shiftKey && 'KeyF' === code)
     ) {
       this.props.focusSearchField();
 
       event.stopPropagation();
       event.preventDefault();
       return false;
+    }
+
+    if (('Escape' === code || 'Esc' === code) && this.props.isSearchActive) {
+      this.props.clearSearch();
     }
 
     if (cmdOrCtrl && shiftKey && 'KeyF' === code) {
@@ -111,9 +122,8 @@ class AppComponent extends Component<Props> {
       return false;
     }
 
-    // prevent default browser behavior for search
-    // will bubble up from note-detail
-    if (cmdOrCtrl && 'KeyG' === code) {
+    // prevent default browser behavior for search and find
+    if (cmdOrCtrl && ('KeyG' === code || 'KeyF' === code)) {
       event.stopPropagation();
       event.preventDefault();
     }
@@ -167,6 +177,7 @@ class AppComponent extends Component<Props> {
 const mapStateToProps: S.MapState<StateProps> = (state) => ({
   autoHideMenuBar: state.settings.autoHideMenuBar,
   hotkeysEnabled: state.settings.keyboardShortcuts,
+  isSearchActive: !!state.ui.searchQuery.length,
   isSmallScreen: selectors.isSmallScreen(state),
   lineLength: state.settings.lineLength,
   showNavigation: state.ui.showNavigation,
@@ -178,6 +189,7 @@ const mapDispatchToProps: S.MapDispatch<DispatchProps> = (dispatch) => {
   return {
     activateTheme: (theme: T.Theme) =>
       dispatch(settingsActions.activateTheme(theme)),
+    clearSearch: () => dispatch(search('')),
     closeNote: () => dispatch(closeNote()),
     createNote: () => dispatch(createNote()),
     focusSearchField: () => dispatch(actions.ui.focusSearchField()),
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -93,11 +93,13 @@ class NoteContentEditor extends Component<Props> {
     const noteChanged = props.noteId !== state.noteId;
     const needsNewContent = contentChanged || noteChanged;
 
-    const content = needsNewContent
+    const content = noteChanged
       ? goFast
         ? props.note.content.slice(0, props.editorSelection[1] + 5000)
         : withCheckboxCharacters(props.note.content)
-      : state.content;
+      : contentChanged
+      ? withCheckboxCharacters(props.note.content)
+      : null;
 
     const editor = noteChanged ? (goFast ? 'fast' : 'full') : state.editor;
 
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -91,15 +91,12 @@ class NoteContentEditor extends Component<Props> {
     const goFast = props.note.content.length > 5000;
     const contentChanged = props.note.content !== state.content;
     const noteChanged = props.noteId !== state.noteId;
-    const needsNewContent = contentChanged || noteChanged;
 
-    const content = noteChanged
-      ? goFast
+    const content = contentChanged
+      ? noteChanged && goFast
         ? props.note.content.slice(0, props.editorSelection[1] + 5000)
         : withCheckboxCharacters(props.note.content)
-      : contentChanged
-      ? withCheckboxCharacters(props.note.content)
-      : null;
+      : state.content;
 
     const editor = noteChanged ? (goFast ? 'fast' : 'full') : state.editor;
 
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -15,6 +15,8 @@ import {
   withCheckboxCharacters,
   withCheckboxSyntax,
 } from './utils/task-transform';
+import IconButton from './icon-button';
+import ChevronRightIcon from './icons/chevron-right';
 
 import * as S from './state';
 import * as T from './types';
@@ -65,6 +67,8 @@ type OwnState = {
   editor: 'fast' | 'full';
   noteId: T.EntityId | null;
   overTodo: boolean;
+  searchQuery: string;
+  selectedSearchMatchIndex: number | null;
 };
 
 class NoteContentEditor extends Component<Props> {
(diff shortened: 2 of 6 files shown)


lib/app.tsx
```diff
@@ -86,7 +86,7 @@ class AppComponent extends Component<Props> {
 
     if (
       (cmdOrCtrl && shiftKey && 'KeyS' === code) ||
-      (isElectron && cmdOrCtrl && !shiftKey && 'KeyF' === code)
+      (cmdOrCtrl && !shiftKey && 'KeyF' === code)
```
dmsnell, line 96: cc: @codebykat

lib/editor/note-content-editor.tsx
```diff
@@ -169,16 +204,18 @@ class NoteContentEditor extends Component<Props> {
       this.state.editor === 'full' &&
       prevProps.searchQuery !== this.props.searchQuery
     ) {
+      this.setState({ selectedSearchMatchIndex: null });
```
dmsnell: for state updates it's more appropriate to set them in `getDerivedStateFromProps()` so that it avoids an additional component re-render and the intermediary partial-update

this often requires us to trap more data in local state as `getDerivedStateFromProps()` is intentionally a static class method.

```ts
static getDerivedStateFromProps(props, state) {
	…
	
	return Object.assign( {},
		props.content !== state.content && { content: … },
		…
		props.searchQuery !== state.searchQuery && { selectedSearchMatchIndex: null },
	);
}
```

codebykat:
Issues found in testing (all in Firefox/OSX):

* ~On load (with a note open) if you start typing a search immediately, the search field often loses focus and you end up editing content in the selected note, I think it happens when the note hasn't loaded before you start typing~ Can't reproduce this now, seems fixed
* ~Performance - results take a long time to show on a long note, looks broken~
* ~Performance - can take a VERY long time to show after switching between a couple of long notes (we need to debounce this I think, it seems to get more bogged down the more you switch between notes)~
* ~Performance - search with Markdown preview turned on is excruciatingly slow/laggy (like a long pause after each character typed)~ Verified all of these as fixed :)

* ~Keyboard combo to switch between note editor and tags can still be triggered when the search results are active, causing editor to lose focus and cursor to get hidden underneath~ fixed in https://github.com/Automattic/simplenote-electron/pull/2313/commits/5749b53b9f6bc3756343678ce586878f33b92157 / https://github.com/Automattic/simplenote-electron/pull/2313/commits/013f29d5db27c176447e17c9c84a5805b933ddfa
* ~If you do a search and it's not contained in the selected note, there's no way to go to the first match. I think Ctrl+G should select the first matching note and go to the first match. @SylvesterWilmott ?~ probably invalid, see below
* Highlight colors are flipped (the dark one should be the current match and all matches should be light)
* Highlight colors overlap and the current match highlight isn't visible in the title
* ~Should ESC dismiss an active search?~ https://github.com/Automattic/simplenote-electron/pull/2313/commits/2aed9fc7158f1fe41dcae322d4d3a518d102a80d
* Scrollbar styling - could be better (Sly's mockups are much prettier with the horizontal lines that match the width of the scrollbar)
* ~The browser Edit menu still triggers browser's find/find again~ There's probably no way to fix this

codebykat:
Bottom of scrollbar gets overlapped by the search results, hiding any highlight at the bottom of the note:

Screen Shot 2020-09-01 at 11 16 48 PM

lib/note-content-editor.tsx
```diff
@@ -169,16 +203,17 @@ class NoteContentEditor extends Component<Props> {
       this.state.editor === 'full' &&
       prevProps.searchQuery !== this.props.searchQuery
     ) {
+      this.setState({ selectedSearchMatchIndex: null });
```
dmsnell: noting so it doesn't get lost in the changing code: this `setState()` shouldn't be here; it's more appropriate inside of `getDerivedStateFromProps()`

dmsnell:
> we need to debounce this I think, it seems to get more bogged down the more you switch between notes

this is why I added the `fast` editor mode. definitely worth examining with the search as well

> Performance… Performance… Performance

if it's really bad we may need to start right from the get-go and do this in a WebWorker. Monaco can invalidate expired search results and we can prevent the search from blocking the UI

> If you do a search and it's not contained in the selected note, there's no way to go to the first match. I think Ctrl+G should select the first matching note and go to the first match

Hm. this seems really surprising to me and I'm sure I'd be confused if the app switched notes on me. Do we have examples of other apps acting like this? My IDEs (file list + opened file) don't change files unless I do a global search.

> Should ESC dismiss an active search?

Yes~! 😄 

> the search field often loses focus and you end up editing content in the selected note

does the editor not update if we don't call `.focus()` on it? maybe we can take out that call to `this.focusEditor()`?

SylvesterWilmott:
>If you do a search and it's not contained in the selected note, there's no way to go to the first match.

Can you clarify what you mean here?

codebykat:
Pushed a few commits, checked off everything that's fixed, here are new items from testing:

* Results bar not shown in Markdown preview mode, so you can't previous/next through matches in preview mode. We should add this to Markdown preview too.
* This is obvious now that I type it out, but since the results bar overlaps the tag bar, it is impossible to ever see or edit tags on a note while you have a search query entered. Are we okay with this?
* [minor but weird] After using next/previous to move selection through results, if you close the search the last selected match remains selected.
* On first note load with editor focused, sometimes Cmd+F doesn't focus the search field. Need to try to reproduce this consistently

Enabling this in preview mode is the only blocker for this PR in my opinion.

codebykat:
To clarify this question for @dmsnell and @SylvesterWilmott: 

> If you do a search and it's not contained in the selected note, there's no way to go to the first match. I think Ctrl+G should select the first matching note and go to the first match

1. Have a note open
2. Do a search that does NOT match in the current note, but has other matches
3. Observe that the note list is filtered to matches, but the current note is still open, with a bar that shows "0 results" at the bottom.
4. Ctrl+G does nothing from here
5. I was proposing you can use Ctrl+G to jump to the first match by selecting the first matching note.

> @dmsnell: Hm. this seems really surprising to me and I'm sure I'd be confused if the app switched notes on me. Do we have examples of other apps acting like this? My IDEs (file list + opened file) don't change files unless I do a global search.

Isn't this a global search though? Maybe we mean something different by global.

In typing this up, I realized that you can use the keyboard to get to the first matching note using the next/prev note J/K shortcuts, so I am willing to retract the comment.

I guess if you're in a note with matches, I wouldn't expect next/previous match to go to matches in the next note, so my intuition here might have been wrong.

lib/icon-button/index.tsx
```diff
@@ -10,15 +10,22 @@ type OwnProps = {
...
-  <Tooltip
-    classes={{ tooltip: 'icon-button__tooltip' }}
-    enterDelay={200}
-    title={title}
-  >
-    <button className="icon-button" type="button" data-title={title} {...props}>
-      {icon}
-    </button>
-  </Tooltip>
+  <span>
```
codebykat: This fixes a console error when it's passed a disabled element. See https://material-ui.com/components/tooltips/#disabled-elements

lib/editor/note-content-editor.tsx
belcherj: done

lib/note-content-editor.tsx
belcherj: done

belcherj:
> Results bar not shown in Markdown preview mode, so you can't previous/next through matches in preview mode. We should add this to Markdown preview too.

This is considerably more work. I think this can push because the browser level find is available on markdown.

belcherj:
> This is obvious now that I type it out, but since the results bar overlaps the tag bar, it is impossible to ever see or edit tags on a note while you have a search query entered. Are we okay with this?

I believe this is expected. @SylvesterWilmott?

belcherj:
> [minor but weird] After using next/previous to move selection through results, if you close the search the last selected match remains selected.

This is expected as you are selecting the text as you navigate through the search results. Where would you expect selection to go? We could capture selection before the search and restore after but I think that would be unexpected.

SylvesterWilmott:
>>This is obvious now that I type it out, but since the results bar overlaps the tag bar, it is impossible to ever see or edit tags on a note while you have a search query entered. Are we okay with this?

>I believe this is expected. @SylvesterWilmott?

Ideally, the tags would appear above the results bar when it is visible.

codebykat:
> > Results bar not shown in Markdown preview mode, so you can't previous/next through matches in preview mode. We should add this to Markdown preview too.
> 
> 
> 
> This is considerably more work. I think this can push because the browser level find is available on markdown.

Except it's sort of not because we've disabled the shortcut keys. But we can do it in a follow-up PR.

belcherj:
> Except it's sort of not because we've disabled the shortcut keys. But we can do it in a follow-up PR.

They are not disabled in the markdown view.

codebykat:
> > [minor but weird] After using next/previous to move selection through results, if you close the search the last selected match remains selected.
> 
> 
> 
> This is expected as you are selecting the text as you navigate through the search results. Where would you expect selection to go? We could capture selection before the search and restore after but I think that would be unexpected.

I just don't expect it to be actually selecting the text I guess. e.g. on Firefox search when you exit out of the search, everything is cleared, there's no lingering selection. I think this is also true in most text editors.

dmsnell:
https://github.com/Automattic/simplenote-electron/pull/2313#issuecomment-686280377

@kat it looks like I understood you and yeah I find that to be unexpected behavior. currently we keep the note open even if it doesn't match the search. this was formerly a problem especially since you could have a note open and the matching search term could disappear from a remote edit. if we move to the next note in cases like this it'd be easy to miss that the note switched or if you do notice, be left confused as to why

> when you exit out of the search, everything is cleared, there's no lingering selection

sounds right and normative to me. existing behavior was also to remove the highlighting of results after the search box was cleared.

---

there's a conflict with `scss/theme.css` - I can try and figure out that out
also I'm going to rework some of the stuff in `getDerivedStateFromProps()`

dmsnell:
I'm a bit confused on the logic @belcherj so I drew a diagram

Screen Shot 2020-09-04 at 11 40 00 AM

It looks like you rearranged the code but left the logic in tact. The only change is that the new `content` returns `null` instead of the previous content. I don't understand why this doesn't clear out the buffer unless that's a happy byproduct of some Monaco detail.

If I have the "desired" column right the mistake in my original code was using `needsNewContent` instead of `contentChanged`.

```ts
const content = needsNewContent
	? goFast
		? props.note.content.slice(0, props.editorSelection[1] + 5000)
		: withCheckboxCharacters(props.note.content)
	: state.content;
```

What was the behavior exactly that was broken? I tried a before/after locally and couldn't remember. That being said, given that your code is logically equivalent to mine (other than the `nulll` vs. `state.content` thing) I'm not surprised. I wonder if there is some other detail influencing the behavior or if I'm missing something.

SylvesterWilmott:
>Observe that the note list is filtered to matches, but the current note is still open, with a bar that shows "0 results" at the bottom.

I see! This does seem strange. I guess there would be two solutions, to either highlight the first note in the search results or to only display the bar when one of the search results is highlighted. I'm with @dmsnell in that I would not expect the note to change without my specific input. Let's see if we can hide the bar for notes with 0 matches.

>when you exit out of the search, everything is cleared, there's no lingering selection

I would expect the selection to disappear as well.

belcherj:
> What was the behavior exactly that was broken? I tried a before/after locally and couldn't remember. That being said, given that your code is logically equivalent to mine (other than the nulll vs. state.content thing) I'm not surprised. I wonder if there is some other detail influencing the behavior or if I'm missing something.

When you have a long note checkboxes don't render.

This is due to:
```
const content = needsNewContent
	? goFast
		? props.note.content.slice(0, props.editorSelection[1] + 5000)
		: withCheckboxCharacters(props.note.content)
	: state.content;
```
needsNewContent is true when content changes, goFast is true when content is long. In this case content will be `props.note.content.slice(0, props.editorSelection[1] + 5000)`

With:
```
    const content = noteChanged
      ? goFast
        ? props.note.content.slice(0, props.editorSelection[1] + 5000)
        : withCheckboxCharacters(props.note.content)
      : contentChanged
      ? withCheckboxCharacters(props.note.content)
      : null;
```
In this case of noteChanged being false and contentChanged being true this will return `withCheckboxCharacters(props.note.content)`

dmsnell:
Updating here for history: in a call with @belcherj he pointed out that my diagram was missing two cases, which should have been obvious since we have three boolean properties (2^3 = 8)

| Note Changed | Content Changed | Is Long (Go Fast) | Choice |
|---|---|---|---|
| N | N| N| State |
| N | N| Y| State |
| N | Y | N | Transformed |
| N | Y | Y | Transformed |
| Y | N| N| State |
| Y | N| Y| State |
| Y | Y | N | Transformed |
| Y | Y | Y | Fast |

With the help of some automated logic reduction that gave us this system of equations:

```ts
const useState = ! contentChanged;
const useFast = noteChanged && contentChanged && isLong;
const useTransformed = ( contentChanged && ! noteChanged ) || ( contentChanged && noteChanged !== isLong )
```

Since we can abort early we reduced further, knowing that once we determine if the content changes that variable becomes irrelevant.

```ts
const useState = ! contentChanged; // assume content changed, `useState === false`
const useFast = noteChanged && isLong;
const useTransformed = ( ! noteChanged ) || ( noteChanged !== isLong )
```

Constructing sequentially we can test in this order and let `useTransformed` be the "catch-all"

```ts
const content = contentChanged
	? noteChanged && isLong
		? fastContent
		: checkboxTransformedContent
	: stateContent;
```

PR #2322, opened 2020-09-03
Icon tooltip: Fix console error with a wrapper span

codebykat:
### Fix

This PR fixes the following error, which has been around for awhile:

Screen Shot 2020-09-03 at 1 25 58 PM

We need to add a wrapping `<span>` to the button which allows the tooltip to still be displayed without an error, whether or not the button is disabled.

See https://material-ui.com/components/tooltips/#disabled-elements

### Test
1. Verify you can still mouseover buttons and get pop-up tooltips
2. Verify this warning is gone from console

### DIFF
--- a/lib/icon-button/index.tsx
+++ b/lib/icon-button/index.tsx
@@ -10,15 +10,22 @@ type OwnProps = {
 type Props = OwnProps;
 
 export const IconButton = ({ icon, title, ...props }: Props) => (
-  <Tooltip
-    classes={{ tooltip: 'icon-button__tooltip' }}
-    enterDelay={200}
-    title={title}
-  >
-    <button className="icon-button" type="button" data-title={title} {...props}>
-      {icon}
-    </button>
-  </Tooltip>
+  <span>
+    <Tooltip
+      classes={{ tooltip: 'icon-button__tooltip' }}
+      enterDelay={200}
+      title={title}
+    >
+      <button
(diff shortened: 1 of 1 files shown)


belcherj (approved):

lib/icon-button/index.tsx
```diff
@@ -10,15 +10,22 @@ type OwnProps = {
...
-    <button className="icon-button" type="button" data-title={title} {...props}>
-      {icon}
-    </button>
-  </Tooltip>
+  <span>
+    <Tooltip
+      classes={{ tooltip: 'icon-button__tooltip' }}
+      enterDelay={200}
+      title={title}
+    >
```
dmsnell, line 18: did you try wrapping the `<button>` from inside the `<Tooltip>`?

I interpreted the error as a wrapper from inside `<Tooltip>` vs. wrapping `<Tooltip>`

codebykat: I thought I tried that first. Lemme see.

codebykat: Oh yeah. I tried that and it leads to the tooltip being misaligned:

Screen Shot 2020-09-03 at 2 23 06 PM

codebykat: @dmsnell Okay I figured it out! https://github.com/Automattic/simplenote-electron/pull/2324

PR #2324, opened 2020-09-03
Icon tooltip: Fix console error with a wrapper span

codebykat:
### Fix

Follow-up to #2322 . Move the `span` inside the `Tooltip`. Icon button elements need to be wrapped in their own span in order to make sure tooltips stay aligned (because the tooltip is spawned as a sibling of the `button`).

See https://github.com/Automattic/simplenote-electron/pull/2322#pullrequestreview-482202141

### DIFF
--- a/lib/icon-button/index.tsx
+++ b/lib/icon-button/index.tsx
@@ -10,12 +10,12 @@ type OwnProps = {
 type Props = OwnProps;
 
 export const IconButton = ({ icon, title, ...props }: Props) => (
-  <span>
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
Nice. I am almost certain I've seen this very same warning/fix/inside/outside thing before in this repo but I can't remember clearly when or where. Glad it's working either way.

codebykat:
Something or other has been throwing this error in the console forever. I believe it is also present in `develop`.

PR #2329, opened 2020-09-09
Delete unused variables

belcherj:
### Fix

These two sass variables are unused. This removes them as they are not necessary.

### Test
1. Ensure app builds
2. Search for the two variables in the codebase
3. They should not exist.


codebykat (approved):
Oh man. We should definitely do something like this for the z-index issues in the future though.

belcherj:
These two sass variables are unused. This removes them as they are not necessary.

PR #2332, opened 2020-09-10
Right align rtl languages

belcherj:
### Fix

When a line is a rtl language such as Arabic it should be aligned to the right. This PR floats it to the right.

### Test
1. Add some Arabic text to a note.
2. It should float to the right

### DIFF
--- a/scss/theme.scss
+++ b/scss/theme.scss
@@ -88,6 +88,9 @@
   display: none !important;
 }
 
+/* Monaco adds the dir='ltr' attribute to rtl languages.
+   This does the opposite of what you think it should do.
+*/
 span[dir='ltr'] {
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
I see no problem here but I also have very little competency to judge it 🙃

codebykat (approved):
After verifying in Slack that it should be `ltr` and not `rtl` in the code, which seems backwards and weird, I tested this with sample text from https://r12a.github.io/scripts/tutorial/summaries/arabic and it seems to work fine.

I think we should comment that line because I am going to see it in a few months and think it's a bug.

Screen Shot 2020-09-10 at 10 15 49 AM

PR #2333, opened 2020-09-14
[staging] Fix editor container padding

codebykat:
### Fix
* This (mostly) fixes a flicker when switching to long notes, by fixing the padding on the plaintext div
* Also removes the larger title size since we removed that in #2298 
* I additionally found a bug with the width calculation within the Monaco container; we want the style to change at a viewport width of 1400px, not a container width, so we need to add the sidebar width to the container width before checking. Yay, more magic numbers!

n.b. There's still a slight flicker when switching to a long note that I haven't tracked down. It's not the font, font size, letter spacing, line height, or anything else I can think of.

### Test
1. Switch between long notes and verify that the note padding stays the same while the note is loading
2. Try it with line lengths of narrow and full
3. Try it with various window widths

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -37,7 +37,10 @@ const getEditorPadding = (lineLength: T.LineLength, width?: number) => {
     return 25;
   }
 
-  if (width <= 1400) {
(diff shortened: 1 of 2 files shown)


belcherj (approved):
LGTM

PR #2334, opened 2020-09-14
[staging] Fix keyboard shortcuts on Edge and non-English layouts

codebykat:
### Fix

Keyboard shortcuts are currently not working on legacy Edge, because it does not include a `code` value.

The docs recommend using `key` instead of `code`, anyhow:

>  Be aware, however, that you can't use the value reported by KeyboardEvent.code to determine the character generated by the keystroke, because the keycode's name may not match the actual character that's printed on the key or that's generated by the computer when the key is pressed.

> For example, the code returned is "KeyQ" for the Q key on a QWERTY layout keyboard, but the same code value also represents the ' key on Dvorak keyboards and the A key on AZERTY keyboards. That makes it impossible to use the value of code to determine what the name of the key is to users if they're not using an anticipated keyboard layout.

> To determine what character corresponds with the key event, use the KeyboardEvent.key property instead.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

**I have NOT YET tested this very heavily so more testing would be appreciated!**

### Test

1. Test all keyboard shortcuts on multiple browsers

### DIFF
--- a/lib/app-layout/index.tsx
+++ b/lib/app-layout/index.tsx
@@ -65,11 +65,12 @@ export class AppLayout extends Component<Props> {
       keyboardShortcutsAreOpen,
       showKeyboardShortcuts,
     } = this.props;
-    const { ctrlKey, code, metaKey } = event;
+    const { ctrlKey, metaKey } = event;
+    const key = event.key.toLowerCase();
 
     const cmdOrCtrl = ctrlKey || metaKey;
 
-    if (cmdOrCtrl && code === 'Slash') {
+    if (cmdOrCtrl && key === '/') {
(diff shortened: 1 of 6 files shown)


belcherj (approved):
lgtm

codebykat:
Tested on OSX/Firefox, OSX/Chrome, OSX/Safari, Windows/Edge (legacy), Windows/Edge (Chromium), Windows/Firefox.

Found some regressions on Edge (Chromium) so I'm debugging that.

codebykat:
Edge (Chromium) should now be fixed

belcherj (approved):

Issue #2331, opened 2020-09-09
Search follow-up tasks


codebykat:
Tracking follow-up tasks from #2313:

* ~Highlight colors are flipped (the dark one should be the current match and all matches should be light)~
* ~Highlight colors overlap and the current match highlight isn't visible in the title~
* Scrollbar styling - could be better (Sly's mockups are much prettier with the horizontal lines that match the width of the scrollbar) **additionally scrollbars do not match the native ones on e.g. Windows** -- tracked in #2553
* ~Bottom of scrollbar gets overlapped by the search results, hiding any highlight at the bottom of the note [#](https://github.com/Automattic/simplenote-electron/pull/2313#issuecomment-685328070) - #2544 -> #2545~
* ~Tag field should appear above the search results bar - #2545~
* ~Selection should be cleared when search is dismissed, instead of leaving a selection of text~ #2335 
* ~ESC does not clear the search if the editor has focus~ #2335
* On first note load with editor focused, sometimes Cmd+F doesn't focus the search field. Need to try to reproduce this consistently. (see #2530 )
* ~Similarly sometimes if you load up and immediately focus the search field you end up typing in the note.~ #2335
* ~Hide the search bar with 0 matches in the note (if you do a search with a selected note that doesn't contain the search query)~

PR #2335, opened 2020-09-14
fix search selection issues

codebykat:
### Fix

Some follow-ups to the new search. #2331 

* Properly unset the selection when search is canceled
* Don't trap ESC in the editor (it clears the selection in Monaco, but this doesn't bubble up)
* Remove the `editor.focus()` that can steal focus from the search field.

### Test
1. Select some text in a note. Verify that pressing ESC clears the selection.
2. Do a search. Jump through matches in a note with Ctrl+G. Press ESC and verify that no lingering selection remains.
3. Do a search that has matches NOT in the current note. Verify that you can still select text in the note with no matches, and clear the selection with ESC.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -625,7 +625,6 @@ class NoteContentEditor extends Component<Props> {
     );
 
     editor.revealLine(start.lineNumber, Editor.ScrollType.Immediate);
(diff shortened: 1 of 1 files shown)


belcherj (approved):
Tested and works as expected.

PR #2337, opened 2020-09-17
Various linting fixes

belcherj:
### Fix

This PR cleans up some lint warnings that have been introduced over time:
- Remove a console.log where it seemed unnecessary
- Changed a console.log to debug where it made sense to keep the logging
- Changed some variable names to distinguish from variables declared in an upper scope

### Test
Does app build and run

### DIFF
--- a/lib/boot.ts
+++ b/lib/boot.ts
@@ -12,7 +12,7 @@ import { boot as bootLoggingOut } from './logging-out';
 const config = getConfig();
 
 const clearStorage = (): Promise<void> =>
-  new Promise((resolve) => {
+  new Promise((parentResolve) => {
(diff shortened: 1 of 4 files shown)


lib/boot.ts
```diff
@@ -12,7 +12,7 @@ import { boot as bootLoggingOut } from './logging-out';
 const config = getConfig();
 
 const clearStorage = (): Promise<void> =>
-  new Promise((resolve) => {
+  new Promise((parentResolve) => {
```
dmsnell: how about something semantic instead of structural?

```ts
new Promise((resolveStorage) => {
```

dmsnell (approved):

PR #2339, opened 2020-09-22
Allow RTL formatting in markdown mode

### LINKED ISSUE #562: Preview mode doesn't support RTL
Hi there.. Thanks a lot to adding RTL support to Simplenote but there is one issues related with that.

When the "Markdown" option was activated for one note, the Preview mode didn't support RTL so text aligned to left and the direction become from Left-to-Right.

Other thing, the font for Headers (h1, h2 etc) changed also in Preview mode

Edit mode: 
edit mode

Preview mode:
privew mode

belcherj:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/562

Allow RTL languages to be right aligned in markdown preview mode.

Before:

Screen Shot 2020-09-21 at 8 22 57 PM

After:

Screen Shot 2020-09-21 at 8 22 42 PM

### Test
1. Have several notes with markdown
2. Does markdown look properly in preview mode

### DIFF
--- a/lib/note-detail/style.scss
+++ b/lib/note-detail/style.scss
@@ -87,6 +87,7 @@
 
   p {
     margin: 0 0 1.5em;
+    unicode-bidi: plaintext;
   }
 
   ul p {


dmsnell (approved):
This tested fine in my test document with mixed-direction content in separate paragraphs but also inline. I don't understand the implications of `unicode-bidi` property but `plaintext` seems reasonable.

belcherj:
https://developer.mozilla.org/en-US/docs/Web/CSS/unicode-bidi

plaintext
This keyword makes the elements directionality calculated without considering its parent bidirectional state or the value of the direction property. The directionality is calculated using the P2 and P3 rules of the Unicode Bidirectional Algorithm.
This value allows the display of data that is already formatted using a tool following the Unicode Bidirectional Algorithm.

PR #2287, opened 2020-08-21
Merge release 1.21.1

mokagio:

PR #2344, opened 2020-09-22
Merge release 1.21.1 with conflicts resolution

mokagio:
**Sorry for letting the 1.21.1 merge become stale and not noticing at all!**

---

Conflicts were on:

- `desktop/menus/format-menu.js`
- `desktop/menus/help-menu.js`
- `lib/boot-without-auth.tsx`
- `lib/dialog-renderer/index.tsx`

I solved them by always applying the version from `HEAD`, under the assumption that because new work went into those files, then that new work should be the one to stay on `develop`.

There were conflicts in the `package.json` and `package-lock.json`, but it was only due to the version change.

Before opening this, I run `make lint build test` everything worked, but I noticed a few lint warnings, although, they didn't seem related to the changes in this branch.

If it helps reviewing, as far as I can see, the only changes that are in this branch but not in develop are from https://github.com/Automattic/simplenote-electron/pull/2285, https://github.com/Automattic/simplenote-electron/pull/2250, and https://github.com/Automattic/simplenote-electron/pull/2249, which is consistent with the fact that those PR have `release/1.21.1` as the base.

For a reference of how the diff looks without conflicts resolution, checkout #2287.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "simplenote",
-  "version": "1.20.0",
+  "version": "1.21.0-beta1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "simplenote",
-  "version": "1.21.0",
+  "version": "1.21.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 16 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "simplenote",
-  "version": "2.0.0-rc3",
+  "version": "2.0.0-rc5",
```
mokagio, line 3: The proper conflict resolution would have been `rc3`, but I thought it appropriate to keep the version in sync between `package.json` and `package-lock.json`.

mokagio:
Closing in favor of #2358.

PR #2345, opened 2020-09-22
Make Electron menus aware of editMode

codebykat:
### Fix

This PR fixes the ability to select all, undo and redo in the non-editor textareas while preview mode is active.

It also makes Electron aware of the editor state and allows it to disable menu options accordingly (though the only one this affects is Insert Checklist).

Follow-up to #2293 

### Test

1. In Electron, go into a Markdown note
2. Type in the tag field and verify that select all, undo and redo affect that field (not the editor)
3. Repeat for the search field
4. Go into preview mode
5. Type in the tag field and verify that it is still possible to select all, undo and redo using the menu or keyboard commands
6. Repeat for the search field

### Review

In the first iteration of this PR, I simply added the Electron listeners to the preview component, but that felt icky. So I went down the rabbithole of figuring out how to make Electron aware of the current edit mode and ended up adding an argument to the existing `settingsUpdate` channel.

The docs claimed more than one argument was supported, but in practice arguments after the second were undefined, so I ended up passing a dictionary.

### Todo

There is still some weirdness here that may or may not need a follow-up.

* SelectAll in edit mode when the editor doesn't have focus doesn't do anything. IMO it would be better to check if the tag or search field have focus and if not, select the text within the note, but I'm not sure this is possible in React.
* If you enter text in the search and/or tag boxes, go into preview mode, then click back on the note preview, selectAll selects all input text: editor text, tag box text AND search box text. It's... odd.

### DIFF
--- a/lib/state/electron/middleware.ts
+++ b/lib/state/electron/middleware.ts
@@ -90,7 +90,10 @@ export const middleware: S.Middleware = ({ dispatch, getState }) => {
     }
   });
 
-  window.electron.send('settingsUpdate', getState().settings);
+  window.electron.send('settingsUpdate', {
+    settings: getState().settings,
+    editMode: getState().ui.editMode,
+  });
 
   return (next) => (action) => {
     const prevState = getState();
@@ -98,7 +101,16 @@ export const middleware: S.Middleware = ({ dispatch, getState }) => {
     const nextState = getState();
 
     if (prevState.settings !== nextState.settings) {
-      window.electron.send('settingsUpdate', nextState.settings);
+      window.electron.send('settingsUpdate', {
+        settings: nextState.settings,
+        editMode: nextState.ui.editMode,
+      });
+    }
+    if (prevState.ui.editMode !== nextState.ui.editMode) {
+      window.electron.send('settingsUpdate', {
+        settings: nextState.settings,
+        editMode: nextState.ui.editMode,
+      });
     }
 
     return result;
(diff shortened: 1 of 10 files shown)


codebykat:
I kiiiiiind of want to rename `settingsUpdate` to `updateMenus` or something now

belcherj (approved):
LGTM, the only thing I might change is the name of the command (`settingsUpdate`). `appStateChange` stored in `appState`? I don't really have strong feelings about it though.

belcherj:
Read this comment right after finishing my review: https://github.com/Automattic/simplenote-electron/pull/2345#issuecomment-696575058

So if it feels icky to both of us we should probably rename.

codebykat:
Renamed, should be good to go

lib/state/electron/middleware.ts
```diff
@@ -90,15 +90,27 @@ export const middleware: S.Middleware = ({ dispatch, getState }) => {
...
+      window.electron.send('appStateUpdate', {
+        settings: nextState.settings,
+        editMode: nextState.ui.editMode,
+      });
+    }
+    if (prevState.ui.editMode !== nextState.ui.editMode) {
+      window.electron.send('appStateUpdate', {
+        settings: nextState.settings,
+        editMode: nextState.ui.editMode,
+      });
```
dmsnell, line 113: there's no need to separate these `if` blocks. we could instead combine the conditions

```ts
if (
	prevState.settings !== nextState.settings ||
	prevState.editMode !== nextState.editMode
) {
	window.electron.send('appStateUpdate', { … } );
}
```

PR #2346, opened 2020-09-22
Adds Stylelint for sass linting

### LINKED ISSUE #1158: Add CSS linting
Here are the rules used in wp-calypso:
https://github.com/Automattic/wp-calypso/blob/master/.stylelintrc

belcherj:
### Fix

Closes https://github.com/Automattic/simplenote-electron/issues/1158

This PR adds stylelint and prettier standard configs with some exceptions for sass.

- I removed the extra whitespace configurations which I believe would be at odds with our prettier configuration.
- I removed [font: inherit ](https://github.com/Automattic/simplenote-electron/compare/develop...add/stylelint#diff-77be6b3031462c4580c10faa0d40ff36L258) from the normalize file as it is invalid CSS. 
- I fixed all the errors where we had no consistent style.
- Removed or added empty lines as required.
- Removed duplicate selectors.

### Test
1. Run app, do fonts look okay
2. Make a change to the css that stylelint will catch
3. Run linting, does it catch the error?
4. Run a commit, if there is an error you should not be able to commit


dmsnell (commented):
Is there a way to get this in with fewer settings? Are our exceptions special enough that they warrant disabling the linter?

What are we trying to catch with this? If it's style then let's have `prettier` handle it and not lint the style. That eliminates the ambiguity. If it's correctness then can we get a small core set of rules that won't enforce as specific of a style as these do?

Should be a good guard against bugs to have this linting in place. I feel like we might be able to remove most of this PR if we rely on `prettier` for formatting issues though. Are we not having `prettier` format our SCSS files?

Makefile
```diff
@@ -165,4 +165,5 @@ format:
 
 .PHONY: lint
 lint:
+	@npx stylelint --ignore-path .gitignore "**/*.scss" --syntax scss
```
dmsnell, line 171: typically in `make` we'd create two new targets and use `lint` to tie them together. this makes the built-in dependency resolution work easier and gives us the ability to lint just the JS or just the SCSS

```makefile
.PHONY: lint
lint: lint-js lint-scss

.PHONY: lint-js
lint-js:
	@npx eslint --ignore-path .gitignore "**/*.{js,jsx,ts,tsx}"

.PHONY: lint-scss
lint-scss:
	@npx stylelint --ignore-path .gitignore "**/*.scss" --syntax scss
```

there are other enhancements we could make but for this PR I think this is the relevant bit. this may allow for parallelism as well

belcherj:
This is ready for review

dmsnell (approved):
Thanks for simplifying the config. I can get behind this particularly since these rules have caught things we can recognize as obvious mistakes or oversights - it's not changing a lot of things that aren't wrong.

PR #2348, opened 2020-09-22
Hides the search results banner on print

belcherj:
### Fix

If you have an active search while printing it will show the search result banner which is not correct. This PR hides the search results banner on print.

Screen Shot 2020-09-22 at 9 56 48 AM

### Test
1. Have an active search
2. Print
3. You should not see the search results bar in the print


dmsnell (approved):
Beautiful

PR #2349, opened 2020-09-22
Display first note in trash when going to trash view

### LINKED ISSUE #2190: Inconsistent note selection with trash/delete/restore actions and switching between note views
### Expected

When I trash, delete, or restore a note, I expect the app to select the note below it in the note list.

I also expect the note selection behavior to be consistent when switching between note views All Notes or Trash.

### Observed

The note selection behavior is different depending on the position of the note in the note list:

1. If you select the first note in the list, and then switch between All Notes and Trash views, no note is open in the note editor. However, if you select any other note in the list, and then switch between All Notes and Trash views, that note remains open in the note editor.
2. If you view the All Notes list, select the first note in the list, and trash it, the next note in the note list is highlighted but it is not opened in the note editor. If you select any other note in the list and trash it, the next note in the note list is highlighted **and** opened in the note editor.
3. If you view the Trash list, select a note in the list, and delete it forever, the next note in the note list is highlighted but it is not opened in the note editor. (This is the only case where it is consistent regardless of which note it is.)
4. If you view the Trash list, select the first note in the list, and restore it, the next note in the note list is highlighted but it is not opened in the note editor. If you select any other note in the list and restore it, the next note in the note list is highlighted **and** opened in the note editor.

**1. Switching between All Notes and Trash**

sn-electron-selection-alltrash

**2. Trashing notes in All Notes**

sn-electron-selection-trashnote

**3/4. Deleting and restoring notes in Trash**

sn-electron-selection-deleterestore

-|-|-|-
VirtualBox VM|Windows|10|1.20.0-beta2

This doesn't seem to be new in 1.20-beta2; I also tested in the 1.19.0 release and see the same behavior there.

belcherj:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/2190

When the trash is opened this change causes the first note to be selected. If no notes are selected it makes the note selection null. This ensures that you cannot be viewing a note that is not deleted while you are viewing the trash.

### Test
1. Switch to trash
2. Observe that it selects the first note in the trash
3. If nothing is in the trash it unselects the current note and selects no new note.

### DIFF
--- a/lib/state/action-types.ts
+++ b/lib/state/action-types.ts
@@ -7,7 +7,7 @@ export type Action<
 > = { type: T } & Args & {
     meta?: {
       analytics?: [string, T.JSONSerializable | undefined][];
-      nextNoteToOpen?: T.EntityId;
+      nextNoteToOpen?: T.EntityId | null;
       searchResults?: {
         noteIds: T.EntityId[];
         tagHashes: T.TagHash[];
--- a/lib/state/ui/middleware.ts
+++ b/lib/state/ui/middleware.ts
@@ -60,9 +60,11 @@ export const middleware: S.Middleware = (store) => (
     }
 
     case 'SHOW_ALL_NOTES':
+    case 'SELECT_TRASH':
       // this middleware runs after the search middleware which is important
       // because we're reading the new search results which came as a result
-      // of changing the note filter from trash or a tag to "all notes"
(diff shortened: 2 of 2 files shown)


dmsnell (approved):
Looks fine. I did not test.

PR #2350, opened 2020-09-22
Change tag autocomplete logic to make it work

### LINKED ISSUE #1854: Right arrow doesn't consistently complete a tag suggestion
Found this while working on https://github.com/Automattic/simplenote-electron/pull/1853
If you are typing a tag for which there is a suggestion, pressing right arrow doesn't usually work to complete it.

### Expected
Pressing right arrow should complete a tag suggestion

### Observed
The only way it works is if you "back up" the cursor a character or two, then hit right arrow repeatedly as fast as you can.

### Reproduced
1. Type out a few characters of a tag that you already use somewhere so that a suggestion pops up
2. Tap the right arrow button. The tag does not complete.
3. Left arrow back a few characters
4. Full speed ahead with the right arrow. Usually you can get it to complete

I believe the issue is in `tag-input`'s [`interceptRightArrow`](https://github.com/Automattic/simplenote-electron/blob/5b501e7c2a64304286056578768f559b8e2341b2/lib/tag-input/index.tsx#L113) handler. `caretPosition` should never equal `value.length` since `value.length` is 1-offset and `caretPosition` is 0-offset. I'm guessing the reason why it works when you press right arrow really fast is that something funky is happening with the cursor position state and it hasn't had time to correct itself.

-|-|-|-
Asus|Zenbook|Linux Mint 19|Current dev branch... App says 1.12.0 but I think it's 1.14.0?

belcherj:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/1854

This updates the logic to get the correct caret (cursor) position inside the contentEditable. The previous logic did not provide the current cursor position.

endOffset is not what you would expect, it does not give you the cursor position. The selection starts and ends at the same place. You have to get the contents of the element and determine where you are within that element.

This PR additionally returns early if window.getSelection does not exist.

### Test
1. Have a tag "blahblah"
2. Go to a not that does not have that tag
3. Start typing "bla"
4. Use the right arrow to autocomplete
5. Does it work?

### DIFF
--- a/lib/tag-input/index.tsx
+++ b/lib/tag-input/index.tsx
@@ -132,11 +132,18 @@ export class TagInput extends Component<Props> {
 
   interceptRightArrow = (event: KeyboardEvent) => {
     const { value } = this.props;
-
+    if (!window.getSelection) {
+      return;
+    }
     // if we aren't already at the right-most extreme
     // then don't complete the suggestion; we could
     // be moving the cursor around inside the input
-    const caretPosition = window.getSelection().getRangeAt(0).endOffset;
+    const origionalRange = window.getSelection().getRangeAt(0);
+    const range = origionalRange.cloneRange();
+    range.selectNodeContents(event.currentTarget);
+    range.setEnd(origionalRange.endContainer, origionalRange.endOffset);
+    const caretPosition = range.toString().length;
+
     if (caretPosition !== value.length) {
       return;
     }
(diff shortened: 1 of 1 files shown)


lib/tag-input/index.tsx
```diff
@@ -132,11 +132,18 @@ export class TagInput extends Component<Props> {
...
     const { value } = this.props;
-
+    if (!window.getSelection) {
+      return;
+    }
     // if we aren't already at the right-most extreme
     // then don't complete the suggestion; we could
     // be moving the cursor around inside the input
-    const caretPosition = window.getSelection().getRangeAt(0).endOffset;
+    const origionalRange = window.getSelection().getRangeAt(0);
```
codebykat: typo "origional" but otherwise this looks good and works

codebykat (changes requested):

codebykat (approved):

PR #2351, opened 2020-09-22
Add support for importing .md files

### LINKED ISSUE #2155: Support .md format on 'Import Notes...' Dialog
Hello,

image

There are only the `.txt` format.
Please add `.md` format.

Thanks.

belcherj:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/2155

Adds support for .md files. Since they are basically plaintext files we can import them using the plaintext editor.

Screen Shot 2020-09-22 at 3 24 20 PM

### Test
1. Test importing markdown files
2. Test importing plain text files.

### DIFF
--- a/lib/utils/import/text-files/index.ts
+++ b/lib/utils/import/text-files/index.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'events';
 import CoreImporter from '../';
-import { endsWith, startsWith } from 'lodash';
+import { startsWith } from 'lodash';
 
 import * as T from '../../../types';
 
@@ -22,13 +22,10 @@ class TextFileImporter extends EventEmitter {
     }
 
     const importTextFile = (file) => {
-      if (
-        !file ||
-        !(
-          endsWith(file.name.toLowerCase(), '.txt') ||
-          endsWith(file.name.toLowerCase(), '.md')
-        )
-      ) {
+      const hasAllowableName = /\.(txt|md)$/.test(
+        file?.name.toLowerCase() ?? ''
+      );
+      if (!hasAllowableName) {
         return;
       }
 
--- a/lib/utils/import/text-files/index.ts
+++ b/lib/utils/import/text-files/index.ts
@@ -17,12 +17,18 @@ class TextFileImporter extends EventEmitter {
     let lastFileName = '';
 
     if (!filesArray) {
-      this.emit('status', 'error', 'No text files to import.');
+      this.emit('status', 'error', 'No files to import.');
       return;
     }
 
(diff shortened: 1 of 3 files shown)


lib/utils/import/text-files/index.ts
```diff
@@ -17,12 +17,18 @@ class TextFileImporter extends EventEmitter {
...
 
     const importTextFile = (file) => {
-      if (!file || !endsWith(file.name.toLowerCase(), '.txt')) {
+      if (
+        !file ||
+        !(
+          endsWith(file.name.toLowerCase(), '.txt') ||
+          endsWith(file.name.toLowerCase(), '.md')
+        )
+      ) {
```
dmsnell: if we want we can combine these tests into a simple RegExp pattern. I like the pattern form and how it emphasizes the match over the transform. take it or leave it

```ts
if (!file || !/\.(txt|md)$/.test(file.name.toLowerCase())) { … }
```

add in optional-chaining…
```ts
if (!/\.(txt|md)$/.test(file?.name.toLowerCase() ?? '')) { … }
```

give it a name…
```ts
const hasAllowableName = /\.(txt|md)$/.test(file?.name.toLowerCase() ?? '');
if (!hasAllowableName) { … }
```

codebykat (changes requested):
This worked great.

I have two tiny nits:

1) I think we should default the "enable Markdown" checkbox to checked when a .md file is selected:

Screen Shot 2020-09-22 at 3 34 42 PM

2) The change in line length makes this line break awkwardly:

Screen Shot 2020-09-22 at 3 34 07 PM

Maybe we can remove the word "plain" or make the line longer by saying "plain-text or Markdown-formatted files":

Screen Shot 2020-09-22 at 3 38 24 PM

Screen Shot 2020-09-22 at 3 39 12 PM

I'm now feeling uncertain if it should be "plain-text" or "plain text" 🤷‍♀️

belcherj:
I changed the text to read `Choose one or more text files (.txt, .md)`

lib/utils/import/text-files/index.ts
belcherj: Done

belcherj:
> I think we should default the "enable Markdown" checkbox to checked when a .md file is selected

There is a bit of complexity in doing this. I don't think the benefit is worth the cruft it adds. I think we should tackle that when we refactor settings as part of spring cleaning.

codebykat:
It seems quite silly to not default to setting Markdown on a Markdown file, but it's not a hill I'm choosing to die on today.

codebykat (approved):

PR #2352, opened 2020-09-22
Make tags in tag list links until they are being edited

### LINKED ISSUE #844: TagList shouldn't render tags as <input> elements when not in Edit mode
The TagList component in the NavigationBar is rendering tags as `<input>` elements, even when they're not in editing mode. When not in editing mode, the tags act like links, not input fields.

There are two problems with this:

1. It is unsemantic HTML and not good for accessibility.
1. An Undo/Redo context menu will be shown on right click, which doesn't make sense for links.

They should be `<a>` (or `<button>`?) elements by default, and only become `<input>` when in editing mode.

belcherj:
### Fix

Closes: https://github.com/Automattic/simplenote-electron/issues/844

In the tag navigation panel, tags are represented as inputs even when not editing.

There are two problems with this:

- It is unsemantic HTML and not good for accessibility.
- An Undo/Redo context menu will be shown on right click, which doesn't make sense for links.

This PR makes them links as they should.

### Test
1. Open chrome dev tools.
2. Inspect tags in the tag list, they should be link tags
3. Ensure that they render correctly
4. Ensure that the selected tag gets rendered correctly
5. Ensure that you can edit a tag

### DIFF
--- a/lib/tag-list/input.tsx
+++ b/lib/tag-list/input.tsx
@@ -39,9 +39,9 @@ export class TagListInput extends Component<Props, OwnState> {
         spellCheck={false}
       />
     ) : (
-      <a className={classes} onClick={onClick}>
+      <button className={classes} onClick={onClick}>
         {value}
-      </a>
+      </button>
     );
   }
 }
--- a/lib/tag-list/input.tsx
+++ b/lib/tag-list/input.tsx
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 type OwnProps = {
   editable: boolean;
   isSelected: boolean;
-  onClick: (event: MouseEvent<HTMLInputElement>) => any;
+  onClick: (event: React.MouseEvent) => any;
   onDone: (event: FocusEvent<HTMLInputElement>) => any;
   value: string;
 };
@@ -28,7 +28,7 @@ export class TagListInput extends Component<Props, OwnState> {
       'is-selected': isSelected,
     });
 
-    return (
+    return editable ? (
       <input
         className={classes}
         readOnly={!editable}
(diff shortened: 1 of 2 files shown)


dmsnell (commented):
If we're not going to use `<button>` or `<input type="button">` (which seem to be [the preferred mechanisms](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role)) then we should at least add in `role="button"` and for the link tag a `href`, I think.

It seems reasonable that styling could prevent us from using the preferred tags. In this case a link seems appropriate but I am far from an authority on accessibility. I also know that this is not an exception in the app and a little better accessibility is going to be better than less.

codebykat (changes requested):
This is introducing a horizontal scroll with long tag names that isn't present in production:

Screen Shot 2020-09-22 at 2 43 11 PM

It's also stripping the leading spaces from a test tag I happened to have that has leading spaces... good or bad, I don't know. Different than production. We should be stripping these whitespace characters out anyway, but maybe not here. The spaces do show up in edit mode.

Screen Shot 2020-09-22 at 2 52 34 PM
Screen Shot 2020-09-22 at 2 52 37 PM


Non-blocker... this incidentally breaks something that maybe shouldn't work, but does: in production you can currently right-click a tag and "select all" to select the tag.

In this PR in Electron, selecting all seems to do nothing, but when you switch to Edit mode, all tags are selected. In Firefox it just does nothing (more fine).

It is not possible to drag the mouse cursor over a tag name (here or in production) in order to copy the text, if desired. The only way to copy a tag name would be to enter Edit mode.

I just found a bug with dragging-and-dropping tag names but it's present in production too 🤦‍♀️

belcherj:
This is ready for review

dmsnell (approved):
Thanks for the updates. This was always weird and I think this makes it better. At some point we probably want to do something better with finishing the tag-name edit session (I think right now it's just an `onBlur`). One step at a time though 👍

PR #2357, opened 2020-09-23
Combine two if statements

belcherj:
### Fix

https://github.com/Automattic/simplenote-electron/pull/2345/files/1a328c423ffc8d6d54078985eb3c73319cfb7aa9..43190c13bbc6a2998d6d9d7f6317f0ba7dc72b9d#r493837621

Combines two if statements that perform the same code. They both do the same thing so this consolidates them.

This was a follow-up to #2345

### Test
https://github.com/Automattic/simplenote-electron/pull/2345

### DIFF
--- a/lib/state/electron/middleware.ts
+++ b/lib/state/electron/middleware.ts
@@ -100,13 +100,10 @@ export const middleware: S.Middleware = ({ dispatch, getState }) => {
     const result = next(action);
     const nextState = getState();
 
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
I approve, but might want to wait for @codebykat

codebykat (approved):
Oh yeah I didn't have a good reason for this, in fact I almost did it this way.

PR #2370, opened 2020-09-30
Electron: Handle all keyboard shortcuts that are in the menu, from the menu

codebykat:
### Fix

When we add menu items in Electron we're passing a keyboard shortcut, but we were clobbering those with our keyboard event listeners in the app. This causes them to fire and be handled without the Electron menu flashing, which makes things feel kind of janky and breaks the illusion of a native app.

This also fixes the missing handler for Find Note, although it does the exact same thing as Search in Notes, so it's kind of odd to have them both in the menu?

### Test

1. Go through the keyboard shortcuts and make sure they all work
2. For the ones that are in the menu, verify that:
a) they can be triggered by clicking in the menu
b) triggering them by keyboard shortcut causes the menu to flash
3. Make sure keyboard shortcuts also work in the browser

### DIFF
--- a/desktop/menus/edit-menu.js
+++ b/desktop/menus/edit-menu.js
@@ -66,7 +66,7 @@ const buildEditMenu = (settings, isAuthenticated, editMode) => {
       {
         label: '&Find in Note',
         visible: isAuthenticated,
-        click: editorCommandSender({ action: 'find' }),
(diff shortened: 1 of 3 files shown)


belcherj (approved):
Tested and works well

PR #2371, opened 2020-09-30
Fix: re-layout editor after focus mode change

codebykat:
### Fix

This fixes an issue where the width doesn't get properly adjusted when going to/from focus mode:

Screen Shot 2020-09-29 at 9 24 23 PM

### Test
1. Toggle focus mode on and off
2. Verify that the width gets adjusted accordingly


belcherj (approved):
Great!

PR #2376, opened 2020-09-30
Internal links: use a Monaco link provider

codebykat:
This PR adds a custom `LinkProvider` that searches the note for `simplenote://` links and linkifies them, then handles the click event by opening the linked note in the editor.

If the note ID does not refer to a note we can open, it will silently fail. I'm not sure what else it can or should do in that case.

Also adds a check to the preview component to skip links that can't be opened (i.e. the noteId isn't in our notes list), as otherwise it closes the selected note in favor of.. nothing.

Screen Shot 2020-09-30 at 12 46 06 AM

### DIFF
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -20,6 +20,7 @@ type StateProps = {
   isFocused: boolean;
   note: T.Note | null;
   noteId: T.EntityId | null;
+  notes: Map<T.EntityId, T.Note>;
   searchQuery: string;
   showRenderedView: boolean;
 };
@@ -37,6 +38,7 @@ export const NotePreview: FunctionComponent<Props> = ({
   isFocused,
   note,
   noteId,
+  notes,
   openNote,
   searchQuery,
   showRenderedView,
@@ -89,7 +91,12 @@ export const NotePreview: FunctionComponent<Props> = ({
             }
 
             const [fullMatch, linkedNoteId] = match;
-            openNote(linkedNoteId as T.EntityId);
+            // if we try to open a note that doesn't exist in local state,
+            // then we annoyingly close the open note without opening anything else
+            // implicit else: links that aren't openable will just do nothing
+            if (notes.has(linkedNoteId as T.EntityId)) {
+              openNote(linkedNoteId as T.EntityId);
+            }
             return;
           }
 
@@ -170,6 +177,7 @@ const mapStateToProps: S.MapState<StateProps, OwnProps> = (state, props) => {
     isFocused: state.ui.dialogs.length === 0 && !state.ui.showNoteInfo,
     note,
     noteId,
+    notes: state.data.notes,
     searchQuery: state.ui.searchQuery,
     showRenderedView:
       !!note?.systemTags.includes('markdown') && !state.ui.editMode,
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -429,7 +429,7 @@ class NoteContentEditor extends Component<Props> {
     monaco.languages.registerLinkProvider('plaintext', {
       provideLinks: (model) => {
         const matches = model.findMatches(
-          'simplenote://note/[a-zA-Z0-9-]*',
+          'simplenote://note/[a-zA-Z0-9-]+',
           true, // searchOnlyEditableRange
           true, // isRegex
           false, // matchCase
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -444,10 +444,7 @@ class NoteContentEditor extends Component<Props> {
         };
       },
       resolveLink: (link) => {
-        const href = editor.getModel()?.getValueInRange(link.range);
-        if (!href) {
-          return;
-        }
+        const href = editor.getModel()?.getValueInRange(link.range) ?? '';
         const match = /^simplenote:\/\/note\/(.+)$/.exec(href);
         if (!match) {
           return;
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -450,13 +450,13 @@ class NoteContentEditor extends Component<Props> {
           return;
         }
 
-        const [fullMatch, linkedNoteId] = match;
+        const [fullMatch, linkedNoteId] = match as [string, T.EntityId];
 
         // if we try to open a note that doesn't exist in local state,
         // then we annoyingly close the open note without opening anything else
         // implicit else: links that aren't openable will just do nothing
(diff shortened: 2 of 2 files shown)


belcherj (approved):

belcherj:
Could you update the PR description with more detail for posterity? Great work!

lib/note-content-editor.tsx
```diff
@@ -425,6 +426,44 @@ class NoteContentEditor extends Component<Props> {
...
+        matches.forEach(function (match) {
+          links.push({
+            range: match.range,
+            // don't set URL here because then Monaco skips resolveLink
+            // @cite: https://github.com/Microsoft/vscode/blob/8f89095aa6097f6e0014f2d459ef37820983ae55/src/vs/editor/contrib/links/getLinks.ts#L43:L65
+          });
+        });
+        return {
+          links: links,
+        };
```
dmsnell, line 444: in some performance-critical code we can use a `for` loop to speed up iteration in JS but `.forEach` runs comparable to `.map` and in this case we're giving up semantic clarify by looping and mutating. how about a one-step transform?

```ts
return {
	// don't set URL here because then Monaco skips resolveLink
	// @cite: https://github.com/Microsoft/vscode/blob/8f89095aa6097f6e0014f2d459ef37820983ae55/src/vs/editor/contrib/links/getLinks.ts#L43:L65
	links: matches.map( ( { range } ) => ( { range } ) )
}
```

lib/note-content-editor.tsx
```diff
@@ -425,6 +426,44 @@ class NoteContentEditor extends Component<Props> {
...
+          return;
+        }
+        const match = /^simplenote:\/\/note\/(.+)$/.exec(href);
+        if (!match) {
+          return;
+        }
+
+        const [fullMatch, linkedNoteId] = match;
+        this.props.openNote(linkedNoteId as T.EntityId);
+        link.url = '#'; // tell Monaco to do nothing and not complain about it
```
dmsnell: let's avoid mutating the object passed-in to this function…

```ts
this.props.openNote(…);

return { ...link, url: '#' };
```

that creates a new copy instead of the "spooky action at a distance" involved when we change a property of the object passed in as an argument

lib/note-content-editor.tsx
```diff
@@ -425,6 +427,42 @@ class NoteContentEditor extends Component<Props> {
...
+          // don't set a URL on these links, because then Monaco skips resolveLink
+          // @cite: https://github.com/Microsoft/vscode/blob/8f89095aa6097f6e0014f2d459ef37820983ae55/src/vs/editor/contrib/links/getLinks.ts#L43:L65
+          links: matches.map(({ range }) => ({ range })),
+        };
+      },
+      resolveLink: (link) => {
+        const href = editor.getModel()?.getValueInRange(link.range);
+        if (!href) {
+          return;
+        }
```
dmsnell: here's a stylistic nit but using the newer optional operators I find it nice to collapse multiple checks into one.

```ts
const href = editor.getModel()?.getValueInRange(link.range) ?? '';
const match = /…/.exec(href);
if (!match) {
	return;
}
```

that is, we don't care so much in this function if the `href` is empty so much as we care if it positively matches the internal link format. we can remove the special-casing for the empty link and preserve the same logic as long as we fallback to the empty string.

as I write and test that out though I realize we don't even need this. `/pattern/.exec(undefined)` and `/pattern/.exec(null)` all return `null` so we don't even  need the `?? ''` to make this focus on positively identifying matches over special-casing non-matches

lib/note-content-editor.tsx
```diff
@@ -425,6 +427,42 @@ class NoteContentEditor extends Component<Props> {
...
+        const match = /^simplenote:\/\/note\/(.+)$/.exec(href);
+        if (!match) {
+          return;
+        }
+
+        const [fullMatch, linkedNoteId] = match;
+
+        if (this.props.notes.has(linkedNoteId as T.EntityId)) {
+          this.props.openNote(linkedNoteId as T.EntityId);
```
dmsnell: do we need to do this here or would it make sense to run it by trapping other links? I don't know the answer, I'm just asking to make sure we're not coupling the linking behavior inside the editor with app behaviors outside of the editor. for instance, do we already trap clicked-on links from the editor to intercept them?

codebykat: No, since the simplenote:// links aren't real links I don't think we can trigger the link interceptor from here.

codebykat: That happens over here from the preview component, but this works because they've been transformed into `<a>` tags for preview: https://github.com/Automattic/simplenote-electron/blob/b325afcf2685286d752195a862f6cad4f12ed908/lib/components/note-preview/index.tsx#L84

codebykat: I guess we probably do need that check within the preview component as well, though

lib/note-content-editor.tsx
codebykat: I think I only did this because TypeScript was warning me about it

codebykat: I don't think there's any way for this function to be called with an empty match

dmsnell: > I don't think there's any way for this function to be called with an empty match

I believe that but we don't carry that information in the type system unfortunately, and hey, it's JavaSript so it could happen. we could assert that it matches but why not handle failures anyway.

`?? ''` should make TypeScript happy as well as clarify that we're matching against a string right?

codebykat:
Updated to address comments, and added a check in the preview component so that it also skips any links that can't be opened in the editor (rather than closing the selected note).

belcherj (approved):
Tested, code looks good.

lib/components/note-preview/index.tsx
```diff
@@ -89,7 +91,12 @@ export const NotePreview: FunctionComponent<Props> = ({
             }
 
             const [fullMatch, linkedNoteId] = match;
```
dmsnell, line 93: if we type-cast here we should be able to avoid it redundantly below

```ts
const [fullMatch, linkedNoteId] = match as [string, T.EntityId];
```

dmsnell (approved):
Excellent work on this PR. I left a comment about making the types more convenient for you and feel free to do that and merge without a re-approval; feel free to ignore it since it's just a small convenience thing.

This was a good find allowing us to avoid the complexity of re-implementing the built-in link state machine. It's a straightforward solution.

souljedi:
Great work! I've been waiting for something like this for years.

Would it be possible to add anchor points to the link (to jump somewhere within a note)? For example the heading titles from markdown could be used?

PR #2383, opened 2020-10-05
Merge 2.0.0-beta1

mokagio:
This does nothing else but changing the version to 2.0.0-beta1.

Semantically, going from an RC (release candidate) version to a beta doesn't make much sense. Betas are usually and earlier stages of the release process.

I think it's okay in this case in the context of why those version numbers were used and our internal release process which uses betas as well as how pre-release versions are distributed using the Electron auto-updater via GitHub.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,42 @@
 # Changelog
 
+## [v2.0.0]
+
+### Enhancements
+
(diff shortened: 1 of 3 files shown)


belcherj (approved):
Splendiferous

mokagio:
The amount of work that went into this release 🤯

PR #2387, opened 2020-10-07
Update the Simperium token to SIMPLENOTE: ELECTRON 2.0

belcherj:
### Fix

Updated the Simperium token. 

### Test
1. Visit CircleCI and download the artifact
2. Run the result of this build
3. Does it work?


codebykat (approved):
Tested the OSX build artifact, works great 👍

PR #2397, opened 2020-10-14
Bump version to 2.0.0

mokagio:


belcherj (approved):
Smerge

mokagio:
I guessed "Smerge" means "Squash & merge", so that's what I did 😄 It was a safe thing to try anyway since the PR only had one commit.

PR #2404, opened 2020-10-21
Fix: zoom shortcuts on numeric keypad

### LINKED ISSUE #2400: Keyboard Shortcuts: Font size doesn't change with numeric keypad
### Expected
_Ctrl_ + _+_  ,  _Ctrl_ + _-_  ,and  _Ctrl_ + _0_  should change the font size

### Observed
It only works if using the +,-, and 0 keys located **outside** the numeric keypad

### Reproduced
Those Keyboard Shortcuts only work when not using the numeric keypad.

**Note:** Ctrl+/ does work when using the numeric keypad

### Where did you see the bug

- System: Windows 10 Home. 
- Simplenote app version: 2.0.0-beta1

codebykat:
### Fix

This fixes #2400, a report that these shortcuts only work when triggered from the top keyboard row, and not from the number pad.

In my testing on OSX, the only one that did not work was Zoom In. But based on the report I've also added additional shortcuts to handle Zoom Out (Ctrl + minus) and Default view (Ctrl + zero).

See:

* https://github.com/electron/electron/issues/3332
* https://www.electronjs.org/docs/api/accelerator#available-key-codes

### Test

1. Find a keyboard with a number pad
2. Verify that the numpad keys work to zoom in, out and restore to default zoom
3. Verify that the original keys still work as well

### Release

Fixed a bug preventing zoom level shortcuts from being triggered by keys on the numeric keypad

### DIFF
--- a/desktop/menus/view-menu.js
+++ b/desktop/menus/view-menu.js
@@ -143,7 +143,7 @@ const buildViewMenu = (settings, isAuthenticated) => {
         role: 'ResetZoom',
       },
 
-      // workarounds for numeric keypad zoom in / zoom out,
+      // backup shortcuts for numeric keypad,
       // see https://github.com/electron/electron/issues/5256#issuecomment-692068367
       {
         label: 'Zoom &In',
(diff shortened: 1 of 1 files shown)


belcherj (approved):
I have numeric keypad to test. However, I can use CMD + "+" and CMD + "-" and CMD + "0" on my Apple keyboard.

codebykat:
Rebased and switched to roles to match #2406

Issue #2403, opened 2020-10-21
Electron: Zoom in/out only affects font size within editor


codebykat:
### Expected
Zooming in or out should increase or decrease all text within the app.

### Observed
Only text within the editor is affected.

### Reproduced
OSX / current develop (2.0.0). In the browser it correctly increases all text.

Issue #1037, opened 2018-11-22
Zoom in only impacts notes, not sidebar sizing, which is a major usability issue for large and high DPI monitors


jkterry1:
The basically title says it all:

Zooming in zooms in on notes, not the list of notes. For people with 4k monitors, other very large monitors, or vision issues, this can still leave the sidebar unreadable, which is a major impediment to using simplenote.

In case this is a bug and not a missing feature, this occurs on Ubuntu 18.04 and 18.10 on the most recent release of simplenote

mirka:
We plan on having dedicated typography settings for the editor, so the editor font size can be set independently (#274).

sprite-1:
I hope you don't mean like actually changing the font size within the individual notes itself because I use the zoom functions only when I'm writing rough drafts in focus mode. Then i reset it back to normal afterwards. 🤔

codebykat:
Zooming in now correctly zooms in both the sidebar/tag list and the note.

PR #2405, opened 2020-10-21
[WIP] Fix for Zoom only working for editor contents

codebykat:
Notes to self:
* Note info is not re-rendering when font size changes, something with the state is wrong

Things that are not yet resizing:
* Search bar input
* In side tray, All Notes / Trash and Settings / About etc
* Tag chips / tag input
* Text within dialogs

### Fix

This is a regression given comments on https://github.com/Automattic/simplenote-electron/issues/1037
Fix for #2403 

### Test

1.
2.
3.


codebykat:
Abandoning for #2406

PR #2406, opened 2020-10-21
Electron: Use native zoom

### LINKED ISSUE #2403: Electron: Zoom in/out only affects font size within editor

codebykat:
### Fix

Electron's menuItems support native Zoom functionality using defined [roles](https://www.electronjs.org/docs/api/menu-item#roles). What we're doing now, for reasons unknown to me (maybe it wasn't previously supported?) is using a hidden `fontSize` setting, which currently has the effect that only the editor content zooms in/out. (#2403)

It works fine in the browser, which uses native zooming.

I explored adjusting the font sizes and styles throughout the app in #2405, which got sticky because we have to set the font size **everywhere**, and still have the accessibility issue where all parts of the app (icons, etc) are not  always zooming.

This is a problem we previously had and fixed (#1037), which makes me think it's a regression introduced by the rewrite, or somewhere else along the line. I haven't tested on older app versions.

Anyhow this PR removes the hidden `fontSize` option and allows Electron to add menu items for the native Zoom functionality, which should work better cross-platform. We pass a fixed font-size and line-height to Monaco (16/24) which matches the current default size, and is also adjusted as the app zooms in and out. 

Needs testing on Windows build.

WIll fix #2403 // supersedes #2405 

### Test
1. On Electron build, zoom in and out using shortcut keys or menu, make sure everything looks good
2. On web, zoom in and out using native browser functionality

### Release

Fixed a bug causing zoom in / zoom out to only apply to the editor contents.

### DIFF
--- a/desktop/menus/view-menu.js
+++ b/desktop/menus/view-menu.js
@@ -127,21 +127,26 @@ const buildViewMenu = (settings, isAuthenticated) => {
       ...(isAuthenticated ? [{ type: 'separator' }] : []),
       {
         label: 'Zoom &In',
+        role: 'ZoomIn',
         visible: isAuthenticated,
         accelerator: 'CommandOrControl+=',
-        click: appCommandSender({ action: 'increaseFontSize' }),
+      },
+      {
+        label: 'Zoom &In',
+        role: 'ZoomIn',
+        visible: false,
+        acceleratorWorksWhenHidden: true,
+        accelerator: 'CommandOrControl+Plus',
       },
       {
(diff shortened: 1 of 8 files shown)


belcherj (approved):
The code looks good. Tested on macOS and Windows. Everything works as expected.

PR #2412, opened 2020-10-22
Adds Internal link references

belcherj:
### Change

This PR adds internal link references to the note info panel. When a user opens the note info panel we display a list of notes that contain internal link references to that note. Those references are also clickable allowing the user to navigate to the notes.

Screen Shot 2020-10-22 at 12 23 02 PM
Screen Shot 2020-10-22 at 12 22 44 PM

### Test
1. Add internal links to a bunch of notes
2. Open the note info panel to view the references to that note
3. Ensure that link count and modification date are correct
4. Ensure references are clickable
5. Ensure light and dark mode look good

### Release

Added internal link references to the note info panel

### DIFF
--- a/lib/note-info/reference.tsx
+++ b/lib/note-info/reference.tsx
@@ -46,7 +46,8 @@ export const Reference: FunctionComponent<Props> = ({
     <button className="reference-link" onClick={openNote}>
       <span className="reference-title note-info-name">{reference.title}</span>
       <span>
-        {reference.count} References, Last modified {formattedDate}
+        {reference.count} Reference{reference.count > 1 ? 's' : ''}, Last
+        modified {formattedDate}
       </span>
     </button>
   );
@@ -63,10 +64,7 @@ const mapDispatchToProps: S.MapDispatch<DispatchProps, OwnProps> = (
   dispatch,
   { noteId }
 ) => ({
-  openNote: () => {
-    dispatch(actions.ui.toggleNoteInfo());
-    dispatch(actions.ui.selectNote(noteId));
-  },
+  openNote: () => dispatch(actions.ui.selectNote(noteId)),
 });
 
 export default connect(mapStateToProps, mapDispatchToProps)(Reference);
--- a/lib/note-info/reference.tsx
+++ b/lib/note-info/reference.tsx
@@ -46,7 +46,7 @@ export const Reference: FunctionComponent<Props> = ({
     <button className="reference-link" onClick={openNote}>
       <span className="reference-title note-info-name">{reference.title}</span>
       <span>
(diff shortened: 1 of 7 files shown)


lib/note-info/reference.tsx
```diff
@@ -0,0 +1,72 @@
...
+import type * as S from '../state';
+import type * as T from '../types';
+
+type OwnProps = {
+  noteId: T.EntityId;
+};
+
+type StateProps = {
+  reference:
+    | {
```
codebykat, line 17: Why are these pipes here

lib/note-info/reference.tsx
```diff
@@ -0,0 +1,72 @@
...
+) => ({
+  reference: getNoteReference(state, noteId),
+});
+
+const mapDispatchToProps: S.MapDispatch<DispatchProps, OwnProps> = (
+  dispatch,
+  { noteId }
+) => ({
+  openNote: () => {
+    dispatch(actions.ui.toggleNoteInfo());
```
codebykat: I think the more React-y way to do this would be to add a case to `TOGGLE_NOTE_INFO` that would return false when the `SELECT_NOTE` action is called, unless we can think of any case where we would want the note info panel to remain open (I can't).

codebykat (changes requested):
This is working great! Handful of small comments and questions...

On the Electron build, the width is bumping the modified date to the next line:

Screen Shot 2020-10-22 at 5 44 49 PM

I'm also wondering if we want to have this section next to the internal link section (whether that means moving it up above, or switching the location of the published link and the references):

Screen Shot 2020-10-22 at 5 49 00 PM

Finally, if there's just one reference, it shouldn't say "1 references" 🙃

lib/note-info/reference.tsx
belcherj: The second pipe is an or. So it is the object or undefined. Prettier added the top one so it could add a line break.

lib/note-info/reference.tsx
belcherj: good call.

codebykat (approved):
Tweaked the width very slightly, calling this good!

codebykat:
Screen Shot 2020-10-23 at 1 18 06 PM

dmsnell:
nice work!

I have two ideas:
 - since we're not using anything in the RegExp beyond direct string literal matching could we consider using `String.indexOf()` instead and avoid all the needless overhead of the RegExp engine?
 - since this information doesn't change except when editing the note could we move it into the same space that search indexing takes place, so that we maintain an index of links between notes? the references would be immediately available and could be a straightforward property in the Redux store.

until then there's no problem if this is working, but we should pay attention on large accounts to make sure this doesn't trigger too much avoidable computation

PR #2415, opened 2020-10-22
Change logic to allow for checking checkboxes in preview

### LINKED ISSUE #2390: Markdown Checkboxes are non interactive
### Expected
When making a markdown checkbox and clicking it, the checkbox should be toggled.  An "x" character should be added to the document.  Like this:

- [ ] click me

### Observed
Nothing happens

### Reproduced

1.  add some markdown `- [ ] Click me`
2. switch to markdown preview mode
3. click the checkbox

### Where did you see the bug

- System Make:
- System Model:
- OS: MacOS
- OS version: 
- Browser (if applicable): Brave
- Browser version (if applicable): Updated recently ~month
- Simplenote app version:

belcherj:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/2390

This changes the logic to allow for checkboxes to be checked in preview mode.

- Fixes regex to use `- [ ]` instead of the Unicode characters.
- Adds `note` to the list of the useEffect dependencies so the value will update.

### Test
1. Add some checkboxes to a note
2. Switch to preview mode
3. Check boxes
4. Can you check and uncheck multiple of them?

### Release

Fixed checkboxes so that they can be checked in markdown mode.

### DIFF
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -44,7 +44,6 @@ export const NotePreview: FunctionComponent<Props> = ({
   showRenderedView,
 }) => {
   const previewNode = useRef<HTMLDivElement>();
-
   useEffect(() => {
     const copyRenderedNote = (event: ClipboardEvent) => {
       if (!isFocused) {
@@ -122,7 +121,6 @@ export const NotePreview: FunctionComponent<Props> = ({
           const content = note.content.replace(
             /(- \[x\]|- \[ \])/g,
             (match) => {
-              debugger;
               return matchCount++ === taskIndex
                 ? match === '- [ ]'
                   ? '- [x]'
@@ -139,7 +137,7 @@ export const NotePreview: FunctionComponent<Props> = ({
     previewNode.current?.addEventListener('click', handleClick, true);
     return () =>
       previewNode.current?.removeEventListener('click', handleClick, true);
-  }, []);
+  }, [note]);
 
   useEffect(() => {
     if (!previewNode.current) {
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -118,12 +118,17 @@ export const NotePreview: FunctionComponent<Props> = ({
           const taskIndex = Array.prototype.indexOf.call(allTasks, node);
 
           let matchCount = 0;
-          const content = note.content.replace(/[\ue000|\ue001]/g, (match) =>
-            matchCount++ === taskIndex
-              ? match === '\ue000'
-                ? '\ue001'
-                : '\ue000'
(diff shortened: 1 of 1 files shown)


codebykat (approved):
Works like a charm!

lib/components/note-preview/index.tsx
```diff
@@ -118,12 +117,16 @@ export const NotePreview: FunctionComponent<Props> = ({
...
           let matchCount = 0;
-          const content = note.content.replace(/[\ue000|\ue001]/g, (match) =>
-            matchCount++ === taskIndex
-              ? match === '\ue000'
-                ? '\ue001'
-                : '\ue000'
-              : match
+
+          const content = note.content.replace(
+            /(- \[x\]|- \[ \])/g,
```
dmsnell, line 122: either by adding an `i` flag here or using `\[xX\]` we should allow for either case of `x`

dmsnell, line 122: note also that we have specific rules that these checkboxes must be the first non-whitespace characters on a line. the rule wasn't implemented with the private use area Unicode characters because they exist as exceptions and should not be found except where there's a checkbox. `- [ ]` is much more readily-entered and we probably want to guard against false positives

lib/components/note-preview/index.tsx
```diff
@@ -134,7 +137,7 @@ export const NotePreview: FunctionComponent<Props> = ({
     previewNode.current?.addEventListener('click', handleClick, true);
     return () =>
       previewNode.current?.removeEventListener('click', handleClick, true);
-  }, []);
+  }, [note]);
```
dmsnell, line 140: good catch! though while we're at it it looks more like the dependencies are `note.content` and `searchQuery` so maybe we should follow-up and set those two specifically. thoughts? obviously if the `note` changes the `note.content` might be different but only indirectly so.

belcherj: searchQuery is not used in that useEffect. I do not think it is needed as a dependency. Changing the dependency to a note.content is a good change.

lib/components/note-preview/index.tsx
belcherj: Good change.

belcherj: https://github.com/Automattic/simplenote-electron/pull/2461

belcherj: https://github.com/Automattic/simplenote-electron/pull/2461

lib/components/note-preview/index.tsx
belcherj: https://github.com/Automattic/simplenote-electron/pull/2461

PR #2418, opened 2020-10-23
Note list: prevent scrollbar slider from being very small with many notes

### LINKED ISSUE #2417: slider button size too small when there are many notes
### What

Please do not reduce the size of the slider button to less than about 1 times the size of a character, or more, when there are many notes.

### Why

When there are many notes, the slider button in the main view becomes too smal (about 1/4 of a character height), scarcely visible and difficult to click with the mouse.

### How

The slider button size apparently scales as the ratio between the number of note titles inside the current view over the total number of notes. Just stop decreasing the slider button size below a reasonable value.

codebykat:
### Fix

This sets a minimum height on the scrollbar slider so that it is more usable with a lot of notes.

Will fix #2417 

This appears to affect Electron builds and Webkit browsers only, from my testing Firefox already has a minimum slider size that makes sense.

### Screenshots

**Before:**

Screen Shot 2020-10-23 at 12 40 52 PM

**After:**

Screen Shot 2020-10-23 at 12 41 36 PM

### Release

Prevent the scrollbar slider from becoming tiny when there are many notes in the note list

### DIFF
--- a/scss/_scrollbar.scss
+++ b/scss/_scrollbar.scss
@@ -12,6 +12,7 @@
   background: $studio-gray-10;
   border-radius: 100px;
   border: 4px solid $studio-white;
+  min-height: 24px;
 }
 // hover effect for scrollbar 'thumb'
 ::-webkit-scrollbar-thumb:hover {


belcherj (approved):
Nice change!

PR #2419, opened 2020-10-23
Fix: Add multiple checkboxes at once

### LINKED ISSUE #2409: Adding multiple checkboxes at once broken on 2.0
### Expected
Previously on Simplenote Electron on Windows and Mac you could highlight several items, click Format> Insert Checklist and all items would get a nice check mark next to them

### Observed
Now if you highlight a few rows of text and do the same thing only the last item  gets a check box

### Reproduced

(Image link: https://d.pr/i/4AVWxO )

### Where did you see the bug

- System Make: Macbook Pro
- OS: Catalina
- OS version: 10.15.7
- Simplenote app version: 2.0

Also reported by a user on Windows #3423060-zen

codebykat:
### Fix

This restores the ability to select multiple lines and toggle checkboxes on all of them.

n.b. this runs **toggle checkbox** on all lines of selection, meaning if you select some lines that already have a checkbox, the checkbox will be removed. Is this the desired behavior?

Fixes #2409 

### Test
1. Select multiple lines, Insert Checklist should run the "toggle checkbox" on each of them (inserting or removing depending on if it had one already)
2. Without a selection, Insert Checklist should toggle checkbox on the current line (where the cursor is)

### Release

Restored ability to select multiple lines and add multiple checkboxes at once with Insert Checklist

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -361,7 +361,6 @@ class NoteContentEditor extends Component<Props> {
 
     // try to get any selected text first
     const selection = editor.getSelection();
-    let lineNumber = null;
     if (selection) {
       for (
         let i = selection.startLineNumber;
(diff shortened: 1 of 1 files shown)


belcherj (approved):
Looks great. The only problem I see is if you have a mix of checkboxes and no checkboxes in a selection. I wonder if we should toggle all lines in a selection the opposite of the first line in the selection. Something to think about for another PR.

PR #2422, opened 2020-10-26
When there are no notes suggest creating one

belcherj:
### Fix

When a user does not have any notes suggest adding a new note. Additionally, if a user types a search term that does not match any notes suggest creating a new not with that query.

This also fixes a bug where if you have no notes "Notes Loading" neve disappears.

Screen Shot 2020-10-26 at 4 58 14 PM
Screen Shot 2020-10-26 at 4 59 17 PM
Screen Shot 2020-10-26 at 4 58 42 PM
Screen Shot 2020-10-26 at 4 58 57 PM

### Test
1. Have no notes, observe note list suggests you add a new note
2. Have a search query that does not match any notes. Observe note list suggest that you create one with that search term

### Release

Suggest creating a new note when none exists or notes match search query.

### DIFF
--- a/lib/note-list/no-notes.tsx
+++ b/lib/note-list/no-notes.tsx
@@ -18,7 +18,11 @@ const NoNotes = () => {
     return (
       <div className="note-list-placeholder">
         No Results{' '}
-        <button onClick={() => dispatch(actions.ui.createNote())}>
+        <button
+          onClick={() =>
+            dispatch(actions.ui.createNote({ content: searchQuery }))
+          }
+        >
           Create a new note titled &quot;{searchQuery}&quot;
         </button>
       </div>
--- a/lib/note-list/no-notes.tsx
+++ b/lib/note-list/no-notes.tsx
@@ -29,7 +29,7 @@ const NoNotes = () => {
     <div className="note-list-placeholder">
       No Notes{' '}
       <button onClick={() => dispatch(actions.ui.createNote())}>
-        Create a new note titled &quot;{searchQuery}&quot;
+        Create a new note
       </button>
     </div>
   );
--- a/lib/note-list/no-notes.tsx
+++ b/lib/note-list/no-notes.tsx
@@ -10,31 +10,34 @@ const NoNotes = () => {
   const searchQuery = useSelector((state: S.State) => state.ui.searchQuery);
   const dispatch = useDispatch();
 
-  if (!hasLoaded) {
-    return <div className="note-list-placeholder">Loading Notes</div>;
-  }
+  const getMessage = () => {
+    return hasLoaded
+      ? searchQuery.length
+        ? 'No Results'
+        : 'No Notes'
+      : 'Loading Notes';
+  };
 
-  if (searchQuery.length) {
-    return (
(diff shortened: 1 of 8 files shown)


lib/note-list/no-notes.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import * as S from '../state';
+
+const NoNotes = () => {
+  const hasLoaded = useSelector((state: S.State) => state.ui.hasLoadedNotes);
+  const searchQuery = useSelector((state: S.State) => state.ui.searchQuery);
+  const dispatch = useDispatch();
+
+  if (!hasLoaded) {
+    return <div className="note-list-placeholder">Loading Notes</div>;
```
belcherj: I need to dry up the return values.

iamthomasbishop:
Looks pretty good at first glance @belcherj, I like this type of suggestion/tip on search. I'm not 100% familiar with y'all's color palette in light vs. dark mode, so I'll leave that to @SylvesterWilmott , but the messaging and placement makes sense to me. A couple of suggestions — both very minor, subjective, and to be taken with a grain of salt:

- Feel free to overrule me if there's an established rule/guideline here, but the text for the "create a new note" button feels rather large to me. Perhaps a "secondary" label size that's just a tad smaller would feel less so?

- I also think it'd benefit from a slight bit of margin between it and the "no results" label. I'm not sure what spacing/grid you're using if any, but just a small unit of margin — something like 8px of margin if on an 8px grid — would probably be enough.

Obviously very minor things, and if established rules should negate either, feel free to bypass 😄

belcherj:
@iamthomasbishop I made the text smaller and added padding between the button and the text.

iamthomasbishop:
@belcherj nice, thanks!

Is it just my eyes, or is the blue color being used for the icons in the top toolbar different from the button text below in dark mode? Light mode _seems_ about right but it's hard to tell for sure. Screenshot:

<a href="https://user-images.githubusercontent.com/1200432/97517473-47ae6400-1963-11eb-83f6-aa759953eedd.png"></a>

codebykat (changes requested):
Code looks good, tests good. 👍 

We definitely need to focus the text within the editor after creating the new note, so you can create a new note and immediately continue typing.

Also, should we consider clearing the search query after the new note is created? Right now you end up here, which is kind of a lot of highlighting:

Screen Shot 2020-10-28 at 9 21 00 PM

belcherj:
We are now focusing the editor when changing notes and creating a new note. (https://github.com/Automattic/simplenote-electron/pull/2422/commits/1281ee975f40e56770761e59a73a4f7755288958)

https://github.com/Automattic/simplenote-electron/pull/2434 will clear the search when a new note is created.

codebykat (approved):

SylvesterWilmott:
>Feel free to overrule me if there's an established rule/guideline here, but the text for the "create a new note" button feels rather large to me. Perhaps a "secondary" label size that's just a tad smaller would feel less so?

That is correct, the label size here should 14px based on the design found in Zeplin. @belcherj can we make sure we update this?

belcherj:
Done

PR #2424, opened 2020-10-27
Note List: Add contextual excerpts for search results

codebykat:
This PR adds search query context to the note list previews.

### Screenshots

Screen Shot 2020-12-10 at 10 22 31 PM

Screen Shot 2020-12-10 at 10 22 20 PM

### Test
Use various search strings in both Comfy and Expanded display mode.

### Release
Added search context to the note list

### DIFF
--- a/lib/utils/note-utils.ts
+++ b/lib/utils/note-utils.ts
@@ -12,6 +12,8 @@ export interface TitleAndPreview {
 export const maxTitleChars = 64;
 export const maxPreviewChars = 200;
 
+const isLowSurrogate = (c: number) => 0xdc00 <= c && c <= 0xdfff;
+
 /**
  * Returns a string with markdown stripped
  *
@@ -69,7 +71,6 @@ const getPreview = (content: string, searchQuery?: string) => {
       preview = matches[0];
 
       // don't return half of a surrogate pair
-      const isLowSurrogate = (c: number) => 0xdc00 <= c && c <= 0xdfff;
       return isLowSurrogate(preview.charCodeAt(0)) ? preview.slice(1) : preview;
     }
   }
--- a/lib/utils/note-utils.ts
+++ b/lib/utils/note-utils.ts
@@ -56,7 +56,7 @@ const getPreview = (content: string, searchQuery: string) => {
       const matches = regExp.exec(content);
       if (matches && matches.length > 0) {
         preview = matches[0];
-        preview = preview.replace(/ +/g, ' ').replaceAll('\n', '').trim();
+        preview = preview.replace(/ +/g, ' ').replace(/\n/g, '').trim();
 
         // @todo: add '...' before/after to indication truncation
         return preview;
--- a/lib/utils/note-utils.ts
+++ b/lib/utils/note-utils.ts
@@ -56,16 +56,13 @@ const getPreview = (content: string, searchQuery?: string) => {
     const firstTerm = terms[0].toLocaleLowerCase();
     const leadingChars = 30 - firstTerm.length;
 
+    // prettier-ignore
     const regExp = new RegExp(
-      /* eslint-disable */
       '(?<=\\s|^)[^\n]' + // split at a word boundary (pattern must be preceded by whitespace or beginning of string)
-      '{0,' +
-      leadingChars +
-      '}' + // up to leadingChars of text before the match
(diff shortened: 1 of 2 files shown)


mokagio:
Bumping this to 2.2.0 because it's time to code freeze 2.1.0.

lib/utils/note-utils.ts
```diff
@@ -42,9 +44,37 @@ export const getTitle = (content) => {
...
+  // contextual note previews
+  if (searchQuery?.trim()) {
+    const terms = getTerms(searchQuery);
+
+    // use only the first term of a multi-term query
+    const firstTerm = terms[0].toLocaleLowerCase();
+    const leadingChars = 30 - firstTerm.length;
+
+    const regExp = new RegExp(
+      /* eslint-disable */
```
codebykat: I can't get auto-linting to stop changing my indentation here :(

belcherj (approved):
🚢  Great work!

lib/utils/note-utils.ts
```diff
@@ -54,23 +54,26 @@ const getPreview = (content: string, searchQuery?: string) => {
...
+      '}' + // up to leadingChars of text before the match
+        escapeRegExp(firstTerm) +
+        '.{0,200}(?=\\s|$)', // up to 200 characters of text after the match, splitting at a word boundary
+      'ims'
+      /* eslint-enable */
+    );
+    const matches = regExp.exec(content);
+    if (matches && matches.length > 0) {
+      preview = matches[0];
+      return preview;
```
dmsnell: a quick fix for split surrogate pairs would be to extend to slice the first character if it's a low-surrogate, meaning it was removed from its high-surrogate. in these cases we'll get garbage displayed or the Unicode replacement character �

```ts
const isLowSurrogate = c => 0xDC00 <= c && c <= 0xDFFF;

…
return isLowSurrogate(preview.charCodeAt(0))
	? preview.slice(1)
	: preview;
```

lib/utils/note-utils.ts
dmsnell: indentation I think is governed by `prettier` not `eslint`, so you can try disabling the auto-formatting with [`// prettier-ignore`](https://prettier.io/docs/en/ignore.html)

```ts
// prettier-ignore
const regExp = new RegExp(
	…
);
```

lib/utils/note-utils.ts
```diff
@@ -42,9 +44,37 @@ export const getTitle = (content) => {
...
+  if (searchQuery?.trim()) {
+    const terms = getTerms(searchQuery);
+
+    // use only the first term of a multi-term query
+    const firstTerm = terms[0].toLocaleLowerCase();
+    const leadingChars = 30 - firstTerm.length;
+
+    const regExp = new RegExp(
+      /* eslint-disable */
+      '(?<=\\s|^)[^\n]' + // split at a word boundary (pattern must be preceded by whitespace or beginning of string)
```
dmsnell, line 63: we might want to verify that this doesn't exclude all searches from languages that don't use spaces

won't it also eliminate building excerpts for partial-word searches?

dmsnell:
Just wanted to drop by and give you some props. This is really extremely hard in the browser where we don't have any support for locale-sensitive search.

I just thought of some idea that potentially could help us out here. We know we can get closer to what we want if we search in a version of the note with the "ignorables" removed, such as diacritics, vowel pointing, and combinations of characters that can be represented by a single one. I think that we could probably also store a copy of the `diff-match-patch` diff between the two versions of the note (its raw content and its simplified form) to rebuild the indices into the raw version based on the indices in the simplified form.

If we could do that then I don't know if we'd need to limit search to spaces and other boundaries. We'd also get significantly more punch for our efforts in supporting non-English text.

lib/utils/note-utils.ts
codebykat: No, the search still matches anywhere in the note, this just trims the preview to a word boundary. e.g.

Screen Shot 2020-12-11 at 10 00 02 AM

codebykat: Here's a kanji search:

Screen Shot 2020-12-11 at 10 00 50 AM

dmsnell (approved):
Looking forward to having better search excerpts!

lib/utils/note-utils.ts
codebykat: THANK YOU

Issue #2341, opened 2020-09-22
Duplicated Characters/Text on Mobile Browser


erricgunawan:
When typing in a text in the title and hitting newline, it will duplicate the title on the next line.

Typing a word in the body content also duplicate the word/characters and produce some other errors related to cursor.

User's exact word of report:

> I typed "BETA" for the title and hit newline. It duplicated "BETA" on the next line. I tried to type "Unusable" and it typed "UNUUNNUUUUNUUUUUN..." as I tapped the Shift Key and tried to enter other characters or control keys like Backspace besides "U" or "N".

Per @dmsnell comment [here](https://github.com/Automattic/simplenote-electron/issues/2030#issuecomment-693747809): 

> looks like maybe something on the mobile browser is creating multiple cursors and then edits apply twice.

### Expected
There should be no duplicate words or lines. 

### Observed
Words/characters become duplicated. Some other cursor-related functions also don't work. 

### Reproduced
1. Start a New Note.
2. Enter "BETA" in the Title. 
3. Tap Newline. 
4. IT (the BETA) entered "BETA" on the new line. 
5. Tap Backspace to (try to) erase "BETA" and IT didn't. Instead IT entered what we see on the screenshot below which resulted from multiple taps on only Backspace.

(The screenshot appears to show several Newlines which I may have entered after the second "BETA" which I then followed with a series of only Backspaces.)

Screenshot: 

![Screenshot_20200916-051326](https://user-images.githubusercontent.com/1813857/93403788-aa1b3b80-f8b2-11ea-9a2a-d2d8f6dfced9.png)

Another screenshot that shows the other errors:

![Screenshot_20200916-054552](https://user-images.githubusercontent.com/1813857/93403763-a38cc400-f8b2-11ea-9794-2fa7c867be0f.png)

### Where did you see the bug
- System Make and Model: Moto g5 Plus
- OS: Android
- OS version: 8.1.0
- Browser version (if applicable): Firefox Daylight 80.1.3 and Chrome 85.0.4183.81
- Simplenote app version: current https://staging.simplenote.com/

User noted that it works on https://app.simplenote.com/ (meaning the issue doesn't happen there).

Reported on 3317478-zen

codebykat:
Just noting I can't duplicate on Firefox / iOS so far, might be something Android specific?

theck13:
[...]

PR #2428, opened 2020-10-27
Disable insert cursor shortcuts

### LINKED ISSUE #2427: Ubuntu: CTRL + SHIFT + UP doesn't work as expected
<kbd>CTRL</kbd> + <kbd>SHIFT</kbd> + <kbd>UP</kbd> keyboard combination doesn't work as expected in Simplenote 2.0.0 on Ubuntu.

### Steps to reproduce
1. Create a list
2. Place mouse pointer somewhere on the last list item
3. Press <kbd>CTRL</kbd> + <kbd>SHIFT</kbd> + <kbd>UP</kbd> multiple times

### Expected

Standard highlighting as seen on the screenshot below (Windows 10).

Screenshot_63

### Observed

Multiple pointer lines appeared.

P.S. I haven't tested this with <kbd>CTRL</kbd> + <kbd>SHIFT</kbd> + <kbd>DOWN</kbd> combination, but I assume it's the same as with <kbd>UP</kbd>.

|OS Version|App Version|
|-|-|
|Ubuntu 18.04 LTS|Simplenote 2.0.0|

codebykat:
### Fix

On Ubuntu it's possible to add multiple cursors with `ctrl+alt+UpArrow` and `ctrl+alt+DownArrow`. This PR disables those.

However simply disabling these key combos did not restore the default selection behavior on Ubuntu. As a hack, I added handlers mapped to CmdCtrl + Shift + up/down ... this does have the side effect of subtly changing the behavior on ~other platforms~ OSX but I think unifying the selection behavior is arguably a good thing.

Fixes #2427 

On Windows this **retains the existing behavior of this key combo**.
On Linux this **restores the missing behavior that should be associated with this key combo**.
On OSX this **changes the existing behavior from select-to-beginning/end-of-document to select-up/select-down**.

### Test

See #2427

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -519,24 +519,22 @@ class NoteContentEditor extends Component<Props> {
 
     editor.addAction({
       id: 'insertCursorAboveNoMulticursor',
-      label: 'Add Cursor Above',
-      precondition: 'editorHasSelection',
+      label: 'Select Up',
       keybindings: [
         monaco.KeyMod.CtrlCmd | monaco.KeyMod.Shift | monaco.KeyCode.UpArrow,
       ],
       run: () => {
-        editor.trigger('shortcuts', 'editor.action.insertCursorAbove', null);
+        editor.trigger('shortcuts', 'cursorUpSelect', null);
       },
     });
     editor.addAction({
       id: 'insertCursorBelowNoMulticursor',
-      label: 'Add Cursor Below',
-      precondition: 'editorHasSelection',
+      label: 'Select Down',
       keybindings: [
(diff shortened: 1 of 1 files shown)


dmsnell:
we should check on mobile browsers with this too since some of our reports look like we're getting multiple cursors even though they are reported as note corruption - such as #2341 possibly

(that is, see if this fixes the other issues)

codebykat:
Yeah, if nothing else this will remove the ability to create multiple cursors via keyboard. I believe there's still a way to create them with the mouse unfortunately.

I didn't include this in the PR, as it's probably not worth it, but there's always the nuclear option. I tested this and it works:

```
    editor.onDidChangeCursorSelection(function (e) {
      editor.trigger('listener', 'removeSecondaryCursors', null);
    });
```

belcherj (approved):

PR #2433, opened 2020-10-29
Focus editor when note editor pane is clicked

belcherj:
### Fix

When in narrow mode if you click on the margins of the editor pane it does not focus the editor. This change focuses the editor when you click into that area.

In this screenshot notice the large area to the left of the text. This PR makes that area clickable.

Screen Shot 2020-10-29 at 11 18 19 AM

### Test
1. In the settings set the line length to narrow.
2. Click into the margin
3. Is the editor focused?
4. Make a selection
5. Click into the margin
6. Selection should not change

### Release

Fix: Made margin in editor clickable to focus editor

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -917,6 +917,7 @@ class NoteContentEditor extends Component<Props> {
         className={`note-content-editor-shell${
           overTodo ? ' cursor-pointer' : ''
         }`}
+        onClick={this.focusEditor}
       >
         <div
           className={`note-content-plaintext${


codebykat (changes requested):
Works, and good catch, but I also just noticed that if you're in narrow view and click on a note, it doesn't focus the editor after note selection. Can we fix that too?

belcherj:
This PR should fix that: https://github.com/Automattic/simplenote-electron/pull/2434

codebykat (approved):
All good then!

PR #2434, opened 2020-10-29
Use the correct action in the UI reducer

belcherj:
### Fix

The CREATE_NOTE action does trigger the reducer because it is caught in the middleware. The correct action is CREATE_NOTE_WITH_ID. 

This updates the case statements to trigger on CREATE_NOTE_WITH_ID not CREATE_NOTE

### Test
1. Have a search, create a note
2. It should clear the search
3. Have revisions open
4. Create a new note
5. Revisions should close

### Release

Fix: Clears search and closes revision panel when creating a new note.

### DIFF
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -104,7 +104,7 @@ const editMode: A.Reducer<boolean> = (state = true, action) => {
     case 'TOGGLE_EDIT_MODE': {
       return !state;
     }
-    case 'CREATE_NOTE':
+    case 'CREATE_NOTE_WITH_ID':
       return true;
     default:
       return state;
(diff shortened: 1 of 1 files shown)


dmsnell:
this is a good catch, but we might want to include both instead of swapping one out for the other.

codebykat:
I'm fine with this if it doesn't break something else but I forget what the difference is between these two actions.

belcherj:
I am not sure why we would add an action that doesn't do anything. To me, that would do more harm than good. Seeing CREATE_NOTE there would make me think that it was doing something when it is not.

dmsnell:
> I am not sure why we would add an action that doesn't do anything. To me, that would do more harm than good. Seeing CREATE_NOTE there would make me think that it was doing something when it is not.

only because it's possible to dispatch `CREATE_NOTE`. not a big deal here since we don't conditionally load the middleware.

PR #2435, opened 2020-10-29
Display cannonical lexical tag name

belcherj:
### Fix

When displaying tags on a note this PR uses the canonical lexical version on the Tag entity and not the lexical tag name on the note. Additionally, this PR makes it so during a rename we no longer change the tags on a note if it is just a lexical change.

### Test
1. Have a tag `A`
2. Add tag `a` to a note
3. After adding the tag it should be displayed as `A`
4. Rename tag to `a` in the tag list.
5. Tag should now display as `a` on the note
6. Rename tag `a` to `b`
7. Tag should now display everywhere as `b`

### Release

Display the canonical lexical version ensuring one capitalization or lexical version of a tag is displayed.

### DIFF
--- a/lib/components/tag-chip/index.tsx
+++ b/lib/components/tag-chip/index.tsx
@@ -1,10 +1,12 @@
 import React, { FunctionComponent } from 'react';
 import classNames from 'classnames';
 
+import type * as T from '../../types';
+
 type OwnProps = {
   onSelect?: (event: React.MouseEvent<HTMLDivElement>) => any;
   selected: boolean;
-  tagName: string;
+  tagName: T.TagName | undefined;
 };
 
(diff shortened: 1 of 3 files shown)


codebykat (approved):
This is working well!

It does seem like it re-ordered the tags on a note with multiple tags, I'm not sure why. (This note had three existing tags and I added "CamelCaseTag" to it; it was displayed as the right-most tag until I renamed it).

The tag autocomplete on a note is also case-sensitive; I think if we're going to treat input in a case insensitive way we should probably match suggested tags that way too.

Neither of these is a blocker but if the tag completion can be fixed easily it would be good to do that here.

belcherj:
I'll do tag auto-complete in another PR.

PR #2439, opened 2020-10-30
Allow window close when logged out

belcherj:
### Fix

We display a dialog when a user has unsynchronized changes and they try to close the window. This logic kept the window from closing. When a user is logged out this kept them from closing the window. This change only prevents window closes when the user is logged in.

### Test
1. Be logged out
2. Close electron window, does it close?
3. Be logged in
4. Close electron window, does it close?
5. Be logged in and have an unsynchronized change
6. Close electron window, does it display dialog?

### Release

Allow for window to be closed when the user is logged out.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -159,7 +159,6 @@ module.exports = function main() {
     // When we receive a close event prevent the window from closing and
     // tell the app to check for unsynchronized notes.
     mainWindow.on('close', (event) => {
-      console.log(isAuthenticated);
       if (isAuthenticated) {
         event.preventDefault();
         mainWindow.webContents.send('appCommand', { action: 'closeWindow' });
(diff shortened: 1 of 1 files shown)


codebykat (approved):

desktop/app.js
```diff
@@ -92,6 +93,8 @@ module.exports = function main() {
     Menu.setApplicationMenu(appMenu);
 
     ipcMain.on('appStateUpdate', function (event, args) {
+      const settings = args['settings'] || {};
```
dmsnell, line 96: any particular reason for the index convention here vs. `args.settings` or a one-liner…

```js
isAuthenticated = 'settings' in args && 'accountName' in args.settings;
```

belcherj: No reason other than consistency: https://github.com/Automattic/simplenote-electron/blob/develop/desktop/menus/index.js#L12

PR #2440, opened 2020-10-30
Fix updater dialog to properly resolve

belcherj:
### Fix

Fixes the updater. The dialog actions were broken because of a deprecated API that was failing silently.

This uses the new async flow to ensure that updates work as expected.

### Test
1. Download the exe from this PR: https://github.com/Automattic/simplenote-electron/pull/2438
2. Install, run app
3. Observe that the update dialog opens
4. Tell it to install

### Release

Fixed the Windows updater

### DIFF
--- a/desktop/app-quit/index.js
+++ b/desktop/app-quit/index.js
@@ -10,12 +10,12 @@ function AppQuit() {
 }
 
 AppQuit.prototype.shouldQuitToBackground = function () {
-  if (this.canQuit === false) {
-    return true;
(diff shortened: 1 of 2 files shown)


mokagio:
Bumping this to 2.2.0 because the code freeze for 2.1.0 is due today.

This seems like a good bug to fix sooner rather than later, though. I'd be happy to ship a new beta for 2.1.0 if you want to get this into it once ready.

codebykat (approved):

PR #2446, opened 2020-11-03
Catch errors when parsing invalid JSON

### LINKED ISSUE #1105: No error is shown when user attempts to import invalid JSON file
#### Steps to reproduce
1. Rename a non-JSON file to `notes.json`.
2. Attempt to import that file from the Simplenote (.json) importer.

#### What I expected

An error to be displayed.

#### What happened instead

The progress bar runs indefinitely.

#### Simplenote version
1.3.3

belcherj:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/1105

If you import an invalid JSON file it would fail silently. This ensures that we catch errors when JSON is invalid. I added the typing and default values because when there is an error we are passing undefined to `<ImportProgress>`.

### Test
1. Have a text file, add a json extension
2. Try and import it
3. See it fail and give you an error message

### Release

Fix: Show error message when trying to import invalid JSON.

### DIFF
--- a/lib/utils/import/simplenote/index.ts
+++ b/lib/utils/import/simplenote/index.ts
@@ -42,7 +42,14 @@ class SimplenoteImporter extends EventEmitter {
         return;
       }
 
-      const dataObj = JSON.parse(fileContent);
+      let dataObj;
+      try {
+        dataObj = JSON.parse(fileContent);
+      } catch (error) {
+        this.emit('status', 'error', 'Invalid json file.');
+        return;
+      }
+
       const noteCount =
         dataObj.activeNotes.length + dataObj.trashedNotes.length;
       const processedNotes = {
--- a/lib/dialogs/import/source-importer/progress/index.tsx
+++ b/lib/dialogs/import/source-importer/progress/index.tsx
@@ -1,10 +1,19 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { FunctionComponent } from 'react';
 
 import ImportProgressBar from './bar';
 import ImportProgressText from './text';
 
-const ImportProgress = ({ currentValue, endValue, isDone }) => {
+type OwnProps = {
+  currentValue?: number;
+  endValue?: number;
+  isDone: boolean;
+};
+
+const ImportProgress: FunctionComponent<OwnProps> = ({
+  currentValue = 0,
+  endValue = 0,
+  isDone,
+}) => {
   return (
     <section>
       <ImportProgressBar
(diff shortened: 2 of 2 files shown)


lib/utils/import/simplenote/index.ts
```diff
@@ -42,7 +42,14 @@ class SimplenoteImporter extends EventEmitter {
...
       }
 
-      const dataObj = JSON.parse(fileContent);
+      let dataObj;
+      try {
+        dataObj = JSON.parse(fileContent);
+      } catch (error) {
+        this.emit('status', 'error', 'Invalid json file.');
+        return;
+      }
```
dmsnell, line 51: while we're at it, since we still leave open the possibility of runtime crashes after this `try/catch` maybe we should consider adding a real parser to the input documents so that we only allow valid documents or valid lists of notes.

```ts
const parseJson = (input: string, default: unknown) => {
	try {
		return JSON.parse(input);
	} catch (e) {
		return default;
	}
}

const parseNote = (data: unknown, defaults: Partial<T.Note>) => {
	if (null === data || 'object' !== typeof data) {h
		return null;
	}

	…

	return {
		...defaults,
		…,
		modificationDate: convertDate(data.modificationDate || data.lastModified),
		…
	}
}

const parseExportData = (input: string) => {
	const data = parseJson(input, null);
	if (null === data || 'object' !== typeof data) {
		return null;
	}

	const {activeNotes, trashedNotes} = data;
	if ((activeNotes && !Array.isArray(activeNotes)) || (trashedNotes && !Array.isArray(trashedNotes))) {
		return null;
	}

	return [
		…(activeNotes || []).map(note => parseNote(note, {isTrashed: false})).filter(a => null !== a),
		…(trashedNotes || []).map(note => parseNote(note, {isTrashed: true})).filter(a => null !== a)
	]
}
```

can't remember if `isTrashed` is the right property of the note. point is, we can do a full parse and prevent errors from creeping deeper into the app if we eliminate them here. we had this problem before with a very old account whose note object didn't contain the expected properties and the app failed to load - whitescreened it.

belcherj: I would like to address this in another PR as we want to get this fix in before Friday and I am constrained on time this week.

codebykat (changes requested):
It still crashes for me trying to import a random .json file (not a Simplenote export). Is that expected?

Screen Shot 2020-11-11 at 12 15 06 PM

Confirmed that it fails on a text file renamed to json.

belcherj:
Yeah, Dennis' comment should probably be addressed to fix that.

codebykat:
I'm okay with that being a follow-up PR too, this fixes one potential error out of two. It seems to me at first thought that it's far more likely that someone would import an actual JSON file that isn't the right format, than that they would import a non-JSON file with a JSON extension.

codebykat:
Merging this, let's follow up on the JSON validation here: https://github.com/Automattic/simplenote-electron/issues/2517

PR #2455, opened 2020-11-06
Queue note updates after tag deletion

belcherj:
### Fix

When a tag is deleted we loop through all the notes and delete the tag. We were not however telling node-simperium to queue a change for that note.

This change queues a note update for all the notes that need to have the tag removed.

### Test
1. Have a tag that is added to several notes
2. Deleted tag
3. Have another device open, does the tag get removed from notes?

### Release

Fixed issue where deleting a tag did not immediately update the note.


codebykat (approved):
This is working, though I am not sure I ever reproduced the bug.

PR #2460, opened 2020-11-09
Update/ignore accents in search

belcherj:
### Test

1.
2.
3.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -24,6 +24,7 @@ import {
 } from './utils/task-transform';
 import IconButton from './icon-button';
 import ChevronRightIcon from './icons/chevron-right';
+import removeAccents from './utils/remove-accents';
 
 import * as S from './state';
 import * as T from './types';
@@ -384,7 +385,9 @@ class NoteContentEditor extends Component<Props> {
       return [];
     }
 
-    const content = model.getValue().normalize().toLowerCase();
+    const content = removeAccents(
+      model?.getValue().normalize().toLowerCase() || ''
+    );
 
     const highlights = terms.reduce(
       (matches: monaco.languages.DocumentHighlight, term) => {
--- a/lib/note-list/decorators.tsx
+++ b/lib/note-list/decorators.tsx
@@ -1,27 +1,96 @@
 import React from 'react';
 import { escapeRegExp } from 'lodash';
-import replaceToArray from 'string-replace-to-array';
+import removeAccents from '../utils/remove-accents';
 
 import { withoutTags } from '../utils/filter-notes';
 
-export const decorateWith = (decorators, text) =>
-  decorators.length > 0
-    ? decorators
-        .reduce((output, { filter, replacer }) => {
-          const searchText = 'string' === typeof filter && withoutTags(filter);
-          const pattern =
-            searchText && searchText.length > 0
-              ? new RegExp(escapeRegExp(searchText), 'gi')
-              : filter;
+function replaceToArray(string, regexp, newValueOrFn) {
+  console.log(string);
+  var output = [];
 
-          return replaceToArray(output, pattern, replacer);
-        }, text)
-        .map((chunk, key) =>
-          chunk && 'string' !== typeof chunk
-            ? React.cloneElement(chunk, { key })
-            : chunk
-        )
-    : text;
+  var replacerIsFn = typeof newValueOrFn === 'function';
+
+  var storedLastIndex = regexp.lastIndex;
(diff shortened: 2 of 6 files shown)


lib/utils/remove-accents.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns a string with accents stripped
+ *
+ * @param {String} inputString string for which to remove accents
+ * @returns {String} string with accents removed
+ */
+const removeAccents = (content: string): string => {
+  return content.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
```
dmsnell, line 8: this is a good start here but have a lot more to do. for instance, `ø` is code point `U+00F8` and things like Hebrew vowel pointing and Japanese dukuten/handakuten should also go. sadly I was unable to find a satisfactory method native to JS to do this. `Intl.Collator` provides an appropriate string comparator but unfortunately doesn't provide the stripped text or provide string matching.

that means we're probably limited practically to _getting closer_ to support than _supporting_ localized search, but we can hit some of those common groups here at the same time. this would be a good place for unit tests that show the extent of our support as well as places we know we aren't in compliance.

might want to cite copied code as well 👍

lib/search/index.ts
```diff
@@ -234,11 +235,11 @@ export const middleware: S.Middleware = (store) => {
       }
 
       const searchText = titleOnly ? getTitle(note.content) : note.content;
-
+      const normalizedSearchText = removeAccents(searchText);
       if (
         searchTerms.length > 0 &&
         !searchTerms.every((term) =>
-          searchText.includes(term.toLocaleLowerCase())
+          normalizedSearchText.includes(term.toLocaleLowerCase())
```
dmsnell, line 242: the `content` in this file is already processed for searching and already does things like normalize and lower-case the text. there's no need to continue to recalculate it for every note on every search when we could add the transformation to the place where the search index is updated in response to note edits.

lib/note-list/decorators.tsx
```diff
@@ -1,27 +1,96 @@
...
+    .reduce((output, { filter, replacer }) => {
+      const searchText =
+        'string' === typeof filter && removeAccents(withoutTags(filter));
+      const pattern =
+        searchText && searchText.length > 0
+          ? new RegExp(escapeRegExp(searchText), 'gi')
+          : filter;
+      console.log(output);
+      return replaceToArray(output, pattern, replacer);
+    }, removeAccents(text))
```
dmsnell, line 68: won't this remove accents from all text?

lib/note-list/decorators.tsx
```diff
@@ -1,27 +1,96 @@
...
+
+    if (result[0] === '') {
+      // When the regexp is an empty string
+      // we still want to advance our cursor to the next item.
+      // This is the behavior of String.replace.
+      regexp.lastIndex++;
+    }
+
+    if (index !== lastIndex) {
+      output.push(string.substring(lastIndex, index));
```
dmsnell, line 29: I believe that this is going to produce some odd results. the act of removing parts of the text will shift the array indices of the search results. if there are lots of accents in the text I think this could get pretty far away from where it should be.

this is hard 😄

lib/note-content-editor.tsx
```diff
@@ -384,7 +385,9 @@ class NoteContentEditor extends Component<Props> {
       return [];
     }
 
-    const content = model.getValue().normalize().toLowerCase();
+    const content = removeAccents(
+      model?.getValue().normalize().toLowerCase() || ''
+    );
```
dmsnell, line 390: see comment below: I believe that notes with lots of removed characters will end up shifting this too far and end up wrong/confusing. there's an open issue I think could already be related to this with checkboxes. (can't find it off hand).

codebykat:
Copying @dmsnell's [comment from a different PR](https://github.com/Automattic/simplenote-electron/pull/2424#issuecomment-743335175) since it would make sense to incorporate it as part of this larger task.

> I just thought of some idea that potentially could help us out here. We know we can get closer to what we want if we search in a version of the note with the "ignorables" removed, such as diacritics, vowel pointing, and combinations of characters that can be represented by a single one. I think that we could probably also store a copy of the diff-match-patch diff between the two versions of the note (its raw content and its simplified form) to rebuild the indices into the raw version based on the indices in the simplified form.

> If we could do that then I don't know if we'd need to limit search to spaces and other boundaries. We'd also get significantly more punch for our efforts in supporting non-English text.

belcherj:
Closing till we have time to focus on this. This is a naive implementation that I think needs more thought.

PR #2461, opened 2020-11-10
Only take into account checkboxes at beginning of line in preview mode

belcherj:
### Fix

Follow up from https://github.com/Automattic/simplenote-electron/pull/2415

Only match checkboxes that occur at the beginning of a line. Users may have `- [ ]`s elsewhere in the text that do not occur at the beginning of a line.

Additionally, this PR also checks for - [X] in addition. to - [x]. It also tightens the dependencies of the useEffect from `note` to `note.content` because we do not use anything else in the note object.

The regex for detecting checkboxes has also changed to ignore checkboxes that are in a list. Checkboxes in lists are not valid GitHub flavoured markdown and the native app clients do not allow them.

### Test
1. Have a note with a checklist and a `- [ ]` somewhere else in the text that is not at the beginning of the line
2. Go into preview mode, do checkboxes check properly?
3. Have a checklist item with - [X]
4. Go into preview mode and uncheck it, does it uncheck?

### Release

- Fixes to checkboxes in preview mode

### DIFF
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -105,35 +105,42 @@ export const NotePreview: FunctionComponent<Props> = ({
 
           return;
         }
+      }
 
-        if (node.className === 'task-list-item') {
-          event.preventDefault();
-          event.stopPropagation();
+      // There are times when showdown will put lists inside of the same
+      // UL as a tasklist. This causes those lists to look like they are
+      // checkboxes. This insures that we are only getting true checkboxes.
+      const element =
+        event?.target?.tagName === 'INPUT'
+          ? event.target.parentElement
+          : event.target;
+      if (element?.children[0]?.tagName === 'INPUT') {
+        event.preventDefault();
+        event.stopPropagation();
 
-          const allTasks = previewNode!.current.querySelectorAll(
-            '[data-markdown-root] .task-list-item'
-          );
-          const taskIndex = Array.prototype.indexOf.call(allTasks, node);
+        const allTasks = previewNode!.current.querySelectorAll(
+          '[data-markdown-root] .task-list-item'
+        );
+        const taskIndex = Array.prototype.indexOf.call(allTasks, element);
 
-          let matchCount = 0;
-          const content = note.content.replace(
-            checkboxRegex,
-            (match, prespace, inside, postspace) => {
-              const newCheckbox =
-                matchCount++ === taskIndex
-                  ? inside === ' '
-                    ? '- [x]'
-                    : '- [ ]'
-                  : inside === ' '
-                  ? '- [ ]'
-                  : '- [x]';
-              return prespace + newCheckbox + postspace;
-            }
-          );
+        let matchCount = 0;
+        const content = note.content.replace(
+          checkboxRegex,
+          (match, prespace, inside, postspace) => {
+            const newCheckbox =
+              matchCount++ === taskIndex
+                ? inside === ' '
+                  ? '- [x]'
+                  : '- [ ]'
+                : inside === ' '
+                ? '- [ ]'
+                : '- [x]';
+            return prespace + newCheckbox + postspace;
+          }
+        );
 
-          editNote(noteId, { content });
-          return;
-        }
+        editNote(noteId, { content });
+        return;
       }
     };
     previewNode.current?.addEventListener('click', handleClick, true);
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -117,7 +117,7 @@ export const NotePreview: FunctionComponent<Props> = ({
           let matchCount = 0;
 
           const content = note.content.replace(
-            /(- \[x\]|- \[ \])/g,
+            /^(- \[x\])|(- \[ \])/gm,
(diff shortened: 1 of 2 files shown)


lib/components/note-preview/index.tsx
```diff
@@ -117,7 +117,7 @@ export const NotePreview: FunctionComponent<Props> = ({
           let matchCount = 0;
 
           const content = note.content.replace(
-            /(- \[x\]|- \[ \])/g,
+            /^(- \[x\])|(- \[ \])/gim,
```
dmsnell: this only allows checkboxes to appear at the beginning of the line but they only need to be the first non-whitespace characters on a line (and apparently inside lists). check out how it's done in the utility function

https://github.com/Automattic/simplenote-electron/blob/8b6b30197c1a3d6ffa7b5cacc9a17b1157b6ee29/lib/utils/task-transform.ts#L1-L6

belcherj: Checkboxes in lists are not valid markdown nor is it the way it works on Android or iOS. I am going to modify that logic to unify it with the way the mobile clients work.

lib/utils/task-transform.ts
```diff
@@ -1,6 +1,15 @@
...
+ *
+ * Checkboxes are valid if they:
+ * - exist as the first non-whitespace on a line
+ * - take one of the following forms ( - [ ] | - [x] | - [X])
+ */
+export const checkboxRegex: RegExp = /^(\s*)- \[( |x|X)\](\s)/gm;
+
 export const withCheckboxCharacters = (s: string): string =>
   s.replace(
-    /^(\s*(?:[-+*\u2022]\s)?)- \[( |x|X)\](\s)/gm,
```
codebykat, line 3: Isn't this going to break detection of some valid checkboxes in the editor? This is also used in the note-content-editor component. I forget what the unicode one is but @dmsnell added them the last time I left them out in a PR.

codebykat (changes requested):

lib/utils/task-transform.ts
belcherj: It is fixing an inconsistency with the native apps and web preview mode. Previous to this change you could have a checkbox in a list `- - [ ]`. This change makes it so those check boxes will no longer be rendered.

belcherj: Test:
- [ ]

- sd
- - [ ]

belcherj: - [ ]

belcherj: Web editing is the only place where you could have checkboxes as part of a list:
Screen Shot 2020-11-11 at 2 56 05 PM

belcherj: Screen Shot 2020-11-11 at 3 04 57 PM

Production is also broken if you have more than one space before the checkox in a list.

codebykat: What is `\u2022`?

I'm not sure I agree we shouldn't support this.

codebykat:
I can't reproduce whatever we're trying to fix here. Behavior seems identical between this branch and develop. A `- [ ]` in the middle of a line does not appear to render as a checkbox on `develop`, and checkboxes after indents or in a list seem to behave identically.

lib/utils/task-transform.ts
dmsnell: `\u2022` is the Unicode character for a bullet: `•`

belcherj:
Use this note:
```
Crème Bruléee

Creme - [x] Brulee

Creme Brule

Creme Brulee

Creme Bruleedsdsdsds

Brulee
 - [ ] sdsdsds
 - [ ] sdsdsdsd
     - [ ] sdsdsds
  - [x] sdsdsdsd
             - [ ] ujyuyu
- [x] dsdsdsds
         - [x] ddfdf


ssss
- - [x] efefdf
- - [ ] sfsfsff
- - [x] cdffdfd
- - [ ] dsdsddsddsdsdsdsdsd
-  - [ ] dfsfdfd
-  - [ ] cddfsfdfvdf
-  - [x] dssfsfdfdf
```

- Go into preview mode
- Observe that the checkbox in the middle of a line is not clickable.

codebykat:
Okay, I've confirmed that, but it's also not rendering that middle-of-line checkbox on `develop` or on the current app.simplenote.com.

Screen Shot 2020-11-12 at 12 15 03 PM

Screen Shot 2020-11-12 at 12 16 42 PM

belcherj:
This PR is still failing to properly account for `- [ ]` and `- [x]` that are not checkboxes in Preview mode. When they appear in the text and not as checkboxes they cause problems.

This note for example is broken:
```
This is a bunch of text

yup
- [x] hi
- [ ] one
- [x] two
- [ ] three
- [x] four
- [ ] five

- [x] hi
	- [x] hello🤷‍♀️😍
	- [x] again	
	- [x] some changes
	- [x] some other changes
 
   - [x] test

Creme - [x] Brulee

- [x] hell

indents
	one hello - [x] testing
		two
			three
```
 The first set of checkboxes is not clickable.

belcherj:
Closing as this appears to be resolved.

PR #2463, opened 2020-11-10
Fix unsynchronized notes log out action

belcherj:
### Fix

There is a typo in the logout action. This bug occurs when you have unsynchronized notes and you click the logout button in the dialog.

### Test
1. Turn off wifi
2. Make a change to a note
3. Log out
4. Can you log out?

### Release

- Fixed log out button when there are unsynchronized changes


codebykat (approved):
I would have thought `REALLY_LOG_OUT` was the correct one (verb not noun?) but it's fine either way.

dmsnell (approved):

PR #2469, opened 2020-11-13
Merge release 2.1.0

mokagio:


mokagio:
I'm going to take the liberty of _smerging_ this, just to make it easier for me to code freeze 2.2.0 next.

I feel this is alright to do because of the two commits on this branch:

- the first is a cherry pick from an [already approved PR](https://github.com/Automattic/simplenote-electron/pull/2440) and it only touches the release notes, so even if this is not appropriate there wouldn't be issues for the users
- the second is a version bump; highly unlikely it would cause an issue

Issue #2425, opened 2020-10-27
Electron: CTRL + Click (follow link) doesn't work


djordjeandrejevic:
Follow link feature isn't working as expected in Simplenote 2.0.0 (outside Preview mode).

**Ubuntu** - nothing happens after clicking on a URL. After few clicks it only underlines some part of the URL, as seen on the screenshot below.

![2020-10-27_19-59](https://user-images.githubusercontent.com/8862464/97353841-d56f4e00-1894-11eb-9576-7d5141b185d6.png)

**Windows** - also wasn't able to open any browser. Instead, the dialog on the screenshot below was shown. Even after selecting _Look for an app in the Microsoft Store_ I wasn't able to select any of the installed browsers (or at least install a new one).

![Screenshot_62](https://user-images.githubusercontent.com/8862464/97354451-b1f8d300-1895-11eb-8575-5dff4dc8c713.png)

|OS Version|App Version|
|-|-|
|Ubuntu 18.04 LTS|Simplenote 2.0.0|
|Windows 10 (build 2004)|Simplenote 2.0.0|

codebykat:
Hi @djordjeandrejevic, thanks for reporting these issues. Do you see the "Follow Link" popup when you hover over a link? Does clicking on that work?

<img width="227" alt="Screen Shot 2020-10-27 at 12 59 06 PM" src="https://user-images.githubusercontent.com/52152/97355499-36c2fd00-1854-11eb-9213-a06ae8440fe7.png">

djordjeandrejevic:
Hi @codebykat, yes I see it, but the behavior is the same in both Ubuntu and Windows.

reginabally:
I'm able to replicate this on macOS 10.15.7 (19H2) with Simplenote 2.0.0.

I see the "Follow Link" popup when I hover over a link + pressing Cmd key. Clicking the link with Cmd key pressed does nothing.

Also, I think it's worth mentioning that Cmd + Click works on the web app.

mokagio:
Bumping this to 2.2.0 because it's time to code freeze 2.1.0.

reginabally:
Another report in 3442967-zen. Windows 10 with Simplenote 2.0.0.

reginabally:
Another report in 3476569-zen. Windows 10 with Simplenote 2.0.0.

codebykat:
Will be fixed in 2.2.0

PR #2470, opened 2020-11-13
Fix: Inability to visit external links on Electron

codebykat:
### Fix

Fix for #2425
This is probably the issue reported upstream here: https://github.com/microsoft/vscode/issues/84265

~First hack: I tried using a custom link provider, but it only works the first time, because Resolve Link is only called once. Also it removed the ability to show a hyperlink in the Follow Link popup (because Monaco didn't know the URL).~

~Plus this would have required defining our own regex for URL detection, boo.~

~Second hack: Add a mouseDown event, check to see if a link was clicked on (or the Follow Link popup), run `viewExternalUrl`. Profit?~

Third hack: JUST SPOOF THE USER-AGENT to fool vscode's platform detection into calling `window.open` with the URL. 😭

### Test
1. Cmd/Ctrl click on links to external sites, verify they open in a browser
2. Use Follow Link popup, ditto
3. Make sure `simplenote://` links still work
4. Verify this still works in the browser as well as Electron build

### Release

Fixed a bug that prevented Follow Link from working within the note editor.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -137,6 +137,11 @@ module.exports = function main() {
 
     mainWindowState.manage(mainWindow);
 
+    // this TERRIBLE HACK forces vscode to call window.open(url) rather than window.open()
+    // see https://github.com/microsoft/monaco-editor/issues/628
+    mainWindow.webContents.userAgent =
+      mainWindow.webContents.userAgent + '/Edge/WebView/FakeUA';
+
     mainWindow.webContents.on('new-window', function (event, linkUrl) {
       event.preventDefault();
       shell.openExternal(linkUrl);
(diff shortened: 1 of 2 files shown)


codebykat:
Well this annoyingly still has the "no handler found" error on Windows, though the link does open.

We should be catching the `window.open` in `app.js`, and yet... 🤔

codebykat:
Welp. After two days of hacking on this, it honestly seems like the best and easiest solution is... adding a couple of strings to the User-Agent so that Monaco will call `window.open(url)`. I just tested on Windows and it works. 😭

belcherj (approved):
Nice find!

codebykat:
Opened an upstream issue reporting this: https://github.com/microsoft/monaco-editor/issues/2220

codebykat:
Noting this has been patched upstream so we will eventually be able to take it out.

PR #2483, opened 2020-11-19
Add appID to fix Windows notification management

### LINKED ISSUE #2482: Windows 10: Can't adjust notification settings
Hi,
I'm a bit of a noob, and don't know if this is even in the right place,
but I'm using **Simplenote v2.1.0** on Windows 10 (v20H2)(but it was the same in v2004),
and every time I make a note change from my phone,
my Windows 10 PC gives me the notification in the corner, along with an annoying system sound every few seconds.
**I would like to disable this noise, and only the noise.**
On the notification is written “electron.app.Simplenote”,
but when I open the Windows 10 notifications settings window,
those words don't appear, and I'm not sure which of the listed apps (if any) is the relevant one.
(Settings, Focus assist, Backup settings, Wireless, AutoPlay, Security and Maintenance, Audio, Battery saver, Microsoft Store, Print Notification, Settings, Suggested)
My only option seems to be to disable sounds for all notifications, which I don't want to do.
Am I being dumb?  How can I achieve this?
Thanks!

codebykat:
### Fix

Looks like we missed a step in setting up notifications for Windows. Right now Simplenote doesn't show up in the notifications manager on Windows, and if you opt-out you can't re-enable them.

Fixes #2482 

> On Windows 10, a shortcut to your app with an Application User Model ID must be installed to the Start Menu. This can be overkill during development, so adding node_modules\electron\dist\electron.exe to your Start Menu also does the trick. Navigate to the file in Explorer, right-click and 'Pin to Start Menu'. You will then need to add the line app.setAppUserModelId(process.execPath) to your main process to see notifications.

https://www.electronjs.org/docs/tutorial/notifications#windows

Related: https://github.com/electron/electron/issues/10864 and https://github.com/electron/electron/issues/24330

### Test

1. After receiving at least one notification from Simplenote
2. On Windows, go to Settings>System>Notifications & actions
3. Verify that Simplenote appears under "Get notifications from these senders"
4. Toggle off sounds and verify that only the banner appears, with no sound

### Release

Fixed a bug causing the app to be missing from Windows 10 notification settings.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -28,7 +28,7 @@ module.exports = function main() {
   let mainWindow = null;
   let isAuthenticated;
 
-  app.setAppUserModelId(process.execPath);
+  app.on('ready', () => app.setAppUserModelId('com.automattic.simplenote'));
 
   app.on('will-finish-launching', function () {
     setTimeout(updater.ping.bind(updater), config.updater.delay);
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -28,8 +28,6 @@ module.exports = function main() {
   let mainWindow = null;
   let isAuthenticated;
 
-  app.on('ready', () => app.setAppUserModelId('com.automattic.simplenote'));
-
   app.on('will-finish-launching', function () {
     setTimeout(updater.ping.bind(updater), config.updater.delay);
     app.on('open-url', function (event, url) {
@@ -249,5 +247,6 @@ module.exports = function main() {
   // This method will be called when Electron has finished
   // initialization and is ready to create browser windows.
   app.on('ready', activateWindow);
+  app.on('ready', () => app.setAppUserModelId('com.automattic.simplenote'));
   app.on('activate', activateWindow);
 };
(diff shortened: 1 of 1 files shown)


codebykat:
Oh hey, my second try actually worked!

Screen Shot 2020-11-19 at 3 09 43 PM

Screen Shot 2020-11-19 at 3 09 55 PM

desktop/app.js
```diff
@@ -28,6 +28,8 @@ module.exports = function main() {
   let mainWindow = null;
   let isAuthenticated;
 
+  app.on('ready', () => app.setAppUserModelId('com.automattic.simplenote'));
```
dmsnell: looks great but would it benefit us to put it down at the end with the other `app.on('ready', …)` hook?

codebykat: HAHA okay

codebykat: I kept it as two function calls for symmetry's sake https://github.com/Automattic/simplenote-electron/pull/2483/commits/31ff235eb907a604ddb1d6be786eca4952918bde

dmsnell (approved):

PR #2490, opened 2020-11-25
Fix keyboard shortcuts on Windows/Linux

codebykat:
## Fix

This fixes sporadically broken shortcut keys on Windows and Linux, including Select All, Undo/Redo, and New Note. Annoyingly I spent awhile debugging this but wasn't able to figure out why some shortcuts are working and others seem to be failing to trigger the Electron accelerators (only on Win/Linux, they all work as expected on the Mac). It doesn't seem related to `editMode`, `roles`, whether the handler is `appCommand` vs `editorCommand`, or anything else I could think of. It might be an obscure Electron bug or an obscure Monaco bug, or a combination of the two.

Anyhow I want to get this fix in, and having redundant handlers for certain functions isn't the worst thing that ever happened, so this is ready for review and I can debug it further later on.

Fixes #2426, #2432, #2436

## Test

On the Windows and Linux app builds, verify that keyboard shortcuts consistently work, especially New Note, Select All, and Undo/Redo. Verify that the context (right-click) menu and clicking on the menu options (e.g. File -> New Note, Edit -> Select All) continue to behave themselves as well.

## Release Notes

Added: Fixed keyboard shortcuts sometimes not working on Win/Linux

Issue #2436, opened 2020-10-30
Windows: Ctrl + Shift + I assigned to both New Note and Debugging Console


reginabally:
Reported in 3429951-zen.

> It looks like Ctrl+Shift+I now is assigned to both "File -> New Note" as well as "Help -> Advanced -> Debugging Console"
And it opens the debugging console and does not create a new note.

### Expected
I would expect the Ctrl + Shift + I hotkey opens up a new note instead of showing the Debugging Console.

### Observed
Using the Ctrl + Shift + I hotkey opens up the Debugging Console.

### Reproduced

1. On the Simplenote app, press Ctrl + Shift + I
2. The app opens up the Debugging Console

### Where did you see the bug

- OS: Windows
- OS version: 10
- Simplenote app version: 2.0.0

I don't have Windows to check this but on Simplenote Elecron 2.0.0 on MacOS, the hotkey for "File > New Note" is ⇧ + ⌘ + I, whereas "Help -> Advanced -> Debugging Console" is ⌥ + ⌘ + I.

I'm asking for some screenshots and will add to this report when we have it.

adyomin:
Updated: Same issue on Ubuntu 20.10 + Simplenote 2.1.0

JavonDavis:
Issue still exists on 2.1-beta-2

codebykat:
Will be fixed in 2.2 #2490

codebykat:
Followed up to remove the label from the menu item in https://github.com/Automattic/simplenote-electron/pull/2493

PR #2493, opened 2020-11-25
Remove keyboard shortcut from debugging console

codebykat:
### Fix

This conflicts with the New Note shortcut. Fixed in #2490 but I realized it still has the label even though it's now handled by New Note. See also #2436 

### Test

1. Verify that Help -> Advanced shows the option for Debugging Console without the keyboard shortcut
2. Verify it works from clicking on the menu*
3. Verify you can still create a new note with Cmd+Shift+I (or Ctrl+Shift+I)

* In my testing it seems like it doesn't work the first time right after building/starting the app, but that is happening on develop as well, so I don't think it's introduced by this PR

### DIFF
--- a/desktop/menus/help-menu.js
+++ b/desktop/menus/help-menu.js
@@ -30,7 +30,7 @@ const buildHelpMenu = (mainWindow, isAuthenticated) => {
       submenu: [
         {
           label: 'Debugging Console',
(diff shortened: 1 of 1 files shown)


nsakaimbo (approved):
👍

PR #2513, opened 2020-12-09
Sidebar icons: studio-blue should be studio-simplenote-blue

codebykat:
### Fix

In #2302 we erroneously used a couple of `studio-blue-50` that should have been `studio-simplenote-blue-50`.

Thanks @SylvesterWilmott for catching this!

Before:

Screen Shot 2020-12-09 at 11 20 41 AM

After:

Screen Shot 2020-12-09 at 11 20 34 AM

### DIFF
--- a/lib/navigation-bar/item/style.scss
+++ b/lib/navigation-bar/item/style.scss
@@ -26,12 +26,12 @@
   }
 
   &.is-selected {
-    background-color: $studio-blue-5;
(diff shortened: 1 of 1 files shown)


SylvesterWilmott:
Thanks @codebykat !

belcherj (approved):
Thank you!

PR #2515, opened 2020-12-11
Fix: Use tab to indent list items from anywhere within the line

### LINKED ISSUE #2468: Unable to create nested list with the Tab key
Reported in 3477730-zen.

> When I use "-" in Markdown, "-" do not move when I press the tab key after pressing enter key. 

### Expected

When I create a list with the `-` or `*` markdown, I would expect to be able to create a nested list item with the Tab key.

### Observed

When I pressed the Tab key, 2 spaces were added without indenting the list item. This happens on notes with and without markdown enabled.

### Reproduced

1. Create a new note on the app
2. Type `-` and some text
3. Press the Enter key and see another `-` character created on the second line, which is expected.
4. Press the Tab key and see that two spaces were created, instead of indenting the list item to create a nested list.

**Screen recording on the Electron app**

Screen Capture on 2020-11-12 at 11-40-31

**Screen recording on the macOS app version 2.3, which is the expected result**

Screen Capture on 2020-11-12 at 11-42-04

### Where did you see the bug

**Information shared in 3477730-zen**

- System Make: Lenovo 
- System Model: ThinkPad E490 
- OS: Windows
- OS version: 10 Pro

**Replicated with the following device**

- System Make: Apple 
- System Model: iMac 27-inch
- OS: macOS
- OS version: 10.15.7 (19H2)

codebykat:
### Fix

This fixes #2468 by adding an action that triggers `editor.action.indentLines` on Tab if the current line is a list or task.

Interestingly, `editor.action.outdentLines` is already being triggered by Shift+Tab. I guess this is because sometimes Tab is expected to insert spacing wherever the cursor is currently located.

### Test

1. Make some lists or checklists
2. Use Tab to indent and shift+tab to outdent
3. Verify that you can still insert normal tab spacing with the Tab key outside of a list item

### Release

Use the Tab key to indent nested list items from anywhere within the line.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -745,6 +745,23 @@ class NoteContentEditor extends Component<Props> {
       },
     });
 
+    // Tab to indent lists and tasks
+    editor.addCommand(monaco.KeyCode.Tab, () => {
+      const lineNumber = editor.getPosition()?.lineNumber;
+      if (!lineNumber) {
+        return;
+      }
+      const thisLine = editor.getModel()?.getLineContent(lineNumber);
+      if (!thisLine) {
+        return;
+      }
+
+      const isList = /^(\s*)([-+*\u2022\ue000\ue001])(\s+)/.test(thisLine);
+      if (isList) {
+        editor.trigger('commands', 'editor.action.indentLines', null);
+      }
+    });
+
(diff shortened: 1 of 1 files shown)


belcherj (approved):
Fixed!

PR #2516, opened 2020-12-11
Fix: guard against passing an empty array to replaceToArray

### LINKED ISSUE #2514: Note list: Crash in decorators with multi-word search query and expanded mode
```
Uncaught TypeError: First argument must be an array or non-empty string
    stringReplaceToArray string-replace-to-array.js:96
    decorateWith decorators.tsx:17
    decorateWith decorators.tsx:7
    render note-cell.tsx:133
    render note-cell.tsx:129
    React 46
    bootWithToken boot-with-auth.tsx:50
    promise callback*bootWithToken boot-with-auth.tsx:28
    run boot.ts:187
    promise callback*run boot.ts:185
    <anonymous> boot.ts:218
    ts app.d03435329e22f6d35a19.js:11821
    __webpack_require__ app.d03435329e22f6d35a19.js:833
    fn app.d03435329e22f6d35a19.js:130
    1 app.d03435329e22f6d35a19.js:12025
    __webpack_require__ app.d03435329e22f6d35a19.js:833
    <anonymous> app.d03435329e22f6d35a19.js:971
    <anonymous> app.d03435329e22f6d35a19.js:974
```

Crash is happening here: https://github.com/Automattic/simplenote-electron/blob/develop/lib/note-list/decorators.tsx#L17

And since it only happens in Expanded mode, seems to be because of the code here:

https://github.com/Automattic/simplenote-electron/blob/develop/lib/note-list/note-cell.tsx#L133

But that's as far as I was able to track it. Does not seem related to any specific search query or note content.

### Reproduced

1. Set Display to "expanded"
2. Type a search query, a space, and any character
3. 💥

codebykat:
### Fix

Fixes the crash in #2514 

### Test
1. Set display to expanded mode
2. Enter a search term with a space and start to type a second word
3. It should not crash

### Release

Fix a crash when entering a multi-word search term in Expanded display mode.

### DIFF
--- a/lib/note-list/decorators.tsx
+++ b/lib/note-list/decorators.tsx
@@ -14,6 +14,9 @@ export const decorateWith = (decorators, text) =>
               ? new RegExp(escapeRegExp(searchText), 'gi')
               : filter;
 
+          if (output?.length === 0) {
+            return [];
+          }
           return replaceToArray(output, pattern, replacer);
         }, text)
         .map((chunk, key) =>
--- a/lib/note-list/decorators.tsx
+++ b/lib/note-list/decorators.tsx
@@ -5,7 +5,7 @@ import replaceToArray from 'string-replace-to-array';
 import { withoutTags } from '../utils/filter-notes';
 
 export const decorateWith = (decorators, text) =>
-  decorators.length > 0
+  decorators.length > 0 && text.length > 0
     ? decorators
         .reduce((output, { filter, replacer }) => {
           const searchText = 'string' === typeof filter && withoutTags(filter);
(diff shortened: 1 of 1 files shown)


lib/note-list/decorators.tsx
```diff
@@ -14,6 +14,9 @@ export const decorateWith = (decorators, text) =>
               ? new RegExp(escapeRegExp(searchText), 'gi')
               : filter;
 
+          if (output?.length === 0) {
+            return [];
+          }
```
dmsnell: it seems like the only way to get an empty string here for `output` is if the input `text` was also an empty string. maybe it makes more sense to abort early and skip all this if we know the input is empty since we can't decorate "nothing"

```ts
export const decorateWith = (decorators, text) =>
	decorators.length > 0 && text.length > 0
		? decorators.reduce(…).map(…)
		: text;
```

belcherj (approved):
Perfect! Fixes the issue.

lib/note-list/decorators.tsx
codebykat: Thanks, updated

PR #2519, opened 2020-12-11
Fix: Weird cursor behavior with lists

### LINKED ISSUE #2336: Web app: Strange cursor behavior with lists
### Expected
When hitting enter after a list item where list items follow, the cursor jumps 2 lines down, only on the web app. This does not happen on the desktop Electron apps or the native Mac app.
On the other apps, hitting enter twice after a list item removes the hyphen and the cursor doesn't move to a new spot.

### Observed
I included video below. If you hit enter twice after the end of one bulleted list item, the cursor jumps down unexpectedly. I can reproduce this consistently.

listt]

### To Reproduce

1. log into the web app (new version)
2. set up a bulleted list with at least 3 items
3. Place cursor at the end of item 1 and hit enter twice
4. The cursor will now be in front of the last item

### Where did you see the bug

I replicated this using: 

- MacOS 10.14.6
- Browser: Chrome Version 84.0.4147.105
- Simplenote 2.0.0-rc3

reported in 3314070-zen:

- Windows 10 home
- Firefox v79.0

codebykat:
### Fix

This fixes the buggy behavior when pressing enter on an empty list item ("lonely bullets").

Fixes #2336

### Test
1. Start a list
2. Hit enter to create a new list item
3. (optional) change the indentation
4. Hit enter without adding text
5. The bullet should be removed and the cursor should end up on the same line
6. Undo should behave sensically

### Release

Fix buggy cursor when hitting enter on an empty list item

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -993,10 +993,15 @@ class NoteContentEditor extends Component<Props> {
           lineNumber: lineNumber + 1,
         });
 
-        // const range = new this.monaco.Range(lineNumber, 0, lineNumber + 1, 0);
-        // const identifier = { major: 1, minor: 1 };
-        // const op = { identifier, range, text: null, forceMoveMarkers: true };
-        // this.editor.executeEdits('autolist', [op]);
+        const range = new this.monaco.Range(
+          lineNumber,
+          0,
+          lineNumber + 1,
+          thisLine.length + 1
+        );
+        const identifier = { major: 1, minor: 1 };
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
Discussed in Slack:

 - thanks for getting this fixed!
 - wow, should have realized that `0` for the line length/column wouldn't cut it
 - clever and nice move combining the editor actions inside the Promise. glad it clears up the duality of making an edit while also returning an edited string for the editor's content

PR #2531, opened 2020-12-17
Fix: only focus editor if search field is not focused

### LINKED ISSUE #2530: Editor steals focus from search field on load
If you load the page and immediately start typing a search, you'll find yourself typing in the note as the editor component calls `focus()` in `componentDidMount`.

Unfortunately without this, nothing at all is focused by default. We should check if any text field has focus and only steal focus if nothing is focused.

### LINKED ISSUE #2613: Focus moved to editor from search field on second search input
### Expected

I'd expect the cursor to stay in the search field when doing a search on the web app.

### Observed

When returning to the search field after doing an initial search, the cursor jumps to the first match of the first character typed and you end up editing the note in the editor instead of writing in the note field. 

### Reproduced

1. Go to Simplenote in Chrome on the web app.
2. Click on CMD+F or CTLR+F to focus on the search field.
3. Go through the search results. 
4. Go back to the search field using CMD+F or CTLR+F and try searching for something different.
5. Notice after the first letter your type the cursor jumps to the note editor field and this takes over from the search field input. 

Sounds related to this: https://github.com/Automattic/simplenote-electron/issues/2530

Screen recording of the issue is available in the ticket, I was also able to reproduce consistently if help is needed to replicate this. 

### Where did you see the bug

- System Make: MacBook Pro
- System Model: 13-inch 2019
- OS: macOS
- OS version: 10.15.7
- Browser (if applicable): Chrome
- Browser version (if applicable): Version 88.0.4324.96 (Official Build) (x86_64)
- Simplenote app version: 2.5.1-Beta1

User reported on Windows 10 in Chrome also. 
3686980-zen

codebykat:
### Fix

Fixes #2530

This one is a bit hard to reproduce as it's a race condition, but it happens fairly frequently to me on local dev.

Also Fixes #2613

If you search for a term and toggle through the results and then search again the focus is brought to the editor. 

### Test

1. Load the app and quickly click into the search field and start typing a search
2. You should never end up in the editor window accidentally typing in your note
3. If the editor has focus it should keep focus as you switch notes, create and delete notes, etc.
4. Perform a search
5. Use shortcut keys to toggle through search results (CmdOrCtrl + g)
6. Go back to search again (CmdOrCtrl + f)
7. Type a new search term
8. Focus should stay in the search element

### Release Notes

- Fixed a couple of bugs where the editor would get focus instead of staying with the search field.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -172,7 +172,9 @@ class NoteContentEditor extends Component<Props> {
         });
       }
     }, SPEED_DELAY);
-    this.focusEditor();
+    if (this.state.searchQuery === '') {
+      this.focusEditor();
+    }
     this.props.storeFocusEditor(this.focusEditor);
     this.props.storeHasFocus(this.hasFocus);
     window.addEventListener('toggleChecklist', this.handleChecklist, true);
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -172,7 +172,7 @@ class NoteContentEditor extends Component<Props> {
         });
       }
     }, SPEED_DELAY);
-    if (this.state.searchQuery === '') {
+    if (document?.activeElement?.id !== 'search-field') {
       this.focusEditor();
     }
     this.props.storeFocusEditor(this.focusEditor);
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -105,7 +105,7 @@ type DispatchProps = {
     direction: 'LTR' | 'RTL'
   ) => any;
   storeNumberOfMatchesInNote: (matches: number) => any;
-  storeSearchSelection: (index: number) => any;
+  storeSearchSelection: (index: number | null) => any;
 };
 
 type Props = OwnProps & StateProps & DispatchProps;
@@ -151,7 +151,7 @@ class NoteContentEditor extends Component<Props> {
     // to avoid, for example, opening a note and having the fourth match selected (or "4 of 1")
     const searchChanged = props.searchQuery !== state.searchQuery;
     if (noteChanged || searchChanged) {
-      props.storeSearchSelection(0);
+      props.storeSearchSelection(null);
     }
 
     return {
@@ -377,9 +377,7 @@ class NoteContentEditor extends Component<Props> {
     if (
       this.editor &&
       this.state.editor === 'full' &&
-      prevProps.selectedSearchMatchIndex !==
-        this.props.selectedSearchMatchIndex &&
-      prevProps.noteId !== this.props.noteId
+      prevProps.selectedSearchMatchIndex !== this.props.selectedSearchMatchIndex
     ) {
       this.setSearchSelection(this.props.selectedSearchMatchIndex);
     }
@@ -1113,6 +1111,7 @@ class NoteContentEditor extends Component<Props> {
     const newIndex = (total + (index ?? -1) + 1) % total;
     this.props.storeSearchSelection(newIndex);
     this.setSearchSelection(newIndex);
+    this.focusEditor();
   };
 
   setPrevSearchSelection = () => {
@@ -1121,16 +1120,16 @@ class NoteContentEditor extends Component<Props> {
     const newIndex = (total + (index ?? total) - 1) % total;
     this.props.storeSearchSelection(newIndex);
     this.setSearchSelection(newIndex);
+    this.focusEditor();
   };
 
   setSearchSelection = (index) => {
-    if (!this.matchesInNote.length) {
+    if (!this.matchesInNote.length || index === null) {
       return;
     }
     const range = this.matchesInNote[index].range;
     this.editor.setSelection(range);
     this.editor.revealLineInCenter(range.startLineNumber);
-    this.focusEditor();
 
     const newDecorations = [];
     this.matchesInNote.forEach((match) => {
@@ -1245,7 +1244,7 @@ const mapStateToProps: S.MapState<StateProps> = (state) => ({
   note: state.data.notes.get(state.ui.openedNote),
   notes: state.data.notes,
   searchQuery: state.ui.searchQuery,
-  selectedSearchMatchIndex: state.ui.selectedSearchMatchIndex ?? 0,
+  selectedSearchMatchIndex: state.ui.selectedSearchMatchIndex,
   spellCheckEnabled: state.settings.spellCheckEnabled,
   theme: selectors.getTheme(state),
 });
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -172,9 +172,7 @@ class NoteContentEditor extends Component<Props> {
         });
       }
     }, SPEED_DELAY);
-    if (document?.activeElement?.id !== 'search-field') {
-      this.focusEditor();
-    }
+    this.focusEditor();
     this.props.storeFocusEditor(this.focusEditor);
     this.props.storeHasFocus(this.hasFocus);
     window.addEventListener('toggleChecklist', this.handleChecklist, true);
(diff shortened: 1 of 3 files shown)


belcherj:
1. Have search query
2. Select a new note
3. Shouldn't it focus the note editor?

codebykat:
@belcherj I updated this to directly check `document.activeElement`, which is not very React-y, but it's pretty hard to use a `ref` for a sibling element.

I'm not sold about triggering `focus()` on a note change honestly, though this now handles that case (which is to say, it focuses the editor text on a note switch). It seems like if your cursor is in the search field, and you're filtering notes, you might expect it to stay there (so you could refine your filtering after checking to see which notes are in the results). At any rate this now behaves in line with the current behavior, minus the race condition bug.

lib/note-content-editor.tsx
```diff
@@ -170,7 +170,9 @@ class NoteContentEditor extends Component<Props> {
         });
       }
     }, SPEED_DELAY);
-    this.focusEditor();
+    if (document?.activeElement?.id !== 'search-field') {
+      this.focusEditor();
+    }
```
dmsnell: there are many ways this can go wrong, far more than the one way it can go right.

we should maybe formalize the rules for changing focus and then see if a more elegant solution drops out. maybe there's some global (Redux) state value tracking what context we're in that changes when we `enter` or `blur` the search field…

maybe we only care about the initial load? that is, we should grab `this.focusEditor()` _unless_ the search bar has already been focused, something we could grab from Redux or from a window-global value - a latch that only turns on the first time the search field is focused.

codebykat: Okay, I will revisit. I agree that we should change the focus as little as possible. It's pretty annoying when the editor fails to get focus, but we really only want to force-focus the editor if there's not another input element focused (in practice this means the search field, although I suppose it might be interesting to keep focus on the tag bar through a note selection change).

That being said:

> there are many ways this can go wrong,

I'm curious if you found any bugs with this implementation or know of a case where checking the `activeElement` would fail.

codebykat: I'm not sure that we only care about initial load, as @belcherj pointed out, right now it also refocuses the editor after you change notes.

dmsnell: it's not about `activeElement` as much as it's `!== 'search-field'` since it's the kind of thing that can end up feeling like spooky action at a distance when we start working with focus on something else, like sort direction or the tag list.

I didn't try this out and I didn't find or see an immediate error with it; just noting that this could be hard to find when the next PR dealing with focus comes around.

to my understanding, re-focusing the editor after changing notes has been intentional since you might want to be opening up a note and then start typing. they keyboard shortcuts still allow you to navigate the note list even when the focus is being taken.

is it only on boot or is it on every note change we care about preserving focus? is there some kind of even that "resets" focus and then the first time to get it keeps it? it sounds as we talk like we might be wanting a global focus latch with a first-come-first-keep rule about it. I'm definitely not sure of this of course.

dcalhoun (commented):
I wanted provide the context that once #2707 merges, the current aggressive focus capture of the editor also creates an issue for the History Revision Selector.

Specifically, opening and closing the Revision Selector (or any dialog) should move focus back to the previously focused element, which may or may not be the editor. If one is using a keyboard or screen reader to open the Revision Selector, the previously focused element will be the button to trigger opening the Revision Selector.

I chose not to address this issue in #2707, as I perceive it to be a bit complex. I do plan on creating a new Issue to capture the details.

I'm not sure if the issue I outlined should be addressed within this PR, but figured another example of focus competition could prove beneficial for planning the approach used in this PR.

lib/note-content-editor.tsx
```diff
@@ -1188,7 +1190,7 @@ const mapStateToProps: S.MapState<StateProps> = (state) => ({
   note: state.data.notes.get(state.ui.openedNote),
   notes: state.data.notes,
   searchQuery: state.ui.searchQuery,
-  selectedSearchMatchIndex: state.ui.selectedSearchMatchIndex,
+  selectedSearchMatchIndex: state.ui.selectedSearchMatchIndex ?? 0,
```
sandymcfadden: Setting the default to 0 as when we were comparing indexes in componentDidUpdate the initial render was comparing null to 0 which was unexpected.

lib/note-content-editor.tsx
```diff
@@ -350,7 +350,9 @@ class NoteContentEditor extends Component<Props> {
     if (
       this.editor &&
       this.state.editor === 'full' &&
-      prevProps.selectedSearchMatchIndex !== this.props.selectedSearchMatchIndex
+      prevProps.selectedSearchMatchIndex !==
+        this.props.selectedSearchMatchIndex &&
+      prevProps.noteId !== this.props.noteId
```
sandymcfadden: If you performed a search then toggled through the search results and then performed a new search the focus would come back to the editor instead of finishing typing the search term.
With help from @codebykat I believe this was only intended for when a new note is selected so I added the additional check.

codebykat: This logic is for deciding when to reset the search selection, isn't it? What does it have to do with the editor focus?

I think the search selection should be updated whenever the search match index changes, not only when changing to a different note.

The actual problem here, which is perhaps being masked, seems to be that `setSearchSelection` makes another call to `this.focusEditor();`. That seems like something it should not do?

lib/note-content-editor.tsx
```diff
@@ -172,9 +172,7 @@ class NoteContentEditor extends Component<Props> {
         });
       }
     }, SPEED_DELAY);
-    if (document?.activeElement?.id !== 'search-field') {
-      this.focusEditor();
-    }
+    this.focusEditor();
```
codebykat: Doesn't this go back to focusing the editor on load without any checks?

codebykat (commented):
So I admit I'm deeply confused by the state of this PR, maybe because I don't remember the context surrounding it. But I just can't follow why this should solve the problem, if it does in fact solve the problem.

Why do we need to focus the editor within `setSearchSelection`?

Why would we want to avoid updating the search match highlights when the selected search index changes, regardless of whether we have changed notes?

@sandymcfadden I'm sorry if I led you astray here 😅

lib/note-content-editor.tsx
sandymcfadden: You're right that is what this does. At first I didn't see any other time than note switches where this would be needed, but that was naive and I did of course find other times. 

You're right again that the problem stems from the call to `this.focusEditor();` in `setSearchSelection`. I had first thought to remove that all together but I'm sure especially if using the keyboard shortcuts to toggle through search matches you would want the focus to end up there so you could immediately edit the text.

I've changed my approach to this now. Instead of calling `this.focusEditor();` in `setSearchSelection` I've moved it to set focus in `setNextSearchSelection` and `setPrevSearchSelection`.

The only weird thing I've seen so far with this is that now the behavior is different if you are using the keyboard shortcuts to toggle through search results or if you are using the search results bar at the bottom to toggle through. If using the search results bar you don't get the focus anymore. The visual selection is still there though.

I think this may be OK as if you are using the keyboard shortcuts you wouldn't want to then have to click to edit, but if you are clicking the next and previous buttons it would likely be OK to click to where you want to edit.

Would love others thoughts on that part as well.

lib/note-content-editor.tsx
sandymcfadden: It does, but I don't think this check is necessary. In most cases, I think we want the editor to have focus on load. We just don't want it to unexpectedly take focus if we've clicked the search field. 
I believe the focus was getting taken not from the initial on load, but the logic check to see if the selected search index had changed was always executing on first load because it was comparing `null` to `0` for the `prevProps.selectedSearchMatchIndex` and `props.selectedSearchMatchIndex`.

I could be wrong about this as I was having a hard time replicating the focus loss as it was happening to you.

I've resolved that by setting the default to be `null`. That fixed another bug as well where the first time you pressed `CtrlOrCmd + g` to go through the search results it would select the second instance instead of the first.

sandymcfadden:
> So I admit I'm deeply confused by the state of this PR, maybe because I don't remember the context surrounding it. But I just can't follow why this should solve the problem, if it does in fact solve the problem.

There were certainly some things I was missing here and I apologize for the confusion. I've updated things now and I think addressed your questions in the other comments.

lib/note-content-editor.tsx
codebykat: This makes sense and I agree with you - I think leaving the focus on the buttons is best when using the buttons.

codebykat:
This is testing well for me, I can't reproduce the focus issue with searches, changing to different notes, reloading and quickly searching, etc. I say we :shipit: and see if anything else arises.

✅ approved (geez I wish GitHub allowed us to change author of a PR)

PR #2536, opened 2020-12-21
Fix: Load configured `app_id` when connecting to Simperium

### LINKED ISSUE #2396: Development: Can't log in to local development server
Hi, I have difficulties to run the app so I can contribute.

### Expected
To be connected after entering username and password

### Observed
After entering my credentials, I can briefly see the home screen, but after less than a second the app is refreshed and I am disconnected from it.

### Reproduced
1. Enter credentials
2. Click "Log in"

I tried it with an instance started with `npm run dev` and a build created with `make package-osx`

gif

### Where did you see the bug

- System Make: GNU Make 3.81
- System Model:
- OS: MacOS
- OS version: 10.15.7 
- Browser (if applicable): 
- Simplenote app version: develop branch
- Node version : v14.13.1
- NPM version : 6.14.8

dmsnell:
Resolves #2396 

In the big rewrite I accidentally instructed the Simperium middleware to always connect
to the production Simperium application and overlook what might be configured in the
`config.json` or `config-local.json` files. This led to people being unable to sign in
to test accounts on the test backend: the successful authorizatio would lead to the app
immediately signing out due to an `Token Invalid` over the WebSocket.

With this fix we're using the configured app name, reading it from the configs, so that
when the repo is cloned and ran it properly connects to `history-analyst-dad` (or any
value set in its place). This fix restores one's ability to sign in.

Props to @tbourrely for valuable diagnosis of this issue in #2396

### DIFF
--- a/lib/state/simperium/middleware.ts
+++ b/lib/state/simperium/middleware.ts
@@ -2,6 +2,7 @@ import { default as createClient } from 'simperium';
 
 import debugFactory from 'debug';
 import actions from '../actions';
+import getConfig from '../../../get-config';
 import { BucketQueue } from './functions/bucket-queue';
 import { NoteBucket } from './functions/note-bucket';
 // import { NoteDoctor } from './functions/note-doctor';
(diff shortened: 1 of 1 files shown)


codebykat (approved):
😱

PR #2542, opened 2020-12-24
Add sorting bar to note list

### LINKED ISSUE #1625: Search: Add "sort by" to search results

codebykat:
### Fix

Adds a "Sort by" bar to the bottom of the notes list. See p2XJRt-2Ct-p2

Designs are at:
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fd77f1846903987b1b40e75 (light mode)
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fd77f196e43637c2cb2889f (dark mode)

Also shortens the search results bar to 44px to match the height of the sorting bar and the new toolbar-height.

Fixes #1625

### Test
1. Use the toggle to change the sort type and direction of the results.
2. Also try it with search filtering
3. Try in different browsers/OSes to make sure the dropdown styling is okay

### Screenshots

Screen Shot 2021-02-02 at 11 26 14 AM

Screen Shot 2021-02-02 at 11 26 20 AM

Screen Shot 2021-02-02 at 11 26 34 AM

Screen Shot 2021-02-02 at 11 26 59 AM

Note that in all browsers except Firefox, the dropdown styling is determined by the system theme, so it looks like this if you have a light system theme and the dark Simplenote theme:

Screen Shot 2021-02-02 at 11 26 44 AM

In Firefox, we have to style the dropdown since otherwise it takes the styles from the parent, so I have added dark mode styles for Firefox that will be shown regardless of the system setting (in Simplenote dark theme):

Screen Shot 2021-02-02 at 11 42 03 AM

The weird thing is that setting a background on the `select` seems to wipe out the padding and rounded edges (only in dark mode), but if we don't set it, we get this:

Screen Shot 2021-02-02 at 12 05 48 PM

Firefox light theme changes the white background to a light gray, which matches default styling:

Screen Shot 2021-02-02 at 11 33 15 AM

Search results bar shortened to match:

Screen Shot 2021-02-02 at 12 02 51 PM

Screen Shot 2021-02-02 at 12 05 11 PM

### Release

Added sort order bar to note list


codebykat:
@SylvesterWilmott Here's where I eventually landed after some experimentation:

- Including the ordering in the text label of the dropdown isn't great, because it is also displayed on dropdown options, and this is gross:
 
Screen Shot 2021-01-07 at 12 26 42 PM

- Using a colon was fraught because of the alignment (the dropdown is the width of the longest text item). `text-align: right` doesn't have support in all browsers. I didn't love it anyway since there's already a colon after `Sort:`.
- Instead I tried putting the text label slightly smaller. This somewhat imitates what we have on iOS, where the order label is on its own line.
- I'm using the ordering labels from iOS (A-Z/Z-A and Oldest/Newest) but kept the sort type labels we're already using on web (Date Modified / Date Created / Alphabetic). This has a slight advantage over the iOS versions (Created / Modified / Alphabetical) in that it keeps the text strings closer to the same length, so we have less awkward space between the dropdown and the label.

lib/sort-order-selector/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+        label: 'Alphabetical',
+        id: 'alphabetical',
+      },
+    ];
+
+    return (
+      <Fragment>
+        {shouldDisplay && (
+          <div className="sort-order-selector">
+            Sort:
```
dmsnell: not that we're generally that accessible, but would it be worth surrounding this in a `<label for="sort-selection">` tag so that there's a link between it and the dropdown?

lib/sort-order-selector/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+              ))}
+            </select>
+            <span className="sort-label">{sortLabel}</span>
+            <span className="sort-button">
+              <IconButton
+                icon={<SortOrderIcon />}
+                onClick={this.props.toggleSortOrder}
+                title="Change Sort Order"
+              />
+            </span>
```
dmsnell: is this `span` necessary for anything?

lib/sort-order-selector/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+};
+
+type DispatchProps = {
+  toggleSortOrder: () => any;
+  setSortType: (sortType: T.SortType) => any;
+};
+
+type Props = StateProps & DispatchProps;
+
+export class SortOrderSelector extends Component<Props> {
```
dmsnell: you may have already made this a function and moved between function and class but it looks like we could go ahead and make it a function component as we don't rely on any state. no need of course but this isn't depending on instance properties.

dmsnell:
Looks fine as it is but I left a few comments that you might consider. Feel free to merge it after responding to or rejecting my suggestions.

codebykat:
Next steps on this:

1. ~Change dropdown labels per new designs~ (p2XJRt-2Ct-p2) https://github.com/Automattic/simplenote-electron/pull/2542/commits/3a9fa0e8624a7178d0f671a01c231dceda268f38

> all 6 sorting options should be listed within each platforms ‘dropdown’ menu to reduce visual prominence of the bar since it will now be displayed all the time.

Dropdown should be: 
Name: A-Z
    Name: Z-A
    Created: Newest
    Created: Oldest
    Modified: Newest
    Modified: Oldest

2. ~Add tracks event~ https://github.com/Automattic/simplenote-electron/pull/2542/commits/cb083c97078549cbb1c1716b6b1d7b155af2f343, https://github.com/Automattic/simplenote-electron/pull/2542/commits/3c286680bf7171f7fd2344a094dc5bad7aa0470a

> sp*_list_sortbar_mode_changed

3. ~Address review feedback~ https://github.com/Automattic/simplenote-electron/pull/2542/commits/8c269463398f75079119d8233df860a5bfa950e3

lib/sort-order-selector/index.tsx
codebykat: Yeah, for styling, though I might be about to refactor it out of existence

codebykat:
Refactored with some data attributes, thanks @dmsnell for the suggestion.

SylvesterWilmott (approved):
The design looks good! Thanks @codebykat

lib/sort-order-selector/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+}) => {
+  const changeSort = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const selectedOption = event.currentTarget.selectedOptions[0];
+    const selectedSortType = selectedOption.dataset.order;
+    const selectedSortReversed = selectedOption.dataset.reversed;
+
+    if (selectedSortType !== sortType) {
+      setSortType(selectedSortType as T.SortType);
+    }
+    if ((selectedSortReversed === 'false') === sortReversed) {
```
dmsnell: this is confusing me, specifically because we're comparing against the string value of `false`

I think we're doing it funny-like because it's a string value and not a real value. in this case _parsing_ it could go a long way for readability. what I expect is to find something like `selectedSortReversed !== sortReversed`

```ts
const selectedSortReversed = selectedOption.dataset.reversed === 'true'; // true if 'true' else false
const selectedSortReversed = JSON.parse(selectedOption.dataset.reversed); // 'true' and 'false' decode properly

if (selectedSortReversed !== sortReversed) {
	…
}
```

lib/sort-order-selector/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+export const SortOrderSelector: FunctionComponent<Props> = ({
+  shouldDisplay,
+  setSortType,
+  sortReversed,
+  sortType,
+  toggleSortOrder,
+}) => {
+  const changeSort = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const selectedOption = event.currentTarget.selectedOptions[0];
```
dmsnell: did you find `HTMLSelectElement.selectedIndex`?

lib/sort-order-selector/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+      order: 'modificationDate',
+      reversed: false,
+    },
+    {
+      label: 'Modified: Oldest',
+      id: 'modificationDate-reversed',
+      order: 'modificationDate',
+      reversed: true,
+    },
+  ];
```
dmsnell: if we want to simplify the flow of data and types through the select component then we can do another thing and that is directly index into this list. consider if this `sortTypes` array were outside of the component and in the module scope and we had a function to find the current setting. our `HTMLSelectElement` could refer to an index into the array as the choice, and although that's less meaningful in itself, it does preserve the types the eliminates the processing and parsing needed to re-connect the data.

```ts
type SortOption = {
	label: string,
	order: T.SortOrder,
	isReversed: boolean
}

const sortTypes: SortOption[] = [{
	label: 'Name: A-Z',
  order: 'alphabetical',
  isReversed: false
}, {
	…
}];

const onChange = ({currentTarget: {selectedIndex}}) => {
	if (-1 === selectedIndex) {
		return;
	}

	const sort = sortTypes[selectedIndex];
	if (sort.sortOrder !== sortOrder) {
		…
	}

	if (sort.isReversed !== sortReversed) {
		…
	}
}

const sortTypesIndex = (sortOrder, isReversed) => sortTypes.findIndex(type => type.order === sortOrder && type.isReversed === isReversed);

<select id="sort-selection" value={sortTypesIndex(sortOrder, reversed)} onChange={…}>
{sortTypes.map((type, index) => (
	<option key={type.label} value={index}>{type.label}</option>
))}
</select>
```

dmsnell: it's _rare_ to use an index as the React key but here since the `sortTypes` array is static the indices will never change and it's safe to use that value

lib/sort-order-selector/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const mapStateToProps: S.MapState<StateProps> = (state) => ({
+  shouldDisplay: state.ui.filteredNotes.length > 0,
+  sortReversed: state.settings.sortReversed,
+  sortType: state.settings.sortType,
+});
+
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
+  setSortType,
+  toggleSortOrder,
```
dmsnell: calling two separate Redux actions introduces another partial update into our app. that is, if we change both sort properties then we're asking React to update after the first Redux dispatch and then again after the second. that intervening time represents a state that was never requested by the customer.

should we examine the `setSortType` action and give it the ability to also change the reversing of the search? if we did that, add an optional `reversed: boolean` property to the action, then we could perform the update in one go and we could eliminate the conditionals at the same time.

lib/sort-order-selector/index.tsx
codebykat: Yes but is dropdown.options[selectedIndex] better in some way than dropdown.selectedOptions[0]?

lib/sort-order-selector/index.tsx
codebykat: Oops, I should have refactored that out actually.

lib/sort-order-selector/index.tsx
codebykat: That works like a charm! https://github.com/Automattic/simplenote-electron/pull/2542/commits/a7d6f62510933d1486c3d79530eb3dfbd2c171f8

lib/sort-order-selector/index.tsx
codebykat: Very cool! I do end up checking for `undefined` a bit, not sure if that's the right way to do this. https://github.com/Automattic/simplenote-electron/pull/2542/commits/3e8bac4c1d38ab1bf7ca5f7ee9d8bd09fde9222b

lib/sort-order-selector/index.tsx
dmsnell: only in the sense that it relies on named things vs. the assumption of 0. I know that I wouldn't expect the selected option to always be the first one in the list, or that's how I read this line. I'm guessing this is a list of all possible selections because a drop down can have multiple selected items?

otherwise we're getting the actual option I think, if using seletedIndex

dmsnell (approved):
the logic is more readable to me after using the array of sort options vs. the data-passing. good detail work making sure the search index updates ✅

pachlava:

PR #2545, opened 2021-01-05
Search bar: Make it into its own component in order to fix layout bugs

### LINKED ISSUE #2544: 'N Results' bar blocks last line of long notes
When I search for something, the last line of a long note (by "long" I mean has a scrollbar) is blocked or partially blocked by the bar telling you how many results are in that particular note (shown in screenshots below).

### Expected

I expect that the note's contents wouldn't be blocked by other UI elements.

### Observed

The last line of the note is blocked by the 'N Results' bar.

### Reproduced

1. Create a longish note (long enough to have a scrollbar)
2. Search for a word from this note in the Search Bar
3. You should now see the last line of the aforementioned note blocked by the 'N Results' bar.

bug

The only fix is to zoom out til there's no scrollbar, which can be a ridiculous amount for long notes:
hack

### Where did you see the bug

- System Make: Dell
- System Model: Dell XPS 13 9360
- OS: Kubuntu (Linux)
- OS version: Kubuntu 20.04.1 LTS
- Simplenote app version: 2.3.0 (desktop app downloaded as deb package from latest release)

codebykat:
### Fix

Fixes #2544 and the issue where the tag bar is overlapped.

### Test
1. Do some searches with matches in long notes
2. Scroll to the bottom of the note and verify that the tags and last lines of the note are still visible
3. Switch notes, change the search, verify that the selected result properly resets to the first.

### Screenshots

Screen Shot 2021-01-04 at 11 29 23 PM

### Release

Fixed layout bugs causing the search results bar to overlap note contents and tag input field

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -22,7 +22,6 @@ import {
   withCheckboxCharacters,
   withCheckboxSyntax,
 } from './utils/task-transform';
-import SearchResultsBar from './search-results-bar';
 
 import * as S from './state';
 import * as T from './types';
@@ -1168,7 +1167,6 @@ class NoteContentEditor extends Component<Props> {
             value={content}
           />
         )}
-        {searchQuery.length > 0 && searchMatches && <SearchResultsBar />}
       </div>
     );
   }
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -145,15 +145,11 @@ class NoteContentEditor extends Component<Props> {
 
     const editor = noteChanged ? (goFast ? 'fast' : 'full') : state.editor;
 
-    // @todo what was this doing and now what should it do instead?
-    // const searchChanged = props.searchQuery !== state.searchQuery;
-    // const selectedSearchMatchIndex =
-    //   noteChanged || searchChanged ? null : props.selectedSearchMatchIndex;
-
-    // maybe something like this to invalidate the pre-existing selection?
-    // if(noteChanged || searchChanged) {
-    // props.storeSearchSelection(null);
-    // }
+    // reset search selection if the search or note has changed
+    const searchChanged = props.searchQuery !== state.searchQuery;
+    if (noteChanged || searchChanged) {
+      props.storeSearchSelection(0);
+    }
 
     return {
       content,
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -145,7 +145,8 @@ class NoteContentEditor extends Component<Props> {
 
     const editor = noteChanged ? (goFast ? 'fast' : 'full') : state.editor;
 
-    // reset search selection if the search or note has changed
+    // reset search selection when either the note or the search changes
+    // to avoid, for example, opening a note and having the fourth match selected (or "4 of 1")
     const searchChanged = props.searchQuery !== state.searchQuery;
     if (noteChanged || searchChanged) {
       props.storeSearchSelection(0);
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -22,8 +22,7 @@ import {
   withCheckboxCharacters,
   withCheckboxSyntax,
 } from './utils/task-transform';
-import IconButton from './icon-button';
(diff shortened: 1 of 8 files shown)


lib/search-results-bar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+type Props = StateProps & DispatchProps;
+
+const SearchResultsBar: FunctionComponent<Props> = ({
+  selectedSearchMatchIndex: index,
+  numberOfMatchesInNote: total,
+  setSearchSelection,
+}) => {
+  const nextButtonRef = useRef<HTMLButtonElement>();
+  const prevButtonRef = useRef<HTMLButtonElement>();
```
dmsnell: what's the difference here between `useRef` and `createRef`? I figured that for adding these to HTML elements we'd use `createRef` (since we're not passing the ref down through anything or using it for a variable.

I'm generally confused on the API for refs…

lib/search-results-bar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+      setSearchSelection(newIndex);
+    };
+    prevButtonRef.current?.addEventListener('click', setPrev, true);
+    nextButtonRef.current?.addEventListener('click', setNext, true);
+
+    return () => {
+      prevButtonRef.current?.removeEventListener('click', setPrev, true);
+      nextButtonRef.current?.removeEventListener('click', setNext, true);
+    };
+  }, [index, total]);
```
dmsnell: Were there any issues in setting these directly on the `<span>`s with `onClick`?

```ts
<span className="search-results-next" onClick={() => setSearchSelection((total + (index ?? -1) - 1) % total)}>
```

lib/search-results-bar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+}) => ({
+  selectedSearchMatchIndex,
+  numberOfMatchesInNote,
+});
+
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = (dispatch) => ({
+  setSearchSelection: (index: number) => {
+    dispatch({ type: 'STORE_SEARCH_SELECTION', index: index });
+  },
+});
```
dmsnell, line 79: for funnies we can reduce this twice if we want:

since it's just dispatching we don't need the curlies
```ts
setSearchSelection: (index: number) => dispatch({…})
```

also since we're not doing anything but creating a dispatchable object I don't think we even need `dispatch` - I might have forgotten though how it works

```ts
const mapDispatchToProps = {
	setSearchSelection: (index: number) => ({type: 'STORE_SEARCH_SELECTION', index })
}
```

lib/search-results-bar/index.tsx
codebykat: Apparently:

> Notice that with a functional component we are using useRef instead of createRef. If you create a ref using createRef in a functional component, React will create a new instance of the ref on every re-render instead of keeping it between renders.

https://medium.com/trabe/react-useref-hook-b6c9d39e2022

(I changed it over because there was a very informative error message that said something like, "createRef won't work here, did you mean useRef?")

lib/search-results-bar/index.tsx
codebykat: I swear I tried this and it didn't work but apparently I just didn't wire it up right 🤦

lib/search-results-bar/index.tsx
dmsnell: ah well actually that makes sense now that I read it

lib/note-content-editor.tsx
```diff
@@ -146,16 +145,17 @@ class NoteContentEditor extends Component<Props> {
 
     const editor = noteChanged ? (goFast ? 'fast' : 'full') : state.editor;
 
+    // reset search selection if the search or note has changed
```
dmsnell: adding a little more here might help us avoid wondering again in the future

```ts
// reset search selection when either the note or the search changes
// to avoid, for example, opening a note and having the fourth match selected
```

dmsnell (approved):
left a note about expanding the comment where we're resetting the selection as I think a little more could help us in the future but this looks fine to me for merging.

pachlava:

Issue #2379, opened 2020-10-02
Bulleted list not auto-adding bullets


rezzap:
Bulleted lists no longer auto- add bullets in the web app when pressing enter after a list item

### Expected

The bullet list should consistently add a bullet point automatically when pressing enter.

### Observed

This isn't happening consistently for some users. 

### Reproduced

@msilbers was able to reproduce this, but I haven't been able to so far:
![list](https://user-images.githubusercontent.com/10325627/94906610-aed91580-0496-11eb-85f9-8bcd53baeec9.gif)

Steps reported by user: 
Steps to reproduce: not-working platform (Simplenote web)
1. Start an unordered list by starting a line with a - or * character, followed by a space.
2. Type a word, such as “one”.
3. Press the Return key to create a new line, which will be prefixed with either - or *, depending on format
4. Press the Tab key
5. Observe that the new line is not indented as it was in the working platform.

They report this working for them in the Electron app 1.21.1 or the MacOS app 2.1.

### Where did you see the bug
- System Make: Mac OS
- System Model:
- OS: macOS 
- OS version: 10.15.5
- Browser (if applicable): Chrome (Melissa) Firefox 81.0 (User)
- Browser version (if applicable): 85.0.4183.121
- Simplenote app version: Simplenote Web 2.0.0-rc5

Reported in 3357474-zen

elibud:
We have a similar report for checklists, Windows 10, Microsoft Edge, and Chrome. I have a video, ping me if you need it.

azollafox:
In addition to no longer adding a bullet, trying to press enter mid list like this now brings your cursor to the bottom of the list
![](https://d.pr/i/UIeWQQ+)
(Image link: https://d.pr/i/UIeWQQ )

Reported in #3476961-zen

codebykat:
That looks like #2336

codebykat:
I've fixed one and maybe two of the 2-3 issues reported here. (#2519)  I think there's still some inconsistent bug where hitting enter from a list item does not always create a new list, but I haven't been able to reproduce it so far.

Would appreciate if folks on this thread could try to test and see if there are still problems ❤️

codebykat:
Closing as I haven't seen this issue or any reports of it lately.

codebykat:
OMG we were able to reproduce it and fix it (see #2548)

pachlava:
Verified on Windows with Electron 2.5.0-beta1

PR #2548, opened 2021-01-05
Fix: List continuations in notes created on Windows

codebykat:
### Fix

Finally! The fix for the issue reported in #2379!

Turns out this was bailing early in the case of Windows EOL characters, as it was only testing for `\n`. Monaco [determines the line ending from the default](https://microsoft.github.io/monaco-editor/api/enums/monaco.editor.endoflinepreference.html) and/or what is currently in the note, so any note originally created on Windows would be buggy forevermore.

Props to @nsakaimbo for figuring out how to reproduce this!

### Test

1. Create a note on Windows (Electron or in browser, it doesn't matter)
2. Insert a checklist item
3. Press enter and confirm that the next line automatically begins with a checklist as well
4. Ditto for bulleted lists
5. Repeat testing steps in a note created on a non-Windows platform

### Release

Fixed a bug preventing checklists and bulleted lists from automatically continuing when a note was created on Windows

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -948,7 +948,8 @@ class NoteContentEditor extends Component<Props> {
       }
 
       const change = event.changes.find(
-        ({ text }) => text[0] === '\n' && text.trim() === ''
+        ({ text }) =>
+          (text[0] === '\r' || text[0] === '\n') && text.trim() === ''
       );
 
       if (!change) {


belcherj:
Will it ever be `\r\n`?

codebykat:
@belcherj It's just looking at the first character so it catches `\r\n` as well as `\r` or `\n`. Kind of takes advantage of the fact that the `trim()` removes any whitespace anyway.

nsakaimbo (approved):
Tested the Windows build from this branch and confirmed this fixes the bug, which was occuring consistently for me. (Checkbox and bullet lists are automatically continued from the previous item after `Enter` is pressed.) Also confirmed that this patch fixes the behavior for notes that were previously created on Windows.

pachlava:

PR #2551, opened 2021-01-06
prevent adding undefined to className when value do not provided

ubaidisaev:
### Fix
Add default value to className param
image

### Test
no test needed

### DIFF
--- a/lib/controls/checkbox/index.tsx
+++ b/lib/controls/checkbox/index.tsx
@@ -6,7 +6,7 @@ type OwnProps = React.HTMLProps<HTMLInputElement> & {
   onChange: () => any;
 };
 
-function CheckboxControl({ className, ...props }: OwnProps) {
+function CheckboxControl({ className = '', ...props }: OwnProps) {
   return (
     <span className={classNames('checkbox-control', className)}>
       <input type="checkbox" {...props} />
--- a/lib/controls/checkbox/index.tsx
+++ b/lib/controls/checkbox/index.tsx
@@ -6,9 +6,9 @@ type OwnProps = React.HTMLProps<HTMLInputElement> & {
   onChange: () => any;
 };
 
-function CheckboxControl({ className = '', ...props }: OwnProps) {
+function CheckboxControl({ className, ...props }: OwnProps) {
   return (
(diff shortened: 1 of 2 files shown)


dmsnell:
Hi @ubaidisaev - thanks for the commit!

I wonder if instead of passing an empty string we could avoid adding the class name altogether. If there's not a more elegant way this looks pretty good IMO.

How'd you find this?

ubaidisaev:
@dmsnell  we can completely remove `className` property because it's not used right now, or we can use `classnames` package  for  conditionally  joining classNames together.

dmsnell:
@ubaidisaev we're already using `classnames` if you look deeper into the code.

still curious, and curiosity is all, how you stumbled upon this - it's a good catch but such a mundane detail 😄

ubaidisaev:
@dmsnell  I know that `classnames` is used in the project and therefore suggested it.
I agree it's funny error but anyway )
I know there a lot more critical problems but i'm gonna go slow and familiarize myself with this project

dmsnell:
@ubaidisaev I meant that in this specific case we're already using `classnames` where you suggested we add it

ubaidisaev:
@dmsnell  just check this  [demo ](https://codesandbox.io/s/simplenote-demo-hfubd?file=/src/App.tsx)
Maybe this way look more legit. It's all about avoiding adding `undefined` to markup.

dmsnell:
@ubaidisaev I don't think I'm coming across clearly. if you look a couple lines under where you made the default value

https://github.com/Automattic/simplenote-electron/pull/2551/files#diff-23a991f90156f9d604d25223e49e6d1caa38297f75b7046fd226ad14fc131327R11

you will see the app is already using `classnames` _as you are suggesting_ and probably the same change would occur without the default values to the arguments but by changing the call from `className` to `[className]`

ubaidisaev:
@dmsnell  yeap you're right. I didn't notice it.

dmsnell (approved):
Thanks for updating with `classNames` usage!

@codebykat I'm turning this over to you to merge since I'm somewhat removed from the release process at the moment.

codebykat:
Should get into 2.5.0 just under the wire. Thanks @ubaidisaev!

PR #2556, opened 2021-01-08
Fix: Add selected tag to new notes by default

### LINKED ISSUE #2423: Tag isn't added automatically when a tag is selected from the menu
When a tag is selected from the hamburger menu to view the list of notes assigned to the tag, the tag wasn't assigned to the new note created under this view. This feature is available in previous version (v 1.5.x, according to 3432482-zen).

### Expected
I would expect the chosen tag will automatically be added to the note when I create a new note after selecting a tag from the hamburger menu.

### Observed
The new note was created without a tag.

### Reproduced

**Note:** The issue was originally reported in 3432482-zen on Windows 10, Simplenote app version 2.0.0. I was able to replicate this on https://app.simplenote.com.

1. Open up the hamburger menu of the app
2. Choose a tag
3. Create a new note
4. See that no tag was added to the note

**Screen recording**

Screenshot
Direct link to the GIF: https://d.pr/i/oeZqVK/raUWKgYlKC

### Where did you see the bug

**Information shared in 3432482-zen**

- OS: Windows
- OS version: 10 v1909
- Browser (if applicable): 

**Information on this test**

- OS: macOS
- OS version: 10.15.7 (19H2)
- Browser (if applicable): Google Chrome
- Browser version (if applicable): Version 86.0.4240.111 (Official Build) (x86_64)

codebykat:
### Fix

Fixes #2423

If you are on a selected tag and choose to add a new note, the new note will now automatically get that tag added to it.

### Test

1. Click a tag in the sidebar to select it
2. Create a new note
3. Verify that the new note has the selected tag
4. Create a new note with no tag selected
5. Verify that no tags are added to that note

### Release

Fix: Apply selected tag to a new note by default

### DIFF
--- a/lib/state/data/middleware.ts
+++ b/lib/state/data/middleware.ts
@@ -2,6 +2,7 @@ import { v4 as uuid } from 'uuid';
 
 import { tagHashOf } from '../../utils/tag-hash';
 import exportZipArchive from '../../utils/export';
+import { tagHashOf as t, withTag } from '../../utils/tag-hash';
 
 import type * as A from '../action-types';
 import type * as S from '../';
@@ -30,10 +31,18 @@ export const middleware: S.Middleware = (store) => (
         systemTags.splice(markdownInNote, 1);
       }
 
+      // apply selected tag by default
+      const selectedTag = state.ui.openedTag;
+      let tags = action.note?.tags ?? [];
+      if (selectedTag) {
+        tags = withTag(tags, selectedTag);
+        // tags.push(t(selectedTag));
+      }
+
       return next({
         type: 'CREATE_NOTE_WITH_ID',
         noteId,
-        note: { ...action.note, systemTags },
+        note: { ...action.note, systemTags, tags },
         meta: {
           nextNoteToOpen: noteId,
         },
--- a/lib/state/data/middleware.ts
+++ b/lib/state/data/middleware.ts
@@ -2,12 +2,12 @@ import { v4 as uuid } from 'uuid';
 
 import { tagHashOf } from '../../utils/tag-hash';
 import exportZipArchive from '../../utils/export';
-import { tagNameOf, withTag } from '../../utils/tag-hash';
+import { withTag } from '../../utils/tag-hash';
 
 import type * as A from '../action-types';
 import type * as S from '../';
 import type * as T from '../../types';
-import { numberOfNonEmailTags } from '../selectors';
+import { numberOfNonEmailTags, openedTag } from '../selectors';
 
 export const middleware: S.Middleware = (store) => (
   next: (action: A.ActionType) => A.ActionType
@@ -32,11 +32,9 @@ export const middleware: S.Middleware = (store) => (
       }
 
       // apply selected tag by default
-      const openedTag = state.ui.openedTag
-        ? tagNameOf(state.ui.openedTag)
-        : null;
+      const selectedTag = openedTag(state);
       const givenTags = action.note?.tags ?? [];
-      const tags = openedTag ? withTag(givenTags, openedTag) : givenTags;
+      const tags = selectedTag ? withTag(givenTags, selectedTag) : givenTags;
 
       return next({
         type: 'CREATE_NOTE_WITH_ID',
--- a/lib/state/data/middleware.ts
+++ b/lib/state/data/middleware.ts
@@ -2,7 +2,7 @@ import { v4 as uuid } from 'uuid';
 
 import { tagHashOf } from '../../utils/tag-hash';
 import exportZipArchive from '../../utils/export';
-import { tagHashOf as t, withTag } from '../../utils/tag-hash';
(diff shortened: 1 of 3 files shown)


lib/state/data/middleware.ts
```diff
@@ -28,10 +29,18 @@ export const middleware: S.Middleware = (store) => (
         systemTags.splice(markdownInNote, 1);
       }
 
+      // apply selected tag by default
+      const selectedTag = state.ui.openedTag;
+      let tags = action.note?.tags ?? [];
+      if (selectedTag) {
+        tags = withTag(tags, selectedTag);
+        // tags.push(t(selectedTag));
+      }
```
dmsnell: might be clearer to combine this into non-mutating expressions

```ts
const { openedTag } = state.ui;
const givenTags = action.note?.tags ?? [];
const tags = openedTag ? withTag(givenTags, openedTag) : givenTags;
```

dmsnell (approved):
Pretty straight-forward code. You know me: I don't like `let` or mutation so I left a comment about removing it but the logic looks good. 👍

lib/state/data/middleware.ts
```diff
@@ -1,6 +1,7 @@
 import { v4 as uuid } from 'uuid';
 
 import exportZipArchive from '../../utils/export';
+import { tagHashOf as t, withTag } from '../../utils/tag-hash';
```
dmsnell: we're not using `t` or `tagHashOf` right? can probably drop it

codebykat:
@dmsnell I appreciate the review but there's a major bug with this as noted in the TODO, which is why it was still set to draft. Do you know what I need to pass to get it to work with a tag with special characters in its name?

dmsnell:
since we're writing the tags into the `tags` array on the note object we shouldn't have to worry about transformation; we should be able to keep it written as typed in the search bar.

`withTag` should be preserving whatever it's given. when you say "empty" tag can you share the data you are referring to? what makes it empty vs. not having the tag at all?

codebykat:
I haven't debugged it deeply yet, but the note shows up with an empty tag pill.

lib/state/data/middleware.ts
sandymcfadden: I've updated based on this recommendation and now using `const`s

sandymcfadden:
This seems to be all working now. I've updated to get the `tagName` of the `openedTag` before sending to `withTag` as it was expecting a `tagName` and not a `tagHash`.

dmsnell:
we need to rebase this against the latest `develop` HEAD. `state.ui.openedTag` is gone and has been replaced by `state.ui.collection` in #2706, but which thankfully [introduced a new `openedTag` selector](https://github.com/Automattic/simplenote-electron/pull/2706/files#diff-662df48571826155fa958190db3aa7146f9dd332ca44151a8ea4f1ad554a7258R49) as well so we should be able to easily adapt this branch.

sandymcfadden:
@dmsnell ah yes you're right. I was going to rebase before I merged. I forgot that change would impact this. I'll work on that now.

sandymcfadden:
Thanks again @dmsnell for that catch. All updated again now.

dmsnell (commented):
there's nothing I see worth pointing out; it's a clean solution

for testing, did you verify that the added tag is the name of the tag as pulled from the tag bucket, not the tag hash, not a trivial transformation from the tag hash to an _unhashed_ version of the tag? (for example, does it properly at `Tag` if `Tag` is opened, because `tag` is the tag hash there).

codebykat (commented):
[Can't request changes because I'm the author apparently 😂 ]

There's an oddity that seems to be resetting the notes list to "all notes" after a new note is added. It's still displaying the tag in the title bar, but the note listing is now showing everything, not just notes with that tag:

Screen Shot 2021-03-25 at 6 23 49 PM

Screen Shot 2021-03-25 at 6 23 55 PM

sandymcfadden:
> for testing, did you verify that the added tag is the name of the tag as pulled from the tag bucket, not the tag hash, not a trivial transformation from the tag hash to an _unhashed_ version of the tag? (for example, does it properly at `Tag` if `Tag` is opened, because `tag` is the tag hash there).

Thanks @dmsnell! The `openedTag` selector provides us with the tagName, `withTag` wants the tagName as well and then tests to see if the hash of it matches the hashes of any other tags in the list already before adding it in.
I did test with the example you gave and `Tag` is added to the note.

sandymcfadden:
>There's an oddity that seems to be resetting the notes list to "all notes" after a new note is added. It's still displaying the tag in the title bar, but the note listing is now showing everything, not just notes with that tag:

Thanks @codebykat! I didn't notice that at all. I've got it updated now so that it displays the note list with the selected tag properly.

codebykat:
This is my PR so I can't approve it, but this is an approval! It's testing great for me now.

PR #2557, opened 2021-01-08
Fix: Note filtering not reset after new note is created from "no results" sidebar

codebykat:
### Fix

Currently the "create a new note titled" callout leaves the search query in a broken state, where the query is removed from the search field, but the filtering is still applied, which makes it appear as though notes are missing. The only way to remove the hidden filtering is by entering a new search query and then clearing it.

This fixes the bug by removing the search query clearing from the dispatch for `CREATE_NOTE_WITH_ID`. As a side effect, the search query will now persist through any creation of a new note, whether created from the sidebar callout or otherwise. It's unclear whether clearing the query when a new note was created was intended behavior but my suspicion is that it was not.

### Test

1. Search for a term with no results
2. Click on "Create a new note titled --"
3. Verify that the search query does not get cleared and you can still view matching notes in the sidebar

### Release

Fixed a bug causing notes to still be filtered after creating a new note from search results

### DIFF
--- a/lib/note-list/no-notes.tsx
+++ b/lib/note-list/no-notes.tsx
@@ -21,9 +21,10 @@ const NoNotes = () => {
   const getButton = () => {
     return searchQuery.length ? (
       <button
-        onClick={() =>
-          dispatch(actions.ui.createNote({ content: searchQuery }))
-        }
+        onClick={() => {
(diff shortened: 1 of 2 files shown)


belcherj (approved):

pachlava:
Filtering is still applied once new note is created from sidebar, as expected.

PR #2575, opened 2021-01-13
Chore: Upgrade Monaco dependencies

codebykat:
### Fix

This updates `monaco-editor-webpack-plugin` and `react-monaco-editor` as far as possible without incurring a required update to React 17.

* 0.40.0 is the last version of `react-monaco-editor` before it begins requesting React 17 as a peer dependency.
* 2.1.0 is the latest version of `monaco-editor-webpack-plugin`, which upgrades us from `monaco-editor@0.20.*` to `monaco-editor@0.21.*`.

This takes us up to `monaco-editor@0.21.2`, the latest released version (unfortunately still four months old) -- which doesn't fix anything in particular for us at the moment, but this stepping-stone will hopefully provide us a smoother upgrade path to the next release.

Props @belcherj for the fix for the regenerator runtime error, which I stole from https://github.com/Automattic/simplenote-electron/pull/2356.

### Test
1. Build and smoke-test the app

### Release

Upgraded Monaco editor package versions


package.json
```diff
@@ -148,7 +149,7 @@
     "jszip": "3.5.0",
     "lodash": "4.17.19",
     "md5.js": "1.3.5",
-    "monaco-editor-webpack-plugin": "1.9.0",
+    "monaco-editor-webpack-plugin": "^2.1.0",
```
dmsnell: did we intentionally start using compatibility versions?

my experience has taught me this is a lie, that no version upgrade should be considered compatible or safe

codebykat: I removed them from this PR, if we want to take them out of the file generally let's do that in a different PR

belcherj (approved):
Approved! Please, check Edge to ensure that is working.

codebykat:
Tested in Windows (Edge, Edge legacy, and CircleCI build of this PR), all looks good.

PR #2581, opened 2021-01-15
Remove perMachine option from Windows build config to allow non admin…

### LINKED ISSUE #1181: Windows installation without admin privileges
Hello!

Can this be installed without admin privileges? I would love to try it.
Technically, that is possible, I am questioning this more from a planning perspective.

Thanks

#### Simplenote version
1.4.0

#### OS version

Windows 7 Enterprise

sandymcfadden:
Remove perMachine option from Windows build config to allow non admin installations

### Fix

When installing on Windows it currently requires administrator privileges to complete the installation. 

Resolve issue in #1181. 

The issue above looks like this is caused by the build option [perMachine](https://www.electron.build/configuration/nsis) being set to true. This forces to install for everyone on the machine which requires admin privileges. The default value for this option is false so I've removed the option and using this PR to build the Windows install so I can test the installation process.

### Current install from exe Screenshots

Installing as Admin user
Installing as Admin user

Installing as Standard user
Installing as Standard user

### Test

1. Download the Windows exe install file generated from this PR
2. Go through the installation as both an Administrator user and Standard Windows user
3. See if it can be installed for current user without requiring admin privileges. 

### Todo

- Add release notes

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -50,7 +50,6 @@
     ]
   },
   "nsis": {
-    "perMachine": true,
     "artifactName": "Simplenote-win-${version}-${arch}.${ext}",
     "deleteAppDataOnUninstall": true
   },


belcherj (approved):
Great change!

pachlava:
[testing] Confirm that standard users is asked for admin privileges while installing 2.5.0, and not asked for them while installing 2.6.0-beta1. Tested on Win10

thomasstache:
Should the merge have closed #1181? Or only when there's a release?

sandymcfadden:
Thanks @thomasstache! You're right, that issue should have been closed. I've done it now.

killerharri:
At the end this is a very bad solution :-(

Now it is no more possible to install the application to c:\program files, where the simplenote.exe would be write protected for non-admin users. Now every user on a computer needs to install this file into his personal user profile, where it is writeable and can be infected by malware. This removed the additional security improvement, which windows delivers when user non-admin accounts for normal work.
Btw. some security options prevent running *.exe files from the user profile, simply because this is often used by malware because it is working without admin rights.

Would it be possible to re-add the per-machine selection page and set the default to user profile? Or add some commandline optione to allow installation to c:\program files (with admin rights, of course)? The installer should require the elevation only if the per-machine installation is selected using this commandline option.

Issue #1396, opened 2019-06-21
Accessibility: Revision selector needs a label


belcherj:
The revision selector input in this [component](https://github.com/Automattic/simplenote-electron/blob/develop/lib/revision-selector/index.jsx) does not have a label.  It needs a label that is hidden

```<input class="slider" type="range" min="0" max="5" value="5">```

### What
Add a hidden lablel to the revision selector input

### Why
To make it accessible to screen readers

### How
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints

<img width="454" alt="Screen Shot 2019-06-21 at 1 54 44 PM" src="https://user-images.githubusercontent.com/6817400/59941953-81489400-942c-11e9-853a-abf78798ce6f.png">

alemagio:
Hi, i would like to work on this issue.

belcherj:
PR's are welcome.  Sorry I didn't respond sooner!

codebykat:
See also #980

alemagio:
I'm sorry i could not work on this but i had some trouble.
If anyone else want to it is ok for me

tbourrely:
Hi, I would be happy to work on this if possible :)

dmsnell:
@tbourrely go for it! what do you need from us to get started?

you are welcome to fork the repo and submit a PR 😄

tbourrely:
Well, once I can get my environment working, I will be good to go 😄  !

sandymcfadden:
Looks like things have been refactored since this was opened. This input is now in the [Slider component](https://github.com/Automattic/simplenote-electron/blob/develop/lib/components/slider/index.tsx).

May want to approach it similar to how we work with the [Search input](https://github.com/Automattic/simplenote-electron/blob/develop/lib/search-field/index.tsx#L81).

PR #2583, opened 2021-01-16
feat(slider): aria-label

tbourrely:
### Fix

Adding an Aria label to the revision slider input to be more accessible specifically for screen readers.
Addresses Issue #1396 

### Test

1. View the history of a note
2. Inspect the slider element
3. See the label is now showing.

### Release

- Add label to note revision slider to make more accessible.

### DIFF
--- a/lib/components/slider/index.tsx
+++ b/lib/components/slider/index.tsx
@@ -16,6 +16,7 @@ export const Slider: FunctionComponent<Props> = ({
   onChange,
 }) => (
   <input
+    aria-label="Select revision"
     className="slider"
     disabled={disabled}
     type="range"


tbourrely:
This should resolve #1396

sandymcfadden (changes requested):
Thanks so much for submitting this!
While taking a closer look we can make this even easier. This is the only instance of us using the Slider component and I don't see any other times currently that we may use it elsewhere. With that, we can instead add the `aria-label` directly to the input without the need of passing in the text for the label as Props. 

If you want to update to that we can get this merged.

tbourrely:
Done 🙂

sandymcfadden (approved):
Looks good. Thanks so much!

pachlava:
[testing] Confirm that slider input now has a new 'aria-label' attribute, with "Select revision" value
Win10 | desktop SN 2.6.0-beta1

PR #2584, opened 2021-01-18
Updating the ui for the editing tags section

### LINKED ISSUE #2567: UI spring cleaning: Edit tags
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5347608905ee4ae2dd3
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc59345ce0c66fcd151a1

* Move delete icon to right / change to trash can instead of x

Screen Shot 2021-01-10 at 10 55 19 PM

Screen Shot 2021-01-10 at 10 55 24 PM

sandymcfadden:
### Fix

As part of the UI Spring cleaning we are doing this looks to take care of updating the Edit Tags section. Resolves #2567.

This changes from using a cross icon on the left side of the tag name to using a trash icon on the right side. 
While editing the styles I also removed a number of lines that don't appear to be in use anymore at all. 

I also updated some of the styles to better use the classes in the themes.scss file instead of colors in the component styles. Not all are removed more refactoring will be needed. 

Removed two icons no longer needed `cross-outline-small` and `cross-outline`.

### Test

1. Go to menu to see tags
2. Tap on edit link
3. Notice the blue icon on the right side
4. Try in both light and dark mode theme

### Screenshots

Screen Shot 2021-01-22 at 3 01 53 PM

Screen Shot 2021-01-22 at 2 58 51 PM

### Release

- Update styles for edit tags panel including changing the cross outline icon to the trash icon.


lib/tag-list/index.tsx
```diff
@@ -78,7 +82,16 @@ const SortableTag = SortableElement(
...
         value={tag.name}
       />
-      {editingActive && allowReordering && <TagHandle />}
+      {editingActive && (
+        <button className="button-borderless">
+          <TrashIcon onClick={() => trashTag(tag.name)} />
+        </button>
+      )}
+      {editingActive && allowReordering && (
+        <button className="theme-color-fg-dim">
```
codebykat: I think both of these buttons should be `icon-button` (and I guess we need to override the colors for now)

lib/tag-list/style.scss
```diff
@@ -64,10 +62,11 @@
   svg {
     cursor: pointer;
   }
+}
 
-  .icon-cross-outline {
-    margin-right: 5px;
-    color: $studio-red-50;
+.tag-list-item {
```
codebykat: There's a linting error from having this same one twice. You can see it by clicking through to the CircleCI error, and you can also run the linter locally with `npm run lint`.

(But also changing this class to an `icon-button` gives us the margin we want for free)

lib/tag-list/style.scss
```diff
@@ -108,138 +107,17 @@ button.tag-list-input {
...
 .icon-reorder {
-  margin-right: 25px;
-}
-
-.editable-list {
-  display: flex;
-  flex-direction: column;
-  overflow: auto;
-  overflow-x: hidden;
-  -webkit-overflow-scrolling: touch;
```
codebykat, line 119: I'm a little nervous about wiping out all of these transition, cursor and z-index styles that were attached to the editable list class. Are we sure they're not needed?

lib/tag-list/style.scss
```diff
@@ -64,10 +62,11 @@
   svg {
     cursor: pointer;
   }
+}
 
-  .icon-cross-outline {
```
codebykat, line 68: Since we don't use this icon anywhere else, I think we can remove it from /icons for now as well.

lib/tag-list/style.scss
```diff
@@ -108,138 +107,17 @@ button.tag-list-input {
...
-      pointer-events: auto;
-    }
-
-    .editable-list-reorder {
-      opacity: 1;
-      pointer-events: auto;
-      transform: translateX(0);
-    }
-  }
+  margin-right: 5px;
```
codebykat: Instead of putting these on the icon, let's put them on the parent button. I would recommend giving it its own class (such as `"button-reorder"`) and any color we give it should propagate to its children.

Setting the margin on the icon moves the icon over within its 22x22 parent container, which isn't what we want (and limits how far we can move it). Setting it on the parent container gives us more leeway to pad the icon out, while still keeping it centered within its container.

I hope that makes sense, it's hard to describe.

Here's the button with the parent div padded out.

Screen Shot 2021-01-21 at 7 24 53 PM

Note that we can give it a much bigger margin now if we want:

Screen Shot 2021-01-21 at 7 26 14 PM

Also worth noting that this margin, since it is on the parent container, is not clickable.

Here's what happens if we try to set a 100px `margin-right` on the `svg.icon-button` instead. It doesn't have anywhere to go, so it ends up overflowing to the right. Additionally, all of the margin (yellow space) is clickable, because it's within the parent element, so we have a lot of clickable whitespace (sometimes we want this, but probably for icon buttons, we don't).

Screen Shot 2021-01-21 at 7 27 06 PM
Screen Shot 2021-01-21 at 7 27 00 PM

lib/tag-list/style.scss
codebykat: A bit later: There's a flicker when dropping a dragged item that might be related to removing some of these transitions, but I haven't tested it. This is in dark mode. When you drop the item the trash icon flashes out of existence briefly.

lib/tag-list/style.scss
```diff
@@ -20,19 +20,21 @@
...
+    border-bottom: 1px solid;
 
     h2 {
-      font-size: 14px;
-      font-weight: 500;
+      font-size: 16px;
+      font-weight: 400;
       margin-bottom: 0;
       text-transform: capitalize;
+      color: $studio-gray-80;
```
codebykat: Annoying that the panel title component is hard-coded as `theme-fg-dim`.

Instead of adding colors to override, let's remove the `PanelTitle` usage and just make it a normal `<h2>` in the tags list:

`<h2 class="theme-color-fg">Tags</h2>`

It looks like the only style we want from `panel-title` is a `margin: 0 0 0.5em;`; we can also remove the `text-transform` override if we remove the `panel-title` class.

lib/tag-list/style.scss
codebykat: We might in fact want a slightly bigger margin on this to avoid interfering with the scrollbar.

Screen Shot 2021-01-21 at 6 56 07 PM

This is 15px:

Screen Shot 2021-01-21 at 7 48 33 PM

lib/tag-list/style.scss
```diff
@@ -108,138 +107,17 @@ button.tag-list-input {
...
-
-.tag-list-item.theme-dark {
-  border-bottom: 1px solid $studio-gray-70;
-
-  svg.icon-reorder {
-    fill: $studio-gray-30;
+  .tag-list-title {
+    .panel-title {
+      color: $studio-white;
+    }
```
codebykat, line 123: Again, we can remove the title styles if we swap the PanelTitle component out for a plain H2.

lib/tag-list/index.tsx
codebykat: Looks like the trash icon should be `$studio-simplenote-blue-30` in dark mode and `$studio-simplenote-blue-50` in light mode. (I know it makes it look like we should use `button-borderless` for these because they're the same color but the borderless buttons are actually something different, updates to that coming soon, sorry!!)

The drag handles are lighter, which you did notice already -- `$studio-gray-50` in dark mode and `$studio-gray-10` in light mode. I pinged Sly to double-check that those should be different from the other icon colors.

codebykat (changes requested):
This is getting there! I left some line comments but they're mostly small (and caused by the general confusion in all of our styles rather than any fault of yours).

In dark mode, the item being dragged is getting a white `bottom-border`. It took me some time to figure out why. The draggable component we're using, react-draggable-hoc, places a clone of the `li` at the very end of the `body`. (You can see this in the inspector by scrolling to the bottom and then clicking to drag a tag item.)

It turns out we can add a class to the item that's being dragged by adding this to the `<SortableTagList` options in `tag-list/index.tsx`:

`helperClass="is-dragging"`

and then adding something like this to the styles:

```
.theme-dark.tag-list-item.is-dragging {
  border-color: $studio-gray-80;
}
```

It doesn't seem like it should be necessary since this item has the `theme-color-border` class on it, so maybe there's a better way. I also didn't test to see if it needs to be addressed in light mode as well.

lib/tag-list/index.tsx
codebykat: Update: Confirmed the handles are intended to be a lighter color than the other gray icons

sandymcfadden: Thanks! I've updated this based on the suggestion and changed them to be icon-button and updated the colors using this in the component.

lib/tag-list/style.scss
sandymcfadden: Thanks for pointing this out. I have this resolved now.

lib/tag-list/style.scss
sandymcfadden: I searched through and I couldn't find these classes being used anywhere.

It looks like they were introduced here https://github.com/Automattic/simplenote-electron/commit/198bc7f1cf450c16124f160f68771ef0733efe40
/lib/editiable-list.jsx

Then they were removed here as part of the re-write https://github.com/Automattic/simplenote-electron/commit/fc8921291b6b702db35b076d65e4632203ff4574

Seems like a lot of the code from lib/editable-list/style.scss was moved to lib/tag-list/style.scss but wasn't needed as those classes weren't used anymore.

I'm still not sure about the flicker you mentioned though, will have to check into this more.

lib/tag-list/style.scss
sandymcfadden: I removed this and the small version as neither seemed to be used anywhere.

lib/tag-list/style.scss
sandymcfadden: This makes a lot of sense, thank you for explaining it so well. I've added classes to the buttons and used those.

Based on these and other changes I'm not seeing any overlap with the scrollbar anymore.

lib/tag-list/style.scss
sandymcfadden: I wondered and was going to ask if it would be ok to remove the `<PanelTitle>`, thanks for this. I've updated it accordingly.

sandymcfadden:
Thanks for all the amazing feedback @codebykat, it was so helpful!

In regards to the border for the items being dragged.
We give it the class `theme-color-border` but that style is defined as `.theme-light .theme-color-border` or `.theme-dark .theme-color-border` in theme.scss. The theme class is added to the div right below the root element. When dragging the temporary element is added to the body so `.theme-light .theme-color-border` doesn't match.

Instead, the element has both classes `theme-dark` and `theme-color-border` so in theme.scss I've added rules like this to catch and apply the border properly.

```
.theme-dark {
 .theme-color-border,
  &.theme-color-border {
    border-color: $studio-gray-80;
  }
}
```

codebykat (approved):
This is testing well for me now, code looks good, I no longer see that flicker in dark mode.

pachlava:
@sandymcfadden can't verify as fixed because sort icons do not appear in Electron app for Windows 10 (but they do appear in Chrome on Windows, and both app/Chrome on macOS):

Screenshot 2021-02-11 at 15 16 48

sandymcfadden:
Thanks for helping me figure out what is going on here @pachlava.
When tags are set to be sorted alphabetically the sort handle doesn't appear.

pachlava:
@sandymcfadden thank you too! 

Verified on Dell Latitude 7400 | Win 10 | Simplenote 2.7.0-beta1

PR #2586, opened 2021-01-19
Updating revision slider position and styling

### LINKED ISSUE #2565: UI spring cleaning: revisions slider
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5329259b45d0a0dddf0
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc594b0ad775786578d24

* Move to bottom of note and re-style

Screen Shot 2021-01-10 at 10 49 55 PM

Screen Shot 2021-01-10 at 10 50 01 PM

sandymcfadden:
### Fix

Part of our UI Spring Cleaning and resolves #2565.
This moves the Slider to the bottom of the note instead of the top and starts to adjust the styles.
This includes updated colors and has the revision date follow the current slider position

Screen Shot 2021-01-25 at 11 24 16 AM
Screen Shot 2021-01-25 at 11 38 36 AM

### Items not yet addressed

- Part of the new design is to have tick marks below the slider one for each revision available. This will be looked at in a different PR later.
- In Chrome the slider handle is down below the slider track, but that seems to be the case in the current live version as well. Does not happen in Electron.
- The design calls to show just the date for revisions and not the time. This makes it awkward for multiple revisions on the same date so for now I've left the time in as well.
- Some colors are still being set in the component and not in the theme.scss file.

### Test

1. Open history for a note
2. Notice the revision slider is now at the bottom
3. Move slider ensure revision date follows current slider position
4. Ensure styling looks correct.
5. Repeat for both dark mode and light mode. 

### Release

- Move note revision slider to the bottom of note and update styles.

### DIFF
--- a/lib/components/slider/index.tsx
+++ b/lib/components/slider/index.tsx
@@ -14,7 +14,6 @@ export const Slider: FunctionComponent<Props> = ({
   min,
   max,
   value,
-  list,
   onChange,
 }) => (
   <input
@@ -25,7 +24,6 @@ export const Slider: FunctionComponent<Props> = ({
     min={min}
     max={max}
     value={value}
-    list={list}
     onChange={onChange}
   />
 );
--- a/lib/components/slider/index.tsx
+++ b/lib/components/slider/index.tsx
@@ -6,7 +6,6 @@ type Props = {
   min: number;
   max: number;
   value: number;
-  list: string;
 };
 
 export const Slider: FunctionComponent<Props> = ({
--- a/lib/components/slider/index.tsx
+++ b/lib/components/slider/index.tsx
@@ -6,6 +6,7 @@ type Props = {
   min: number;
   max: number;
   value: number;
+  list: string;
 };
 
 export const Slider: FunctionComponent<Props> = ({
@@ -13,6 +14,7 @@ export const Slider: FunctionComponent<Props> = ({
   min,
   max,
   value,
+  list,
   onChange,
 }) => (
   <input
--- a/lib/components/slider/index.tsx
+++ b/lib/components/slider/index.tsx
@@ -25,6 +25,7 @@ export const Slider: FunctionComponent<Props> = ({
     min={min}
     max={max}
     value={value}
+    list={list}
     onChange={onChange}
   />
 );
--- a/lib/components/slider/style.scss
+++ b/lib/components/slider/style.scss
@@ -12,17 +12,17 @@
     display: flex;
     align-items: center;
     height: 2px;
-    background: $studio-white;
+    background: $studio-gray-30;
     border-radius: 5px;
   }
   &::-moz-range-track {
     height: 2px;
-    background: $studio-white;
+    background: $studio-gray-30;
     border-radius: 5px;
   }
   &::-ms-track {
     height: 2px;
-    border-color: $studio-simplenote-blue-50;
+    border-color: $studio-gray-30;
     border-width: 7px 0;
     color: transparent;
   }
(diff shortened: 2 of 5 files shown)


lib/components/slider/index.tsx
```diff
@@ -6,24 +6,34 @@ type Props = {
...
+      disabled={disabled}
+      type="range"
+      min={min}
+      max={max}
+      value={value}
+      list={list}
+      onChange={onChange}
+    />
+    <datalist id={list}>
+      {_.range(min, max + 1).map((value, index) => {
```
dmsnell: let's avoid assuming that a global `_` exists. I didn't realize this is the case - does this work or crash?

if you want `lodash.range` it can be imported at the top of the module

```ts
import {range} from 'lodash';
```

sandymcfadden: Thanks @dmsnell. This did work I used the autocomplete, but I do see a warning in VS Code about importing as well. I'm going to end up removing this from this PR before it's ready to go. In the future I'll be sure to explicitly import.

lib/components/slider/index.tsx
```diff
@@ -6,6 +6,7 @@ type Props = {
   min: number;
   max: number;
   value: number;
+  list: string;
```
belcherj: Remove as this is no longer used.

sandymcfadden: Thanks! Removed that now.

dmsnell:
I'll leave this note here since there's a particularly detail-centric need with the revision slider that's confusing. In addition to now showing the blank note at revision `0` we have a few other things to consider.

 - if there are more than 60 revisions for a note then the revision at the `0` index won't be a blank note and thus it will be a legitimate selectable revision
 - the position on the slider at the far right should be one more than the number of available revisions but it shouldn't select a revision; it should select "no revisions" and appear as the current working copy of the note. this may be content the server isn't aware of yet and it's not possible to "revert" to it, but rather "reverting to it" would imply cancelling the dialog without reverting to any given version

dmsnell:
Didn't mean to close - I clicked the wrong button

sandymcfadden:
Hi @dmsnell I need a bit more clarification on your comment above. Was this directed at this PR or something to consider for the future?
I haven't changed any of the functionality of the revision slider with this one at all. 
Currently, the way it is working is it seems to show all revisions we have for a note, at least more than 60.
The leftmost position on the slider would be the first version of the note which had content it looks like. The far-right position shows the current version of the note. If the current version is selected the Restore button is not accessible and you are only able to cancel the dialog.

dmsnell:
If this is limited to the HTML structure and accessibility concerns then we don't have to fix the existing bugs. I wanted to ear-mark the issues though as we come around to the revision slider so we're aware of them.

The "current version" for example can be misleading, since most folks will consider the current version to be whatever they have written in their note but if I'm not wrong, the revision slider says it's "the most-recent version of the note we got from the server"

codebykat:
As part of the UI spring cleaning project, I think we should keep the changes to style/layout updates only, not functionality (unless something is really broken, or the design changes imply deeper functional changes). Let's open a new issue to log the existing problems with the revision slider and try to get them fixed up.

codebykat:
Opened an issue here for follow-up work: https://github.com/Automattic/simplenote-electron/issues/2630

codebykat (approved):
This tests great! I was able to view and restore history, looks good in light and dark modes.

pachlava:
[testing] fix/feature verified on MBP 16" | macOS 11.2. No access to Windows meanwhile ;(

PR #2587, opened 2021-01-19
Add/email verification

belcherj:
### Fix

Adds the Email Verification UI to either update or send verification email and dialog reminding users to click the verification link that is in the email we sent them.

### Test
1. Have an unverified account
2. Log in to the app
3. Click the button to change email
4. Change email
5. Refresh app
6. Click button to send email
7. Observe that dialog changes to the secondary dialog asking you to check your email
8. Click link in email
9. Observe dialog disappears
10. Log out
11. Log in
12. Refresh
13. Observe dialog does not appear again

### Release

- Added email verification UI

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -34,6 +34,7 @@ type StateProps = {
   isSmallScreen: boolean;
   lineLength: T.LineLength;
   isSearchActive: boolean;
+  showEmailVerification: boolean;
   showNavigation: boolean;
   showNoteInfo: boolean;
   theme: 'light' | 'dark';
@@ -161,6 +162,7 @@ class AppComponent extends Component<Props> {
     const {
       isDevConfig,
       lineLength,
+      showEmailVerification,
       showNavigation,
       showNoteInfo,
       theme,
@@ -180,7 +182,7 @@ class AppComponent extends Component<Props> {
 
     return (
       <div className={appClasses}>
-        <EmailVerification />
+        {showEmailVerification && <EmailVerification />}
         {isDevConfig && <DevBadge />}
         <div className={mainClasses}>
           {showNavigation && <NavigationBar />}
@@ -199,6 +201,7 @@ const mapStateToProps: S.MapState<StateProps> = (state) => ({
   isSearchActive: !!state.ui.searchQuery.length,
   isSmallScreen: selectors.isSmallScreen(state),
   lineLength: state.settings.lineLength,
+  showEmailVerification: selectors.shouldShowEmailVerification(state),
   showNavigation: state.ui.showNavigation,
   showNoteInfo: state.ui.showNoteInfo,
   theme: selectors.getTheme(state),
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -6,6 +6,7 @@ import NavigationBar from './navigation-bar';
 import AppLayout from './app-layout';
 import DevBadge from './components/dev-badge';
 import DialogRenderer from './dialog-renderer';
+import EmailVerification from './email-verification';
 import { isElectron, isMac } from './utils/platform';
 import classNames from 'classnames';
 import {
@@ -179,6 +180,7 @@ class AppComponent extends Component<Props> {
 
     return (
       <div className={appClasses}>
+        <EmailVerification />
         {isDevConfig && <DevBadge />}
         <div className={mainClasses}>
           {showNavigation && <NavigationBar />}
--- a/lib/email-verification/index.tsx
+++ b/lib/email-verification/index.tsx
@@ -1,0 +1,146 @@
+import React, { Fragment, FunctionComponent, useState, useEffect } from 'react';
+import classNames from 'classnames';
+import Modal from 'react-modal';
+import { connect } from 'react-redux';
+
+import CrossIcon from '../icons/cross';
+import MailIcon from '../icons/mail';
+import WarningIcon from '../icons/warning';
+
+import actions from '../state/actions';
+
+import * as S from '../state';
+import * as T from '../types';
+
+type StateProps = {
+  accountVerification: T.VerificationState;
+  dismissed: boolean;
+  email: string | null;
+  theme: string;
+};
+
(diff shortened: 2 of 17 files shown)


lib/state/account/actions.ts
```diff
@@ -0,0 +1,23 @@
...
+});
+
+export const setEmailVerified: A.ActionCreator<A.SetEmailVerified> = (
+  verified: boolean
+) => ({
+  type: 'SET_EMAIL_VERIFIED',
+  verified,
+});
+
+export const setSendEmailVerifyURL: A.ActionCreator<A.SetEmailVerifyURL> = (
```
belcherj: Delete this

dmsnell (commented):
Thanks for pushing this out so quickly. Here are a few initial reactions if you find them useful.

We're storing separate boolean parameters for the email verification status but I believe that it could be collapsed into a single state value, and if we did that, the actions and updates would also become simpler, requiring fewer actions and reducers.

```ts
type VerificationState = 'verified' | 'pending' | 'dismissed' | 'unverified' | 'unknown';

const accountVerification: A.Reducer<VerificationState> = (state = 'unknown', action) =>
	action.type === 'UPDATE_ACCOUNT_VERIFICATION' && state !== 'dismissed'
		? action.state
		: state;
```

Another thing it appears we're doing here is attempting to store the entire account bucket in state and I don't believe we need to do that or care about doing that. As of our known needs now, we will need to update the account info on every load and all we need is one value from it. To that end I don't think we need to shove any values from Simperium into Redux state. I think we can rely instead on the default in-memory bucket store provided by `node-simperium`. That would eliminate the new `BucketStore` and allow us to focus solely on trapping the update to the `email-verification` entity.

```ts
const parseVerificationToken = (token: string) => {
	if (!token) {
		return null;
	}

	try {
		const {username, verified_at: verifiedAt} = JSON.parse(token);
		return {username, verifiedAt };
	} catch (e) {
		return null;
	}
}

const updateVerificationState = ({token, pending}) => {
	const parsedToken = parseVerificationToken(token);
	const hasValidToken = parsedToken && parsedToken.username === username;
	const hasPendingEmail = pending?.username === username;

	const state =
		hasValidToken ? 'verified' : 
		hasPendingEmail ? 'pending' :
		'unknown';

	return dispatch(setVerificationState(state));
}

const accountBucket = client.bucket('account');
accountBucket.channel.on('update', (entityId, entity) => {
	if ('email-verification' == entityId) {
		udpateVerificationState(entity);
	}
});
```

We're adding specifics about the structure of the data in the API inside of our `lib/types` file, which corresponds to how we have some note and tag data, but in this case it's not necessary and we don't have the new `account` bucket setup as a table with homogeneous entity types. To that end I wonder if it would be more helpful to eliminate these types altogether. It's not exactly accurate to say that the type of the `account` bucket is this new `Account` type (because it's a more standard Simperium bucket it's type is more appropriately `JSONValue`) and we shouldn't need that bucket at all and we don't reference these types anywhere else.

Finally, we briefly touched on this in a chat, but the `EmailVerification` component makes a major divergence from the pattern established for React components in the project and with TypeScript and is more complicated and harder to test or extract into separate contexts with the heavy use of hooks. What problem are we trying to solve by moving to hooks?

How is is more complicated?

It's more complicated because the explicit contracts that are present with typed React components is now gone and one has to read through the component and interpret the component in order to understand its API. that API isn't "fixed" anymore and can change simply by essence and not by explicit intention (that is, by adding a new hook buried deep in the component function).

The hook lifecycle is different from the `connect` lifecycle. With `connect` it's obvious that the entire component atomically updates as app state changes. With hooks we have re-introduced opportunities for partial updates since we don't have the guarantee that the hooks all update together in one operation. For example, what happens when we call `setIsShown()` after two seconds if the component isn't mounted? A dispatch to dismiss the UI would only affect the data in app state and be a safe operation; when calling closures inside components it's less obvious if we'll have a crash or not due to a component no longer being mounted.

The `useState` `isShown` status here also might be more adequately modeled if we lean more on app state design with something like the proposed `accountVerification` status. Using the declarative design takes us away from these imperative partial updates and I think "set is shown" could be replaced more aptly by a conditional render based on app state.

```ts
{accountVerification === 'unverified' && <EmailVerification />}
```

How is it harder to test?

Because of the magic behind the scenes of the hooks, where the component _looks_ like a straightforward function but depends on machinery behind the scenes, we have to add additional setup in our test code, especially if we want to test specific states of our UI.

```ts
// with a defined/wrapped component interface
import {EmailVerification} from '../'

<EmailVerification email="test@example.com" />
```

The defined component interface, beyond explicitly indicating in one place the data a component depends on, also lifts the need for plenty of tests. That is, TypeScript tests all of that contract for us and we know that as long as we fill in the right types then we can be assured that the component works as expected.

If we're worried about not testing the fully-connected component in the state tree then that's evidence we're trying to do integration tests inside of our unit tests. We have an end-to-end test suite wherein we can verify that stuff, but TypeScript already ensures we don't hook up state that doesn't exist or which doesn't match the expectations sets by the component API.

Take that for what you will. I wanted to flesh out in specifics what about it is more complicated since that term can be tossed around subjectively. I didn't mean "it's harder for me to understand" (though it's definitely harder to understand what expectations the component has since they are no longer listed in one place).

codebykat (approved):
There's one remaining issue, which is the welcome note not showing up on new account creation until the page is reloaded. Let's merge this and address that problem in a follow-up PR.

codebykat:
Okay @dmsnell helped figure out the bug with the note bucket not getting reindexed after the welcome note was added. I also added analytics in cf540191cdedcdfc2942a9b21b5004ccf3baf124 (as defined in p2XJRt-2AX-p2). I have tested the flows on this one and it is looking good to me, I would like at least one other person to review and test this before we merge.

belcherj (commented):
Looks great

pachlava:
[testing] implementation verified on Win10 | SN desktop 2.6.1-beta1

PR #2591, opened 2021-01-20
Remove ?from=react from Settings URL

codebykat:
### Fix

We can deploy this once the associated PR is deployed on App Engine. (It won't break or anything but this parameter is no longer required.)

### Test
1. Go to Settings
2. Ensure you see the abbreviated Settings screen (without the display settings for the old web app)
3. Click on Back to Notes and ensure you are directed back to the new app

### Release

Updated the external Settings URL


belcherj (approved):
Great!

pachlava:
[testing] Confim that '/?from=react' is no longer present in the URL once user clicks 'Edit Account'. Checked on Win10 | SN 2.6.0-beta1

Issue #2563, opened 2021-01-11
UI spring cleaning: Top menu bar


codebykat:
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5303183bb6483b4e9c8
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc59220aab754ed6ecd18

* !Add new note icon to top left when sidebar is collapsed~ Resolved in #2596
* ~Add checklist icon to menu~ Addressed in #2603
* ~Add new info and ... buttons~ Let's call this part of #2569

<img width="235" alt="Screen Shot 2021-01-10 at 11 02 38 PM" src="https://user-images.githubusercontent.com/52152/104153759-07725580-5398-11eb-8779-9918e25fab2a.png">

<img width="188" alt="Screen Shot 2021-01-10 at 11 03 58 PM" src="https://user-images.githubusercontent.com/52152/104153805-27097e00-5398-11eb-96d2-8a9b7aceabd2.png">

<img width="891" alt="Screen Shot 2021-01-10 at 10 45 32 PM" src="https://user-images.githubusercontent.com/52152/104152847-95007600-5395-11eb-9173-b85b02ba4241.png">

<img width="713" alt="Screen Shot 2021-01-10 at 10 44 17 PM" src="https://user-images.githubusercontent.com/52152/104152802-77cba780-5395-11eb-8412-778c19483c88.png">

PR #2596, opened 2021-01-25
Add new note icon to top toolbar when in focus mode

sandymcfadden:
### Fix

Adding the new note icon to the note toolbar when we are in focus mode.
This addresses part of the issue #2563 as part of the UI Spring Cleaning.

Screen Shot 2021-01-25 at 3 01 04 PM
Screen Shot 2021-01-25 at 3 01 26 PM

### Test

1. Enter focus mode by toggling the left sidebar so it is hidden 
2. Click the new note icon in the upper left
3. Ensure the new note is created
4. Try this in both light and dark mode to ensure styles appear correct. 

### Release

- Add the new note icon to the toolbar when in focus mode.

### DIFF
--- a/lib/note-toolbar/index.tsx
+++ b/lib/note-toolbar/index.tsx
@@ -17,7 +17,6 @@ import ShareIcon from '../icons/share';
 import SidebarIcon from '../icons/sidebar';
 import TrashIcon from '../icons/trash';
 import actions from '../state/actions';
-import { createNote } from '../state/ui/actions';
 
 import * as S from '../state';
 import * as T from '../types';
@@ -204,7 +203,7 @@ const mapStateToProps: S.MapState<StateProps> = ({
 
 const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
   deleteNoteForever: actions.ui.deleteOpenNoteForever,
-  newNote: createNote,
+  newNote: actions.ui.createNote,
   restoreNote: actions.ui.restoreOpenNote,
   shareNote: () => actions.ui.showDialog('SHARE'),
   toggleEditMode: actions.ui.toggleEditMode,
--- a/lib/note-toolbar/index.tsx
+++ b/lib/note-toolbar/index.tsx
@@ -22,7 +22,6 @@ import * as S from '../state';
 import * as T from '../types';
 
 type StateProps = {
-  newNote: Function;
   editMode: boolean;
   hasRevisions: boolean;
   isOffline: boolean;
--- a/lib/note-toolbar/index.tsx
+++ b/lib/note-toolbar/index.tsx
@@ -3,10 +3,7 @@ import { connect } from 'react-redux';
 import { isMac } from '../utils/platform';
 
 import BackIcon from '../icons/back';
-<<<<<<< HEAD
 import ChecklistIcon from '../icons/check-list';
-=======
->>>>>>> Add new note icon to top toolbar when in focus mode
 import IconButton from '../icon-button';
 import InfoIcon from '../icons/info';
 import NewNoteIcon from '../icons/new-note';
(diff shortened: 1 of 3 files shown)


codebykat (changes requested):
Looks and works great! There's a teensy alignment issue on the Mac build (I think we're adding some extra toolbar padding for this platform):

Screen Shot 2021-01-25 at 12 06 30 PM

I'm also wondering (and this is an honest question, not me being Socratic) if there's a way to use the same button in the note toolbar and in the menu bar, and use styles to change its placement... or even if those two menu bars should now be the same component. 🤔

Does that seem possible or too much of a pain?

lib/note-toolbar/index.tsx
```diff
@@ -25,6 +28,7 @@ type StateProps = {
 
 type DispatchProps = {
   deleteNoteForever: () => any;
+  onNewNote: () => any;
```
dmsnell: while common, and common historically in this repo, the `onBlahBlah` naming can be verbose beyond value. our types ensure that something is a function or a value, so let's consider naming the prop more directly, something like `createNote` or `newNote` or `createNote` or `createNewNote`

lib/note-toolbar/index.tsx
```diff
@@ -180,6 +191,7 @@ const mapStateToProps: S.MapState<StateProps> = ({
 
 const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
   deleteNoteForever: actions.ui.deleteOpenNoteForever,
+  onNewNote: () => createNote(),
```
dmsnell: fun fact! there's no need to create a new function closure here unless we specifically want to create a function which ignores its arguments (as can be more useful when passing it as above in a click callback, which otherwise would pass the click event as the first argument to `onNewNote`)

for this we can pass the function directly and the type system makes sure we don't accidentally pass a non-function value.

```ts
onNewNote: createNote,
```

lib/note-toolbar/index.tsx
sandymcfadden: Thanks Dennis! I've updated the naming now.

lib/note-toolbar/index.tsx
sandymcfadden: This is really helpful, thank you again!

sandymcfadden:
Thanks for pointing me in the right direction with the padding @codebykat and catching that. I had tested on macOS and then made some changes and obviously didn't check closely again. 

I did take a quick look to see before I started to see if I could use that same button but didn't see a good way to do it. I didn't look at refactoring to make the menu bars one component though. I'll investigate that a bit.

sandymcfadden:
@codebykat I took a look at moving the two menus into one component. I think it could be done. It would likely make for a big component and refactor the layout of the app. I don't think it is something I could do as part of this PR, but would be happy to look at it more together at some point if you are up for it.

lib/note-toolbar/index.tsx
```diff
@@ -1,21 +1,24 @@
...
 import InfoIcon from '../icons/info';
+import NewNoteIcon from '../icons/new-note';
 import PreviewIcon from '../icons/preview';
 import PreviewStopIcon from '../icons/preview-stop';
 import RevisionsIcon from '../icons/revisions';
 import TrashIcon from '../icons/trash';
 import ShareIcon from '../icons/share';
 import SidebarIcon from '../icons/sidebar';
 import actions from '../state/actions';
+import { createNote } from '../state/ui/actions';
```
dmsnell: note that this should already be available from `actions` as `actions.ui.createNote`

sandymcfadden: Thank you! I updated based on this feedback.

codebykat (approved):
Looks and works great now, including on OSX 👍

pachlava:

Issue #2549, opened 2021-01-06
Keyboard shortcuts are displayed as "Ctrl" in browser, including on the Mac


codebykat:
We could test this in a more sophisticated way 

https://github.com/Automattic/simplenote-electron/blob/develop/lib/dialogs/keybindings/index.tsx#L43

and 

https://github.com/Automattic/simplenote-electron/blob/develop/lib/utils/platform.ts#L4

### Expected
Keybindings should display with Cmd on Mac browsers

### Observed
Keybindings are displayed as Ctrl unless using Mac on Electron

codebykat:
#2601 is a partial fix but I want to think more about this because it still seems like a bug, bumping this to 2.8.0 for further consideration 🤔

PR #2601, opened 2021-01-27
Fix the isMac variable so it is accurate on web and electron

sandymcfadden:
### Fix

We were using the variable isMac to determine if we should use show Ctrl or Cmd for keyboard shortcuts and a couple other places. But isMac was only defined on Electron. It The keyboard shortcut display worked, but was a side effect.

- This fixes the isMac variable so it reliably will tell us if a system is Mac. 
- Updates places where isMac was used but meant for only Mac & Electron to check both isMac and isElectron.
- Fixes the Insert checklist keyboard shortcut so it works with both Ctrl and Cmd on Mac Browsers
- Introduces a new variable CmdOrCtrl which allows us to display keyboard shortcuts appropriately.
- Updates icon tool tips to use this variable so they wills how the right keyboard shortcut.

### Test

1. Open keyboard shortcuts with either `Ctrl + /` or `Cmd + /` keyboard shortcut or by going to the menu and clicking the keyboard shortcut link
2. Ensure on Mac Electron the keyboard shortcuts are showing as Cmd
3. Ensure on Windows and Linux they are showing as Ctrl
4. Ensure top navigation bar is the correct height on Mac Electron
5. Ensure top navigation bar is the correct height in browsers.
6. Ensure the insert checklist keyboard shortcut works with either `Ctrl+Shift+C` or `Cmd+Shift+C` in Mac Browsers

### Release

- Update keyboard shortcut keys to reflect correctly based on platform

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -177,7 +177,7 @@ class AppComponent extends Component<Props> {
       'note-info-open': showNoteInfo,
       'navigation-open': showNavigation,
       'is-electron': isElectron,
-      'is-macos': isMac,
+      'is-macos': isElectron && isMac,
     });
 
     return (
--- a/lib/dialogs/keybindings/index.tsx
+++ b/lib/dialogs/keybindings/index.tsx
@@ -143,7 +143,7 @@ export class AboutDialog extends Component<DispatchProps> {
                   </li>
                 )}
                 <li>
-                  <Keys keys={[CmdOrCtrl, 'Shift', 'P']}>
+                  <Keys keys={['Ctrl', 'Shift', 'P']}>
                     Toggle Markdown preview
                   </Keys>
                 </li>
--- a/lib/dialogs/keybindings/index.tsx
+++ b/lib/dialogs/keybindings/index.tsx
@@ -2,7 +2,7 @@ import React, { Component, Fragment } from 'react';
 import { connect } from 'react-redux';
 import Dialog from '../../dialog';
 import { closeDialog } from '../../state/ui/actions';
-import { isElectron, isMac } from '../../utils/platform';
+import { CmdOrCtrl, isElectron } from '../../utils/platform';
 
 import * as S from '../../state';
 
@@ -40,7 +40,6 @@ const Keys = ({
 export class AboutDialog extends Component<DispatchProps> {
   render() {
     const { closeDialog } = this.props;
-    const CmdOrCtrl = isMac ? 'Cmd' : 'Ctrl';
 
     return (
       <div className="keybindings">
@@ -143,7 +142,7 @@ export class AboutDialog extends Component<DispatchProps> {
                   </li>
                 )}
                 <li>
-                  <Keys keys={['Ctrl', 'Shift', 'P']}>
+                  <Keys keys={[CmdOrCtrl, 'Shift', 'P']}>
                     Toggle Markdown preview
                   </Keys>
                 </li>
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -70,7 +70,7 @@ export class Auth extends Component<Props> {
 
     return (
       <div className={mainClasses}>
-        {isMac && <div className="login__draggable-area" />}
+        {isElectron && isMac && <div className="login__draggable-area" />}
         <SimplenoteLogo />
         <form className="login__form" onSubmit={this.onSubmit}>
           <h1>{buttonLabel}</h1>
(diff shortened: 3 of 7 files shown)


lib/app.tsx
```diff
@@ -177,7 +177,7 @@ class AppComponent extends Component<Props> {
       'note-info-open': showNoteInfo,
       'navigation-open': showNavigation,
       'is-electron': isElectron,
-      'is-macos': isMac,
+      'is-macos': isElectron && isMac,
```
belcherj, line 180: Why add the catch for isElectron here?

sandymcfadden: From what I can tell the only place we are really using the `is-macos` class in styles is meant for the Mac Electron app around the size of the note toolbar and menu toolbar.

codebykat:
This would also fix #2549

lib/dialogs/keybindings/index.tsx
```diff
@@ -143,7 +143,7 @@ export class AboutDialog extends Component<DispatchProps> {
                   </li>
                 )}
                 <li>
-                  <Keys keys={[CmdOrCtrl, 'Shift', 'P']}>
+                  <Keys keys={['Ctrl', 'Shift', 'P']}>
                     Toggle Markdown preview
```
codebykat: Is this change meant to address the fact that the Cmd shortcut is bound to something else in the browser? It should still display as Cmd (and this works on Electron).

In fact that reminds me why I'm uncertain about this fix, right now Cmd doesn't work in the browser for some shortcuts, but Ctrl is supported is well (you can use either). So on the Mac you can use Ctrl as a backup, and in the browser you often have to.

sandymcfadden: Ah I see. The only one in the list that I saw that wouldn't work with Cmd in the browser was the Toggle Markdown preview, which is why I updated it to be only `Ctrl`.

I noticed this because on Mac `Ctrl + Shift + C` doesn't work for inserting the checklist, only `Cmd + Shift + C` does.

sandymcfadden: @codebykat what do you think the right approach here is?
Should we work to have the insert checklist work with `Ctrl + Shift + C` and only list keyboard shortcuts with the Ctrl key?
Or should we show mostly Cmd key and Ctrl only where Cmd doesn't work?

codebykat: I think the insert checklist should respond to both/either, if possible (I believe that's what the other shortcuts do).

I'm still not sure about how best to document them, to be honest. I wonder if it wouldn't be easier to just document with `CmdOrCtrl` on the Mac and just Ctrl on Windows.

codebykat: The problem being that in the browser, different sets of them work with Cmd (depending on whether the browser overrides them with something else). They all work on Electron.

sandymcfadden: Thanks, @codebykat I've just committed a change that allows inserting the checklist with Ctrl or Cmd + Shift + C on Mac in browser.

For documenting I just want to be sure I understand how you think it should work. 
- On Electron Mac document all keyboard shortcuts using Cmd. 
- On Mac Browser document all keyboard shortcuts using Ctrl.
- On other platforms document all keyboard shortcuts using Ctrl. 

Do I have that right?

There also seem to be some other keyboard shortcuts in the editor which currently only respond to Cmd on Mac browser [like this one](https://github.com/Automattic/simplenote-electron/blob/develop/lib/note-content-editor.tsx#L752). Do we want those to work with Ctrl as well? My first thought was that these are a bit different than keyboard shortcuts and more keyboard controls so it would be fine to have them as Cmd only.

sandymcfadden: @SylvesterWilmott Would be curious about your thoughts as well on how we should document keyboard shortcuts across the different platforms.

sandymcfadden: Checked with Sly on this and he agreed that the above way I mentioned above is a good way to go forward. I've updated this now to reflect that.

codebykat (approved):
This tests well for me on Mac. I still think it's weird to display Ctrl in the Mac browser but I need to think about that more (and probably build a table of which shortcuts work where).

pachlava:
[testing] fix/feature verified on 
Dell Latitude 7400 | Win 10 | Simplenote 2.7.0-beta1 
and on 
MBP 16" | macOS 11.2 | Simplenote 2.7.0-beta1

PR #2602, opened 2021-01-27
Refactor tag-list/input components to allow better control of renaming tags

### LINKED ISSUE #2207: Tag renaming has various issues
Various tag renaming issues that we should fix.

### Expected
1. Consistency is all I ask.

### Observed
1. You can rename a tag to nothing in the sidebar (just delete all the characters). This should be disallowed.
2. You can rename a tag to only whitespace in the sidebar. This should be disallowed.
3. We should trim extra spaces from tags. (Trim whitespace from beginning/end and disallow more than one space in a row).
4. ~Adding a tag in the note editor allows spaces, but splits on spaces when you hit comma or enter. That's weird.~ #2607
5. You can rename a tag in the sidebar to include commas, but in the note editor a comma submits the tag.
6. ~You can rename a tag to an existing tag name and now you have a duplicate.~ #1128

See p2XJRt-24v-p2

sandymcfadden:
### Fix

Started pairing with @belcherj to look at issues in #2207 
We ended up refactoring the TagList and TagListInput components. We moved the `onDone` function from the TagList to TagListInput instead. We also converted the TagListInput to be a function component instead of a class.

This allowed us to address:
- You can no longer rename a tag to be blank.
- You can no longer rename a tag to be only whitespace.
- You can [no longer use spaces](https://github.com/Automattic/Simplenote-United/blob/trunk/docs/tags.md) in a tag.
- You can no longer use commas in a tag.

### Test

1. Open tag list.
2. Choose to edit tags.
3. Try to remove all characters from the tag. Ensure the tag resets to the original name if you do.
4. Try to rename a tag to only have whitespace. Ensure the tag resets to the original name if you do.
5. Try to add spaces to a tag. Ensure no space can be added.
6. Try to add a comma to a tag. Ensure no comma can be added. 

### Release

- Update renaming tag functionality to be more consistent in the app and across platforms.

### DIFF
--- a/lib/tag-list/input.tsx
+++ b/lib/tag-list/input.tsx
@@ -38,7 +38,7 @@ export const TagListInput: FunctionComponent<Props> = ({
 }) => {
   const [value, setValue] = useState(tagName);
   const onChange = (event: ChangeEvent<HTMLInputElement>) => {
-    setValue(event.target.value.replace(/\s/g, ''));
+    setValue(event.target.value.replace(/[\s,]/g, ''));
   };
   const onDone = (event: FocusEvent<HTMLInputElement>) => {
     const newTagName = event.target?.value.trim() as T.TagName;
--- a/lib/tag-list/input.tsx
+++ b/lib/tag-list/input.tsx
@@ -1,9 +1,4 @@
-import React, {
-  ChangeEvent,
-  FocusEvent,
-  FunctionComponent,
-  useState,
-} from 'react';
+import React, { FunctionComponent, useState } from 'react';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
 
@@ -20,7 +15,7 @@ type OwnProps = {
 };
 
 type OwnState = {
-  value: string;
+  enteredTagName: string;
 };
 
 type DispatchProps = {
@@ -36,17 +31,17 @@ export const TagListInput: FunctionComponent<Props> = ({
   renameTag,
   tagName,
 }) => {
-  const [value, setValue] = useState(tagName);
-  const onChange = (event: ChangeEvent<HTMLInputElement>) => {
-    setValue(event.target.value.replace(/[\s,]/g, ''));
+  const [enteredTagName, setTagName] = useState(tagName);
+  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTagName(event.target.value.replace(/[\s,]/g, ''));
   };
-  const onDone = (event: FocusEvent<HTMLInputElement>) => {
+  const onDone = (event: React.FocusEvent<HTMLInputElement>) => {
     const newTagName = event.target?.value.trim() as T.TagName;
 
     if (newTagName && newTagName !== tagName) {
       renameTag(tagName, newTagName);
     } else {
-      setValue(tagName);
+      setTagName(tagName);
     }
   };
   const classes = classNames('tag-list-input', 'theme-color-fg', {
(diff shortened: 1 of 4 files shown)


lib/tag-list/input.tsx
```diff
@@ -1,49 +1,78 @@
-import React, { Component, FocusEvent, MouseEvent } from 'react';
+import React, {
+  ChangeEvent,
+  FocusEvent,
+  FunctionComponent,
+  useState,
+} from 'react';
```
dmsnell: take this with a grain of salt since the code already exists before this PR but I find it convenient to reference the React events directly from the `React` object instead of importing them and referencing them in _unqualified_ ways. this is because the DOM platform already defines most of the events with the same name and it can be tricky to distinguish which one is in play or which one is supposed to be in play.

for example, with no import statement `FocusEvent` is already a defined type but it has different properties than `React.FocusEvent`

this has caused enough confusion for me that I have started using `FocusEvent` (and its friends) only when I mean "the DOM event" and `React.FocusEvent` when I am dealing with a React event.

lib/tag-list/input.tsx
```diff
@@ -1,49 +1,78 @@
...
+type Props = OwnProps & DispatchProps;
+
+export const TagListInput: FunctionComponent<Props> = ({
+  editable,
+  isSelected,
+  onClick,
+  renameTag,
+  tagName,
+}) => {
+  const [value, setValue] = useState(tagName);
```
dmsnell: this here is a fairly appropriate place to use `useState`, but I find the name a bit generic. everything is a value. (again this was an existing issue so it's not particularly a problem with this PR, but this PR does offer a chance to clean it up, especially since we _are_ changing the way it's stored)

could we name it something meaningful like `tagName` and `setTagName` or `updateTagName`? maybe even `enteredTagName` or `partialTagName` or `tagFieldValue`?

lib/tag-list/input.tsx
```diff
@@ -1,49 +1,78 @@
...
+  ) : (
+    <button className={classes} onClick={onClick}>
+      {value}
+    </button>
+  );
+};
+
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
+  renameTag: (oldTagName, newTagName) =>
+    actions.data.renameTag(oldTagName, newTagName),
```
dmsnell: we may not need to create a new function here - did you try only passing `renameTag: actions.data.renameTag`?

there's value seeing the arguments, but those are also provided by the type signature and IDE

lib/tag-list/input.tsx
sandymcfadden: That makes a lot of sense. I updated this for both FocusEvent and also ChangeEvent which was done the same way.

lib/tag-list/input.tsx
sandymcfadden: Good call! I've updated the naming for this now.

lib/tag-list/input.tsx
sandymcfadden: That does seem to work fine and looks cleaner. I've updated it now.

sandymcfadden:
Based on [this comment](https://github.com/Automattic/simplenote-electron/pull/2607#pullrequestreview-579104529) on #2607 I'm not sure if we should currently disallow tag names to include commas. MacOS and iOS both allow you to rename tags to include commas and iOS allows you to create new tags with commas in them.

sandymcfadden:
After discussion commas are going to be used as delimiters so this should be all set.

codebykat (approved):
Works great! I even tried to paste in a string with commas and spaces and they were removed 😇 Awesome cleanup 🧹

pachlava:

PR #2603, opened 2021-01-27
Add Checklist icon to note toolbar

sandymcfadden:
### Fix

This addresses part of #2563 and adds a checklist icon to the left of the note toolbar. This allows clicking the icon to start a checklist instead of relying on keyboard shortcuts or electron menu items.

A big part of this is thanks to @codebykat for coming up with a way to communicate with the editor instance from outside its component.

Screen Shot 2021-01-27 at 1 14 00 PM
Screen Shot 2021-01-27 at 1 13 44 PM

### Test

1. In a note click into the note where you want to start a checklist
2. Click the checklist icon in the top note toolbar.
3. Notice a checklist is started.
4. Highlight a few lines in a note.
5. Click the checklist icon again.
6. Notice all selected lines are turned in checklist items.

### Release

- Add checklist icon to note toolbar to allow easily starting checklists.

### DIFF
--- a/lib/note-toolbar/index.tsx
+++ b/lib/note-toolbar/index.tsx
@@ -1,15 +1,17 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
+import { isMac } from '../utils/platform';
 
-import IconButton from '../icon-button';
 import BackIcon from '../icons/back';
+import ChecklistIcon from '../icons/check-list';
+import IconButton from '../icon-button';
 import InfoIcon from '../icons/info';
 import PreviewIcon from '../icons/preview';
 import PreviewStopIcon from '../icons/preview-stop';
 import RevisionsIcon from '../icons/revisions';
-import TrashIcon from '../icons/trash';
 import ShareIcon from '../icons/share';
 import SidebarIcon from '../icons/sidebar';
+import TrashIcon from '../icons/trash';
 import actions from '../state/actions';
 
 import * as S from '../state';
@@ -59,6 +61,7 @@ export class NoteToolbar extends Component<Props> {
       note,
       toggleNoteInfo,
     } = this.props;
+    const CmdOrCtrl = isMac ? 'Cmd' : 'Ctrl';
     return !note ? (
       <div className="note-toolbar-placeholder theme-color-border" />
     ) : (
@@ -81,6 +84,12 @@ export class NoteToolbar extends Component<Props> {
         </div>
         {isOffline && <div className="offline-badge">OFFLINE</div>}
         <div className="note-toolbar__column-right">
+          <div className="note-toolbar__button">
+            <IconButton
+              icon={<ChecklistIcon />}
+              title={`Insert Checklist • ${CmdOrCtrl}+Shift+C`}
+            />
+          </div>
           {markdownEnabled && (
             <div className="note-toolbar__button">
               <IconButton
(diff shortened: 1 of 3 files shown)


lib/note-toolbar/index.tsx
```diff
@@ -81,6 +84,13 @@ export class NoteToolbar extends Component<Props> {
         </div>
         {isOffline && <div className="offline-badge">OFFLINE</div>}
         <div className="note-toolbar__column-right">
+          <div className="note-toolbar__button">
+            <IconButton
+              icon={<ChecklistIcon />}
+              onClick={() => window.dispatchEvent(new Event('toggleChecklist'))}
+              title={`Insert Checklist • ${CmdOrCtrl}+Shift+C`}
```
dmsnell, line 91: I believe we already have a Redux action for adding a checklist. If you use `INSERT_TASK` "the system" should handle the machinery of communicating with the editor for you and then we can avoid the events and listeners

codebykat: If only, if only

[edited to be less unintentionally mysterious: this action sadly only tracks analytics and queues the note for edit, it doesn't communicate back to Monaco]

codebykat: (Sandy asked me how to do this last night and I was like "oh that seems straightforward, surely there's a simple way to--- ....oh.)

dmsnell: was there a problem with the Redux action?

sandymcfadden: It looks like[ Kats message](https://github.com/Automattic/simplenote-electron/pull/2603#discussion_r565535022) could have been edited after the last reply, wanted to comment again pointing to it.

codebykat: Apparently I should not use GitHub comments as a chat client

dmsnell: modified Zawinski's law

> Every program attempts to expand until it can ~read mail~ chat.

codebykat (approved):
I picked @dmsnell's brain about this a bit, here is some copy-pasting:

Dennis:
> one way to explore, not sure if it’d be nice, is to take that event dispatching code from the PR and put it into somewhere like how we have the search-field middleware, possibly rename that file and put them both there.
> fire the redux action, middleware sees action, calls window.dispatchEvent yada yada
> inside the editor we call the editor function directly. outside of the editor we fire the action and the middleware calls the window event which the editor traps
> not the most important since we only have one place for inserting a task, though we could also make the keyboard hotkey call this, call it in tests, etc…

I noted that we would need to add a new action, because right now the editor dispatches `INSERT_TASK` after running the toggle checklist action (so using that same action would cause it to call itself)

Dennis:
> we could change that so the editor calls the action and then all updates happen in response to the Redux action

(I don't love this idea because right now we're handling it with a Monaco command, which doesn't have to bounce out to the rest of the app at all)

Dennis:
> or just leave it for now in the PR and add mega comments

(mega comment added ✅ )

As far as next steps on this PR go, IMO it can be merged as-is. We are only calling this from a single place outside the editor, and the Electron menus already use their own communication channel (all the `appCommand` and `editorCommand` stuff).

If we want to add another action to the middleware, I would advocate for renaming the existing action to clarify that it is triggered *by* the toggle checklist action, rather than triggering it.

sandymcfadden:
Thank you both very much for your thoughts and help on this!

pachlava:

PR #2605, opened 2021-01-28
Fix loading of the analytics preference so it will send events if user has opted in

belcherj:
### Fix

Ensure that we set the analyticsAllowed value on window when page loads and the preferences bucket is ready.

### Test
1. Sign in
2. Have analytics enabled
3. Ensure window.analyticsEnabled is set
4. Refresh
5. Ensure window.analyticsEnabled is set

### DIFF
--- a/lib/state/simperium/middleware.ts
+++ b/lib/state/simperium/middleware.ts
@@ -60,7 +60,6 @@ export const initSimperium = (
     ghostStoreProvider: (bucket) => {
       switch (bucket.name) {
         case 'account':
-        case 'preferences':
           return new InMemoryGhost();
 
         default:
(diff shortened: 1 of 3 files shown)


lib/state/simperium/middleware.ts
```diff
@@ -223,6 +223,12 @@ export const initSimperium = (
       });
     }
   });
+  preferencesBucket.channel.once('ready', () => {
+    dispatch({
+      type: 'REMOTE_ANALYTICS_UPDATE',
+      allowAnalytics: getState().data.analyticsAllowed,
```
dmsnell: I think here we want `allowAnalytics: !!(await preferencesBucket.get('preferences-key'))?.analytics_enabled` or whatever it takes to do that. maybe this works too and it won't matter much.

belcherj: Resolved

codebykat (approved):
Do we want this as a hotfix into 2.6.0?

Code looks sensible to me, I did not test it. I'm happy to do so tomorrow, but I empower you to merge before then if you're confident.

PR #2604, opened 2021-01-28
Dependencies: revert to monaco-editor@0.20.0

codebykat:
**Hotfix for 2.6.0**

### Fix

This restores cut/copy/undo/redo functionality, fixes #2600.

Partial revert of #2575. I had to add `monaco-editor` to `package.json` to get it to stop installing the latest, even after removing `node_modules` and `package-lock.json` and downgrading `react-monaco-editor` and `monaco-editor-webpack-plugin`, I'm not 100% sure why, but I think it's actually better to have it explicitly specified in there.

The regenerator-runtime fix doesn't seem to cause any issues on this version so we can leave it in.

We should be able to redo this upgrade after the next `monaco-editor` release (after 0.21.3), as it has been fixed upstream.

### Test

1. Make sure the app builds and there are no console errors
2. Make sure you can type and undo/redo text edits (using menu, keyboard, etc.)

### Release

Release notes updated: Modified the line for the prior PR, now partially-reverted.

PR #2606, opened 2021-01-28
Dependencies: revert to monaco-editor@0.20.0 (#2604)

codebykat:
This is a cherry-pick of #2604 for `develop` so we don't deploy it with broken commands.


codebykat:
#2610

Issue #2589, opened 2021-01-19
Tag input not cleared when switching notes


pachlava:
### Expected
The root cause is that tag is not submitted in Electron as soon as tag editor looses focus (clicked out).

### Observed
See video (make sure to have player controls hidden - they cover the area where fun takes place). It's possible to enter a text as a tag, not submit it, switch between notes, and have this text shown instead of 'Add a tag...'

### Reproduced

1. In any note, start adding a new tag
2. Don't submit a tag
3. Open / create other note ~> the not submitted text will be shown in all notes instead of 'Add a tag...'

https://user-images.githubusercontent.com/73365754/105096505-d8eb2d80-5aaf-11eb-8a2c-d85edd99b1bf.mov

### Where did you see the bug
- System Make: PC | Mac
- System Model: Some DELL laptop | MBP 15"
- OS: Win | macOS
- OS version: Win10 | macOS 10.15.7
- Browser (if applicable): X
- Browser version (if applicable): X
- Simplenote app version: Electron app 2.5.0-beta1

codebykat:
Huh, interesting find. I'm not sure if this is a bug or not. If you've entered something in that text input, it isn't cleared on switching notes. I guess we should probably clear it.

codebykat:
I definitely don't think that we should **submit** any unfinished tag as a tag when you switch notes, that seems a bit unexpected to me.

pachlava:
I agree it's hard to consider this a bug, I'd say this just looks a bit weird. In native and iOS the tag is submitted in this case though...

sandymcfadden:
I was taking a look at this and agree it feels weird the way it is. Working as native and iOS do makes sense to me. If I have typed a tag on a note, but not submitted, and then click out of the tag field it seems like that tag should be submitted. To me clicking away means I'm done inputting now.

codebykat:
When in doubt, making it consistent with the other platforms is usually a good bet.

pachlava:
[testing] fix/feature verified on Dell Latitude 7400 | Win 10 | Simplenote 2.7.0-beta1

PR #2607, opened 2021-01-28
Update tag submission from note editor

sandymcfadden:
### Fix

When adding tags from the note editor you can add spaces and then when you press enter or a comma tags are submitted and separated on the spaces. This is brought up in #2207.

In addition when entering a tag if you switch to another note, or click out of the tag input area the unfinished tag stays as described in #2589 

This PR will submit the new tag when you enter a space or leave the tag input area.

### Test

1. Open a note.
2. Enter a tag from the tag input at the bottom.
3. Try to type a space character.
4. Notice the tag is added instead of a space character.
5. Enter a tag but don't submit.
6. Click into the note above or to switch to a different note
7. Notice the tag is added to the original note.

### Release

- Fix tag input from the note editor to insert tags when a space is used or when clicking outside of the input area.

### DIFF
--- a/lib/tag-input/index.tsx
+++ b/lib/tag-input/index.tsx
@@ -13,6 +13,7 @@ import type * as T from '../types';
 
 const KEY_TAB = 9;
 const KEY_ENTER = 13;
+const KEY_SPACE = 32;
 const KEY_RIGHT = 39;
 const KEY_COMMA = 188;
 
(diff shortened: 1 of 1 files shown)


codebykat (approved):
Great fix!

Is it correct that we use comma as a submit key? I haven't checked what the other platforms do. Are commas supposed to be allowed in tag names? 🤔

sandymcfadden:
Thanks @codebykat! I've taken a look at native MacOS and iOS and they seem to have different behaviors. On MacOS entering a comma submits the tag, but also allows you to rename tags to include commas. On iOS you can directly add a tag with a comma and also rename tags to have commas included. This will have an impact on #2602 as well.

sandymcfadden:
Based on some internal discussion we are going to continue to use comma as a separator for tags now.

pachlava:

PR #2608, opened 2021-01-28
Update tag input styling

### LINKED ISSUE #2564: UI spring cleaning: tag input
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5303183bb6483b4e9c8
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc59220aab754ed6ecd18

* (x) on hover tag pill
* change Add a tag... to Add tag...
* verify color values

Screen Shot 2021-01-10 at 10 47 45 PM

Screen Shot 2021-01-10 at 10 48 23 PM

sandymcfadden:
### Fix

This PR resolves #2564.

Addresses the following changes:
- Updates placeholder text to be "Add tag...".
- Updates colors for both light and dark mode.
- Adds a cross icon indicator when you hover over an added tag to better show this will delete the tag from the note.

Screen Shot 2021-02-01 at 11 14 15 AM

Screen Shot 2021-02-01 at 11 03 02 AM

### Test

1. Open a note.
2. Add at least one tag.
3. Hover over tag.
4. Notice (x) indicator appears.
5. Ensure clicking removes tag from note.
6. Verify colors are correct in both dark and light mode.

### Release

- Added delete tag indicator when hovering over tags in note editor and updated styles.

### DIFF
--- a/lib/components/tag-chip/__snapshots__/test.tsx.snap
+++ b/lib/components/tag-chip/__snapshots__/test.tsx.snap
@@ -6,5 +6,20 @@ exports[`TagChip should not introduce visual regressions 1`] = `
   data-tag-name="spline"
 >
   spline
+  <span
+    className="remove-tag-icon"
+  >
+    <svg
+      className="icon-cross-small"
+      height="22"
+      viewBox="0 0 22 22"
+      width="22"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        d="M16.707 6.707l-1.414-1.414L11 9.586 6.707 5.293 5.293 6.707 9.586 11l-4.293 4.293 1.414 1.414L11 12.414l4.293 4.293 1.414-1.414L12.414 11z"
+      />
+    </svg>
+  </span>
 </div>
 `;
--- a/lib/components/tag-chip/index.tsx
+++ b/lib/components/tag-chip/index.tsx
@@ -1,4 +1,5 @@
 import React, { FunctionComponent } from 'react';
+import SmallCrossIcon from '../../icons/cross-small';
 import classNames from 'classnames';
 
 import type * as T from '../../types';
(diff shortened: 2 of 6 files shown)


codebykat (changes requested):
Looks and tests great!

I noticed one oddity that seems to have been introduced in this PR. If you position the cursor in the tag box and hit backspace, it highlights the last tag with the X icon. You can then type in the tag box, and if you hit backspace it will delete that tag ~rather than~ **in addition to** backspacing in your inputted text.

For example:
1. Type a tag named "test" and use comma or space to add it to the note
2. Hit backspace once. An X will show over tag "test".
3. Type something else, like "atypoedtagg" in the tag field.
4. Press backspace with the intent of deleting the extra "g". Tag "test" will be deleted ~rather than the text you've typed~ Update: In addition to the text you typed!.

Screen Shot 2021-02-04 at 2 25 37 PM
Screen Shot 2021-02-04 at 2 25 40 PM


There's also something up with the tag suggestions that I don't quite grok. If I type the letter "a" I get the suggestion "anewtag", but if I hit space it adds "tag23". If I type the letter "b" it suggests "butno", but a space or comma adds a tag just called "b" (I think this is the correct behavior?)

Screen Shot 2021-02-04 at 2 23 54 PM
Screen Shot 2021-02-04 at 2 23 57 PM

This second one is also happening on develop.

sandymcfadden:
Thanks @codebykat. The first one seems to have been around, I could duplicate it in the current live version. I've taken a look at resolving it though in my latest commit. I'm not confident on what `hiddenTag` is. Maybe we can look at this together? Specifically [here in the unselect](https://github.com/Automattic/simplenote-electron/blob/develop/lib/tag-field/index.tsx#L221), it always seems to be undefined and therefore not ever unselecting the tag?

For the second issue you mentioned, I'm not able to duplicate that at all. Maybe we could look at both together.

codebykat (approved):
Welp, I don't know what the hidden tag is either. It is creating an invisible `input` element, but it doesn't look like anything ever gets set in it?

Your update does seem to have fixed the first issue I was seeing, and I'm suspecting that the second has something to do with messed up tags in my test account (due to renaming, duplicates, etc.). I'm seeing it on production too anyway so I don't think it has anything to do with this PR.

Issue #2579, opened 2021-01-14
Mac dictation results in duplicated text using Firefox in web app


sandymcfadden:
When looking at https://github.com/Automattic/simplenote-electron/issues/1567 I tried to test to see if it was something with voice to text in general. I tried to use the built in Mac dictation option.

### Expected

I expected as I spoke the text would appear in the note.

### Observed

The text did appear. As I finished with the text I wanted to input the text would suddenly duplicate. 

### Reproduced

1. Using a Mac start a new note
2. Press the FN key twice to start dication
3. When finished click done on the dictation control or return/enter on the keyboard.
4. Notice the text which had appeared in the note is duplicated.

![Screen Capture on 2021-01-14 at 10-44-48](https://user-images.githubusercontent.com/1326294/104606068-ad9eb380-5655-11eb-9be3-385520dfee62.gif)

### Where did you see the bug
- System Make: Mac
- System Model: MacBook Pro
- OS: MacOS Catalina
- OS version: 10.15.7
- Browser (if applicable): Firefox
- Browser version (if applicable):  84.0.2
- Simplenote app version: Web

codebykat:
I suspect this to be the same issue as our old friend #2394, I guess we'll see if they ever release that upstream fix.

Issue #2327, opened 2020-09-07
Duplicated tildes (áá éé, etc.) on Firefox and Seamonkey on Crunchbang++ Linux (Openbox)


erricgunawan:
### Expected
There should just be a single character and it's not duplicated.

### Observed
This bug occurs only in Firefox and Seamonkey but not in Google Chrome, and only, as far as it's known, in Crunchbang++ Linux (Openbox).

In this particular condition (Crunchbang++ and Firefox), typing letters with a tilde, as in `adiós`, duplicate those special
characters: ``adióós``.

It does, work, however, in other Linux Operating Systems such as Linux Mint and Ubuntu.

A particular thing to notice with the input method of Crunchbang is that these special characters need two keystrokes (\` and a for á) and when typing the first character, "`" appears, and then it is replaced with "á" after pressing the a. This may confuse the script that receives the text.

### Where did you see the bug
- System Make: -
- System Model: -
- OS: Linux
- OS version: Crunchbang++ (Openbox)
- Browser (if applicable): Firefox
- Browser version (if applicable): (confirming) 
- Simplenote app version: (confirming)

leogermani:
It also happens on Mac OS. I just opened a duplicate issue now (closing #2330)

I'd like to add that this makes the app Unusable!!!

codebykat:
Thanks for the reports; we're working on this upstream here: https://github.com/microsoft/vscode/issues/106392

codebykat:
Reported in 3562935-zen

> when typing special characters, like "á" (éíóú), "^", "—" ... they are automatically duplicated.
> I guess it's not a big issue for english users, but for spanish it's really annoying.

leogermani:
Yes, please fix this. It's not just annoying, it's almost unusable!

dmsnell:
thanks to the work that @belcherj did this is now fixed upstream in Monaco and as soon as we update that dependency (not entirely sure if it's out in the release yet as it was merged at the end of November) Simplenote should also have this fixed.

this was broken everywhere Monaco was in use and will now be fixed everywhere across the internet because it was a problem for Simplenote 😄

codebykat:
Will be finally fixed in 2.7.0!

pachlava:
@codebykat I'm not sure if it's a separate thing or a side effect of the fix. When I try entering à via long hold on 'a' key, and then pressing number button in the popup, it results in 'aà' in Firefox (macOS):


[...]

PR #2611, opened 2021-01-29
Chore: Update monaco editor (again), to 0.22.0

codebykat:
### Fix

This was just released and includes the fix for the missing action triggers from 0.21.x (see #2604) AND the fix for the double emoji insertion in Firefox (#2327) !!!

I had to explicitly import `Range` to avoid an error, this needs VERY THOROUGH testing in case anything else has been subtly broken. There are literally thousands of vscode commits in the past couple of versions.

### Release

Updated Monaco editor to 0.22.0


sandymcfadden:
So far I've gone through and testing this in the web on Mac, Windows, and Linux. I also tested in the Mac Electron build.

The things I tested were:
- Search highlighting
- Note linking
- External linking
- Lists
- Task items
- Keyboard shortcuts
- Editor keyboard commands

All seem to be working well save a few edge case keyboard shortcuts like
- Ctrl + Shift + U Doesn't work on Linux
- Ctrl + Shift + P Doesn't work on Linux Firefox

Hoping to test the Linux and Windows builds as well but so far having issues getting the Linux build to complete in CircleCI so I can't download the artifacts.

sandymcfadden:
It also looks like this resolves #2579.

codebykat:
I merged in the compression format change from #2620 and this seems to build on Linux now! Props @sandymcfadden for that fix

codebykat:
I downloaded the .deb file and got this installed on Debian (Parallels), smoke-tested, it seems to work fine!

sandymcfadden (approved):
Tested this out again on Linux and Windows builds and everything looks good.
Thanks for getting these updates made!

PR #2618, opened 2021-02-01
Update behavior and view when adding new note from trash

### LINKED ISSUE #2612: Disallow adding new note when in trash.
Currently if you enter the trash adding a new note is confusing.

### Expected
When in trash you shouldn't be able to add a new note. 

### Observed
The button in the toolbar is more faded, but still changes to a pointer when you hover and gives the keyboard shortcut to add a new note. If you click the button nothing happens.

If you use the keyboard shortcut it creates a new note. Changes the heading from Trash to All Notes, but still shows the notes which are in the trash.

If the trash is empty you get a create new note link which behaves the same as above.

### Reproduced

1. Open Simplenote
2. Click to enter the Trash
3. Empty trash and click add new note link or use keyboard shortcut `Ctrl + Shift + i` to insert a new note
4. Notice the heading change
5. Notice the new note doesn't appear

https://user-images.githubusercontent.com/1326294/106280264-10767a00-6214-11eb-980d-300128683e68.mov

### Where did you see the bug
This really applies everywhere for Web/Electron

- MacBook Pro 16-inch 2019
- OS: MacOS Catalina
- OS version: 10.15.7
- Simplenote app version: 2.5.0

### LINKED ISSUE #2627: UI Spring cleaning: Update empty note list screens
There are four scenarios where the note list will be empty. These views need to be updated.

- All notes, but no notes are added.
- Trash, but trash is empty.
- Clicked on a tag, but no notes have that tag.
- Search, but no notes match the query.

https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a894f2aa69a20f6e8011b
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a89507b48133e4141b077
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a8950fedeec29013c1bf1
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a89526aeb5c21d9971c79
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a89511bd494221930a53a
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a8951eb530a215ea61748
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a8951cc821e25a7a1b113
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/601a895275979a26b2597049

Screen Shot 2021-02-04 at 3 05 51 PM
Screen Shot 2021-02-04 at 3 05 42 PM
Screen Shot 2021-02-04 at 3 05 35 PM
Screen Shot 2021-02-04 at 3 05 23 PM
Screen Shot 2021-02-04 at 3 05 07 PM
Screen Shot 2021-02-04 at 3 04 56 PM
Screen Shot 2021-02-04 at 3 04 48 PM
Screen Shot 2021-02-04 at 3 04 37 PM

sandymcfadden:
### Fix

Currently when you are viewing the trash adding a new note is confusing.
This PR resolves #2612 by updating the view to all notes when adding a note from trash.

It also resolves #2627 by updating the views which appear when there are no notes in a list.

no-tags
no-search-results
empty-trash
empty-note-list

### Test

1. Visit trash screen.
2. Notice New Note icon appears disabled.
3. Clicking New Note icon does not create new note.
4.  Pressing keyboard shortcut `Cmd or Ctrl + Shift + i` does create a new note
5. After note is created you have the all notes view.
6. Remove all notes from trash.
7. Notice there is no link to create a new note.

Additional testing

1. Visit All Notes
2. Trash all notes
3. Notice updated empty note list view
4. Search for a non existent term
5. Notice updated empty note list view
6. Visit a tag which have no notes
7. Notice updated empty note list view

### Release

- Update the messaging and view when no notes show in a specific list.
- Fixes the view when adding a new note when viewing the trash.

### DIFF
--- a/lib/menu-bar/index.tsx
+++ b/lib/menu-bar/index.tsx
@@ -9,6 +9,7 @@ import { connect } from 'react-redux';
  */
 import { CmdOrCtrl } from '../utils/platform';
 import IconButton from '../icon-button';
+import { isMac } from '../utils/platform';
 import NewNoteIcon from '../icons/new-note';
 import MenuIcon from '../icons/menu';
 import { withoutTags } from '../utils/filter-notes';
@@ -48,6 +49,8 @@ export const MenuBar: FunctionComponent<Props> = ({
     : openedTag
     ? 'Notes With Selected Tag'
     : 'All Notes';
+
+  const CmdOrCtrl = isMac ? 'Cmd' : 'Ctrl';
 
   return (
     <div className="menu-bar theme-color-border">
--- a/lib/note-list/no-notes.tsx
+++ b/lib/note-list/no-notes.tsx
@@ -1,22 +1,26 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
+
+import NotesIcon from '../icons/notes';
+import TagIcon from '../icons/tag';
+import TrashIcon from '../icons/trash';
 
 import actions from '../state/actions';
 
 import * as S from '../state';
 
+type EmptyNoteListPlaceholder = {
+  message: String;
+  icon: JSX.Element | null;
+  button: JSX.Element | null;
+};
+
 const NoNotes = () => {
   const hasLoaded = useSelector((state: S.State) => state.ui.hasLoadedNotes);
   const searchQuery = useSelector((state: S.State) => state.ui.searchQuery);
+  const showTrash = useSelector((state: S.State) => state.ui.showTrash);
+  const openedTag = useSelector((state: S.State) => state.ui.openedTag);
   const dispatch = useDispatch();
-
-  const getMessage = () => {
-    return hasLoaded
-      ? searchQuery.length
-        ? 'No Results'
-        : 'No Notes'
-      : 'Loading Notes';
-  };
 
   const getButton = () => {
     return searchQuery.length ? (
@@ -29,15 +33,54 @@ const NoNotes = () => {
       </button>
     ) : (
       <button onClick={() => dispatch(actions.ui.createNote())}>
-        Create a new note
+        Create your first note
       </button>
     );
   };
 
+  const placeholderInfo = ({
+    searchQuery,
+    openedTag,
+    showTrash,
+  }): EmptyNoteListPlaceholder => {
+    if (searchQuery.length > 0) {
+      return { message: 'No Results', icon: null, button: getButton() };
+    }
+
+    if (openedTag !== null) {
+      return {
+        message: `No notes tagged "${openedTag}"`,
+        icon: <TagIcon />,
+        button: null,
+      };
+    }
+
+    if (showTrash) {
+      return {
+        message: 'Your trash is empty',
+        icon: <TrashIcon />,
+        button: null,
+      };
+    }
+
+    return {
+      message: '',
+      icon: <NotesIcon />,
+      button: getButton(),
+    };
+  };
+
+  const { message, icon, button } = placeholderInfo({
+    searchQuery,
+    openedTag,
+    showTrash,
+  });
+
   return (
     <div className="note-list-placeholder theme-color-fg">
-      {getMessage()}
-      {hasLoaded && getButton()}
+      <div className="no-notes-icon">{icon}</div>
+      <div className="no-notes-message">{message}</div>
+      <div className="no-notes-button">{hasLoaded && button}</div>
     </div>
   );
 };
--- a/lib/search/index.ts
+++ b/lib/search/index.ts
@@ -364,6 +364,11 @@ export const middleware: S.Middleware = (store) => {
       }
 
       case 'CREATE_NOTE_WITH_ID':
+        searchState.showTrash = false;
+        searchState.notes.set(action.noteId, toSearchNote(action.note ?? {}));
+        indexNote(action.noteId);
+        queueSearch();
+        return next(action);
       case 'IMPORT_NOTE_WITH_ID':
       case 'REMOTE_NOTE_UPDATE':
       case 'RESTORE_NOTE_REVISION':
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -37,6 +37,7 @@ type StateProps = {
   showEmailVerification: boolean;
   showNavigation: boolean;
   showNoteInfo: boolean;
+  showTrash: boolean;
   theme: 'light' | 'dark';
 };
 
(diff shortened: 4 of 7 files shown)


lib/menu-bar/index.tsx
```diff
@@ -48,6 +49,9 @@ export const MenuBar: FunctionComponent<Props> = ({
     ? 'Notes With Selected Tag'
     : 'All Notes';
 
+  const CmdOrCtrl = isMac ? 'Cmd' : 'Ctrl';
+  const title = showTrash ? '' : `New Note • ${CmdOrCtrl}+Shift+I`;
```
dmsnell: considering that this title is specific to the new note button I know I would find it more straightforward if we put the value inside of the JSX itself so I don't have to wander in order to find it.

```ts
title={showTrash ? '' : `New Note…`}
```

otherwise if we keep it here let's try picking a more specific name than `title` since there are multiple "title" values in mind with the menu bar. an explanatory comment would help considerably too since my first impressions were that we were doing strange things in this component vs. in its parent, until I realized this has to be this way.

```ts
// when the trash is open we want to disable creating new notes
// but we want the icon to remain, so hide the title to make it disappear
const newNoteTitle = showTrash ? … : …;
```

that being said, is there a reason we need to hide the button title?
the history button has a disabled state that adds "unavailable" to explain why the button is disabled. is there a reason not to just always show the title with or without an explanation? seems like it could be confusing to someone trying to figure out what that button is and why it's disabled.

sandymcfadden: You're right @dmsnell it makes a lot more sense to keep the title in the jsx. When I added the `CmdOrCtrl` I thought at first it may be cleaner to move the title by it as well. 

>that being said, is there a reason we need to hide the button title?

I should have followed the example of the history button. I've done that now leaving a title but showing unavailable. Doing that and moving it back into the jsx I think makes it more understandable without the need for comments, but I can add that as well if you think it would still help.

lib/note-list/no-notes.tsx
```diff
@@ -1,43 +1,62 @@
...
-      : 'Loading Notes';
-  };
-
-  const getButton = () => {
-    return searchQuery.length ? (
+  const getSearchButton = () => {
+    return (
       <button
         onClick={() =>
           dispatch(actions.ui.createNote({ content: searchQuery }))
```
dmsnell: 🤨 nothing like deep coupling - how'd that get in there?

lib/search/index.ts
```diff
@@ -364,6 +364,11 @@ export const middleware: S.Middleware = (store) => {
       }
 
       case 'CREATE_NOTE_WITH_ID':
+        searchState.showTrash = false;
+        searchState.notes.set(action.noteId, toSearchNote(action.note ?? {}));
+        indexNote(action.noteId);
+        queueSearch();
+        return next(action);
```
sandymcfadden, line 371: I'm not confident this is the best way or best place to do this. It seems to work but would love thoughts on if there is a better way. @dmsnell

lib/note-list/no-notes.tsx
sandymcfadden: Yeah. I'm not sure if my approach to this refactoring of the empty note list views is the best way to go. Any thoughts would be much appreciated.

dmsnell: best policy is to not change code styling in an unrelated PR so I wouldn't recommend doing anything differently. that being said I occasionally point out things in code review that are unrelated but which stand out to me. this is one of those because I saw `dispatch` inside a UI component, which is something that sets off all the alarms in my head

lib/search/index.ts
dmsnell: yeah this is something we have to be careful about since we're changing the search property for trash. we would only want to do this if we know that the `showTrash` reducer is also turning off the trash view when creating a note

sandymcfadden: Thanks @dmsnell If I understand things properly it does look like we are doing that [here](https://github.com/Automattic/simplenote-electron/blob/update/disallow-creating-note-from-trash/lib/state/ui/reducer.ts#L305). Do I have that right?

dmsnell: yep that's right 👍

lib/note-list/no-notes.tsx
```diff
@@ -1,43 +1,62 @@
...
+    case showTrash:
+      message = 'Your trash is empty';
+      icon = <TrashIcon />;
+      button = '';
+      break;
+    default:
+      message = 'Create your first note';
+      icon = <NotesIcon />;
+      button = '';
+  }
```
dmsnell: this is clever, but we don't typically use the pattern of `switch (true)` and do'm reluctant to introduce it. it's definitively not what the purpose of the structure is for, plus this relies on mutating variables.

a simple approach to this scenario is to create a separate function and use `return` statements to abort or pick the values. a separate function also gives the benefit of making it easier to test the behavior

```ts
const placeholderInfo = ({searchQuery, openedTag, showTrash}) => {
	if (searchQuery.length > 0) {
		return {message: 'No Results', icon: null, button: getSearchButton()}
	}

	if (openedTag != null) {
		return {message: `No notes tagged "${openedTag}"`, icon: <TagIcon />, button: ''};
	}

	if (showTrash) {
		return {message: 'Your trash is empty', icon: <TrashIcon />, button: ''};
	}

	return {message: 'Create your first note', icon: <NotesIcon />, button: ''};
}

const {message, icon, button} = placeholderInfo({searchQuery, openedTag, showTrash});
```

adding types gives us additional safety, but I did not introduce those in this example

sandymcfadden: This makes a lot of sense. I've updated to use this approach and added some types as well.

codebykat (changes requested):
This looks good and works well!

One small request. [Similar to iOS](https://github.com/Automattic/simplenote-ios/issues/1132), I think the "Create your first note" text should be a link that creates a note.

sandymcfadden:
Good catch, thanks @codebykat. I followed the designs, but that makes sense to have it as a link. I have it updated now.

codebykat (approved):
Works great 👍

lib/note-list/no-notes.tsx
```diff
@@ -29,15 +28,60 @@ const NoNotes = () => {
...
+        Create your first note
       </button>
     );
   };
 
+  type EmptyNoteListPlaceholder = {
+    message: String;
+    icon: JSX.Element | null;
+    button: JSX.Element | null;
+  };
```
dmsnell: let's keep our type definitions in the module-global scope and towards the top of the module, under the imports

that will help avoid "hunting" for them in the code

sandymcfadden: Thanks! I moved that now.

PR #2619, opened 2021-02-02
Email verification: lowercase username from verification token

codebykat:
**2.6.0 Hotfix**

### Fix

This prevents email verification from erroneously thinking an account is not verified if we receive a mixed-case username from Simperium.

n.b. We are already forcing usernames to lowercase on login, so we only need to apply this transform to the username from the Simperium token. https://github.com/Automattic/simplenote-electron/blob/e805fd665efbcb34576e8a66d78cc80d4e7e9730/lib/boot-without-auth.tsx#L63

### Test

I'm not sure how to get a mixed-case email address in the system, possibly by creating an account in iOS? But either way...

1. Log in with an unverified account and make sure it still prompts for verification
2. Verify, or log in with a verified account, and make sure the verification dialog is not shown

### Release

Updated email verification logic to handle a mixed-case email address.


codebykat:
Closing - we know that Simperium is lowercasing emails, and we're already lowercasing them on login/signup, so we don't need to take action here.

PR #2623, opened 2021-02-03
Updating icon pack with newest versions and removing unused icons

### LINKED ISSUE #2539: UI spring cleaning: new icon pack
New icon pack is here https://github.com/Automattic/Simplecons
All icons we use will need to be updated to the newest versions there.

sandymcfadden:
### Fix

As part of the UI Spring cleaning this updates all the used icons to the latest version.
Resolves #2539.
There were a few older icons that are not used in the app anywhere so they were removed. 

### Test

1. Smoke-test in both light and dark mode and ensure icons look consistent and the correct sizes. 

### Release

- Updated the icon set used to the latest versions.


codebykat (commented):
I'll leave it to @SylvesterWilmott to approve, but I gave the code and the app a once-over (both light and dark themes) and didn't see anything amiss!

lib/search-field/style.scss
```diff
@@ -22,6 +22,10 @@
     flex: 0 0 auto;
   }
 
+  .icon-search-small {
+    max-width: 16px;
+  }
```
codebykat: I think what we actually want here is a `flex: 0 0 auto` which will tell it not to resize itself, but if you tried that and it didn't work, ignore me.

lib/note-list/style.scss
```diff
@@ -114,6 +114,11 @@
     color: transparent;
     height: 16px;
     width: 16px;
+
+    .icon-pinned-small {
+      height: 16px;
```
codebykat: Again - these sizes should inherit properly from the parent, it might need a `flex: 0 0 auto` to say "do not squish or stretch me"

codebykat: (and on the parent, probably, she says without testing it)

SylvesterWilmott:
Looking good.

Couple of things:

- In the search bar, the icon used for clearing search should be the 16px `cross.svg`
- In settings > tools Import/export notes, these icons should be 16px `chevron-right.svg`

lib/search-field/style.scss
sandymcfadden: I tried this and a few variations but still couldn't get the sizing right. It did lead me to look at what I think is a better option for setting the size instead of having to do it in multiple places. I've updated the approach now.

lib/note-list/style.scss
sandymcfadden: I couldn't get this to work, I'm likely missing something, but I did come up with what I think is a better way to set the sizes and have updated it with that now.

sandymcfadden:
Thanks @SylvesterWilmott I've added the chevron right icon now to be used in that settings > tools area.
The clearing search icon does seem to be the 16px `cross.svg`. Maybe I'm misunderstanding or missing something?

SylvesterWilmott:
>The clearing search icon does seem to be the 16px cross.svg.

Ah yeah, you are right. The icon is the correct one, but it's being displayed larger than it's natural size.
Screenshot 2021-02-04 at 12 15 41

sandymcfadden:
Ah thank you @SylvesterWilmott  I think the latest styling updates I committed should resolve that. Are you still seeing it? If so what browser etc are you using so I can test and fix there too.

SylvesterWilmott:
After pulling the latest changes the cross icon is appearing at the correct size.

Thanks @sandymcfadden

SylvesterWilmott (approved):

codebykat (approved):
Great update! :shipit:

pachlava:
[testing] fix/feature verified indirectly by running smoke test and seeing no issues with icons. Dell Latitude 7400 | Win 10 | Simplenote 2.7.0-beta1

Issue #2560, opened 2021-01-11
UI spring cleaning: notes list


codebykat:
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5d768735855a07651d9
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5e81bc37159baf42a47

- [x] Verify all colors
- [ ] There's a timestamp shown?
- [x] Three lines instead of four for Expanded mode
- [x] Blue highlight on selected note [dark mode]
- [x] Text is lighter on selected note [dark mode]
- [x] Add mouseover state

<img width="1049" alt="Screen Shot 2021-01-10 at 10 33 42 PM" src="https://user-images.githubusercontent.com/52152/104152283-f3c4f000-5393-11eb-9cd9-ae12642ba5a2.png">

<img width="1037" alt="Screen Shot 2021-01-10 at 10 36 15 PM" src="https://user-images.githubusercontent.com/52152/104152398-456d7a80-5394-11eb-885b-cfe8045c4916.png">

codebykat:
~@SylvesterWilmott pointed out that the blue color is currently wrong for the pinned icon as well. Should be `blue30` not `blue20`.~ This has been fixed

codebykat:
Another note from Slack:

> We have updated the selection colour in the notes lost in dark mode on macOS from blue50 to blue50 at 40% opacity.

Edit: Included this in #2631

sandymcfadden:
From slack. The idea of a timestamp being displayed was going to be conditional based on what sorting option was selected. If notes were sorted by created date, or modified date we would display the timestamp.
More thought is going to go into this before making any changes to add the timestamp.

codebykat:
Cool - let's close this parent issue for now, then.

PR #2631, opened 2021-02-05
UI spring cleaning: Some small style fixes

codebykat:
### Fix

As per @SylvesterWilmott:

* In the top toolbar the notes list title eg. All Notes should be 16px with 600 weight
* The text in the search bar should be 16px (including placeholder)
* Can we align the search icon in the search bar with the hamburger icon above?
* Can we add weight: 600 to the note titles in the notes list?
* The entire notes list panel should have a new width of 380px

Also as mentioned [here](https://github.com/Automattic/simplenote-electron/issues/2560#issuecomment-771347658):
* We have updated the selection colour in the notes lost in dark mode [...] from blue50 to blue50 at 40% opacity.

### Screenshots

Screen Shot 2021-02-04 at 11 48 06 PM

Screen Shot 2021-02-04 at 11 48 19 PM

Screen Shot 2021-02-04 at 11 48 35 PM

At medium width the note list is still 300px wide:

Screen Shot 2021-02-04 at 11 50 08 PM

Dark mode:

Screen Shot 2021-02-05 at 12 09 35 AM

### Release

Adjusted note list width and font weights

### DIFF
--- a/lib/app-layout/style.scss
+++ b/lib/app-layout/style.scss
@@ -105,7 +105,7 @@
 }
 
 .app-layout__source-column {
-  width: 380px;
+  width: $note-list-width;
   display: flex;
   flex-direction: column;
   flex-shrink: 0;
--- a/lib/app-layout/style.scss
+++ b/lib/app-layout/style.scss
@@ -105,7 +105,7 @@
 }
 
 .app-layout__source-column {
-  width: 328px;
+  width: 380px;
   display: flex;
   flex-direction: column;
   flex-shrink: 0;
--- a/lib/menu-bar/index.tsx
+++ b/lib/menu-bar/index.tsx
@@ -55,7 +55,7 @@ export const MenuBar: FunctionComponent<Props> = ({
         onClick={toggleNavigation}
         title="Menu • Ctrl+Shift+U"
       />
-      {placeholder}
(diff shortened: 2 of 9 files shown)


SylvesterWilmott:
All looks good.

The only thing is that the `font-weight:600` looks a bit too heavy in comparison with the design. I think for the notes list title and note row titles we can use `font-weight:500` instead which looks closer to the designs.

Screenshot 2021-02-05 at 09 41 22

I also noticed that in the `.note-list-item` class we are setting the font to `font-family: 'Simplenote Tasks', sans-serif;` which presumably is the icon font that is handling the checklist icons. However, we should also include the rest of our font stack here since the `sans-serif` is causing the font to be rendered incorrectly in the notes list.

SylvesterWilmott:
I think this could also be a good time to update the font size of tags in a search eg "tag:blog" to 16px as well as changing the height of those rows to 36px. See below for illustration.

Screenshot 2021-02-05 at 09 52 14

codebykat:
Okay, feedback should all be addressed!

Screen Shot 2021-02-05 at 11 17 59 AM

Screen Shot 2021-02-05 at 11 17 45 AM

belcherj (approved):
Tested and works as described.

Issue #2464, opened 2020-11-11
History icon greyed out when the note was created on the Windows app


reginabally:
Reported in 3475633-zen.

> Basically, in a small number of my notes, this History button is greyed out, and the alt text is History (unavailable) when I hover over it.

> I’ve tried a couple of things. If I create a new note in the Android version, history works on that note in the Windows app. If I create a new note in the Windows app, history is unavailable in Windows app but available in the Android app.

### Expected

I would expect the History icon works/clickable for the note created with the Electron app.

### Observed

The History icon greyed out when I created a note with the Electron app. But if I created a note in the Android/iOS apps, I can click on the History icon.

### Reproduced

1. Create a new note with the Electron app.
2. Notice the History icon greyed out.
3. Create a new note with the Android/iOS app.
4. The History icon of the note created in Step 3 can be clicked.
5. Edited the note created in Step 1 with the Android/iOS app, the History icon still greyed out on the Electron app but I was able to access the History of the same note with the Android/iOS app.

![Screen Capture on 2020-11-11 at 11-41-43](https://user-images.githubusercontent.com/40906847/98762842-8d633600-2413-11eb-9cfe-1f7e6ed8bacd.gif)

### Where did you see the bug

*As reported in 3475633-zen*
- OS: Windows
- OS version: 10 Home version 2004
- Simplenote app version: 2.0

*Reproduced with the following device*
- OS: macOS
- OS version: 10.15.7 (19H2)
- Simplenote app version: 2.0

dmsnell:
The history icon is grayed out when the app has no revisions to revert. It grabs history when changing notes. We can improve this by making sure that after a note is open for a certain period of time we grab the history anyway, but so far we have limited this to keep things simple in the code and prevent thrashing the network with requests for note history as it operates differently than the rest of the data in Simplenote.

That is, either there are no revisions or the app hasn't requested them. In the meantime, try switching to another note and back and you should find that the history loads. This change was introduced to eliminate some of the confusion when opening the history panel while revisions were still loading.

reginabally:
Thanks for taking a look, @dmsnell! I tried editing the note I created on the Electron app with the Android app yesterday and switched back to view the note on the Electron app, the History icon still greyed out. When I looked at it again today, I can click on the History icon and see the revisions.

[...]

PR #2633, opened 2021-02-05
[WIP] Only load revisions when history is opened

codebykat:
### Fix

@belcherj started this PR, I think it needs a loading state for the history slider (it also needs a rebase)

Fix for #2464 

To consider: This probably changes offline behavior of revisions (history would no longer be accessible if the note was simply opened)

### Test
 TBD


dmsnell:
can we try and fix the behavior where the history doesn't load before we rip out all the revision persistence and go back to somewhat that was actively confusing folks?

PR #2639, opened 2021-02-08
Bump version to 2.7.0-beta1

mokagio:


package-lock.json
```diff
@@ -14686,6 +14688,26 @@
...
+        "loader-utils": {
+          "version": "2.0.0",
+          "resolved": "https://registry.npmjs.org/loader-utils/-/loader-utils-2.0.0.tgz",
+          "integrity": "sha512-rP4F0h2RaWSvPEkD7BLDFQnvSf+nK+wr3ESUjNTyAGobqrijmW92zc+SO6d4p4B1wh7+B/Jg1mkQe5NYUEHtHQ==",
+          "requires": {
+            "big.js": "^5.2.2",
+            "emojis-list": "^3.0.0",
+            "json5": "^2.1.2"
+          }
+        }
```
mokagio, line 14710: As you can see from the `package.json` diff, I didn't make a change that would have resulted in this.

I'm guessing this came from a previous change that wasn't committed?

belcherj: This is probably due to a minor version update for a package that uses the ^. What do you think we should do? I am happy to merge this down.

belcherj (approved):

package-lock.json
mokagio: I'm fine to merge it, too. Was mainly curious what was behind it.

I now traced it down to `426d4388375c6443530f56775ba6a883b262c4ce`. If I run `npm install` from that commit, I get the same change here.

I wonder why it didn't happen at the time of #2610? 🤔 I've got lots to learn still to wield `npm` properly.

Thanks for the prompt review @belcherj 🙌

PR #2642, opened 2021-02-08
Pin dependencies and remove a duplicate

belcherj:
### Fix

Pin all dependencies and remove `@types/clipboard` from the dependencies since it is already in `devDependencies`.

We should be updating dependencies manually.

### Test
1. npm install
2. Run app

There are no changes here that could break anything unless we were depending on a minor update from one of the dependencies where the hat was removed.


sandymcfadden (approved):
This looks good to me. As you mentioned it pins the dependencies. 
It works well in browser, and Electron builds on Mac, Linux, and Windows.

codebykat (approved):
👍 NO MORE HATS

PR #2645, opened 2021-02-09
UI Spring Cleaning: Navigation bar blue should match note list blue

codebykat:
### Fix

As per @SylvesterWilmott the selected item in the navigation bar / tag list in dark mode should be the same as in the note list, `$studio-simplenote-blue-50` at 40% opacity.
    
Screen Shot 2021-02-08 at 9 28 42 PM


SylvesterWilmott (approved):
Thanks @codebykat

Issue #2616, opened 2021-02-01
UI spring cleaning: Update dark mode surface color palette 


sandymcfadden:
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5db2c14aba30fd54e7149b20
![Screen Shot 2021-02-01 at 10 23 31 AM](https://user-images.githubusercontent.com/1326294/106471127-b675ee80-6477-11eb-94a6-12f08a4d17fa.png)

For dark mode, we have a new color palette to indicate elevation. These colors will need to be updated throughout the app.

PR #2622, opened 2021-02-03
UI Spring Cleaning: Note info panels

codebykat:
### Fix

* Split note info component into two components, NoteInfo and NoteActions.
* Remove publish note info from Share dialog
* Remove redundant icons from the note toolbar

Fixes #2569 

Known issues:
- There's a bug where if you check a checkbox repeatedly. This has to do with updates to the note coming back from GAE. The timing can be off and GAE will send an update basically undoing one of the changes in-app. This isn't something we are going to fix right now as it would involve substantial work in GAE. Thanks, @dmsnell for helping track this down.
- Dark mode colors aren't 100% and will be updated separately. 

**Screenshots**

**Before:**
<details>
<summary>Light Mode</summary>
Screen Shot 2021-03-22 at 9 56 02 AM
</details>
<details>
<summary>Dark Mode</summary>
Screen Shot 2021-03-22 at 9 56 15 AM
</details>

**After:**

**Light Mode**
<details>
<summary>Note info w/ references</summary>
Screen Shot 2021-03-22 at 9 50 56 AM
</details>
<details>
<summary>Note info no references</summary>
Screen Shot 2021-03-22 at 9 51 09 AM
</details>
<details>
<summary>Note Actions</summary>
Screen Shot 2021-03-22 at 9 52 21 AM
</details>

**Dark Mode**
<details>
<summary>Note info w/ references</summary>
Screen Shot 2021-03-22 at 9 50 39 AM
</details>
<details>
<summary>Note info no references</summary>
Screen Shot 2021-03-22 at 9 51 23 AM
</details>
<details>
<summary>Note Actions</summary>
Screen Shot 2021-03-22 at 9 51 56 AM
</details>

### Test

1. Open a note
2. Notice new note toolbar 
3. Click on the new info `(i)` icon
4. Notice the new info dialog 
5. Click the new actions `(...)` icon
6. Ensure all options work as expected
7. Smoke test in light and dark mode ensuring styles are correct

### Release Notes

- Updates the note toolbar and note info sidebar to be two dialogs, one for note info, and one for note actions.

PR #2649, opened 2021-02-10
Update the navigation bar and taglist styles

### LINKED ISSUE #2566: UI spring cleaning: tags list
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5352163646a2d1b37ea
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc5944b88d656997f593a

* Re-style section headers
* Remove blue / verify color values
* background color in dark mode
* Tag list should now appear on top of the content rather than pushing everything over

Screen Shot 2021-01-10 at 10 51 49 PM

Screen Shot 2021-01-10 at 10 53 54 PM

sandymcfadden:
### Fix

As part of the UI spring cleanup project, this PR works to resolve #2566.

This causes the navigation bar to open on top of other content rather than push it to the side. 
It updates the styles and colors used. 
It also adds an overlay to the other content when the navigation bar is open instead of reducing opacity.
It does not address the background color in dark mode as it will be addressed separately in #2616.

### Test

1. Test in both light and dark mode.
2. Ensure colors all look correct.
3. Ensure the navigation bar loads over top of content instead of shifting everything over.
4. Ensure there is an overlay on the rest of the content.

### Release

- Update navigation bar style and have it appear over content rather than push it to the side.

### DIFF
--- a/lib/app-layout/style.scss
+++ b/lib/app-layout/style.scss
@@ -30,9 +30,25 @@
   &.is-navigation-open {
     pointer-events: none;
 
+    &:before {
+      content: '';
+      background-color: rgba($studio-black, 0.2);
+      display: block;
+      height: 100%;
+      left: 0;
+      position: absolute;
+      top: 0;
+      width: 100%;
+      z-index: 1;
+    }
+
     @media only screen and (max-width: $single-column) {
       .app-layout__note-column {
         opacity: 0;
+      }
+
+      &:before {
+        display: none;
       }
     }
   }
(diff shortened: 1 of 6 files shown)


codebykat (approved):
Tests well in light and dark mode on web/Electron. I had a few similar changes in #2622 - I was going to just back them out but realized I can't do it easily so I updated it to match yours for an easier rebase. (I had a `z-index: 1` rather than `2` but I'm not going to quibble about it.)

PR #2652, opened 2021-02-10
Make keyboard shortcuts for searching always active regardless of keyboard shortcut preference

### LINKED ISSUE #2651: Keep Ctrl/Cmd + F keyboard shortcut even when keyboard shortcuts disabled
### What
When using the built-in browser search it will only capture the content in the note editor which is currently visible on the screen. If there is content matching the outside of the view which is then scrolled into view it won't show as selected.

### Why
The browser search isn't supported in the note editor and we should default to our built in search

### How
When keyboard shortcuts are disabled we should still keep the Ctrl/Cmd + F keyboard shortcut to use our search option.

sandymcfadden:
### Fix

This PR resolves #2651.
Currently, the browser's built-in search doesn't work well with the note editor. For content not in view in the editor, the search is not able to find and select matching content.
For this reason, we are going to keep the keyboard shortcut for searching always on regardless of the keyboard shortcut preference.

### Test

1. Disable keyboard shortcuts from Settings > Tools.
2. Ensure Ctrl or Cmd + F take you to the search field in Simplenote instead of browser search.
3. Ensure other keyboard shortcuts are disabled.
4. Enable keyboard shortcuts from Settings > Tools.
5. Ensure all keyboard shortcuts still work.

### Release

- Enable keyboard shortcut for search even when keyboard shortcut preference is set to be disabled.

### DIFF
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -77,7 +77,7 @@ class AppComponent extends Component<Props> {
   handleShortcut = (event: KeyboardEvent) => {
     const { hotkeysEnabled } = this.props;
 
-    // Handle search shortcuts even if keyboardshortcuts are disabled.
+    // Handle search shortcuts even if keyboard shortcuts are disabled.
     if (!window.electron || !isMac) {
       this.handleBrowserSearchShortcut(event);
     }
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -76,6 +76,12 @@ class AppComponent extends Component<Props> {
 
   handleShortcut = (event: KeyboardEvent) => {
     const { hotkeysEnabled } = this.props;
+
+    // Handle search shortcuts even if keyboardshortcuts are disabled.
+    if (!window.electron || !isMac) {
+      this.handleBrowserSearchShortcut(event);
+    }
+
     if (!hotkeysEnabled) {
       return;
     }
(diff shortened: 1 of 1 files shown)


lib/app.tsx
```diff
@@ -76,6 +76,12 @@ class AppComponent extends Component<Props> {
 
   handleShortcut = (event: KeyboardEvent) => {
     const { hotkeysEnabled } = this.props;
+
+    // Handle search shortcuts even if keyboard shortcuts are disabled.
+    if (!window.electron || !isMac) {
```
codebykat: Hah - I was going to comment that this conditional is weird and bad, and then I realized it's a copy of the one I wrote, and it's like this because I never figured out why the shortcuts don't fire on Win/Linux. Nice.

codebykat (approved):
Good change, tests well. I didn't verify on Windows build.

lib/app.tsx
sandymcfadden: Should we maybe change these to `if ( !(window.electron && isMac) )` instead to make it easier to read?

codebykat: I'd rather leave it as-is; ultimately I want to debug why they don't get fired on Win/Linux but it has been knocked off my list by more urgent work. I suppose since we now have it in two places we could conceivably create a new variable, like `shouldHandleBrowserShortcuts = !window.electron || !isMac`. I leave that to your discretion.

PR #2653, opened 2021-02-10
Add a delete tag confirmation

### LINKED ISSUE #2568: Add delete tag confirmation
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbdd884623ca5a8c957125
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbdd887d9c865c6d783840

Screen Shot 2021-01-10 at 10 57 31 PM

Screen Shot 2021-01-10 at 10 57 35 PM

sandymcfadden:
### Fix

Currently, when deleting a tag if you click the delete icon it immediately is deleted. This PR resolves #2568 by adding a confirmation dialog where you must confirm the deletion.

Screen Shot 2021-03-04 at 7 37 48 AM
Screen Shot 2021-03-04 at 7 38 12 AM

### Test

1. Test in both light and dark mode.
2. Open tag list.
3. Click to edit the tags.
4. Click on the trash icon to delete a tag.
5. Notice the dialog appears asking for confirmation.
6. Click the delete button.
7. Notice the tag is deleted.
8. Follow through tests again, but instead of clicking the delete button, close the dialog.
9. Notice the tag isn't deleted.

### Release

- Adds a confirmation dialog before deleting a tag.

### DIFF
--- a/lib/app-layout/index.tsx
+++ b/lib/app-layout/index.tsx
@@ -144,7 +144,7 @@ export class AppLayout extends Component<Props> {
 const mapStateToProps: S.MapState<StateProps> = (state) => ({
   hasRevisions:
     state.ui.showRevisions && state.data.noteRevisions.has(state.ui.openedNote),
-  keyboardShortcutsAreOpen: state.ui.dialogs.includes('KEYBINDINGS'),
+  keyboardShortcutsAreOpen: selectors.isDialogOpen(state, 'KEYBINDINGS'),
   keyboardShortcuts: state.settings.keyboardShortcuts,
   isFocusMode: state.settings.focusModeEnabled,
   isNavigationOpen: state.ui.showNavigation,
--- a/lib/dialog-renderer/index.tsx
+++ b/lib/dialog-renderer/index.tsx
@@ -10,6 +10,7 @@ import KeybindingsDialog from '../dialogs/keybindings';
 import LogoutConfirmation from '../dialogs/logout-confirmation';
 import SettingsDialog from '../dialogs/settings';
 import ShareDialog from '../dialogs/share';
+import TrashTagConfirmation from '../dialogs/trash-tag-confirmation';
 import { closeDialog } from '../state/ui/actions';
 
 import * as S from '../state';
@@ -35,36 +36,40 @@ export class DialogRenderer extends Component<Props> {
   static displayName = 'DialogRenderer';
 
   render() {
-    const { theme, closeDialog } = this.props;
-
+    const { theme, closeDialog, dialogs } = this.props;
     return (
       <Fragment>
-        {this.props.dialogs.map((dialog) => (
+        {dialogs.map((dialog) => (
           <Modal
-            key={dialog}
+            key={dialog.type}
             className="dialog-renderer__content"
-            contentLabel={dialog}
+            contentLabel={dialog.type}
             isOpen
             onRequestClose={closeDialog}
             overlayClassName="dialog-renderer__overlay"
             portalClassName={`dialog-renderer__portal theme-${theme}`}
           >
-            {'ABOUT' === dialog ? (
+            {'ABOUT' === dialog.type ? (
               <AboutDialog key="about" closeDialog={closeDialog} />
-            ) : 'BETA-WARNING' === dialog ? (
+            ) : 'BETA-WARNING' === dialog.type ? (
               <BetaWarning key="beta-warning" />
-            ) : 'CLOSE-WINDOW-CONFIRMATION' === dialog ? (
+            ) : 'CLOSE-WINDOW-CONFIRMATION' === dialog.type ? (
               <CloseWindowConfirmation key="close-window-confirmation" />
-            ) : 'IMPORT' === dialog ? (
+            ) : 'IMPORT' === dialog.type ? (
               <ImportDialog key="import" />
-            ) : 'KEYBINDINGS' === dialog ? (
+            ) : 'KEYBINDINGS' === dialog.type ? (
               <KeybindingsDialog key="keybindings" />
-            ) : 'LOGOUT-CONFIRMATION' === dialog ? (
+            ) : 'LOGOUT-CONFIRMATION' === dialog.type ? (
               <LogoutConfirmation key="logout-confirmation" />
-            ) : 'SETTINGS' === dialog ? (
+            ) : 'SETTINGS' === dialog.type ? (
               <SettingsDialog key="settings" />
-            ) : 'SHARE' === dialog ? (
+            ) : 'SHARE' === dialog.type ? (
               <ShareDialog key="share" />
+            ) : 'TRASH-TAG-CONFIRMATION' === dialog.type ? (
+              <TrashTagConfirmation
+                key="trash-tag-confirmation"
+                tagName={dialog.tagName}
+              />
             ) : null}
           </Modal>
         ))}
--- a/lib/dialogs/trash-tag-confirmation/index.tsx
+++ b/lib/dialogs/trash-tag-confirmation/index.tsx
@@ -1,0 +1,54 @@
+import React, { FunctionComponent } from 'react';
+import { connect } from 'react-redux';
+
+import actions from '../../state/actions';
+import Dialog from '../../dialog';
+
+import type * as S from '../../state';
+import type * as T from '../../types';
+
+type OwnProps = {
+  tagName: T.TagName;
+};
+
+type DispatchProps = {
+  closeDialog: () => any;
+  trashTag: (tagName: T.TagName) => any;
+};
+
+type Props = OwnProps & DispatchProps;
+
+const TrashTagConfirmation: FunctionComponent<Props> = ({
+  closeDialog,
+  tagName,
+  trashTag,
+}) => (
+  <Dialog
+    className="trash-tag-confirmation"
+    onDone={closeDialog}
+    title="Delete Tag"
+  >
+    <div className="dialog-inner-content">
+      Are you sure you want to delete &quot;
+      <span className="tag-name">{tagName}</span>&quot;?
+    </div>
+    <div className="dialog-buttons">
+      <button
+        className="button-primary delete-tag"
+        onClick={() => trashTag(tagName)}
+      >
+        Delete Tag
+      </button>
+    </div>
+  </Dialog>
+);
+
+const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
+  trashTag: (tagName) => ({
+    type: 'TRASH_TAG',
+    tagName,
+  }),
+  closeDialog: actions.ui.closeDialog,
+};
+
+export default connect(null, mapDispatchToProps)(TrashTagConfirmation);
--- a/lib/state/action-types.ts
+++ b/lib/state/action-types.ts
@@ -101,7 +101,13 @@ export type SetUnsyncedNoteIds = Action<
   { noteIds: T.EntityId[] }
 >;
 export type ShowAllNotes = Action<'SHOW_ALL_NOTES'>;
-export type ShowDialog = Action<'SHOW_DIALOG', { dialog: T.DialogType }>;
+export type ShowDialog = Action<
+  'SHOW_DIALOG',
+  {
+    name: T.DialogType['type'];
+    data?: object;
+  }
+>;
 export type StoreEditorSelection = Action<
   'STORE_EDITOR_SELECTION',
   { noteId: T.EntityId; start: number; end: number; direction: 'RTL' | 'LTR' }
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -108,9 +108,23 @@ const dialogs: A.Reducer<T.DialogType[]> = (state = [], action) => {
   switch (action.type) {
     case 'CLOSE_DIALOG':
       return state.slice(0, -1);
-
-    case 'SHOW_DIALOG':
-      return state.includes(action.dialog) ? state : [...state, action.dialog];
+    case 'TRASH_TAG':
+      return state.filter((dialog) => dialog.type !== 'TRASH-TAG-CONFIRMATION');
+    case 'REMOTE_TAG_DELETE':
+      return state.filter(
+        (dialog) =>
+          !(
+            dialog.type === 'TRASH-TAG-CONFIRMATION' &&
+            tagHashOf(dialog.tagName) === action.tagHash
+          )
+      );
+    case 'SHOW_DIALOG': {
+      const { type, name, ...data } = action;
+      // This ensures we only show one dialog of each type at a time.
+      return state.find((dialog) => dialog.type === name)
+        ? state
+        : [...state, { type: name, ...data }];
+    }
 
     default:
       return state;
@@ -278,7 +292,7 @@ const showNavigation: A.Reducer<boolean> = (state = false, action) => {
     case 'SHOW_ALL_NOTES':
       return false;
     case 'SHOW_DIALOG':
-      if (action.dialog === 'SETTINGS') {
+      if (action.name === 'SETTINGS') {
         return false;
       }
       return state;
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -65,14 +65,18 @@ export type ConnectionState = 'green' | 'red' | 'offline';
 // Application Types
 ///////////////////////////////////////
 export type DialogType =
-  | 'ABOUT'
-  | 'BETA-WARNING'
-  | 'CLOSE-WINDOW-CONFIRMATION'
-  | 'IMPORT'
-  | 'KEYBINDINGS'
-  | 'LOGOUT-CONFIRMATION'
-  | 'SETTINGS'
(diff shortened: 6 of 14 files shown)


lib/types.ts
```diff
@@ -66,7 +66,8 @@ export type DialogType =
   | 'KEYBINDINGS'
   | 'LOGOUT-CONFIRMATION'
   | 'SETTINGS'
-  | 'SHARE';
+  | 'SHARE'
+  | 'TRASH-TAG-CONFIRMATION';
```
dmsnell: nice work finding this and adding the new dialog to the type 👍

lib/dialogs/trash-tag-confirmation/index.tsx
```diff
@@ -0,0 +1,60 @@
+import actions from 'lib/state/actions';
```
dmsnell: it probably feels arbitrary but this, for consistency sake, belongs below third-party imports. it looks like your IDE might have auto-included this and the imports in line 5. we can replace both import statements with the more standard-in-this-app `import actions from '../../state/actions'` and then call using the fully-qualified action access, like `actions.ui.closeDialog`

lib/dialogs/trash-tag-confirmation/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    className="trash-tag-confirmation"
+    onDone={closeDialog}
+    title="Delete Tag"
+  >
+    <div>Are you sure you want to delete this tag?</div>
+    <button
+      className="button-primary delete-tag"
+      onClick={() => {
+        trashTag(tagToTrash);
+        closeDialog();
```
dmsnell: we want to avoid this kind of "double-dispatch" because it introduces _partial updates_ to our app, where the UI state is in transition. much better is to look at the dialog reducer and ask ourselves, "what changes this data?" we can answer that by saying, "if we trash a tag, we imply that we should close the tag-trash-confirmation dialog."

in that case we only need to dispatch the single action which expresses the user-oriented intent "yes, trash this tag," and let the application _react_ to that however it wants.

from time to time we may only want to conditionally do something like this, like close the dialog on trashing a tag. when those cases arise we might want to include metadata on the Redux action to indicate to the appropriate reducer that yes indeed this implies closing the dialog. here I'm not sure that extra metadata would be warranted because why would we continue showing that confirmation dialog if the tag was trashed.

that raises another point concerning the dialog reducer. _if_ it's open and the tag is trashed from another device we might want to close the dialog, or if we want to keep it open then we could have the dialog change to reflect that the tag was just remotely trashed

lib/state/ui/middleware.ts
```diff
@@ -91,6 +93,10 @@ export const middleware: S.Middleware = (store) => (
       }
       return next(action);
 
+    case 'TAG_TO_TRASH':
+      store.dispatch(showDialog('TRASH-TAG-CONFIRMATION'));
+      return next(action);
```
dmsnell: we should not need this. although as we think about it we think, "when we want to trash a tag then open the dialog"  but the dialog reducer can listen and react on its own without a second dispatch. ping me in Slack if you want to talk through this

lib/dialogs/trash-tag-confirmation/index.tsx
sandymcfadden: Thanks for catching this!

dmsnell (commented):
looking good @sandymcfadden 

I'm noting how almost all of the places we open a dialog it's purely a dialog name we need. I wonder if we can leave `actions.ui.showDialog` as it was, or add an optional second parameter that only `trashTag` need supply? that would leave all of the other existing calls unchanged, and the same could apply to `appCommandSender`, since it doesn't (currently) trash a tag. might be the case where if we need the additional data we could find a way to do it without exposing the internals of the action to the callers, as we can see is done when we require them to pass `{type: ""}` vs. just the name.

sandymcfadden:
Thank you @dmsnell. I've updated things based on the feedback, but still don't have the types all set correctly as I'm getting errors in VS Code. Things all do seem to be working though.

lib/state/action-types.ts
```diff
@@ -101,7 +101,13 @@ export type SetUnsyncedNoteIds = Action<
   { noteIds: T.EntityId[] }
 >;
 export type ShowAllNotes = Action<'SHOW_ALL_NOTES'>;
-export type ShowDialog = Action<'SHOW_DIALOG', { dialog: T.DialogType }>;
+export type ShowDialog = Action<
+  'SHOW_DIALOG',
+  {
+    dialogName: T.DialogType['type'];
+    dialogData: object;
```
dmsnell: generally speaking we can omit the prefix-as-namespace since the action type itself indicates we are talking about a dialog. in the previous code we had to have something there to indicate which one, and we probably could have used `name` but in the absence of a clear word for it, `dialog` communicated "which dialog"

here we can probably communicate just as effectively if we store `name` and `data`

lib/state/ui/reducer.ts
```diff
@@ -90,9 +90,20 @@ const dialogs: A.Reducer<T.DialogType[]> = (state = [], action) => {
   switch (action.type) {
     case 'CLOSE_DIALOG':
       return state.slice(0, -1);
-
+    case 'TRASH_TAG':
+      return state.filter(
+        (element) => element.type !== 'TRASH-TAG-CONFIRMATION'
+      );
```
dmsnell: do you have `prettier` and the git hooks setup? I thought it would have added a newline between the `case` lines

lib/state/ui/reducer.ts
```diff
@@ -90,9 +90,20 @@ const dialogs: A.Reducer<T.DialogType[]> = (state = [], action) => {
...
+    case 'REMOTE_TAG_DELETE':
+      return state.filter(
+        (element) =>
+          element.tagName !== tagHashOf(action.tagHash) ||
```
dmsnell: I'm a bit lost in the negative-or logic with the filter. would it be clearer if we stated the specific dialog we want to get rid of instead of the qualifications for which ones can stay?

```ts
return state.filter(element => !(
	element.type === 'TRASH-TAG-CONFIRMATION' &&
	element.tagName === tagHashOf(action.tagHash)
))
```

lib/state/action-types.ts
sandymcfadden: Sounds good and makes sense to me.

lib/state/ui/reducer.ts
sandymcfadden: Thank you! I was confusing myself with this logic as well. This suggestion is so much more readable.

lib/state/ui/reducer.ts
sandymcfadden: I believe I have it all set up properly, but it doesn't add a newline between the `case` lines. 
It looks like there is a mix of styles used after a quick search where in some places there are spaces but not everywhere.

dmsnell: interesting. thanks for checking

lib/app-layout/index.tsx
```diff
@@ -144,7 +144,9 @@ export class AppLayout extends Component<Props> {
 const mapStateToProps: S.MapState<StateProps> = (state) => ({
   hasRevisions:
     state.ui.showRevisions && state.data.noteRevisions.has(state.ui.openedNote),
-  keyboardShortcutsAreOpen: state.ui.dialogs.includes('KEYBINDINGS'),
+  keyboardShortcutsAreOpen:
+    state.ui.dialogs.find((element) => element.type === 'KEYBINDINGS') !==
+    undefined,
```
dmsnell: simply because this is a big complicated and specific to the dialogs reducer I wonder if we should add a selector in `selectors.ts` to make it clearer in the normal code:

```ts
export const isDialogOpen = (state: S.State, name: T.DialogType) =>
	state.ui.dialogs.find(({type}) => type === name) !== undefined
```

```ts
import * from selectors;

…

keyboardShortcutsAreOpen: selectors.isDialogOpen(state, 'KEYBINDINGS'),
```

normally I wouldn't encourage writing a selector or function for a single use like this but that's kind of quirky code _necessary_ to get the right answer and moving it into its own testable place might be helpful

lib/dialog-renderer/index.tsx
```diff
@@ -36,35 +37,39 @@ export class DialogRenderer extends Component<Props> {
 
   render() {
     const { theme, closeDialog } = this.props;
-
     return (
       <Fragment>
         {this.props.dialogs.map((dialog) => (
```
dmsnell: is this happy with the types? does it know that `tagName` exists below? sometimes it has problems with `map()`

dmsnell: also you can add `dialogs` above in line 38 and then refer to it here without `this.props` - looks like that was something that already existed, and maybe there's a reason (shadowing?) to avoid `dialogs` but if not you can it

lib/state/ui/reducer.ts
```diff
@@ -90,9 +90,22 @@ const dialogs: A.Reducer<T.DialogType[]> = (state = [], action) => {
   switch (action.type) {
     case 'CLOSE_DIALOG':
       return state.slice(0, -1);
-
+    case 'TRASH_TAG':
+      return state.filter(
+        (element) => element.type !== 'TRASH-TAG-CONFIRMATION'
```
dmsnell: at this point I believe we are dealing with `dialog`s - might be more helpful than the generic "element"

lib/state/ui/reducer.ts
```diff
@@ -90,9 +90,22 @@ const dialogs: A.Reducer<T.DialogType[]> = (state = [], action) => {
...
+      return state.filter(
+        (element) =>
+          !(
+            element.type === 'TRASH-TAG-CONFIRMATION' &&
+            tagHashOf(element.tagName) === action.tagHash
+          )
+      );
     case 'SHOW_DIALOG':
-      return state.includes(action.dialog) ? state : [...state, action.dialog];
+      return state.find((element) => element.type === action.name)
```
dmsnell: this prevents overlapping multiple tag trash dialogs over each other, which I believe is the right move anyway - just noting it. might be worth a comment in the code to tell future us that it's not a mistake

dmsnell (commented):
Left a few final observations you can choose to respond to; you don't need to take them all but I think those are the only things that I see as being considerations before a merge. Otherwise the code looks good and assuming you have tested it then it looks good to me. I did not test this.

lib/app-layout/index.tsx
sandymcfadden: This certainly makes it cleaner to read when in `app-layout.tsx` I end up with a type error for `'KEYBINDINGS'` though `Argument of type 'string' is not assignable to parameter of type 'DialogType'.ts(2345)`

There are a couple of other places where I've run into similar as well. I'm not sure how to best deal with it. Another example is in the [reducer](https://github.com/Automattic/simplenote-electron/blob/add/delete-tag-confirmation-middleware/lib/state/ui/reducer.ts#L108). The last `tagName` there gives the error 
`Property 'tagName' does not exist on type 'Action<"SHOW_DIALOG", { name: "ABOUT" | "BETA-WARNING" | "CLOSE-WINDOW-CONFIRMATION" | "IMPORT" | "KEYBINDINGS" | "LOGOUT-CONFIRMATION" | "SETTINGS" | "SHARE" | "TRASH-TAG-CONFIRMATION"; data: object; }>'.ts(2339)`

It does all work though, any ideas or pointers you have?

lib/dialog-renderer/index.tsx
sandymcfadden: Types seem to be fine here. 
I added `dialogs` as you mentioned and that works fine, thank you.

lib/state/ui/reducer.ts
sandymcfadden: Very true. I'll update this.

lib/state/ui/reducer.ts
sandymcfadden: Good call. 
I do have a question about this section @dmsnell 

`return state.find((dialog) => dialog.type === action.name)
        ? state
        : [...state, { type: action.name, tagName: action.tagName }];`

Is there some way I can make this more generic? I tried a few things but couldn't get it to work.
If we add more dialogs that have extra data it would have to be manually added here. 
I figure there should be a way to do it because in the `ShowDialog` [type](https://github.com/Automattic/simplenote-electron/blob/add/delete-tag-confirmation-middleware/lib/state/action-types.ts#L108) we are using the generic `data` property. But I can't seem to figure out how to get this to work.

lib/app-layout/index.tsx
dmsnell: well my code examples are rarely correct or without bugs. `T.DialogType` was updated in this PR and I was reusing the old type. you can get just the `type` name by using `T.DialogType['type']`

lib/state/ui/reducer.ts
dmsnell: I don't know how much more generic we want it. that's a question at least this PR doesn't raise: the generic question would likely be "can we have more than one dialog of a specific type open if they have different data?" for tag deletion it seems reasonable to not allow that, and we don't currently have any other dialogs with associated data.

were you thinking of another kind of generalization?

lib/state/ui/reducer.ts
```diff
@@ -90,9 +90,21 @@ const dialogs: A.Reducer<T.DialogType[]> = (state = [], action) => {
...
+            dialog.type === 'TRASH-TAG-CONFIRMATION' &&
+            tagHashOf(dialog.tagName) === action.tagHash
+          )
+      );
     case 'SHOW_DIALOG':
-      return state.includes(action.dialog) ? state : [...state, action.dialog];
+      // This ensures we only show one dialog of each type at a time.
+      return state.find((dialog) => dialog.type === action.name)
+        ? state
+        : [...state, { type: action.name, tagName: action.tagName }];
```
dmsnell: we don't need `tagName` called out specifically here - maybe I misread your comment about generalization.

note that we have access to associated data via the action. we can accomplish this in a couple ways, but if we want to start by blindly accepting each additional bit of data, we can spread the rest of the arguments.

```ts
case 'SHOW_DIALOG': {
	const {type, name, meta, ...data} = action;

	…
		? state
		: [...state, {type: action.name, ...data}];
}
```

sandymcfadden: This is exactly what I was trying to get at and figure out in my other comment, thank you!

lib/state/ui/reducer.ts
sandymcfadden: You got at what I was looking for in your other comment. Thank you!

lib/app-layout/index.tsx
sandymcfadden: Of course, I totally missed that.

belcherj:
Be sure to tag @SylvesterWilmott for design review.

belcherj:
I think the corners are more rounded in the design.

Screen Shot 2021-03-01 at 12 13 04 PM
Screen Shot 2021-03-01 at 12 12 42 PM

sandymcfadden:
You're absolutely right @belcherj. I've fixed that now. It does still need design review though as I added the tag name to the text of the dialog which wasn't included in the design. I wanted to get your opinion on that @SylvesterWilmott 
Screen Shot 2021-03-01 at 2 14 01 PM

lib/dialogs/trash-tag-confirmation/index.tsx
```diff
@@ -28,7 +28,9 @@ const TrashTagConfirmation: FunctionComponent<Props> = ({
     onDone={closeDialog}
     title="Delete Tag"
   >
-    <p>Are you sure you want to delete &quot;{tagName}&quot;?</p>
+    <div className="dialog-inner-content">
+      Are you sure you want to delete &quot;{tagName}&quot;?
```
dmsnell: if it's giving you a headache with the formatting you can also produce the entire string in JavaScript

```ts
<div className="dialog-inner-content">{`Are you sure you want to delete '${tagName}'?`</div>
```

strings like this will be properly escaped by React for HTML

sandymcfadden: Thanks. For once this is something I learned already :-)
The changes here were more about getting the spacing to be consistent with the designs.
I used the `&quot;` characters following another example I had found in the codebase.
Do you prefer one way over the other?

SylvesterWilmott:
After some back and forth on slack, the design looks good!

lib/dialogs/trash-tag-confirmation/index.tsx
dmsnell: no preference from me!

codebykat:
This is looking good.

I did have to test with this 😂

Screen Shot 2021-03-03 at 11 43 37 PM

Similar to the fix we need for the email verification screen, I bet. `word-break: break-all` ONLY on the tag name itself. #2712

sandymcfadden:
Thanks, @codebykat I have that fixed now.

codebykat (approved):
LGTM 👍

dmsnell (approved):
:shipit:

PR #2661, opened 2021-02-11
Re-add so that when adding a new tag to a note a space will submit the tag

### LINKED ISSUE #2659: [Tags] Using a space in note editor tags area should submit a tag
### What
Comma and space are both mentioned as tags separators [here](https://github.com/Automattic/Simplenote-United/blob/trunk/docs/tags.md). When the user types a comma, all text before comma will be submitted as a tag. Why is this not so for a space?

### Why
Will provide more transparent experience for user.

### LINKED ISSUE #2660: [Tags] Possible to break app layout by entering many spaces in tags input
### Expected
Leading spaces should not be allowed in tags input in note editor. This bug will be fixed if #2659 is implemented.

### Observed
See video:

https://user-images.githubusercontent.com/73365754/107640369-da58e180-6c7a-11eb-8763-4d24da674518.mov

### Reproduced
1. Open any note
2. Set focus to tags input in note editor
3. Hold space until the amount of spaces starts breaking the app layout.

### Where did you see the bug
- System Make: Mac, Dell
- System Model: MBP 16", Latitude 7400
- OS: Both Win10 and Big Sur 11.2
- Browser (if applicable): both Chrome and Electron app
- Browser version (if applicable): Chrome 88.0.4324.150

sandymcfadden:
### Fix

This was added previously in PR #2607 but then I overwrote those changes in PR #2602 by mistake.
Adding it back here so that when you type a Space key when adding tags to a note it will submit the tag.
This resolves #2659 and resolves #2660.

### Test

1. Open a note.
2. Enter a tag from the tag input at the bottom.
3. Try to type a space character.
4. Notice the tag is added instead of a space character.

### Release

- Fix tag input from the note editor to insert tags when a space is used

### DIFF
--- a/lib/tag-input/index.tsx
+++ b/lib/tag-input/index.tsx
@@ -13,6 +13,7 @@ import type * as T from '../types';
 
 const KEY_TAB = 9;
 const KEY_ENTER = 13;
+const KEY_SPACE = 32;
 const KEY_RIGHT = 39;
 const KEY_COMMA = 188;
 
@@ -122,6 +123,7 @@ export class TagInput extends Component<Props> {
     invoke(
       {
         [KEY_ENTER]: this.submitTag,
+        [KEY_SPACE]: this.submitTag,
         [KEY_COMMA]: this.submitTag,
         [KEY_TAB]: this.interceptTabPress,
         [KEY_RIGHT]: this.interceptRightArrow,


codebykat (approved):
I rebased this to `release/2.7.0` and changed the base of the PR. Looks good!

PR #2667, opened 2021-02-11
Fix: refresh note list after revision restored

### LINKED ISSUE #2664: Unpinning note via history leaves unpinned note above pinned ones
### Expected
Unpinned note should not mix with pinned ones.

### Observed
If a note in unpinned via history, its position in the note list will not change:
Screenshot_2021-02-11 20 12 24_SW8YnN

### Reproduced
1. Make sure you have an unpinned note
2. Have 'Sort by Modified: Newest' enabled for better display
3. Pin the note ~> It should appear at the very top of notes list
4. Open history for the note and revert to previous revision ~> note is no longer pinned, but still is placed at the very top (see pic above)

### Where did you see the bug
- OS version: both Win10 and macOS 11.2
- Browser (if applicable): Chrome and also Electron

codebykat:
### Fix

Fixes #2664 

### Test

1. Have an unpinned note
2. Pin it
3. Restore to prior version with History slider
4. Note that it correctly moves in the notes list

### Release

Fixed a bug where the notes list was not updated after restoring a revision

### DIFF
--- a/lib/search/index.ts
+++ b/lib/search/index.ts
@@ -366,6 +366,7 @@ export const middleware: S.Middleware = (store) => {
       case 'CREATE_NOTE_WITH_ID':
       case 'IMPORT_NOTE_WITH_ID':
       case 'REMOTE_NOTE_UPDATE':
+      case 'RESTORE_NOTE_REVISION':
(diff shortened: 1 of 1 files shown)


sandymcfadden (approved):
This looks good and tests well for me.

PR #2669, opened 2021-02-12
Fix: Remove unwanted context menu items

### LINKED ISSUE #2668: Context menu items shown that shouldn't be there
Not sure when this started happening, maybe with the latest Monaco update. We attempt to remove the default Cut and Copy actions so we can add ours with the shortcuts, but it's now showing both, as well as stuff like Command Palette which we had intended to be hidden.

Looks like our hacky CSS has failed us and needs to be updated.

Screen Shot 2021-02-11 at 2 13 21 PM

codebykat:
### Fix

**2.7.0 Hotfix.** Resolves #2668

### Test

Right-click. Verify no odd options are shown:

Screen Shot 2021-02-11 at 4 07 20 PM

### Release

Fixed a bug causing duplicate and unwanted items to appear in the context menu.


sandymcfadden (approved):
Nice work. This certainly seems like a better approach than hiding them with CSS.

PR #2678, opened 2021-02-16
Updating the windows build configuration to allow chosing of installation path

sandymcfadden:
### Fix

This disables the `oneClick` installation option in the build configuration.
When it is disabled during the installation process it allows the option to choose to install either for All users (Requires administrator privileges) or only for the current user (doesn't require administrator privileges).

If installed for all users it is installed to "C:\Program Files\Simplenote\Simplenote.exe"
If installed for current user it is installed to "C:\Users\USERNAME\AppData\Local\Programs\Simplenote\Simplenote.exe"

This is an update to PR #2581 to address the issue mentioned in [this comment](https://github.com/Automattic/simplenote-electron/pull/2581#issuecomment-778281512).

**Options during installation**

Screen Shot 2021-02-16 at 8 19 55 AM

**If installing for all users when not logged in as an Administrator**

Screen Shot 2021-02-16 at 8 21 47 AM

**If installing for all users when logged in as an Administrator**

Screen Shot 2021-02-16 at 8 27 28 AM

### Test

Use the exe build to install on Windows.

1. Install with an Administrator Windows user.
2. Ensure you can install it for all users on the computer.
3. Ensure you can also install for just this logged-in user.
4. Verify the installation location for each. (Should match the locations listed above)
5. Install with a Standard Windows user. 
6. Ensure you can install just for this logged-in user with no security prompt.
7. Ensure you can install for all users by entering an Administrator password.
8. Verify the installation location for each user again. (Should match the locations listed above)

### Release

- Updates the Windows installation process to be able to choose to install for all users on the computer or just the current user.

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -51,7 +51,8 @@
   },
   "nsis": {
     "artifactName": "Simplenote-win-${version}-${arch}.${ext}",
-    "deleteAppDataOnUninstall": true
+    "deleteAppDataOnUninstall": true,
+    "oneClick": false
   },
   "linux": {
     "artifactName": "Simplenote-linux-${version}-${arch}.${ext}",


codebykat (approved):
Nice sleuthing. I did not test this but the fix makes sense.

Issue #2570, opened 2021-01-11
UI spring cleaning: Settings


codebykat:
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc52ff862f4675ba2fd81
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc58e0a996a61cb2a2dbd

Basically a total overhaul, see designs:

https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc607629dbc5680a3485c
https://app.zeplin.io/project/5d795a3d8ed5261aa146e057/screen/5fbbc609b2baf1561c28e1bd

<img width="408" alt="Screen Shot 2021-01-10 at 11 08 11 PM" src="https://user-images.githubusercontent.com/52152/104154067-caf32980-5398-11eb-9294-9b06c42237fb.png">

<img width="404" alt="Screen Shot 2021-01-10 at 11 08 15 PM" src="https://user-images.githubusercontent.com/52152/104154079-cf1f4700-5398-11eb-8c83-b7b763f37280.png">

PR #2685, opened 2021-02-22
Update importers to use a single dialog

sandymcfadden:
### Fix

First PR to look at #2570. This is starting to combine the importers to one Dialog so that any of the supported file types can be chosen and imported. This also now allows the ability to import more than one Simplenote or Evernote export file at the same time.

**Screenshots**
<details><summary>Light mode select files</summary>
Screen Shot 2021-03-08 at 10 23 46 AM
</details>
<details><summary>Light mode files added</summary>
Screen Shot 2021-03-17 at 12 20 50 PM
</details>
<details><summary>Light mode complete</summary>
Screen Shot 2021-03-17 at 12 20 19 PM
</details>

<details><summary>Dark mode select files</summary>
Screen Shot 2021-03-08 at 10 28 45 AM
</details>
<details><summary>Dark mode files added</summary>
Screen Shot 2021-03-17 at 12 19 22 PM
</details>
<details><summary>Dark mode complete</summary>
Screen Shot 2021-03-17 at 12 19 54 PM
</details>

### Test
- Go to Settings > Tools > Import Notes.
- Notice you can select any supported files. 
- Choose multiple combinations of txt, md, Simplenote, and Evernote files.
- Ensure they import as expected.
- Ensure selecting to enable markdown for all notes does enable markdown.
- Smoke test in light and dark mode to ensure styling looks correct.

### Release
- Updating imports to a single dialog where any supported file type can be imported at the same time.
- Updating the styles of the import dialogs.


lib/dialogs/import/dropzone/index.tsx
```diff
@@ -27,33 +20,11 @@ function ImporterDropzone({
...
-        count: 0,
-        multiple: true,
-      },
-    ];
-
     const filteredFiles = [];
+    const fileCounts = new Map();
+    importers.forEach((importer) => {
+      fileCounts.set(importer.name, 0);
+    });
```
dmsnell: it took me a minute to understand the flow of data here. there are a few ways that I think it would be clearer.

_move the import finder into the new imports module_:

```ts
import * as importers from '../importers';

const seenImporters = new Set();
for (const file of acceptedFiles) {
	const importer = importers.forFilename(file);
	if (!importer) {
		setError('unrecognized_format');
		break;
	}

	if (!importer.multiple && seenImporters.has(importer)) {
		setError('too_many_files', importer.errorMessage);
		break;
	}

	seenImporters.add(importer);
	filteredFiles.push(file);
}
```

I believe that some of the complexity here is pointing to a mismatch with the model. If we step back a few levels, what is the goal of this function? What is wrong with importing multiple JSON files, or multiple ENEX files? Is the `multiple` complexity introduced because we had to always permit the upload dialog to allow multiple selections? It looks like if we can currently set multiple errors and keep on going; also that we can theoretically fail to find an importer for the dropped file and continue on from the `importers.find` statement.

lib/utils/import/multiple/index.ts
```diff
@@ -33,81 +30,45 @@ class MultipleImporter extends EventEmitter {
...
-      } else if (fileExtension === 'json') {
-        this.simplenoteFiles.push(file);
-      } else if (fileExtension === 'enex') {
-        this.evernoteFiles.push(file);
-      }
+      // Get the required importer based on the current files extension
+      // Add the file to this importers list of files
+      const theImporter = importers.find((importer) => {
+        return importer.fileTypes.indexOf(fileExtension) !== -1;
+      });
```
dmsnell: another place where having the function in the `importers` module would be nice.

```ts
const importer = importers.forFilename(file);
```

dmsnell: I'm noting here that with this one there's an implicit rule that the first importer to support importing the file extension is the one we'll choose. we can declare this rule when we move the logic into `importers.ts` or we can set up an explicit mapping between filename/extension and chosen importer to remove the ambiguity, for example, a switch statement on file extension.

lib/utils/import/multiple/index.ts
```diff
@@ -33,81 +30,45 @@ class MultipleImporter extends EventEmitter {
...
-          importedNoteCount += arg;
-          if (totalImporters === totalComplete) {
-            this.emit('status', 'complete', importedNoteCount);
+    // Loop through each of the files to import and process them with the appropriate importer
+    this.importFiles.forEach(async (files, name) => {
+      if (files.length > 0) {
+        totalImporters++;
+        const { default: Importer } = await import(
+          /* webpackChunkName: 'utils-import-[request]' */ `../../../utils/import/${name}`
+        );
```
dmsnell, line 69: this probably doesn't do what you think it does. it would be nice to move this into a function in the new `importers` module you created where the import paths can be explicit and static.

```ts
export const load = (name: ImporterName): Promise<???> => {
	switch (name) {
		case 'simplenote':
			return import(/* webpackChunkName: 'utils-import-simplenote' */ '../../../utils/import/simplenote');

		case 'evernote':
			return import(/* webpackChunkName: 'utils-import-evernote' */ '../../../utils/import/evernote');

		case 'text-files':
			return import(/* webpackChunkName: 'utils-import-text-files' */ '../../../utils/import/text-files');
	}

	throw new ValueError(`Unrecognized importer named ${name}`);
}
```

```ts
const { default: Importer } = await importers.load(name);
```

This allows us to find those import paths by string search and will prevent web pack from doing what it does, which I think would be to create a chunk/bundle for every file in the `utils/import/` path, possibly including the README. It additionally protects us against dynamic names exposing things we don't expect.

lib/dialogs/import/dropzone/index.tsx
sandymcfadden: Thank you @dmsnell. The idea of using a Set here and moving the finder into the imports module makes total sense.

In the current iteration of the importers, there are three different import screens, one for each type, text files, Simplenote export, and Evernote export. The only one that allows multiple files to be uploaded is the text files. The Simplenote and Evernote importers are set to process only the first file of their type even if multiple are managed to be uploaded. I'm not sure what the reasoning for this was, but I'm guessing maybe performance concerns, especially around the Evernote exports. The interface would only allow you to choose one file though.

With this design change, there is only one import screen that should be able to accept any of the supported file types. So previously we depended on the dropzone to only allow one Simplenote or Evernote export file to be selected and uploaded at a time. I don't see a way to do that while also allowing multiple text files at the same time. That's why I chose this way of filtering the files and then only processing one of each and discarding the rest.

lib/utils/import/multiple/index.ts
sandymcfadden: This all makes sense and I'll work on it. Thank you!

lib/utils/import/multiple/index.ts
sandymcfadden: You're right, that's not what I expected. I thought using the `[request]` would use the variable passed to the path based on examples I looked at. Sure seems you're right though looking at it again. Thank you!

lib/dialogs/import/dropzone/index.tsx
dmsnell: maybe we should "see what happens" if we allow queueing up multiple imports for Simplenote and Evernote exports. it might work if we queue them up to run one-after-another.

then again, if that doesn't "just work" then it would probably be best to stop the process in its tracks and alert that we received multiple `.enex` files but can only use one yada yada

lib/utils/import/multiple/index.ts
dmsnell: it might actually interpolate that `request`, but my point was more on the side of web pack preemptively creating chunks that we don't expect, because it uses some code analysis and tries to guess "what are the possible files that could be referenced by a variable in that path?" and that boils down to something like `../../../utils/import/*`

lib/dialogs/import/dropzone/index.tsx
```diff
@@ -16,43 +16,43 @@ function ImporterDropzone({
...
-      } else {
-        filteredFiles.push(file);
+
+      const importer = importers.forFilename(file.name);
+      if (!importer) {
+        setErrorMessage([
+          ...errorMessage,
+          `The file type for "${file.name}" is not recognized`,
+        ]);
+        continue;
```
dmsnell, line 32: nice job updating the error so that they all come through.

lib/dialogs/import/dropzone/index.tsx
```diff
@@ -107,17 +89,22 @@ function ImporterDropzone({
         {file.name}
       </li>
     ));
+
+    const errorList = errorMessage.map((error) => (
+      <div key={error}>
+        <WarningIcon />
+        {error}
```
dmsnell, line 92: I don't know if it matters but if it's a list we might consider generating that HTML for the sake of screenreaders. it's not like this app is in a good shape for accessibility, but if we're adding new code we might be able to help.

```ts
{errorMessages.length && (
	<ul>
		{errorMessages.map(error => (
			<li key={error}><WarningIcon />{error}</li>
		)}
	</ul>
)}
```

lib/dialogs/import/importers.ts
```diff
@@ -26,4 +28,46 @@ const importers: Array<importer> = [
...
+
+    case 'txt':
+    case 'md':
+      return importers.find((importer) => importer.name === 'text-files');
+  }
+
+  throw new Error(`No importer found for file ${file}`);
+};
+
+export const load = (name: ImporterName): Promise<object> => {
```
dmsnell, line 69: it's probably worth putting an explanation here - "here be dragons" - to warn people about the dynamic imports. it doesn't need to be much

```ts
/** Async-loads importer JS bundle
 *
 * Warning! Don't replace the static strings with string-interpolation
 * or else they won't be as easy to find and web pack will generate
 * more than we expect.
 */
```

lib/dialogs/import/importers.ts
```diff
@@ -26,4 +28,46 @@ const importers: Array<importer> = [
   },
 ];
 
-export default importers;
+export const getImporters = () => {
+  return importers;
+};
```
dmsnell, line 21: we should be able to expose `importers` directly - was there an issue with that?

lib/dialogs/import/importers.ts
```diff
@@ -26,4 +28,46 @@ const importers: Array<importer> = [
...
+    case 'json':
+      return importers.find((importer) => importer.name === 'simplenote');
+
+    case 'enex':
+      return importers.find((importer) => importer.name === 'evernote');
+
+    case 'txt':
+    case 'md':
+      return importers.find((importer) => importer.name === 'text-files');
+  }
```
dmsnell, line 58: given that there's only a few possibilities we shouldn't need to find. `importers` is right above in this same module

```ts
switch (fileExtension) {
	case 'json':
		return simplenoteImporter;

	case 'enex':
		return evernoteImporter;

	case 'txt':
	case 'md':
		return textfilesImporter;
}
```

this may demand that we extract the individual importer definitions from the list…

```ts
const simplenoteImporter: Importer { … }
…

const importers = [
	simplenoteImporter,
	evernoteImporter,
	textfilesImporter
];
```

lib/dialogs/import/importers.ts
sandymcfadden: This works fine, and I've changed it to this now. The reason I added the getter function was just because when using this it read weird to me with the repeated importers name like `importers.importers.forEach`. It is only used once so far though so not really an issue.

lib/dialogs/import/importers.ts
sandymcfadden: Sounds good. I'll update this!

lib/dialogs/import/dropzone/index.tsx
sandymcfadden: You're right, I'll change it back to a list. I had it that way at first and then changed to divs temporarily to avoid some styling issues while making sure it worked.

lib/dialogs/import/importers.ts
sandymcfadden: Good idea!

lib/dialogs/import/dropzone/index.tsx
sandymcfadden: I took a look today at allowing multiple SImplenote and Evernote files at once and it would require a refactor of the importers themselves to allow that. Maybe it is something we can look at in a separate PR.

For now I think the flow works pretty well. If they add more than one Simplenote or Evernote file it doesn't stop the flow and then have them go back and re-add the files they do want to import. Instead it shows them a list of the files which will be imported and gives a warning showing the files which won't be imported. 
This at least allows them to get some notes imported, then they could import separately the ones included in the warning.

My guess, but have not looked for data yet to back this up, is that people wouldn't normally have multiple Simplenote or Evernote backups that they are looking to import so likely not something that will impact anyone very often.

lib/dialogs/import/importers.ts
dmsnell: ha yeah well that can be a valid reason to rename things.

fun fact, you can also rename an import when you import them, but I honestly can't think of a better name.

maybe something like `availableImporters` but I'm not sure that's any better.

lib/dialogs/import/dropzone/index.tsx
dmsnell: I'm terrible with styles. when I build something I toss in some garbage and ask @belcherj to make it pretty. he's quite good at that.

lib/dialogs/import/dropzone/index.tsx
```diff
@@ -49,29 +73,42 @@ function ImporterDropzone({
...
+      <li key={error}>
+        <WarningIcon />
+        {error}
+      </li>
+    ));
+
+    return (
+      <Fragment>
+        <div className="accepted-files-header">
+          Import File{acceptedFile.length > 1 ? 's' : ''}
```
dmsnell: when dealing with string literals it can be very advantageous to keep them whole, even at the cost of duplicating code. some day someone is going to want to find where in the app it's printing "Import Files" but this wouldn't turn up.

```ts
{acceptedFile.length > 1 ? 'Import Files' : 'Import File'}
```

this is a common issue with translations as well.

dmsnell (commented):
what is left in this @sandymcfadden that you think needs work? is it ready? are you happy with it? what would you like from review?

lib/dialogs/import/dropzone/index.tsx
sandymcfadden: That makes total sense. In fact, I've run into that in the past when searching for strings in code. I've fixed this now.

lib/dialogs/import/dropzone/index.tsx
sandymcfadden: I've been taking another look at allowing multiple json and enex files to be imported at the same time. The Simplenote importer seems to work fine as expected. But the Evernote one I'm struggling with.
I am doing simple tests with two enex files each with two notes in them. If I upload one at a time it works as expected. If I try uploading them both sometimes all the notes will duplicate and eight notes will be imported instead of the expected four.
I'm pretty sure it is something in the [desktop importer](https://github.com/Automattic/simplenote-electron/blob/develop/desktop/evernote-import/index.js) stream.
Do you have any theories or ideas @dmsnell?

sandymcfadden: Thanks for looking at this with me today @dmsnell It was super helpful.

sandymcfadden:
@dmsnell this is ready for review now. I've tested on all platforms and all seems to be working as expected.

lib/dialogs/import/dropzone/index.tsx
```diff
@@ -63,14 +63,21 @@ function ImporterDropzone({
...
+      'Drop files here'
+    ) : (
+      <div className="drop-instructions">
+        Drag and drop to to select files or <span>browse</span> to choose
+      </div>
+    );
+    return (
+      <Fragment>
+        {errorMessage.length > 0 ? <WarningIcon /> : <CloudIcon />}
+        {instructions}
```
dmsnell: personally I don't find it helpful to split this into a separate variable. did you extract it here because of the new `<div>`?

sandymcfadden: Sort of. I was struggling to get HTML, the `div` and `span`, to render without extracting it.

lib/dialogs/import/dropzone/style.scss
```diff
@@ -1,37 +1,86 @@
...
+  background-color: transparent;
 
   &:focus-within {
     overflow: hidden;
-    outline: $focus-outline;
-  }
-
-  @at-root .theme-dark & {
-    background: $studio-gray-70;
+    outline: none;
```
dcalhoun: IMO it is a poor practice to disable `outline` styles for interactive elements. Keyboard users rely upon these styles to track focus selection in the UI.

In my experience, browsers have gotten better about not applying outlines to elements purely because they were clicked, which I believe is why many people tend to disable `outline`. Regardless, I believe creating a product with an accessible design benefits all and I view focus outlines as a postive.

What do you think about not setting `outline: none`?

lib/dialogs/import/dropzone/index.tsx
dmsnell: it should work if you inline it all, it's okay to mix strings and React components

```ts
{isDragActive ?
	'Drop files here'
: (
	<div className="drop-instructions">
		Drag and drop to select files or <span>browse</span> to choose
	</div>
)}
```

also I just noticed a typo "Drag and drop **to to** select files"

belcherj:
Something feels wrong about the UI here, it seems too bunched up: 

Screen Shot 2021-03-10 at 11 25 46 AM

lib/dialogs/import/dropzone/index.tsx
sandymcfadden: Thank you on both of these points. I've fixed them up.

lib/dialogs/import/dropzone/style.scss
sandymcfadden: I have no issues with it no. I have removed that line now.

sandymcfadden:
Thanks @belcherj I believe I have it matching the designs. Except for the elevation colors which I'll be working on in a separate PR as part of the UI spring cleaning.

@SylvesterWilmott  what are your thoughts.

belcherj:
Looks like it matches designs, seemed a bit squooshed.

belcherj (approved):

SylvesterWilmott:
This looks great @sandymcfadden, thanks!

A few points of feedback, mostly related to the dialog component itself

1. I think we need to include a max-height for the dialog component to avoid times like these where it can grow too tall. I would suggest 420px which will also be the width of our new larger dialogs eg. settings.
2. I don't think dialogs should be dismissed when clicking outside of it. These dialogs are mostly accessible from the settings and so accidentally closing one can be quite frustrating. In my opinion, the "x" is a good enough click target to close them. What do you think?
3. It looks like you've grabbed the `file` icon from the designs which is fine for now but we do need our own in Simplecons. I'll keep you updated on that, hopefully I can do that before this PR gets merged.

sandymcfadden:
Thanks so much, @SylvesterWilmott. 
1. I did have a max height set but it was pretty large. I basically made sure it didn't go outside of the screen size. I've updated that now to be the 420px you've suggested.
2. That sounds good to me. I've updated the dialogs so they don't close by clicking outside of them.
3. We had an existing File icon in the app already and I used it. I can certainly update it with a  new one. Just let me know  🙂

SylvesterWilmott:
Looks great @sandymcfadden 

One last thing, in dark mode, can we use SNBlue30 for the "Browse" text?

SylvesterWilmott:
I also wonder if in these dialog components, the bottom area containing the button should be in fixed position. Perhaps something to iterate on in a separate PR.

sandymcfadden:
Thanks @SylvesterWilmott! I've updated the browse color in dark mode now.

>I also wonder if in these dialog components, the bottom area containing the button should be in fixed position. Perhaps something to iterate on in a separate PR.

I had almost asked you about this and thought the same thing. I think I'll go through in a separate PR and update all the dialogs to act that way. 

Edit: Opened issue here #2777

SylvesterWilmott (approved):

dmsnell (approved):
It appears that this is largely in the styling stage. As long as no major code changes occur I'm happy with what's in this PR. This approval means "no need to check back with me; merge when the styling issues are resolved and you are satisfied with the state of the branch."

pachlava:
Tested on 2.9.0-beta1 under different setups (Windows 10 Home + Chrome, Edge, Desktop | macOS 11.2.3 Chrome + Desktop)

✔️ Importing md, enex, txt and json types simultaneously
❌ Importing 'alien' json (created #2806)
✔️ Selecting both supported and unsupported files types imports only the supported ones
✔️ Import via drag and drop as well as via browsing 
✔️ Drag and drop of folder imports all supported files in this folder
✔️ Drag and drop of enex file to Web app is handled right (forbidden)
✔️ Using both light and dark themes
❌ Import of big number of files creates minor layout issue (created #2801)
✔️ Checking that `Enable Markdown for all notes` really works

PR #2687, opened 2021-02-23
Adding Untagged Notes to Sidebar

### LINKED ISSUE #1230: Search: Add Untagged Notes to the sidebar
#### Steps to reproduce
Open the tag list

#### What I expected
Select untagged notes

#### What happened instead
This function is not yet available with simplenote-electron,
but with simplenote-android it is available:
_**Added a sidebar tab to view untagged notes**_
"https://github.com/Automattic/simplenote-android/pull/527"

#### Simplenote version
1.5.0

#### OS version
Windows 10

39345814-8c681f5a-499f-11e8-99de-464fdff7206a

illusaen:
Fixes #1230 

### Fix
* Adds "Untagged Notes" filter to sidebar.
* Incidentally, found bug where filtering Trash notes and then filtering by a tag results in searching for that tag in Trash notes only. I'm assuming that's not the desired behaviour and corrected it; however I can easily change it back.  **_NOTE_: This bug has been fixed with state refactor.**

#### Untagged Notes
1. Create note with no tags and a note with tags.
2. Open sidebar.
3. Click on "Untagged Notes".
4. See that only the note with no tags shows up in list.

### Screenshots
<details>
  <summary>Before</summary>
  
</details>

<details>
  <summary>No untagged notes</summary>
  
no-untagged
</details>

<details>
  <summary>Delete tags</summary>
  
delete-tags-reset-untagged

</details>

<details>
  <summary>Long list of tags</summary>
  
tag-lots
</details>

### Release
```
### Enhancements

- Added Untagged Notes filter to sidebar.

### Fixes

- Fixed bug where selecting Trash filter and then filtering on a tag resulted in a filter where only notes with that tag in the Trash were shown.
```

### DIFF
--- a/lib/state/selectors.ts
+++ b/lib/state/selectors.ts
@@ -1,5 +1,6 @@
 import * as S from './';
 import * as T from '../types';
+import isEmailTag from '../utils/is-email-tag';
 
 /**
  * "Narrow" views hide the note editor
@@ -54,3 +55,8 @@ export const showTrash: S.Selector<boolean> = ({ ui: { collection } }) =>
   collection.type === 'trash';
 export const isDialogOpen = (state: S.State, name: T.DialogType['type']) =>
   state.ui.dialogs.find(({ type }) => type === name) !== undefined;
+
+export const numberOfNonEmailTags: S.Selector<number> = ({ data }) =>
+  [...data.tags.values()]
+    .filter((tag) => !isEmailTag(tag.name))
+    .map((tag) => tag.name).length;
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -1,5 +1,6 @@
 import { combineReducers } from 'redux';
 
+import { tagHashOf } from '../../utils/tag-hash';
 import {
   withCheckboxCharacters,
   withCheckboxSyntax,
@@ -98,8 +99,15 @@ const collection: A.Reducer<T.Collection> = (
       return { type: 'all' };
     case 'SHOW_UNTAGGED_NOTES':
       return { type: 'untagged' };
-    case 'TRASH_TAG':
-      return action.tagName === state.tagName ? { type: 'all' } : state;
+    case 'TRASH_TAG': {
+      const openedTagIsGone =
+        state.type === 'tag' &&
+        tagHashOf(state.tagName) === tagHashOf(action.tagName);
+      const lastTagDisappeared =
+        state.type === 'untagged' && action?.remainingTags === 0;
+
+      return openedTagIsGone || lastTagDisappeared ? { type: 'all' } : state;
+    }
     default:
       return state;
   }
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -138,6 +138,7 @@ const editingTags: A.Reducer<boolean> = (state = false, action) => {
     case 'OPEN_TAG':
     case 'SELECT_TRASH':
     case 'SHOW_ALL_NOTES':
+    case 'SHOW_UNTAGGED_NOTES':
     case 'NAVIGATION_TOGGLE':
       return false;
     default:
@@ -276,6 +277,7 @@ const showNavigation: A.Reducer<boolean> = (state = false, action) => {
     case 'OPEN_TAG':
     case 'SELECT_TRASH':
     case 'SHOW_ALL_NOTES':
+    case 'SHOW_UNTAGGED_NOTES':
       return false;
     case 'SHOW_DIALOG':
       if (action.dialog === 'SETTINGS') {
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -1,6 +1,5 @@
 import { combineReducers } from 'redux';
 
-import { tagHashOf } from '../../utils/tag-hash';
 import {
   withCheckboxCharacters,
   withCheckboxSyntax,
@@ -97,6 +96,8 @@ const collection: A.Reducer<T.Collection> = (
       return { type: 'trash' };
     case 'SHOW_ALL_NOTES':
       return { type: 'all' };
+    case 'SHOW_UNTAGGED_NOTES':
+      return { type: 'untagged' };
     case 'TRASH_TAG':
       return action.tagName === state.tagName ? { type: 'all' } : state;
     default:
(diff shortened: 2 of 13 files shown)


lib/state/ui/reducer.ts
```diff
@@ -206,6 +208,20 @@ const openedTag: A.Reducer<T.TagHash | null> = (state = null, action) => {
...
 };
 
+const showUntaggedNotes: A.Reducer<boolean> = (state = false, action) => {
+  switch (action.type) {
+    case 'SHOW_UNTAGGED_NOTES':
+      return true;
+    case 'SELECT_TRASH':
+    case 'SHOW_ALL_NOTES':
+    case 'OPEN_TAG':
+    case 'TRASH_TAG':
```
dmsnell: why would trashing a tag close out the untagged view?

dmsnell:
@illusaen this is great! it’s also something that’s long been on our list. nice work submitting code that’s very much in the style of the project.

one of the reasons I have personally not attended to this yet is a small state refactor I was interested in doing. as you noted, we have a few booleans describing mutually-exclusive states, whereas we could very well represent the same information with a single state value that cannot be represented in contradictory ways, as is currently the case with `showTrash`, `showUntagged`, and `openedTag`. This would likely be something akin to this:

```ts
type Collection
  = {type: ‘all-notes’}
  | {type: ‘trash’}
  | {type: ‘tag’, tagName: string}
  | {type: ‘untagged’};

const collection: A.Reducer<Collection> = (state = {type: ‘all-notes’}, action) => {
    switch (action.type) {
        case ‘OPEN_TRASH’:
            return {type: ‘trash’};

        case ‘OPEN_TAG’:
            return {type: ‘tag’, tagName: action.tagName};

        case ‘SHOW_ALL_NOTES’:
            return {type: ‘all-notes’};

        case ‘SHOW_UNTAGGED’:
            return {type: ‘untagged’};

        default:
            return state;
    }
}
```
(please overlook any defects in my code example - I’m rarely posting examples that don’t have errors in them because I’m only really concerned about getting the point across)

with this we never have to worry about “closing” the state, such as by monitoring for `SHOW_UNTAGGED` inside of the `showTrash` reducer or attending to those states in the search middleware.

now it may be that this state refactor could turn into a major job and I don’t know if it would be worth it. this PR looks pretty good as-is, but what are your thoughts about such a refactor and would you be interested in trying to perform such a feat for this task?

lib/state/ui/reducer.ts
illusaen: Good point, removing that!

illusaen:
> @illusaen this is great! it’s also something that’s long been on our list. nice work submitting code that’s very much in the style of the project.

Thanks!

> one of the reasons I have personally not attended to this yet is a small state refactor I was interested in doing...
> now it may be that this state refactor could turn into a major job and I don’t know if it would be worth it. this PR looks pretty good as-is, but what are your thoughts about such a refactor and would you be interested in trying to perform such a feat for this task?

I think that refactor would definitely be worth it because of how many chances there are to accidentally break it as it is right now. However, I think there are probably a lot of small edge cases that need to tested so I'd like to split it off from this PR if possible. WDYT?

dmsnell:
> I'd like to split it off from this PR if possible. WDYT?

come on now, you're just speaking my language and trying to butter me up 😉 

my only concern with this, which is why I asked your thoughts, is that I don't want a refactor of existing state to stand in the way of your contribution. if you would like to attempt the state refactor in another PR and then come back to this one after that (which should leave this PR in a fairly terse and clear state) then I'm all for that. what I don't want to do is set goalposts that are unreasonable for someone coming in and volunteering some help to the app. this is a long-requested feature and a feature-parity issue with out other platforms.

we will prioritize review for any PRs you submit and be happy to interact with you the way we'd interact with each other on it.

illusaen:
> my only concern with this, which is why I asked your thoughts, is that I don't want a refactor of existing state to stand in the way of your contribution. if you would like to attempt the state refactor in another PR and then come back to this one after that (which should leave this PR in a fairly terse and clear state) then I'm all for that. 

Yes, i think this would be the simplest way to do things. :)

> what I don't want to do is set goalposts that are unreasonable for someone coming in and volunteering some help to the app. this is a long-requested feature and a feature-parity issue with out other platforms.
> 

That's fair! I do have a week and a half left with y'all so hopefully the refactor will go faster than that, haha.

> we will prioritize review for any PRs you submit and be happy to interact with you the way we'd interact with each other on it.

Awesome, thanks much! I've noticed y'all are very on the ball about reviewing PRs hehe.

SylvesterWilmott:
Thanks for tackling this @illusaen 

Design-wise there's a few things to mention:

- Since this particular filter is only relevant when the user has at least 1 tag, the untagged notes filter should be hidden if there are 0 tags in the tags list.
- You can find the `untagged` icon [here](https://github.com/Automattic/Simplecons/blob/trunk/SVG/24px/untagged.svg) in the Simplecons repo.
- The untagged notes row should appear below tags in the UI in a separate section. See below for illustration.

Tags

lib/state/ui/reducer.ts
illusaen: No longer pertinent due to state refactor.

SylvesterWilmott:
Looks good and works as expected. Thanks @illusaen

codebykat (approved):
This is testing out great for me now! Let's :shipit:

lib/state/selectors.ts
```diff
@@ -54,3 +55,8 @@ export const showTrash: S.Selector<boolean> = ({ ui: { collection } }) =>
   collection.type === 'trash';
 export const isDialogOpen = (state: S.State, name: T.DialogType['type']) =>
   state.ui.dialogs.find(({ type }) => type === name) !== undefined;
+
+export const numberOfNonEmailTags: S.Selector<number> = ({ data }) =>
+  [...data.tags.values()]
+    .filter((tag) => !isEmailTag(tag.name))
+    .map((tag) => tag.name).length;
```
dmsnell, line 62: No need for this map - it’s redundant

illusaen: You are 1000% correct. See https://github.com/Automattic/simplenote-electron/pull/2723

dmsnell: Also the selector may even be overkill. It’s fine as is, but I don’t see us needing to know often how many tags there are

illusaen: That's fair. I was actually going to make it return an array of `TagName`s (thus why I had that map in there) so it could be used elsewhere but then I read the comment 🤷‍♀️

PR #2694, opened 2021-02-25
Fixing duplicate file name issue in exports

sandymcfadden:
### Fix

When trying to export notes there are some instances where notes with the same title will overwrite each other. This is caused by some characters in the titles that can break our method for checking for duplicate file names. This can be demonstrated by using markdown images as the first line, or title of notes.

For example, if you have two notes with this content:
```
![Simplenote.com](https://github.com/Automattic/simplenote-electron/raw/develop/resources/images/icon_512x512.png)

File 1
```

```
![Simplenote.com](https://github.com/Automattic/simplenote-electron/raw/develop/resources/images/icon_512x512.png)

File 2
```

Only one txt file will be exported because the file names would end up the same and overwrite each other.

The root cause identified (lodash.update was interpreting the "filename" as a property path, thus interpreting. and array-like indexing with brackets. This removes the update function and uses a Map to check for duplicates instead.

### Test

1. Create multiple notes with the same markdown image as the first line.
2. Try to export the notes.
3. Ensure all notes are exported.

### Release

- Fix bug in note export to avoid duplicate filenames when certain characters were used in the note title.

### DIFF
--- a/lib/utils/export/to-zip.ts
+++ b/lib/utils/export/to-zip.ts
@@ -86,9 +86,8 @@ const appendTags = (note) => {
  * @returns {[Array, Object]} final note list and accumulating filename counts
  */
 const toUniqueNames = ([notes, nameCounts], note) => {
-  nameCounts.set(note.fileName, (nameCounts.get(note.fileName) ?? -1) + 1);
-  const count = nameCounts.get(note.fileName);
-
+  const count = nameCounts.get(note.fileName) ?? 0;
+  nameCounts.set(note.fileName, count + 1);
   const fileName = count > 0 ? `${note.fileName} (${count})` : note.fileName;
 
   return [[...notes, { ...note, fileName }], nameCounts];
--- a/lib/utils/export/to-zip.ts
+++ b/lib/utils/export/to-zip.ts
@@ -1,5 +1,5 @@
 import sanitize from 'sanitize-filename';
-import { identity, update } from 'lodash';
+import { identity } from 'lodash';
 
 import { LF_ONLY_NEWLINES } from './export-notes';
 
@@ -81,19 +81,17 @@ const appendTags = (note) => {
  * // when given `Yummy Recipe` and `Yummy Recipe` as duplicates
  * // will return `Yummy Recipe` and `Yummy Recipe (1)`
  *
- * @param {[Array, Object]} accumulator list of note objects for export and filename counts
+ * @param {[Array, Map]} accumulator list of note objects for export and filename counts
  * @param {Object} note note object
  * @returns {[Array, Object]} final note list and accumulating filename counts
  */
 const toUniqueNames = ([notes, nameCounts], note) => {
-  const encodedFileName = btoa(note.fileName);
-  const newNameCounts = update(nameCounts, encodedFileName, (n) =>
-    n || 0 === n ? n + 1 : 0
-  );
-  const count = newNameCounts[encodedFileName];
+  nameCounts.set(note.fileName, (nameCounts.get(note.fileName) ?? -1) + 1);
+  const count = nameCounts.get(note.fileName);
+
   const fileName = count > 0 ? `${note.fileName} (${count})` : note.fileName;
 
-  return [[...notes, { ...note, fileName }], newNameCounts];
+  return [[...notes, { ...note, fileName }], nameCounts];
 };
 
 export const noteExportToZip = (notes) => {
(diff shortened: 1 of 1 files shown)


dmsnell:
> some characters in the titles that can break our method for checking for duplicate file names

what's the mechanism for breaking the duplicate check? I don't understand how those characters do that. are we sure it isn't the special characters causing trouble in the zip library, as if it's not expecting those and breaking?

an important point I guess is that we cannot have the path separator as part of the filename because it will be indistinguishable from a nested path. we should definitely avoid `/` and `\` (in ASCII, UTF-8, _and_ even UTF-16 because it's just the same character with a leading zero byte). Apart from that the filenames should be fine, but do we really want a filename from an image?

I'm curious to confirm if the duplicate check is failing or some other mechanism and then maybe there are things we should harden based on that. I'm not sure that arbitrarily replacing characters here is an optimal solution.

sandymcfadden:
Thanks @dmsnell. I admit I went for the quick fix that I could think of here. 

The problem is happening in the update function being used in [newNameCounts](https://github.com/Automattic/simplenote-electron/blob/fix/export-duplicate-filenames/lib/utils/export/to-zip.ts#L90)

With my sample notes above it should return an object like this.
```
{
  "!-Simplenote-com--httpsgithub-comAutomat": 1
}
```

Instead, it is returning this without the characters stripped out.

```
{
  "!": {
    "Simplenote": {
      "com": {
        "(httpsgithub": {
          "comAutomat": 1
        }
      }
    }
  }
}
```

The right way would be to fix this instead. I'm not sure right off how to do that, but can look into it.

sandymcfadden:
I've pushed an update now that instead of replacing characters from the file name I encode the file name and use the encoding to check for duplicates.

dmsnell:
> The right way would be to fix this instead. I'm not sure right off how to do that, but can look into it.

that's it. it's `lodash.update` interpreting the filename as an object path selector. we can rip it out and it should work fine in browsers. additionally why not make it a `Map` since that's available too now.

```ts
const toUniqueNames = ([notes, nameCounts], note) => {
	nameCounts.set(note.fileName, (nameCounts.get(note.fileName) ?? 0) + 1);
	const count = nameCounts.get(note.fileName);

	…
}
```

should have no problems. if we want to clean up the filenames beyond that we can but we shouldn't need to. this was a type error, where `lodash.update` is expecting a path-specifier and we gave it a raw string value (the types aren't present or indicated but I'm saying the problem was fundamentally that we have been passing in the wrong kind of information, and yeah I believe I wrote this code so it was my error)

sandymcfadden:
That sounds like an even better plan. Thank you yet again @dmsnell! I have it updated to use a Map now.

lib/utils/export/to-zip.ts
```diff
@@ -81,18 +81,17 @@ const appendTags = (note) => {
...
+ * @param {[Array, Map]} accumulator list of note objects for export and filename counts
  * @param {Object} note note object
  * @returns {[Array, Object]} final note list and accumulating filename counts
  */
 const toUniqueNames = ([notes, nameCounts], note) => {
-  const newNameCounts = update(nameCounts, note.fileName, (n) =>
-    n || 0 === n ? n + 1 : 0
-  );
-  const count = newNameCounts[note.fileName];
+  nameCounts.set(note.fileName, (nameCounts.get(note.fileName) ?? -1) + 1);
```
dmsnell: why start at -1? is it to avoid creating a file with `-0`?

could we reverse the order of these queries?

```ts
count = nameCounts.get(note.fileName) ?? 0;
noteCounts.set(note.fileName, count + 1);
```

sandymcfadden: Yup, that is why I used the -1. Is there a reason you prefer not to use the -1? Is it for readability or something else I'm missing?

dmsnell: `-1` stands out to me because it's a value we know we'll never want. there is no occasion where we'll want a negative number appended to a filename. appearing in the source code then stands out as a mismatch; it breaks the symmetry between the concept and implementation.

put another way, it's a clue that maybe we overlooked something about how we were doing it. in this case I think the discrepancy was related to the ordering of our data dependencies and something as simple as swapping those lines resulted in an interface that doesn't suffer from the question that led us to introduce `-1` and it eliminates the need to read _twice_ from the map to get the count.

because of these things it does seem more readable to me, because it doesn't prompt me to ask, "now why is that there?"

dmsnell (approved):
the fix looks straightforward; the root cause identified (`lodash.update` was interpreting the "filename" as a property path, thus interpreting `.` and array-like indexing with brackets); and nothing should be affected by this.

I did not test this so please lean on your own testing or rely on other review.

lib/utils/export/to-zip.ts
sandymcfadden: Thank you!

PR #2695, opened 2021-02-25
Update account creation flow

belcherj:
### Fix

Updates signup flow to new, secure version. See p2XJRt-2LG-p2

n.b. This is in the 2.8.0 milestone as we want to include it in that code freeze, but we'll also want to cut a 2.7.1 hotfix so that we can ship this fix ASAP.

### Test

1. While signed out of the app, sign up for a new account.
2. Verify that after clicking the verification email and choosing a password, you are properly redirected to the app and signed in with the newly-created account.
3. While signed in on the app, sign up for a different account on the web and click on the verification email.
4. The app should prompt you to log out.

### Release

Updated new account signup flow


lib/state/electron/middleware.ts
```diff
@@ -7,6 +7,11 @@ import * as S from '../';
 const debug = debugFactory('electron-middleware');
 
 export const middleware: S.Middleware = ({ dispatch, getState }) => {
+  window.electron.receive('tokenLogin', (url) => {
+    const { searchParams } = new URL(url);
+    const email = searchParams.get('email');
```
dmsnell: expect this to be base64-encoded, meaning we'll have to decode it

lib/state/ui/actions.ts
```diff
@@ -91,6 +91,20 @@ export const showDialog: A.ActionCreator<A.ShowDialog> = (
   dialog,
 });
 
+export const showTokenLoginLogout: A.ActionCreator<A.ShowTokenLoginLogout> = (
+  email: string
+) => ({
+  type: 'SHOW_TOKEN_LOGIN_LOGOUT',
```
dmsnell: this action name is pretty confusing to me. I don't know if there are other options but I'm not sure what it's purpose is, what interaction it's matching.

lib/app.tsx
```diff
@@ -198,6 +199,7 @@ class AppComponent extends Component<Props> {
     return (
       <div className={appClasses}>
         {showEmailVerification && <EmailVerification />}
+        {showTokenLoginLogout && <TokenLoginLogout />}
```
dmsnell: is it viable to leave this all in `boot` so that the magic link doesn't need to handled from "inside" the authenticated part of the app? or were you intentionally wanting to be able to show all the notes in the background with this modal? I was thinking that if we handle this in `boot` then we don't need a feedback loop for the token from inside the app and we can handle it all from a prompt, "Stay logged in as 'Dale'?   |   Login as 'Chip' instead"

if in-app, we're dropping the token altogether, right? it's just "Stay logged in as 'Dale'?   |   Logout"?

desktop/app.js
```diff
@@ -211,16 +212,18 @@ module.exports = function main() {
     // Protocol handler for platforms other than macOS
     // argv: An array of the second instance’s (command line / deep linked) arguments
     // The last index of argv is the full deeplink url (simplenote://SOME_URL)
+    throw 'blah';
```
dmsnell: blarg

lib/state/electron/middleware.ts
codebykat: Weirdly, it seems to be working as-is...

lib/app.tsx
codebykat: I experimented a bit with this and I think I like it where it is. You're already logged in to this account, so it makes sense that it is a modal alert similar to the email verification one.

Correct that we're dropping the token in this case and just prompting the user to log out if desired. They would then have to manually enter the new username and password.

lib/state/electron/middleware.ts
codebykat: I think this is okay, as per https://github.com/Automattic/simplenote-ios/pull/1182#discussion_r584943922

lib/state/ui/actions.ts
codebykat: I changed this component to `AlternateLoginPrompt` and the action to`SHOW_ALTERNATE_LOGIN_PROMPT`

lib/alternate-login-prompt/style.scss
```diff
@@ -0,0 +1,77 @@
...
+  justify-content: center;
+  align-items: center;
+  background: rgba($studio-gray-20, 0.75);
+}
+
+.alternate-login__dismiss {
+  display: flex;
+  height: 60px;
+  justify-content: flex-end;
+  line-height: 60px;
```
dcalhoun:
```suggestion
  align-items: center;
```

We should likely apply the fix from #2711 here to avoid the cut off close icon in macOS Safari.

safari-modal-close

lib/alternate-login-prompt/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <WarningIcon />
+      </span>
+      <h2>Logout?</h2>
+      <p>
+        You are already logged in as <strong>{email}</strong>.
+      </p>
+      <p>
+        You tried logging in with <strong>{alternateLoginEmail}</strong>. Would
+        you like to log out?
+      </p>
```
dcalhoun: We should likely apply the fix from #2712 for email addresses listed here to avoid long email addresses breaking the modal design.

dcalhoun (commented):
I know this is a work in progress, but figured I'd share some knowledge regarding issues that were copied into this branch.

lib/alternate-login-prompt/style.scss
```diff
@@ -0,0 +1,77 @@
...
+    a {
+      padding-right: 12px;
+    }
+    .button-borderless {
+      color: $studio-simplenote-blue-50;
+    }
+  }
+
+  &:focus {
+    outline: 0;
```
dcalhoun, line 70: One additional note — what do you think about not setting `outline: 0`? Keyboard users rely upon focus styles to track focus selection in the UI. Alternatively, we could design our own focus style that does not rely upon the default `outline` styling. 

I also posed this question in https://github.com/Automattic/simplenote-electron/pull/2685#discussion_r590577004.

lib/auth/index.tsx
```diff
@@ -60,14 +62,58 @@ export class Auth extends Component<Props> {
...
+          {isElectron && isMac && <div className="login__draggable-area" />}
+          <div className="accountRequested">
+            <MailIcon />
+            <p className="accountRequested__message theme-color-fg">
+              We&apos;ve sent an email to{' '}
+              <strong>{this.props.emailSentTo}</strong>. Please check your inbox
+              and follow the instructions.
+            </p>
+            <p className="accountRequested__footer theme-color-fg-dim">
+              Didn&apos;t get an email? Please contact{' '}
```
belcherj: This copy should be updated to: 

`Didn't get an email? You may already have an account. Contact support@simplenote.com for help.`

lib/alternate-login-prompt/style.scss
codebykat: This seems out of scope for this PR. If we choose to remove outlines or design our own focus style, we can update this along with all the other extant dialogs.

belcherj (commented):
The electron app logged in fine however it asked me to verify even though the account was verified upon creation.

I got the popup asking me if I would like to logout and that worked well.

From the PR description, I am unsure what bugs I should be looking for.

lib/auth/index.tsx
```diff
@@ -60,14 +62,58 @@ export class Auth extends Component<Props> {
     const helpMessage = isCreatingAccount
       ? 'Already have an account?'
       : "Don't have an account?";
-    const errorMessage = isCreatingAccount
-      ? 'Could not create account. Please try again.'
-      : 'Could not log in with the provided email address and password.';
+
+    const errorMessage = isCreatingAccount ? (
+      <>
```
sandymcfadden, line 67: This doesn't really matter, but for my own curiosity. I've seen in this PR both `<>` and `<Fragment>` used. Does that matter, or is there a preference one over the other?

codebykat: I learned about `<>` like last week. I don't know which we prefer but I guess I can make it consistent..

I think it's the case that `<>` doesn't require importing `Fragment`? So maybe that's better

belcherj: We should decide what to use and stick with it. I may have used `<>` since I was fast coding and not really thinking about where I was and code style. For now, let's change it to use Fragment since that is the code standard. It all compiles to the same thing so it is whatever feels right to yous.

lib/state/action-types.ts
```diff
@@ -101,15 +102,18 @@ export type SetUnsyncedNoteIds = Action<
   { noteIds: T.EntityId[] }
 >;
 export type ShowAllNotes = Action<'SHOW_ALL_NOTES'>;
-export type ShowUntaggedNotes = Action<'SHOW_UNTAGGED_NOTES'>;
```
sandymcfadden, line 104: Is this going to cause regression and remove the untagged notes option which was added?

sandymcfadden: Nevermind. I see added back. I didn't go far enough.

PR #2701, opened 2021-02-28
[feature] Add setting to exclude tags from default view

### LINKED ISSUE #2729: Exclude notes from default view by tag name
### What

This is a feature request for my pull request, #2701.

Adds 'Hidden tags' to settings. Any note that contains at least one of the tags listed under Hidden tags will not be visible in the default 'All Notes' view. They will still be visible in the tag view.

### Why

Users may want to be able to exclude certain notes from the default All Notes view if they have a large number of notes. This allows specific filtering of notes based on tags.

### How

See my pull request.

Symbitic:
### Fix

This adds a feature to allow certain tags to be excluded from the default view (All Notes). When the default view is shown, any note that contains at least one tag listed under hidden tags will not be visible in the notes list.

This only affects the notes list under All Notes. The tags are still visible in the tag list, and any items that would be hidden under All Notes will be visible when viewing that tag.

The list of hidden tags can be changed in Display Settings.

### Test

1. Go to settings. Click on Display.
2. Scroll down until you see Hidden tags.
3. Add at least one tag name and hit 'Enter'.
4. All notes that contain that tag will no longer be visible under All Notes.
5. Click the trash icon to delete the tag, and it will be listed again.

### Release

* Add feature to allow notes to be hidden from All Notes by tag name.

### DIFF
--- a/lib/dialogs/list-settings-group.tsx
+++ b/lib/dialogs/list-settings-group.tsx
@@ -1,0 +1,86 @@
+import React, { FunctionComponent, useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import PanelTitle from '../components/panel-title';
+
+type ItemsType = string[];
+
+type Props = {
+  description?: string;
+  items: ItemsType;
+  onChange: Function;
+  pattern?: string;
+  placeholder?: string;
+  title: string;
+};
+
+const ListSettingsGroup: FunctionComponent<Props> = ({
+  description,
+  items,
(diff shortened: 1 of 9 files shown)


sandymcfadden:
Thanks again for the PR and the interest in contributing. As I mentioned on [the feature request ](https://github.com/Automattic/simplenote-electron/issues/2729#issuecomment-804825565)this isn't something we are looking to add at this time.

PR #2705, opened 2021-03-02
Keyboard shortcuts: Fix Ctrl+G and other intermittently-firing shortcuts on Win/Linux

### LINKED ISSUE #2420: Search highlight isn't working with Ctrl + G (Find Next/Again)
Before updating the app to version 2.0.0, pressing Ctrl + G after performing a search on the app will highlight the search keyword in a darker color (e.g. brown), but this isn't working now.

### Expected
I would expect the highlighted keyword will show a distinct color from the other search keywords.

### Observed
The highlighted keyword shows the same color with the other search keywords.

### Reproduced

**Note:** The issue was originally reported in 3424483-zen but I was able to replicate this on the Simplenote web app.

1. Go to https://app.simplenote.com
2. Type a keyword to search the notes on the search box
3. Open up a note under the search result
4. Press Ctrl + G to highlight the search keyword
5. See that the color of the highlighted search keyword is the same with the other keywords

**Screen recording**

Screen Capture on 2020-10-26 at 13-48-58

**Screenshot shared in 3424483-zen**

hhcbbpndidnnncln

### Where did you see the bug

**Information shared in 3424483-zen**
- OS: Windows
- OS version: Windows 10

**Information for the test on Simplenote web app**
- OS: macOS
- OS version: 10.15.7
- Browser version: Version 86.0.4240.111 (Official Build) (x86_64)

codebykat:
### Fix

Fixes #2420 

Some keyboard shortcuts were silently failing on Win/Linux builds. After adding logging and debugging this I was able to determine that the `if focusedWindow` check was failing. I still don't know exactly why, as this works on OSX.

Changing it to explicitly check if the `focusedWindow` is `undefined` seems to fix the issue in my testing. I have only tested this on Windows, not Linux.

After some research, I think it would be safe to remove this guard entirely. The [documentation](https://github.com/electron/electron/blob/master/docs/api/menu-item.md) does state that it can be `undefined`, but only in the case where no window is open, which only applies to OSX. (confirmed by a comment [here](https://github.com/electron/electron/issues/24144))

### Test

1. Download the Windows and/or Linux builds from CircleCI
2. Try all keyboard shortcuts, including Ctrl+G, and verify they all work with no errors in the console

### Release

Fixed a bug that caused Ctrl+G and some other shortcuts to fail on Windows/Linux builds

### DIFF
--- a/desktop/menus/utils.js
+++ b/desktop/menus/utils.js
@@ -24,7 +24,7 @@ const editorCommandSender = (arg) => {
 
 const commandSender = (commandName, arg) => {
   return (item, focusedWindow) => {
-    if (focusedWindow) {
+    if (typeof focusedWindow !== 'undefined') {
       focusedWindow.webContents.send(commandName, arg);
     }
   };
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -77,7 +77,7 @@ class AppComponent extends Component<Props> {
 
   handleShortcut = (event: KeyboardEvent) => {
     const { hotkeysEnabled } = this.props;
-    const shouldHandleBrowserShortcuts = !window.electron || !isMac;
+    const shouldHandleBrowserShortcuts = !window.electron;
 
     // Handle search shortcuts even if keyboard shortcuts are disabled.
     if (shouldHandleBrowserShortcuts) {
@@ -116,7 +116,6 @@ class AppComponent extends Component<Props> {
   // handle all keyboard shortcuts that are duplicated in the Electron menus
   // this listener is only called in browsers, as otherwise the
   // menu will trigger them via the provided Accelerator, so we don't need a listener
-  // n.b. we're also running this on Win/Linux builds as they seem to be broken otherwise
   handleBrowserShortcut = (event: KeyboardEvent) => {
     const { ctrlKey, metaKey, shiftKey } = event;
     const key = event.key.toLowerCase();


sandymcfadden (changes requested):
I've been testing on Linux and the only one that doesn't seem to be working as expected is the open tag list `Ctrl + Shift + U`. If you are focused on the editor this doesn't fire while all others seem to.

It does work on the Mac build as expected. Should this be moved into the `handleBrowserShortcut` function?

codebykat:
> I've been testing on Linux and the only one that doesn't seem to be working as expected is the open tag list `Ctrl + Shift + U`. If you are focused on the editor this doesn't fire while all others seem to.

Hmm. It works on Windows for me, regardless of what has focus.

Are you on Ubuntu? Is it possible Ubuntu is eating it? https://superuser.com/questions/358749/how-to-disable-ctrlshiftu-in-ubuntu-linux

Also, how does this compare to the functionality in the 2.7.0 and/or 2.6.0 releases? I don't know if this shortcut gets much use, so maybe it's never worked properly..

sandymcfadden:
You're right this seems to be the case that it is a global keyboard shortcut on some versions of Linux. I tried on 2.6.0 and had the same issue. Should we look at changing this shortcut, or be OK with it not working on Linux?

codebykat:
I'm going to merge this for 2.8.0 to get the Windows shortcuts fixed, and open a follow-up issue. Thanks @sandymcfadden!

PR #2706, opened 2021-03-02
Refactoring reducer states showTrash and openedTag with showCollection

illusaen:
### Fix
Since `showTrash`, `openedTag`, and the [upcoming] `untaggedNotes` states are mutually exclusive, we now combine the booleans into one, `showCollection`. This stops bugs like the one where filtering by trashed notes and then filtering by tags ends up finding only notes with that tag that are deleted.

### Screenshots

  <details>
    <summary>Simple filter usage</summary>

simple
  </details>

<details>
    <summary>Renaming tag</summary>

renaming
  </details>

<details>
  <summary>Delete current tag</summary>

  delete-current-tag
</details>

<details>
  <summary>Delete other tag</summary>

  delete-other-tag
</details>

#### Bug
1. Create note and delete it and create another note with a tag.
2. Open sidebar.
3. Click Trash, see that deleted note only shows up.
4. Click filter for that tag, see that no notes show up (with bug), and that the note with the tag should show up (bug fixed).

#### General
1. Create note with tag A.
2. Create note with tag B.
3. Create note and delete it.
4. Open sidebar.
5. Click Trash, see that deleted note shows up.
6. Click tag A, see that tag A note shows up.
7. Click All Notes, see that all notes are there again.
8. Rename tag A to C, see that search list does not change but tag is successfully updated.
9. Click tag B. Rename tag B to D, see that search list updates to show current note with tag changed.

### Release
No user-facing changes here.

### DIFF
--- a/lib/navigation-bar/index.tsx
+++ b/lib/navigation-bar/index.tsx
@@ -17,7 +17,7 @@ import * as T from '../types';
 type StateProps = {
   autoHideMenuBar: boolean;
   isDialogOpen: boolean;
-  openedTag: T.TagEntity | null;
+  openedTag: T.TagHash | null;
   showNavigation: boolean;
   showTrash: boolean;
 };
@@ -126,13 +126,16 @@ export class NavigationBar extends Component<Props> {
 
 const mapStateToProps: S.MapState<StateProps> = ({
   settings,
-  ui: { dialogs, openedTag, showNavigation, showTrash },
+  ui: { dialogs, showNavigation, showCollection },
 }) => ({
   autoHideMenuBar: settings.autoHideMenuBar,
   isDialogOpen: dialogs.length > 0,
-  openedTag,
+  openedTag:
+    showCollection.type === 'tag' && showCollection.tagHash
+      ? showCollection.tagHash
+      : null,
   showNavigation,
-  showTrash,
+  showTrash: showCollection.type === 'trash',
 });
 
 const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
--- a/lib/navigation-bar/index.tsx
+++ b/lib/navigation-bar/index.tsx
@@ -16,8 +16,8 @@ import * as T from '../types';
 
 type StateProps = {
   autoHideMenuBar: boolean;
-  isDialogOpen: boolean;
   collection: T.Collection;
+  isDialogOpen: boolean;
   showNavigation: boolean;
 };
 
@@ -126,11 +126,11 @@ export class NavigationBar extends Component<Props> {
 
 const mapStateToProps: S.MapState<StateProps> = ({
   settings,
-  ui: { dialogs, showNavigation, collection },
+  ui: { collection, dialogs, showNavigation },
 }) => ({
   autoHideMenuBar: settings.autoHideMenuBar,
-  isDialogOpen: dialogs.length > 0,
   collection,
+  isDialogOpen: dialogs.length > 0,
   showNavigation,
 });
 
--- a/lib/navigation-bar/index.tsx
+++ b/lib/navigation-bar/index.tsx
@@ -17,7 +17,7 @@ import * as T from '../types';
 type StateProps = {
   autoHideMenuBar: boolean;
   isDialogOpen: boolean;
-  showCollection: T.Collection;
+  collection: T.Collection;
   showNavigation: boolean;
 };
 
@@ -60,7 +60,7 @@ export class NavigationBar extends Component<Props> {
   }: {
     selectedRow: 'all' | 'trash' | 'untagged';
   }) => {
-    return this.props.showCollection.type === selectedRow;
+    return this.props.collection.type === selectedRow;
   };
 
   render() {
@@ -126,11 +126,11 @@ export class NavigationBar extends Component<Props> {
 
 const mapStateToProps: S.MapState<StateProps> = ({
   settings,
-  ui: { dialogs, showNavigation, showCollection },
+  ui: { dialogs, showNavigation, collection },
 }) => ({
   autoHideMenuBar: settings.autoHideMenuBar,
   isDialogOpen: dialogs.length > 0,
-  showCollection,
+  collection,
   showNavigation,
 });
 
--- a/lib/navigation-bar/index.tsx
+++ b/lib/navigation-bar/index.tsx
@@ -17,9 +17,8 @@ import * as T from '../types';
 type StateProps = {
   autoHideMenuBar: boolean;
   isDialogOpen: boolean;
-  openedTag: T.TagHash | null;
+  showCollection: T.Collection;
   showNavigation: boolean;
-  showTrash: boolean;
 };
 
 type DispatchProps = {
@@ -56,11 +55,12 @@ export class NavigationBar extends Component<Props> {
   };
 
   // Determine if the selected class should be applied for the 'all notes' or 'trash' rows
-  isSelected = ({ isTrashRow }: { isTrashRow: boolean }) => {
-    const { showTrash, openedTag } = this.props;
-    const isItemSelected = isTrashRow === showTrash;
-
-    return isItemSelected && !openedTag;
+  isSelected = ({
+    selectedRow,
+  }: {
+    selectedRow: 'all' | 'trash' | 'untagged';
+  }) => {
+    return this.props.showCollection.type === selectedRow;
   };
 
   render() {
@@ -75,13 +75,13 @@ export class NavigationBar extends Component<Props> {
           />
           <NavigationBarItem
             icon={<TrashIcon />}
-            isSelected={this.isSelected({ isTrashRow: true })}
+            isSelected={this.isSelected({ selectedRow: 'trash' })}
             label="Trash"
             onClick={this.onSelectTrash}
           />
           <NavigationBarItem
             icon={<NotesIcon />}
-            isSelected={this.isSelected({ isTrashRow: false })}
+            isSelected={this.isSelected({ selectedRow: 'all' })}
             label="All Notes"
             onClick={onShowAllNotes}
           />
@@ -130,12 +130,8 @@ const mapStateToProps: S.MapState<StateProps> = ({
 }) => ({
   autoHideMenuBar: settings.autoHideMenuBar,
   isDialogOpen: dialogs.length > 0,
-  openedTag:
-    showCollection.type === 'tag' && showCollection.tagHash
-      ? showCollection.tagHash
-      : null,
+  showCollection,
   showNavigation,
-  showTrash: showCollection.type === 'trash',
 });
 
 const mapDispatchToProps: S.MapDispatch<DispatchProps> = {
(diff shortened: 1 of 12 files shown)


lib/state/ui/reducer.ts
```diff
@@ -192,15 +192,21 @@ const openedRevision: A.Reducer<[T.EntityId, number] | null> = (
...
-      return null;
+      return { type: 'all' };
     case 'OPEN_TAG':
-      return tagHashOf(action.tagName);
+      return { type: 'tag', tagHash: tagHashOf(action.tagName) };
     case 'TRASH_TAG':
-      return tagHashOf(action.tagName) === state ? null : state;
+      return tagHashOf(action.tagName) === state.tagHash
+        ? { type: 'all' }
+        : state;
```
dmsnell: love how we can make this clear update based on the deletion of a tag without having to worry about updating a few different elements in different places.

illusaen: It definitely is more descriptive :)

dmsnell (commented):
love this work @illusaen - thanks for tackling it!

there are a few things that stand out to me at first glance over this I'd like your thoughts on.

 - "showTrash" seems clear since it's on or off, but `showCollection` makes me think it's some kind of boolean toggle and not a selector. would something like `openedCollection`, `visibleCollection`, `noteCollection`, or just `collection` be clearer?

 - we have `selectors.ts` where some complicated state queries can go in order to remove the duplication of the same complex process in various files, as well as remove the risk of introducing little errors in those duplicate places. I could easily imagine two such new selectors being valuable here, `collection(state) -> 'trash' | 'all' | 'tag' | 'untagged'` (the name of this could match whatever name we settle on for the value in state, but it doesn't have to), and `openedTag(state) -> string | null`, which can be used for those components which are only supposed to be visible when we're in an opened tag view.
 - it looks like we've updated `mapStateToProps` to prevent breaking the app, but some of the uses are now a bit awkward. there are places where passing in the new `collection` value could simplify the UI components themselves, places where we currently pass in the set of flags.

```ts
const sidebarTitle = (collection: Collection) => {
	switch (collection.type) {
		case 'trash':
			return 'Trash';

		case 'tag':
			return 'Notes With Selected Tag'; // or return `Notes with tag "{collection.tagName}"` - left for another PR

		case 'all':
			return 'All Notes';

		case 'untagged':
			return 'Untagged Notes';
	}
}

<div className="notes-title">{sidebarTitle(collection)}</div>
```

I would have left line-level comments but I think these things cover a more broad aspect touching many files so I left them here.

lib/note-list/no-notes.tsx
```diff
@@ -18,8 +18,12 @@ type EmptyNoteListPlaceholder = {
...
   const hasLoaded = useSelector((state: S.State) => state.ui.hasLoadedNotes);
   const searchQuery = useSelector((state: S.State) => state.ui.searchQuery);
-  const showTrash = useSelector((state: S.State) => state.ui.showTrash);
-  const openedTag = useSelector((state: S.State) => state.ui.openedTag);
+  const showTrash = useSelector(
+    (state: S.State) => state.ui.showCollection.type === 'trash'
+  );
+  const openedTag = useSelector(
+    (state: S.State) => state.ui.showCollection.tagHash
+  );
```
dmsnell: these update together and I don't see any reason to create two hooks for them. if we remodel the data (as noted in the PR comment) then we can get away with replacing the two with a single one.

```ts
const collection = useSelector((state: S.State) => state.ui.collection);

switch (collection.type) {
	case 'tag':
		return {message: `No notes tagged "${collection.tagName}"`, icon: <TagIcon />, button: null};

	case 'trash':
		return {message: 'You trash is empty', icon: <TrashIcon />, button: null};

	case …
}
```

illusaen: I think passing in the one filter makes sense! I honestly didn't even think of it after seeing how many changes I already had, lol. 

What do you think about `collectionFilter`?

lib/types.ts
```diff
@@ -50,6 +50,12 @@ export type VerificationState =
   | 'unverified'
   | 'verified';
 
+export type Collection =
+  | { type: 'all' }
+  | { type: 'trash' }
+  | { type: 'tag'; tagHash: TagHash | null }
```
dmsnell: previously I might have hinted that this should be the tag hash, so if I did I apologize. based on how we use this I think we'll want to store the `tagName` instead, since we display it.

if we _do_ want to store the `tagHash` to make sure we're consistent all over, then we need to perform a lookup to display the tag name, something we could use a selector for, if it doesn't already exist.

`No notes tagged "${tagName(collection.tagHash)}"`

illusaen: To be honest you didn't; I just set it to `TagHash` because `openedTag` was a `TagHash`. I think selectors make sense, with it as a `Tag` instead.

dmsnell: it's easily to collapse a tag name into a tag hash than to go backwards, since collapsing to the hash can be done with just the name but going from the hash to the name involves a lookup in state.

lib/note-list/no-notes.tsx
dmsnell: let's see if @codebykat has thoughts on the name. I'm not as excited about that because it doesn't feel like a "filter" to me.

there's one aspect I haven't commented on, which might be "the elephant in the room" but I haven't raised because it implies somewhat more extensive refactoring, and that's the search query. as it stands it seems proper to keep the search query separate since it applies in all opened collections/views; there's a chance it belongs inside this state description. let's leave that for now though since it might still be best as-is.

that being said the search query is really what "filters" the notes and these are more like drawers we are opening. it's true that they filter, but they represent different spaces or collections.

sorry this must sound pedantic, naming stuff is hard.

lib/types.ts
dmsnell: > I think selectors make sense, with it as a Tag instead.

Note that `Tag` is a type we don't use too often inside the UI. `TagName` is what we use for its namesake. Fun fact! it's a "branded type" meaning that the type system won't let you assign arbitrary strings (or a tag hash) to a value that's typed as a `TagName` unless you declare it `as TagName`. This was introduced to remove the confusion and opportunities for bugs around treating the two as if they are the same.

lib/note-list/no-notes.tsx
illusaen: Okay! I'm honestly not too bothered by the name, renamed to `collection`.

I also combined the different selectors into the file.

lib/types.ts
illusaen: Set as `TagName` instead. :)

lib/note-list/no-notes.tsx
codebykat: I'm fine with `collection`.

lib/navigation-bar/index.tsx
```diff
@@ -126,13 +126,12 @@ export class NavigationBar extends Component<Props> {
 
 const mapStateToProps: S.MapState<StateProps> = ({
   settings,
-  ui: { dialogs, openedTag, showNavigation, showTrash },
+  ui: { dialogs, showNavigation, collection },
```
codebykat: Just a weird nit, we tend to (try to) put these in alphabetical order

lib/navigation-bar/index.tsx
```diff
@@ -126,13 +126,12 @@ export class NavigationBar extends Component<Props> {
...
 }) => ({
   autoHideMenuBar: settings.autoHideMenuBar,
   isDialogOpen: dialogs.length > 0,
-  openedTag,
+  collection,
```
codebykat, line 132: Ditto on the alphabetical order comment - and a bunch of other places too, which I am not going to be super annoying by leaving a line comment for each one 😂

codebykat (commented):
This is a beautiful refactor (in terms of reading the code) and is working well for me. I have a very minor request which is alphabetizing arguments/props, and I know the codebase is terribly inconsistent about that, but I do try to fix them when I can 😐 

Given that I'd love to get this in so we can move forward on the untagged note search!

lib/navigation-bar/index.tsx
illusaen: Will do, fixing :)

illusaen: @codebykat Alphabetized!

lib/search-field/index.tsx
```diff
@@ -13,9 +14,8 @@ import type * as T from '../types';
 const KEY_ESC = 27;
 
 type StateProps = {
-  openedTag: T.Tag | null;
+  openedTag: T.TagName | undefined;
```
dmsnell: `null` is a clearer signal than `undefined` so it could be nice to have the selector return that if no tag is open.

fun fact! `undefined` is ironically a value, which in earlier versions of JavaScript could be reassigned. it can still come about in two separate conditions: one from the absence of a value, and two in the passing of the `undefined` value itself.

```ts
const o = {a: undefined};

o.a // undefined
o.b // undefined
o.hasOwnProperty('a') // true
o.hasOwnProperty('b') // false
```

lib/search/index.ts
```diff
@@ -445,8 +447,14 @@ export const middleware: S.Middleware = (store) => {
...
 
-        if (searchState.openedTag === oldHash) {
-          searchState.openedTag = newHash;
+        if (
+          searchState.collection.type === 'tag' &&
+          searchState.collection.tagName === action.oldTagName
+        ) {
+          searchState.collection = {
+            type: 'tag',
+            tagName: action.newTagName,
```
dmsnell, line 456: very nice catch!

though here in the search middleware it should be fine to rely on the `tagHash`, since we're not _displaying_ the tag. everything compared against for searching and filtering will rely on the `tagHash`

lib/tag-list/index.tsx
```diff
@@ -196,13 +196,13 @@ const mapStateToProps: S.MapState<StateProps> = (state) => {
   const {
     data,
     settings: { sortTagsAlpha },
-    ui: { editingTags, openedTag },
+    ui: { editingTags },
   } = state;
   return {
     editingTags,
+    openedTagHash: selectors.openedTagHash(state),
```
dmsnell: it would be preferable to rely on `tagHash` here than creating another selector, since the selectors are best when they are slim and specific to getting data out of state, calculating it _if necessary_

lib/search/index.ts
illusaen: Good point; I wanted to keep the signature the same in both though in case maybe later all the searchState could be placed in redux.

lib/tag-list/index.tsx
illusaen: Fixed!

lib/search-field/index.tsx
illusaen: Fixed!

codebykat (approved):
Checked with @dmsnell on this, it's looking great. Merge at will! 🙇

dmsnell:
whoops! looks like we didn't squash before merging. repo history is now a lie again. oh well.

PR #2707, opened 2021-03-02
Improve navigation accessibility

### LINKED ISSUE #979: Improve accessibility for sidebars
The NavigationBar and NoteInfo sidebars should basically act like a modal, in that:

1. <kbd>Esc</kbd> should close it
1. Keyboard focus should jump to the first focusable element when the sidebar is activated
1. Keyboard focus should be trapped inside the sidebar
1. Keyboard focus should return to the original element when the sidebar is closed

We should also consider `tablist` / `tabpanel` semantics for the NavigationBar so it makes sense in screen readers.

### LINKED ISSUE #980: Improve accessibility for Revision Selector
The RevisionSelector should act like a modal, in that:

1. It should have role="dialog"
1. It should be `aria-hidden` when not visible on screen
1. <kbd>Esc</kbd> should close it
1. Keyboard focus should jump to the first focusable element when the dialog is activated
1. Keyboard focus should be trapped inside the dialog
1. Keyboard focus should return to the original element when the dialog is closed

dcalhoun:
### Fix

Fixes #979 and fixes #980. Improve navigation accessibility via the following:

- Add focus trap to navigation sidebar and revisions selector. 
- Hide unrelated content while navigation sidebar or revisions selector is open. 
- Reorder navigation elements to match visual presentation. 
- Adjust styles to improve active focus indication. 

<details>
<summary>VoiceOver Interaction: Navigation</summary>

https://user-images.githubusercontent.com/438664/109722627-90289900-7b72-11eb-97a9-fb9b47e3d0e6.mov

</details>

<details>
<summary>VoiceOver Interaction: Revisions Selector</summary>

https://user-images.githubusercontent.com/438664/109820267-75016c00-7bfa-11eb-884c-c2beb85dbd2e.mov

</details>

| Styles Before | Styles After |
| --- | --- |
| styles-before | styles-after |

#### Scenario 1

1. Enable macOS VoiceOver or similar screen reader technology.
1. Open the navigation sidebar by pressing the menu button in the top-left
   corner of the screen.

Expected Outcome:

* The first element in the navigation has focus.
* Moving selection backwards does not move focus past "All Items" menu item.
* Moving selection forwards does not move focus past "About" menu item.
* Pressing <kbd>Esc</kbd> dismisses the navigation, deactivates the focus
  trap, and returns focus to the "Menu" button.
* Clicking outside of the navigation dismisses the navigation, deactivates the focus
  trap, and returns focus to the "Menu" button.

#### Scenario 2

1. Enable macOS VoiceOver or similar screen reader technology.
1. Open the navigation sidebar by pressing the menu button in the top-left
   corner of the screen.
1. Open "About" dialog by pressing the "About" link at the end of the navigation
   list.

Expected Outcome:

* Navigation focus trap is disabled while dialog is open.
* Dialog contents are focused.
* Dismissing the dialog with <kbd>Esc</kbd> causes "About" link to regain focus.

#### Scenario 3

1. Enable macOS VoiceOver or similar screen reader technology.
1. Open the revisions selector by pressing the history button in the top-right
   corner of the screen.

Expected Outcome:

* The first element in the navigation has focus.
* Moving selection backwards does not move focus past "History" heading.
* Moving selection forwards does not move focus past "Restore" button.
* Pressing <kbd>Esc</kbd> dismisses the navigation, deactivates the focus
  trap, and returns focus to the note editor.
* Clicking outside of the navigation dismisses the navigation, deactivates the focus
  trap, and returns focus to the note editor.

### Release

- Improve navigation sidebar and revision selector accessibility for keyboards and screen readers.

### DIFF
--- a/lib/app-layout/index.tsx
+++ b/lib/app-layout/index.tsx
@@ -115,10 +115,16 @@ export class AppLayout extends Component<Props> {
       </TransitionDelayEnter>
     );
 
+    const hiddenByRevisions = showRevisions ? true : undefined;
+
     return (
-      <div className={mainClasses}>
+      <div
+        className={mainClasses}
+        aria-hidden={isNavigationOpen ? true : undefined}
+      >
         <Suspense fallback={placeholder}>
           <aside
+            aria-hidden={hiddenByRevisions}
             aria-label="Notes list"
             className="app-layout__source-column theme-color-bg theme-color-fg"
           >
@@ -132,9 +138,13 @@ export class AppLayout extends Component<Props> {
               aria-label="Note editor"
               className="app-layout__note-column theme-color-bg theme-color-fg theme-color-border"
             >
-              <NoteToolbar />
+              <NoteToolbar aria-hidden={hiddenByRevisions} />
               {showRevisions ? (
-                <NotePreview noteId={openedNote} note={openedRevision} />
+                <NotePreview
+                  aria-hidden={hiddenByRevisions}
+                  noteId={openedNote}
+                  note={openedRevision}
+                />
               ) : (
                 <NoteEditor />
               )}
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -30,9 +30,13 @@ type DispatchProps = {
   openNote: (noteId: T.EntityId) => any;
 };
 
-type Props = OwnProps & StateProps & DispatchProps;
+type Props = OwnProps &
+  StateProps &
+  DispatchProps &
+  React.HTMLProps<HTMLDivElement>;
 
 export const NotePreview: FunctionComponent<Props> = ({
+  'aria-hidden': ariaHidden,
   editNote,
   isFocused,
   note,
@@ -162,7 +166,7 @@ export const NotePreview: FunctionComponent<Props> = ({
   }, [note?.content, searchQuery, showRenderedView]);
 
   return (
-    <div className="note-detail-wrapper">
+    <div aria-hidden={ariaHidden} className="note-detail-wrapper">
       <div className="note-detail note-detail-preview">
         <div
           ref={previewNode}
--- a/lib/app.tsx
+++ b/lib/app.tsx
@@ -39,6 +39,7 @@ type StateProps = {
   showEmailVerification: boolean;
   showNavigation: boolean;
   showNoteInfo: boolean;
+  showRevisions: boolean;
   theme: 'light' | 'dark';
 };
 
@@ -182,6 +183,7 @@ class AppComponent extends Component<Props> {
       showEmailVerification,
       showNavigation,
       showNoteInfo,
+      showRevisions,
       theme,
     } = this.props;
 
@@ -201,7 +203,11 @@ class AppComponent extends Component<Props> {
       <div className={appClasses}>
         {showEmailVerification && <EmailVerification />}
         {showAlternateLoginPrompt && <AlternateLoginPrompt />}
-        {isDevConfig && <DevBadge />}
+        {isDevConfig && (
+          <DevBadge
+            aria-hidden={showNavigation || showRevisions ? true : undefined}
+          />
+        )}
         <div className={mainClasses}>
           {showNavigation && <NavigationBar />}
           <AppLayout />
@@ -223,6 +229,7 @@ const mapStateToProps: S.MapState<StateProps> = (state) => ({
   showEmailVerification: selectors.shouldShowEmailVerification(state),
   showNavigation: state.ui.showNavigation,
   showNoteInfo: state.ui.showNoteInfo,
+  showRevisions: state.ui.showRevisions,
   theme: selectors.getTheme(state),
 });
 
--- a/lib/components/dev-badge/index.tsx
+++ b/lib/components/dev-badge/index.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
 
-const DevBadge = () => (
-  <div aria-hidden="true" className="dev-badge">
-    DEV
-  </div>
-);
+const DevBadge = (props: React.HTMLProps<HTMLDivElement>) => {
+  return (
+    <div className="dev-badge" {...props}>
+      DEV
+    </div>
+  );
+};
 
 export default DevBadge;
--- a/lib/components/slider/index.tsx
+++ b/lib/components/slider/index.tsx
@@ -1,6 +1,6 @@
 import React, { ChangeEventHandler, FunctionComponent } from 'react';
 
-type Props = {
+type OwnProps = {
   disabled: boolean;
   onChange: ChangeEventHandler<HTMLInputElement>;
   min: number;
@@ -8,7 +8,10 @@ type Props = {
   value: number;
 };
 
+type Props = OwnProps & React.HTMLProps<HTMLInputElement>;
+
 export const Slider: FunctionComponent<Props> = ({
+  'aria-valuetext': ariaValueText,
   disabled,
   min,
   max,
@@ -16,6 +19,7 @@ export const Slider: FunctionComponent<Props> = ({
   onChange,
 }) => (
   <input
+    aria-valuetext={ariaValueText}
     aria-label="Select revision"
     className="slider"
     disabled={disabled}
--- a/lib/navigation-bar/index.tsx
+++ b/lib/navigation-bar/index.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import onClickOutside from 'react-onclickoutside';
+import FocusTrap from 'focus-trap-react';
 
 import ConnectionStatus from '../connection-status';
 import isEmailTag from '../utils/is-email-tag';
@@ -26,7 +26,7 @@ type StateProps = {
 
 type DispatchProps = {
   onAbout: () => any;
-  onOutsideClick: () => any;
+  onFocusTrapDeactivate: () => any;
   onSettings: () => any;
   onShowAllNotes: () => any;
   onShowUntaggedNotes: () => any;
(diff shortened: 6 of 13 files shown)


dmsnell (commented):
can you share a quick synopsis of the difference that `focus-trap-react` brings from `react-onclickoutside`?

also, should we be removing the `react-onclickoutside` dependency? if it's still in use in other modules, should they also be updated?

dcalhoun:
@dmsnell absolutely. This is still a bit of a work in progress, but I'll be sure to include information on the packages utilized.

dcalhoun:
> can you share a quick synopsis of the difference that `focus-trap-react` brings from `react-onclickoutside`?
> 
> also, should we be removing the `react-onclickoutside` dependency? if it's still in use in other modules, should they also be updated?

@dmsnell I went ahead and refactored the Revisions Selector to improve its accessibility and remove its usage of `react-onclickoutside`. The remaining location using `react-onclickoutside` is the notes Info sidebar, but @sandymcfadden mentioned that area is actively being refactored to a dialog, so I left it as-is. Once that Info sidebar is a dialog, `react-onclickoutside` could be removed from the project.

| Package | Gzipped Size | Click Outside | Focus Trap | Escape Dismiss |
| --- | --- | --- | --- | --- |
| [`react-onclickoutside`](https://github.com/Pomax/react-onclickoutside) | [1.6kb](https://bundlephobia.com/result?p=react-onclickoutside@6.10.0) | ✅ | ❌ | ❌ |
| [`focus-trap-react`](https://github.com/focus-trap/focus-trap-react)| [4.2kb](https://bundlephobia.com/result?p=focus-trap-react@8.4.2)  | ✅ | ✅ | ✅ |
| [`react-focus-lock`](https://github.com/theKashey/react-focus-lock) | [5.4kb](https://bundlephobia.com/result?p=react-focus-lock@2.5.0) | ❌ | ✅ | ❌ |
| [`react-focus-on`](https://github.com/theKashey/react-focus-on) | [8.7kb](https://bundlephobia.com/result?p=react-focus-on@3.5.1) | ✅ | ✅ | ✅ |

As represented above, I chose to leverage `focus-trap-react` because it provides functionality to "trap" focus within a portion of the UI, which improves keyboard and screen reader navigation. I opted against `react-focus-lock` due to its larger size<sup>1</sup>. Please let me know if believe an alternative approach may be better. Thanks!

---

1. `react-focus-lock` technically offers better tree shaking and supports [lazy loading some of its capabilities](https://github.com/theKashey/react-focus-lock#separated-usage). However, I decided introducing that complexity was not worth it at this time.

---

**Edit:** Updated chart to correct misrepresentation `react-focus-lock`. It does not include escape/click outside to dismiss. We must either build that ourselves or include `react-focus-on` instead, which includes `react-focus-lock`.

dcalhoun:
>We should also consider `tablist` / `tabpanel` semantics for the NavigationBar so it makes sense in screen readers.

One additional note is that I did not address the above from #979. I believe that will likely require a larger investigation and discussion regarding to how to best organized the navigation items and their relation to the note list and note editor. I.e. how to associate navigation items with specific states for the notes list and note editor. I plan to create a separate issue for exploring this piece.

dmsnell:
@dcalhoun we're not talking about a big difference in code size between those libraries _beyond_ `react-onclickoutside`. Looks like you noticed that `react-focus-lock` may end up compiling smaller if it's architecture is better suited to tree-shaking.

What are your thoughts on the quality and maintainership of those libraries? Is one clearly better for accessibility? I don't have experience with either.

dcalhoun:
> What are your thoughts on the quality and maintainership of those libraries? Is one clearly better for accessibility? I don't have experience with either.

@dmsnell great questions!

| Package | Weekly Downloads | Maintenance | Quality |
| --- | --- | --- | --- |
| [`focus-trap-react`](https://github.com/focus-trap/focus-trap-react) | [226k](https://www.npmjs.com/package/focus-trap-react) | [Maintainers changed](https://github.com/focus-trap/focus-trap-react/pull/92) in Aug 2020. | Source quality appears good. Has test coverage. Provides accessibility features we need. Not as feature rich as competing packages, which means source is a bit more straightforward.  |
| [`react-focus-lock`](https://github.com/theKashey/react-focus-lock) | [1,646k](https://www.npmjs.com/package/react-focus-lock) | Author appears active in package and community, particularly on this topic. | Source quality appears very good. Has test coverage. Provides accessibility features we need. It covers more functionality and thus the source is a bit more complex (e.g. offers [`react-focus-on`](https://github.com/theKashey/react-focus-on) to compose all features together).  |

I think the two packages are fairly comparable in terms of quality. However, it is very evident that `react-focus-lock` is _significantly_ more popular in terms of downloads. I believe I opted for `focus-trap-react` because (1) I was familiar with it and (2) the smaller size without the need to side load portions of its source. 

That said, after restating these thoughts out for consideration, it does seem like `react-focus-lock` may be a better longterm choice. What do you think? 

Relatedly, one day [React may includes its own focus management](https://github.com/reactjs/rfcs/pull/109). However, I do not believe that is likely to occur soon.

dmsnell:
thanks for providing such helpful info @dcalhoun 

> What do you think?

sounds like both are probably fine given what you said. you can pick the direction and set it 😉 I feel like I understand the context much better now though, and hopefully soon we'll be able to get rid of the `react-outsideclick` as well.

this is great work you've been providing. I'm sure it's helping out in some of the worst usability areas with our app

dcalhoun:
After considering the available options again, I have chosen to continue using `focus-trap-react`.

* `focus-trap-react` ([~4.2kb](https://bundlephobia.com/result?p=focus-trap-react@8.4.2)) includes click outside and escape to close features. `react-focus-lock` ([~5.4kb](https://bundlephobia.com/result?p=react-focus-lock@2.5.0)) does not which means we would need to either (A) build that logic ourselves or (B) use `react-focus-on` ([~8.7kb](https://bundlephobia.com/result?p=react-focus-on@3.5.1)) instead, which includes features we do not need. 
* In order to reduce the initial load size of `react-focus-on` we would need to leverage `sidecar` to lazy load dependencies. From my testing the resulting variation in size bundle between `focus-trap-react` and lazy loaded `react-focus-on` is negligible.
* From my perspective, using `focus-trap-react` is more straightforward and provides the accessibility improvements we set out to accomplish.

Plan to move forward with `focus-trap-react` unless I receive guidance otherwise. 👍🏻

lib/app-layout/index.tsx
```diff
@@ -115,20 +116,32 @@ export class AppLayout extends Component<Props> {
...
             {showSortBar && <SortOrderSelector />}
           </div>
           {editorVisible && (
             <div className="app-layout__note-column theme-color-bg theme-color-fg theme-color-border">
-              <NoteToolbar />
+              <NoteToolbar aria-hidden={hiddenByRevisions} />
               {showRevisions ? (
-                <NotePreview noteId={openedNote} note={openedRevision} />
+                <NotePreview
+                  aria-hidden={hiddenByRevisions}
```
belcherj, line 144: why use `aria-hidden` as opposed to `ariaHidden`?

lib/components/note-preview/index.tsx
```diff
@@ -29,9 +29,13 @@ type DispatchProps = {
   openNote: (noteId: T.EntityId) => any;
 };
 
-type Props = OwnProps & StateProps & DispatchProps;
+type Props = OwnProps &
+  StateProps &
+  DispatchProps &
+  React.HTMLProps<HTMLDivElement>;
```
belcherj, line 36: Why not specify aria-hidden as true or null?

lib/components/note-preview/index.tsx
```diff
@@ -29,9 +29,13 @@ type DispatchProps = {
...
 
 export const NotePreview: FunctionComponent<Props> = ({
+  'aria-hidden': ariaHidden,
```
belcherj, line 39: Seems odd to have the prop be `aria-hidden` just to set it to ariaHidden to set it to the attribute `aria-hidden`

lib/app-layout/index.tsx
dcalhoun: Largely for consistency between [DOM elements and React elements](https://reactjs.org/docs/accessibility.html#wai-aria) (i.e. use kebab-case in both instances).

That said, I don't have a strong opinion. Happy to change the key for React elements to be camelcase. Would you like me to do that?

lib/components/note-preview/index.tsx
dcalhoun: In TypeScript, I generally default to leveraging types provided by the ecosystem rather than defining custom types. I.e. `React.HTMLProps<HTMLDivElement>` includes `aria-hidden`.

Not a strong stance by any means, though. Would you like me to replace `React.HTMLProps<HTMLDivElement>` with something like...

```ts
type OwnProps = {
  ariaHidden?: 'true' | 'false';
  noteId: T.EntityId;
  note?: T.Note;
};
```

lib/components/note-preview/index.tsx
dcalhoun: Agreed. It's a little odd. The cause for this originates from https://github.com/Automattic/simplenote-electron/pull/2707#discussion_r591678420. If we decided to replace kebab-case for camelcase there, this will go away.

lib/app-layout/index.tsx
belcherj: @codebykat @sandymcfadden Any thoughts? I don't feel super strongly about it.

lib/components/note-preview/index.tsx
belcherj: I don't have strong opinions about it. I think what you have is fine.

lib/app-layout/index.tsx
sandymcfadden: I think the kebab-case is fine, I don't have strong feelings either.

sandymcfadden (approved):
This might be good to have someone else look over as well, but from my look and testing it all seems to be working well as expected.
Love the accessibility improvements!

PR #2712, opened 2021-03-03
Fix overflowing dialog content caused by long email address

### LINKED ISSUE #2709: Long email causes broken overflow in Review Your Account dialog
### Expected
A long email wraps to remain within the dialog width.

### Observed
A long email overflows the dialog width making the text less legible.

### Reproduced
1. Visit the Simplenote desktop or web app.
2. Sign up with a fairly long email address.

modal-overflowing-email

### Where did you see the bug
- System Make: Mac
- System Model: MacBook Pro
- OS: macOS
- OS version: 11.2.2
- Browser (if applicable): Chrome
- Browser version (if applicable): 88.0.4324.192

dcalhoun:
### Fix
Fixes #2709. Allow the dialog to wrap the email address to prevent the content
from overflowing the dialog container.

| Before | After |
| --- | --- |
| 109825032-07a40a00-7bff-11eb-8895-837555f564af | Screen Shot 2021-03-03 at 09 59 03 |

### Test
1. Visit the Simplenote desktop or web app.
1. Sign up with a fairly long email address.

Expected outcome: The email address wraps within the dialog rather than
overflowing the dialog.

### Release

- Fix overflowing dialog content when long email address is used during sign up.

### DIFF
--- a/lib/email-verification/index.tsx
+++ b/lib/email-verification/index.tsx
@@ -60,9 +60,9 @@ const EmailVerification: FunctionComponent<Props> = ({
         <WarningIcon />
       </span>
       <h2>Review Your Account</h2>
-      <p>
-        You are registered with Simplenote using the email{' '}
-        <strong>{email}</strong>.
+      <p>You are registered with Simplenote using the email:</p>
+      <p className="email-verification__email">
+        <strong>{email}</strong>
       </p>
       <p>
         Improvements to account security may result in account loss if you no
--- a/lib/email-verification/style.scss
+++ b/lib/email-verification/style.scss
@@ -45,6 +45,7 @@
   p {
     margin: 0 26px;
     margin-block-start: 0;
+    word-break: break-all;
   }
 
   p:not(:last-of-type) {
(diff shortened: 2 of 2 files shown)


codebykat (changes requested):
Unfortunately, it does this in Firefox (or presumably anywhere the window/zoom size doesn't work out right). I think we need to add a span around the email address and just put `break-all` on that.

Screen Shot 2021-03-03 at 10 07 35 PM

codebykat:
Somewhat absurdly, putting the style on the `<strong>` even works. I think we should make the break before the email an explicit `<br>` as well, that will help matters.

Screen Shot 2021-03-03 at 10 10 16 PM

Screen Shot 2021-03-03 at 10 10 20 PM

dcalhoun:
@codebykat good catch. Just pushed a change. Is the image below what you were envisioning? New line for the email to avoid likelihood of wrapping.

Screen Shot 2021-03-04 at 08 33 28

codebykat (approved):
Looks great now 👍 

Screen Shot 2021-03-04 at 9 25 05 AM

PR #2713, opened 2021-03-03
Set Unusual Line Terminators option to 'Auto'

### LINKED ISSUE #2698: Handle "unusual line terminators"
A user in 3768648-zen got the following message from Monaco:

```
	Detected unusual line terminators

This file contains one or more unusual line terminator characters, like Line Separator (LS) or Paragraph Separator (PS).

It is recommended to remove them from the file. This can be configured via `editor.unusualLineTerminators`.
```

Looks like we can set that option to `auto` instead of the default of `prompt`: https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandaloneeditorconstructionoptions.html#unusuallineterminators

illusaen:
### Fix
Fixes #2698

Allows Monaco editor to auto remove unusual line terminators.

### Release
> Added `unusualLineTerminators: auto` to Monaco Editor options to auto remove line terminators like line separators (LS) or paragraph separator (PS) [#2698](https://github.com/Automattic/simplenote-electron/issues/2698)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [v2.8.0]
+
+### Fixes
+
+- Added `unusualLineTerminators: auto` to Monaco Editor options to auto remove line terminators like line separators (LS) or paragraph separator (PS) [#2698](https://github.com/Automattic/simplenote-electron/issues/2698)
+
 ## [v2.7.0]
 
 ### Enhancements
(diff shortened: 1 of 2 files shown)


codebykat (approved):
Awesome, I was able to create a file with this weird line endings and confirm that they are silently removed with this fix.

1. `php -r 'echo "foo\u{2028}bar\u{2029}baz";' > file.txt`
2. Import this file into Simplenote
3. Select the imported note

PR #2715, opened 2021-03-03
Add error boundary

### LINKED ISSUE #879: Add error boundaries to prevent white screens of death
We can add React [error boundaries](https://reactjs.org/docs/error-boundaries.html) to provide more graceful fallbacks when bugs like #838 pop up.

dcalhoun:
### Fix
Fixes #879. Add error boundary to provide the user a helpful message and action,
rather than an empty white page.

| Small Screens | Large Screens |
| --- | --- |
| error-boundary-small | error-boundary-large |

| Dark Theme | Analytics Opt-In |
| --- | --- |
| error-boundary-large-dark | error-boundary-analytics |

### Test
Testing requires introducing an intentional error to the code. After doing so,
the new error boundary should be displayed.

It is important to note that the app has two initialization paths: authenticated
and unauthenticated. To fully test, one must introduce an error in each of these
paths.

### Release
Improve error handling and messaging within the app.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -139,6 +139,10 @@ module.exports = function main() {
       importNotes(filePath, mainWindow);
     });
 
+    ipcMain.on('reload', function () {
+      mainWindow.reload();
+    });
+
     mainWindowState.manage(mainWindow);
 
     // this TERRIBLE HACK forces vscode to call window.open(url) rather than window.open()
--- a/desktop/preload.js
+++ b/desktop/preload.js
@@ -9,6 +9,7 @@ const validChannels = [
   'importNotes',
   'noteImportChannel',
   'reallyCloseWindow',
+  'reload',
   'setAutoHideMenuBar',
   'tokenLogin',
   'wpLogin',
--- a/lib/boot-with-auth.tsx
+++ b/lib/boot-with-auth.tsx
@@ -4,7 +4,7 @@ if (__TEST__) {
 
 import React from 'react';
 import App from './app';
-import ErrorBoundary from './error-boundary';
+import { ErrorBoundaryWithAnalytics } from './error-boundary';
 import Modal from 'react-modal';
 import getConfig from '../get-config';
 import { makeStore } from './state';
@@ -45,11 +45,13 @@ export const bootWithToken = (
     });
 
     render(
-      <ErrorBoundary isDevConfig={isDevConfig(config?.development)}>
-        <Provider store={store}>
+      <Provider store={store}>
+        <ErrorBoundaryWithAnalytics
+          isDevConfig={isDevConfig(config?.development)}
+        >
           <App isDevConfig={isDevConfig(config?.development)} />
-        </Provider>
-      </ErrorBoundary>,
+        </ErrorBoundaryWithAnalytics>
+      </Provider>,
       document.getElementById('root')
     );
   });
--- a/lib/boot-with-auth.tsx
+++ b/lib/boot-with-auth.tsx
@@ -4,6 +4,7 @@ if (__TEST__) {
 
 import React from 'react';
 import App from './app';
+import ErrorBoundary from './error-boundary';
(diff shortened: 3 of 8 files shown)


dmsnell:
This is great! I'm noting that this should only catch errors in the React code or in the main UI thread, which I think are extremely rare. I was concerned about it taking over the whole page, but in those rare cases I think the page freezes or empties out.

> Rest assured that our systems have captured the details of the error so that we may work to fix the issue

I don't believe this is true. Is that something that you were planning on building into this PR as well?

dcalhoun:
> This is great! I'm noting that this should only catch errors in the React code or in the main UI thread, which I think are extremely rare. I was concerned about it taking over the whole page, but in those rare cases I think the page freezes or empties out.

@dmsnell just to make sure I understand correctly — you are not asking for a change here, correct? I interpreted the spirit of #879 to be to improve the "white screen of death" that occurs from errors, which is why the fullscreen message was used at the top level of the application..

> > Rest assured that our systems have captured the details of the error so that we may work to fix the issue
> 
> I don't believe this is true. Is that something that you were planning on building into this PR as well?

The copy (and the design) of the error message is definitely something I wanted to discuss with the team. So, it's not final. That said, the statement regarding "systems have captured the details" was a reference to Sentry. Based on the [existing usage of Sentry](https://github.com/Automattic/simplenote-electron/blob/da1ba4fdab26b5e6831597cb5ea241b628c88659/lib/state/analytics/middleware.ts#L63-L71) and this PRs usage of `Sentry.ErrorBoundary`, I presumed we would receive at least some information regarding any error that causes this boundary to trigger. Do you believe that to be inaccurate?

Related to the Sentry topic, I now notice that I likely need to [conditionally leverage Sentry](https://github.com/Automattic/simplenote-electron/blob/da1ba4fdab26b5e6831597cb5ea241b628c88659/lib/state/analytics/middleware.ts#L60). Is that accurate?

dmsnell:
>  you are not asking for a change here, correct?

that's correct. I was relating my thoughts as I processed it in case anyone else has the same ones.

> That said, the statement regarding "systems have captured the details" was a reference to Sentry.

very good. I had forgotten about that. I'm not sure if any of us are actively watching that 🙃

dmsnell:
> I now notice that I likely need to conditionally leverage Sentry. Is that accurate?

this is true. we should not be tracking anything about a customer without their express permission, which comes by opting in to tracking

dcalhoun:
Hey @SylvesterWilmott. 👋🏻 Whenever you have time, I'd appreciate your review of and input on the copy and design of this new screen. Thank you!

SylvesterWilmott:
Thanks @dcalhoun. Looking good!

I think we can reuse the max-width that is found in the sign up/log in screens for the layout as well as centering the content in the viewport.

I've tried to shorten the copy a little too, let me know what you think:

`Oops!`
`We’re sorry - something went wrong.
Our team has been notified so that we can work to fix the issue. Please refresh or try again later.`
`[Refresh application]`
`If the problem persists, please email us at support@simplenote.com.`

error

dcalhoun:
@SylvesterWilmott thank you for the guidance. I have adapted the page to match your design.

I noticed that the type and button styles for the client differ slightly from the server rendered authentication pages. So, I attempted to match your design as best I could using the existing patterns in the client app. Will you please let me know what you think of the images below? Thanks!

| Small Screens | Large Screens |
| --- | --- |
| error-boundary-small | error-boundary-large |

| Dark Theme | Analytics Opt-In |
| --- | --- |
| error-boundary-large-dark | error-boundary-analytics |

SylvesterWilmott:
Thanks @dcalhoun looks great.

Sorry for the delay here.

SylvesterWilmott (approved):

sandymcfadden (approved):
I wasn't familiar with Error Boundaries before this, except for a bit of reading. 
This all makes sense though and tests well.

PR #2719, opened 2021-03-04
[Tooling] Mark *.enc configure files as binary

AliSoftware:
## Why?

So that GitHub does not mislead us about them being empty when they are actually not or showing us useless diffs of their content.

## To test

 - Create a branch on top of this one.
 - Modify one of the `.configure-files/*.enc` files. You can do that by modifying the hash in the `.configure` file to point to an old hash, then run `fastlane run configure_apply` for example.
 - Create a (draft) PR on GitHub to supposedly merge this branch.
 - Check that the diff on GitHub shows the `*.enc` file as something like "binary file differs" – instead of showing a textual diff with garbage content, or suggesting that the file is empty, like it GitHub has been doing in the past with those files recently.
 - Close the draft PR.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+.configure-files/*.enc binary


codebykat:
I've been wondering why this happened!!!

mokagio:
@codebykat if I'm reading your question correctly, you're wondering why we have this PR?

We use encrypted files in the iOS and Android apps and we've been baffled recently by GitHub showing changes to one of those as "empty file" when the file wasn't actually empty. @AliSoftware figured out why, and kindly opened PRs to make GitHub diff properly by explicitly telling Git to treat any `.enc` file as a binary, e.g.: https://github.com/wordpress-mobile/WordPress-iOS/pull/16010.

The Electron client doesn't use the tooling that fetches those binary files, though. So this PR can be closed 😄 .

codebykat:
@mokagio No, I was wondering why the diffs looked wrong for encoded files. I just remembered seeing that on some GitHub PR and being totally baffled by it.

codebykat:
Huh. I think we *do* want this change, actually? I had to go hunting for where I remembered seeing this and it came up in https://github.com/Automattic/simplenote-electron/pull/2686/files .

@mokagio Thoughts?

PR #2721, opened 2021-03-05
When deleting last tag when inside the untagged view, switch to all notes view

dmsnell:
See #2687 

this patch removes the double-dispatch introduced to jump to the all-notes view
when deleting the last tag. that dispatch introduced the unwanted side-effect
of closing the navigation bar.

in this version of the code we introduce a new data point to the `TRASH_TAG`
action which represents the number of remaining tags in the account _after_
trashing the one being removed. the collection reducer and any other reducer
can then use that count to make decisions on how to update without needing
to introduce update cycles.


illusaen:
See GIF of jump here: 
untagged-jump

PR #2724, opened 2021-03-08
Fix unnecessary separators in login

Klauswk:
### Fix

While on login screen, a few separators were add unnecessary on File and Edit menu.

image

image

This changes fix these issue checking if the separator is necessary before adding.

image

image

### Release
- Fix unnecessary separators in the Electron builds File and Edit menus when not yet logged in.

### DIFF
--- a/desktop/menus/file-menu.js
+++ b/desktop/menus/file-menu.js
@@ -5,38 +5,42 @@ const { appCommandSender } = require('./utils');
 const buildFileMenu = (isAuthenticated) => {
   isAuthenticated = isAuthenticated || false;
 
-  const submenu = [
-    {
-      label: '&New Note',
-      visible: isAuthenticated,
-      accelerator: 'CommandOrControl+Shift+I',
-      click: appCommandSender({ action: 'newNote' }),
-    },
-    ...(isAuthenticated ? [{ type: 'separator' }] : []),
-    {
-      label: '&Import Notes…',
-      visible: isAuthenticated,
-      click: appCommandSender({
-        action: 'showDialog',
-        dialog: 'IMPORT',
-      }),
-    },
-    {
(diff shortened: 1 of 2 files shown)


desktop/menus/file-menu.js
```diff
@@ -5,38 +5,42 @@ const { appCommandSender } = require('./utils');
...
+        visible: isAuthenticated,
+        accelerator: 'CommandOrControl+Shift+E',
+        click: appCommandSender({
+          action: 'exportNotes',
+        }),
+      },
+      ...(isAuthenticated ? [{ type: 'separator' }] : []),
+      {
+        label: '&Print…',
+        visible: isAuthenticated,
```
sandymcfadden: Since we are inside an if statement based on `isAuthenticated` we should be able to remove all the `isAuthenticated` checks from inside. 
This would be applicable to the edit menu code as well.

sandymcfadden (changes requested):
Thanks so much for noticing this and submitting a PR to resolve it.
I've left one comment so far on the code.

desktop/menus/file-menu.js
Klauswk: Thank you for the review.

Fix in the lastest commit.

sandymcfadden: Thanks again. It still looks like inside the `isAuthenticated` if blocks we could change all these lines `...(isAuthenticated ? [{ type: 'separator' }] : []),` to be just `{ type: 'separator' }` instead.

Klauswk: fix in the last commit

sandymcfadden (approved):
This is looking good and tests well.
Thanks again for submitting and making some changes along the way!

PR #2725, opened 2021-03-08
Fix so that notes imported with tags will have the tags sync properly

sandymcfadden:
### Fix

Currently, when importing Simplenote or Evernote files if any notes have tags, the tags will not sync properly.
They will appear only in the app which imports the notes.

This resolves that and syncs the tags properly when notes are imported.

### Test

1. Import notes which have tags.
2. Check the same account on other versions of the app.
3. Ensure tags have been added properly to both the note and tag list.

### Release

- Fix to ensure that tags on imported notes are added and synced properly.

### DIFF
--- a/lib/state/simperium/middleware.ts
+++ b/lib/state/simperium/middleware.ts
@@ -374,7 +374,6 @@ export const initSimperium = (
 
       // other note editing actions however
       // should trigger an immediate sync
-      case 'IMPORT_NOTE_WITH_ID':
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
this looks safe and to-the-point.

we might miss some tag updates if they already exist in app state, but those are the ones that are harder to update without triggering unwanted side-effects and this should cover most cases.

Issue #1935, opened 2020-02-26
Follow Web Content Accessibility Guidelines (WCAG) on the webapp


liviopv:
Reproducing feature suggestion from a visually-impaired user, I'm not entirely sure if it applies.

Translated from Portuguese:

> In the web app, it's already possible to use a screen reading software (used by visually-impaired people) using keyboard commands. However, it's necessary to implement the guidelines from WCAG to improve your user experience, promoting accessibility. To give examples, 1) when going through the editing fields with the Tab key and screen reading software on, the names of fields are not announced; 2) when you open note to edit, the screen reading software doesn't announce that new content was loaded on the page.

### What
Make the app fully compliant to https://www.w3.org/WAI/standards-guidelines/wcag/

### Why
Because it aligns with Automattic's core values and makes Simplenote accessible to a new set of users.

User-report: #2106972-zen

dcalhoun:
@belcherj I believe achieving WCAG compliance could be a larger project that is likely worth splitting up into multiple tickets. We would likely also need to decide which version of WCAG to target (e.g. WCAG 2.1) and what level of compliance (e.g. AA) we would like to achieve. In the meantime, I wanted to capture some resources I am familiar with that could help us approach the larger project.

[Deque](https://www.deque.com) provides a variety of services that can help access and improve accessibility of products.

* [axe DevTools](https://www.deque.com/axe/browser-extensions/): Free Chrome extension that can automatically detect certain accessibility issues providing tooling and information on how to address them. The paid version allows integrating the automated tooling into CI suites.
* [Deque University](https://www.deque.com/training/online-deque-university/): Online training resources for self-paced accessibility learning. This can be helpful for understanding the nuances of WCAG and how to apply them to a product.
* [Audits & Training Sessions](https://www.deque.com/services/): Deque also provides audits and training sessions to help teams assess their status and equip themselves for building accessible products.
* [axe Auditor](https://www.deque.com/axe/auditor/): Tool to help define and run manual accessibility test cases.

There are obviously other companies that provide similar tooling and services, but Deque just happens to be the one with which I am most familiar.

From my perspective, a good first step could be leveraging the axe DevTools to address the issues that it finds.

<img width="1540" alt="axe-devtools-example" src="https://user-images.githubusercontent.com/438664/110178284-d62c6980-7dcb-11eb-9602-b139232e454c.png">

Hope this information is helpful moving forward!

PR #2726, opened 2021-03-08
Improve keyboard and screen reader accessibility

dcalhoun:
### Fix
Related to https://github.com/Automattic/simplenote-electron/issues/1935#issuecomment-791742151, this addresses several issue that inhibited keyboard 
and screen reader users from navigation the application easily.

Summary of changes:

- Add ARIA roles for virtualized list
- Improve accessibility of notes list
- Reduce noise in screen reader audio
- Increase accuracy of search field descriptions
- Adjust search button styles to improve tap area
- Increase accuracy of tag list accessibility labels
- Allow scaling to improve accessibility
- Improve ARIA descriptions for application layout
- Add ARIA label for all icon buttons
- Set language for HTML document

| Issue Count Before | Issue Count After |
| --- | --- |
| axe-devtools-before | axe-devtools-after |

The remaining issues identified by the axe DevTools fall into three categories:

1. Issues that require design input (e.g. text color legibility).
1. Issues with the third-party Monaco editor (e.g. lack of ARIA attributes).
1. Issues that are not applicable to the Simplenote web application (e.g. lack of level-one heading).

### Test
1. Launch Simplenote web or desktop app.
1. Enable macOS VoiceOver or screen reader.
1. Navigate the application.

_Expected Outcome:_ No major issues block navigating the application.

### Release

Improved accessibility of the application for keyboard and screen reader users.

### DIFF
--- a/lib/app-layout/index.tsx
+++ b/lib/app-layout/index.tsx
@@ -118,14 +118,20 @@ export class AppLayout extends Component<Props> {
     return (
       <div className={mainClasses}>
         <Suspense fallback={placeholder}>
-          <div className="app-layout__source-column theme-color-bg theme-color-fg">
+          <aside
+            aria-label="Notes list"
+            className="app-layout__source-column theme-color-bg theme-color-fg"
+          >
             <MenuBar />
             <SearchField />
             <NoteList />
             {showSortBar && <SortOrderSelector />}
-          </div>
+          </aside>
           {editorVisible && (
-            <div className="app-layout__note-column theme-color-bg theme-color-fg theme-color-border">
+            <main
+              aria-label="Note editor"
+              className="app-layout__note-column theme-color-bg theme-color-fg theme-color-border"
+            >
               <NoteToolbar />
               {showRevisions ? (
                 <NotePreview noteId={openedNote} note={openedRevision} />
@@ -133,7 +139,7 @@ export class AppLayout extends Component<Props> {
                 <NoteEditor />
               )}
               {hasRevisions && <RevisionSelector />}
-            </div>
+            </main>
           )}
         </Suspense>
       </div>
--- a/lib/components/dev-badge/index.tsx
+++ b/lib/components/dev-badge/index.tsx
@@ -1,7 +1,11 @@
 import React from 'react';
 
 const DevBadge = () => {
-  return <div className="dev-badge">DEV</div>;
+  return (
+    <div aria-hidden="true" className="dev-badge">
+      DEV
+    </div>
+  );
 };
 
 export default DevBadge;
--- a/lib/components/dev-badge/index.tsx
+++ b/lib/components/dev-badge/index.tsx
@@ -1,11 +1,9 @@
 import React from 'react';
 
-const DevBadge = () => {
-  return (
-    <div aria-hidden="true" className="dev-badge">
-      DEV
-    </div>
-  );
-};
+const DevBadge = () => (
+  <div aria-hidden="true" className="dev-badge">
+    DEV
+  </div>
+);
 
 export default DevBadge;
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -149,6 +149,19 @@ const createCompositeNoteList = (
   ];
 };
 
+const sidebarTitle = (collection: T.Collection) => {
+  switch (collection.type) {
+    case 'tag':
+      return 'Notes With Selected Tag';
+    case 'trash':
+      return 'Trash';
+    case 'untagged':
+      return 'Untagged Notes';
+    default:
+      return 'All Notes';
+  }
+};
+
 export class NoteList extends Component<Props> {
   static displayName = 'NoteList';
 
@@ -213,19 +226,6 @@ export class NoteList extends Component<Props> {
     this.toggleShortcuts(false);
   }
 
-  computedLabel() {
-    switch (this.props.collection.type) {
-      case 'tag':
-        return 'Notes With Selected Tag';
-      case 'trash':
-        return 'Trash';
-      case 'untagged':
-        return 'Untagged Notes';
-      default:
-        return 'All Notes';
-    }
-  }
-
   handleShortcut = (event: KeyboardEvent) => {
     if (!this.props.keyboardShortcuts) {
       return;
@@ -280,6 +280,7 @@ export class NoteList extends Component<Props> {
 
   render() {
     const {
+      collection,
       filteredNotes,
       noteDisplay,
       onEmptyTrash,
@@ -326,10 +327,10 @@ export class NoteList extends Component<Props> {
                 {({ height, width }) => (
                   <List
                     // Ideally aria-label is changed to aria-labelledby to
-                    // reference the existing notes title element instead of
+                    // reference the existing #notes-title element instead of
                     // computing the label, but is not currently possible due to
                     // a limitation with react-virtualized. https://git.io/JqLvR
-                    aria-label={this.computedLabel()}
+                    aria-label={sidebarTitle(collection)}
                     ref={this.list}
                     estimatedRowSize={24 + 18 + 21 * 4}
                     height={height}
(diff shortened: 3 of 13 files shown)


lib/search-field/index.tsx
```diff
@@ -73,21 +73,24 @@ export class SearchField extends Component<Props> {
   render() {
     const { openedTag, searchQuery } = this.props;
     const hasQuery = searchQuery.length > 0;
-    const placeholder = openedTag ?? 'Search notes and tags';
 
-    const screenReaderLabel =
-      'Search ' + (openedTag ? 'notes with tag ' : '') + placeholder;
+    const description =
+      'Search ' +
+      (openedTag ? 'notes with tag ' + openedTag : 'notes and tags');
```
dcalhoun: I wanted to draw attention to this change. Previously the placeholder value and the ARIA label were different strings. From reviewing the history of the UI, this at one point made sense because it appears the search field was utilized as the "title" of the list of notes, but also served a search input.

However, several changes were made to this portion of the app — the UI was moved around, the structure of the state, etc. These various changes led to a screen reader description that read "Search Search notes and tags Search notes and tags, search text field." 😂 Needless to say, this description is a bit...confusing.

So, these changes are an attempt to fix this label, but it also involves changing the _visual_ placeholder for the search input. I would appreciate if others would please look carefully at this piece to ensure the current state matches the desired experience.

lib/app-layout/index.tsx
```diff
@@ -118,22 +118,28 @@ export class AppLayout extends Component<Props> {
...
+            >
               <NoteToolbar />
               {showRevisions ? (
                 <NotePreview noteId={openedNote} note={openedRevision} />
               ) : (
                 <NoteEditor />
               )}
               {hasRevisions && <RevisionSelector />}
-            </div>
+            </main>
```
dmsnell, line 142: nice - I like the changes to `aside` and `main`

lib/components/dev-badge/index.tsx
```diff
@@ -1,7 +1,11 @@
 import React from 'react';
 
 const DevBadge = () => {
-  return <div className="dev-badge">DEV</div>;
+  return (
+    <div aria-hidden="true" className="dev-badge">
+      DEV
+    </div>
+  );
```
dmsnell: while we're at it, can we have this return the value directly instead of opening up a function with return here?

```ts
const DevBadge = () => (
	<div aria-hidden="true" className="dev-badge">
		DEV
	</div>
);
```

lib/note-list/index.tsx
```diff
@@ -212,6 +213,19 @@ export class NoteList extends Component<Props> {
...
+      case 'tag':
+        return 'Notes With Selected Tag';
+      case 'trash':
+        return 'Trash';
+      case 'untagged':
+        return 'Untagged Notes';
+      default:
+        return 'All Notes';
+    }
+  }
```
dmsnell: we don't need to refer to the class instance. can we move this into the module-global scope?

```ts
const sidebarTitle = (collection: Collection) => {
	switch (collection.type) {
		…
	}
}
```

on this I raise two questions:
 - should we extract this function to another file where this module and the list item module can use it together?
 - should we be interpolating the selected tag name here or leave it at "selected tag"?

lib/note-list/index.tsx
```diff
@@ -320,6 +339,7 @@ export class NoteList extends Component<Props> {
                     rowHeight={heightCache.rowHeight}
                     rowRenderer={renderNoteRow}
                     scrollToIndex={selectedIndex}
+                    tabIndex={null}
```
dmsnell, line 348: what's the difference between `tabIndex={-1}` and `tabIndex={null}`?

dcalhoun: The difference being that `tabIndex={null}` unsets a `tabIndex="0"` [set by default from `react-virtualized`](https://github.com/bvaughn/react-virtualized/issues/275#issuecomment-226077606) and `tabIndex={-1}` disables the element from being focused by keyboard tabbing while still allowing programmatic focus.

Given that the contents of our virtualized list are focusable (buttons to pin or edit a note), IMO it doesn't make sense for the entire list to be a focusable element. We also, to my knowledge, have no need to programmatically focus the list, so `null` makes more sense than `-1`.

lib/note-list/index.tsx
dcalhoun: Moving it from a class method to the module scope makes sense. ✅ 

> should we extract this function to another file where this module and the list item module can use it together?

I originally avoided doing this because (1) it is only used in two places currently and (2) I hope its [usage here eventually goes away](https://github.com/Automattic/simplenote-electron/blob/0797ce03a44e2493fb2ef06cb5cf0bd958325c13/lib/note-list/index.tsx#L328-L331). That said, I'm happy to abstract it if you'd like. What do you think?

> should we be interpolating the selected tag name here or leave it at "selected tag"?

Yes, it may be a improvement for the title to state the tag name rather than a generic statement. One caveat is that that tag is likely to be visually truncated. Would you like me to change the visual title and ARIA label to be the tag value itself?

Screen Shot 2021-03-09 at 09 05 25

dmsnell: thanks for the screenshot. now that I'm staring at it "Notes With …" seems confusing. "Notes Tagged …" would seem clearer, but also the truncation looks like it will hit us regularly. I think we used to show the tag name without the additional titling, which didn't have this problem.

this is beyond me at the moment. feel free to reach out to others or make a call for what you think is best.

lib/note-list/index.tsx
dmsnell: interesting. thanks! so the list is no longer focusable: does that have any implications for scrolling the list with the keyboard?

dcalhoun: Good question. I had not explicitly tested scrolling the list with the keyboard until just now. It would appear the scrolling the list is still possible and matches the experience as when `tabindex="0"` is set. 👍🏻

lib/note-list/index.tsx
dcalhoun: >I think we used to show the tag name without the additional titling, which didn't have this problem.

Indeed. I confirmed that. That fact is also related to https://github.com/Automattic/simplenote-electron/pull/2726#discussion_r589807280.

>this is beyond me at the moment. feel free to reach out to others or make a call for what you think is best.

Alright. Hopefully someone else will also find time to review and provide feedback.

belcherj: @SylvesterWilmott what say you?

SylvesterWilmott: The title in the title bar when a tag is selected should be `tag name` to be consistent with `all notes` or `trash`.

Regarding the search placeholder, electron is the only remaining platform where search is not global and so we have this need to imply search is contained only within the selected tag. Once search becomes global (searches all notes all the time) then the search placeholder will read "Search notes and tags". Until then we can try "Search `tag name`" or "Search in `tag name`" to avoid that truncation that looks like it would be a fairly common occurance.

dcalhoun: Thanks for the feedback. I updated the search placeholder to be `Search in <tag_name>`. I also applied an ellipsis when the placeholder overflows the input width.

Screen Shot 2021-03-11 at 11 43 15

sandymcfadden (approved):
This looks good and works well. Seem much better than what we had previously. Thanks for taking the time to do this.
The only confusing part I had while testing with voiceover was when clicking the button to open the left navigation. Trying to get to the items in it once I clicked the button wasn't straight forward, but I think we can look at that in some future enhancements.

dcalhoun:
> The only confusing part I had while testing with voiceover was when clicking the button to open the left navigation. Trying to get to the items in it once I clicked the button wasn't straight forward, but I think we can look at that in some future enhancements.

@sandymcfadden thanks for taking time to review. #2707 _should_ address all issues with navigating the left navigation with a keyboard or screen reader. So, until both PRs are merged, one has to test those areas independently, unfortunately. Please let me know if I am misunderstanding or there are specific issues you encountered.

sandymcfadden:
Yes @dcalhoun yeah, of course, you're right. I even looked at and tested the other one. I think I got confused with the two, sorry for that confusion.

codebykat (approved):
With the caveat that I don't have a lot of domain knowledge here, so in general I trust your expertise. This all looks great!

PR #2735, opened 2021-03-10
Check if email is set when dismissing alternate login prompt

### LINKED ISSUE #2734: [2.7.1] `Cancel` and [x] buttons do not work in the logout prompt
### Expected
All onscreen elements are functional

### Observed
`Cancel` and `[x]` buttons do not work here:

Screenshot 2021-03-10 at 13 05 27

### Reproduced
1. Be logged in on the app
2. Sign up for another account via Web
3. Set up password via email link
4. When asked to open Simplenote app, agree ~> you will see the dialog from above.
5. Try clicking `Cancel` / `[x]`

**Side question**: should clicking `Logout` result in just logout (current), or also in a7c login of new user?

### Environment
- System Model: Parallels Desktop on MBP 16"
- OS: Windows 10
- OS version: 10 Home 20H2 19042.508
- Browser (if applicable): -
- Browser version (if applicable): -
- Simplenote app version: Desktop app 2.7.1

sandymcfadden:
### Fix

When dismissing the alternate login prompt the email is set to false. This caused an error when trying to decode the email with `atob`.
This adds a check to ensure the email is set before decoding.

Fixes #2734

### Test

1. Be logged into the app.
2. Sign up for a new account on web.
3. In the prompt choose to open the link in the app.
4. On the logout prompt in the app choose to cancel.
5. Notice the dialog is dismissed properly.

### Release

- Fixes bug when dismissing the logout view from the alternate login prompt.

### DIFF
--- a/lib/state/ui/reducer.ts
+++ b/lib/state/ui/reducer.ts
@@ -92,7 +92,7 @@ const alternateLoginEmail: A.Reducer<string | null> = (
 ) => {
   switch (action.type) {
     case 'SHOW_ALTERNATE_LOGIN_PROMPT':
-      return atob(action.email);
+      return action.email ? atob(action.email) : null;
     case 'HIDE_ALTERNATE_LOGIN_PROMPT':
       return null;
     default:


belcherj (approved):
Seems like a reasonable check to me.

pachlava:
[testing] fix/feature verified on Windows 10 | Simplenote Desktop 2.7.1-beta2

PR #2737, opened 2021-03-10
Improve note Action and Info focus management

dcalhoun:
### Fix
- Remove redundant focus management within note Info. React Modal provides its own focus trap logic, so wrapping with `FocusTrap` is unnecessary.
- Support ClipboardButton within focus trap for note actions. In order for clipboard.js to function properly within a focus trap, we [must set a `container`](https://clipboardjs.com/#advanced-usage) that is "reachable" from within the focus trap.

### Test
1. Open Simplenote.
1. Click note Actions button (three dots in circle) in the top-right.
1. Click "Copy Internal Link."
1. Click "Copy Link."

_Expected Outcome:_ Both internal and public links should be within clipboard
history.

### Release

n/a

### DIFF
--- a/lib/components/clipboard-button/index.tsx
+++ b/lib/components/clipboard-button/index.tsx
@@ -1,8 +1,13 @@
 import React, { useEffect, useRef, useState } from 'react';
-import PropTypes from 'prop-types';
 import Clipboard from 'clipboard';
 
-function ClipboardButton({ linkText, text }) {
+type Props = {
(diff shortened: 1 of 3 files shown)


dcalhoun:
@sandymcfadden this PR represents what I found when investigating usage of clipboard.js within a focus trap. Please feel free to merge/cherrypick/modify/discard any of these changes as you see fit.

sandymcfadden (approved):
This looks great, thank you so much for your help with this.

PR #2739, opened 2021-03-10
Improve recognizability of inline code elements

### LINKED ISSUE #1891: Code elements are hard to distinguish

dcalhoun:
### Fix
Fixes #1891. Increase recognizability of inline code elements in Markdown
previews by applying a subtle background color to them.

| Before | After | After (Dark) |
| --- | --- | --- |
| before | after | after-dark |

### Test
1. Create a new note.
1. Click note Info button in the top-right.
1. Enable Markdown formatting.
1. Add an inline code block to the note content.
1. Click Preview button in the top-right.

_Expected Outcome:_ The inline code element is styled with a monospace typeface
and a background color different than rest of the content.

### Release
Improve recognizability of inline code elements.

### DIFF
--- a/lib/note-detail/style.scss
+++ b/lib/note-detail/style.scss
@@ -173,6 +173,12 @@
     font-size: 85%;
   }
 
+  p > code {
+    background-color: $studio-gray-5;
+    border-radius: 2px;
+    padding: 3px;
+  }
+
   table {
     border-collapse: collapse;
     border-spacing: 0;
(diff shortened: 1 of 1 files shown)


sandymcfadden (approved):
Straightforward, works well and makes it much nicer to read inline code.

PR #2742, opened 2021-03-11
Merge 2.8.0 beta 2 into develop

mokagio:
This is to closes the loop on the "check if email is set when dismissing prompt" fix. See #2734.


sandymcfadden (approved):
I know this is still a draft, but had the review requested and it looks ready to go, so adding my approval.

mokagio:
Thanks @sandymcfadden .

I usually wait for the tests to run before turning the PR "ready for review", just in case I get a failure and the following "hey the PR looks good but did you notice there's a failure?"

I opened this but didn't see the tests result because then I had to run to cook dinner 😅

PR #2746, opened 2021-03-11
Fix adding note from trash so it takes you back to All Notes

### LINKED ISSUE #2743: Creating a new note from Trash via shortcut should be forbidden
### Expected
If the `new-note` button is disabled when in `Trash`, the shortcut should not work too (same for menu item).

### Observed
`Ctrl+Shift+I` shortcut will switch the app to weird state: there will be still `Trash` header, but notes will be shown from `All Notes`, and also there will be the `Empty Trash` button:

https://user-images.githubusercontent.com/73365754/110797256-91984680-8281-11eb-9212-ca345c842c82.mov

### Reproduced
1. Open `Trash`
2. Press `Ctrl+Shift+I`

### Where did you see the bug
- System Model: Parallels Desktop on MBP 16"
- OS: Windows
- OS version: 10 Home 20H2 build 19042.508
- Browser (if applicable): x
- Browser version (if applicable): x
- Simplenote app version: 2.8.0-beta2 desktop

sandymcfadden:
### Fix

Fixes #2743 which was a regression.

This enables the new note button when in the trash and ensures you are switched to the All Notes view when creating a new note from the trash.

### Test

1. Open trash.
2. Ensure you can click the new note button.
3. Ensure when you create the new note it takes you to the All Note view.

### Release

- Fixes a regression allowing the adding of a new note from the trash view.

### DIFF
--- a/lib/menu-bar/index.tsx
+++ b/lib/menu-bar/index.tsx
@@ -69,7 +69,6 @@ export const MenuBar: FunctionComponent<Props> = ({
       />
       <div className="notes-title">{placeholder}</div>
       <IconButton
-        disabled={collection.type === 'trash'}
         icon={<NewNoteIcon />}
         onClick={() => onNewNote(withoutTags(searchQuery))}
         title={`New Note • ${CmdOrCtrl}+Shift+I`}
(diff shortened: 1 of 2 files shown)


dmsnell (approved):
Lovely seeing how this is centralized after @illusaen's work to collapse all the flags into the new `collection` state.

PR #2751, opened 2021-03-11
Fixing tooltip so it leaves properly for disabled buttons

### LINKED ISSUE #2745: `History (unavailable)` hint is not hidden automatically
### Current
The `History (unavailable)` hint will be hidden only after a repeated hover above history icon:

https://user-images.githubusercontent.com/73365754/110802240-ad521b80-8286-11eb-84ad-bb4915f344a6.mov

Happens on Windows, macOS and Ubuntu.

### Proposed
Should be hidden automatically as happens for hints for neighbouring buttons.

sandymcfadden:
### Fix

Fixes #2745 

Currently, when a button is disabled the tooltip isn't dismissed properly in Chrome. This fixes that so it is dismissed as expected.

### Test

1. Test in Chrome.
2. Visit a note with no revisions
3. Hover over the History button in the note toolbar. 
4. Ensure when you leave the button the tooltip dismisses.

### Release

- Fix dismissing the tooltip for disabled buttons.

### DIFF
--- a/lib/icon-button/style.scss
+++ b/lib/icon-button/style.scss
@@ -9,6 +9,7 @@
 
   &:disabled {
     opacity: 0.4;
+    pointer-events: none;
   }
 }
 


codebykat (approved):
This isn't just a Chrome bug in my experience, but a bit hard to get it to happen consistently when I want it to... (in Firefox)... I don't understand why this fixes the issue, but it seems to, or at least, I couldn't get the sticky tooltip to happen on this branch with a bit of testing.

sandymcfadden:
Thanks @codebykat I had [found some documentation](https://material-ui.com/components/tooltips/#disabled-elements) that suggested it needed to be added for native disabled elements so that is what led me here. 

>If you're not wrapping a Material-UI component that inherits from ButtonBase, for instance, a native <button> element, you should also add the CSS property pointer-events: none; to your element when disabled:

sandymcfadden:
Fixes #2745
Currently, when a button is disabled the tooltip isn't dismissed properly in Chrome. This fixes that so it is dismissed as expected.

pachlava:
@sandymcfadden I'm testing this in 2.9.0-beta1 (Win and macOS desktop + Chrome on macOS): the hint for history button is now never shown, in case if it's disabled. Is this expected?

sandymcfadden:
@pachlava it is yes, sorry my testing steps weren't great on this. Using the component we are this is the recommended way and the note toolbar is going to be changing very soon removing that button altogether.

pachlava:
@sandymcfadden thanks!

PR #2756, opened 2021-03-12
Parse multiple pasted tags and edit pending tags

### LINKED ISSUE #2199: Tags: Can't edit by placing cursor as it always jumps to the end
I'm adding this as an enhancement as it doesn't seem quite like a bug.

### What
A Simplenote user reported strange behavior when trying to add tags. Upon a closer look, I believe it's just a way we didn't expect them to be added.

## Observe

1. Copy a few words with spaces between (like a sentence with no punctuation)
2. Paste into the tag area
3. Try to click into a word to edit it
4. The cursor will jump to the end and it's not possible to edit the words without just deleting from the right side, so any edits end up at the end and look like a bug

User reported this for Windows 10 desktop app and the web.
I can reproduce in Mac Electron 1.19 and on the web in Chrome

dcalhoun:
### Fix
Fixes #2199. These changes look to improve the UX for pasting multiple tags
and editing pending tags.

Previously, any click on or input into the "Add tags" input resulted in the
caret being placed at the end of the current value. This inhibited the ability to
modify the current value of the "Add tags" input.

Additionally, pasting multiple tags (i.e. a string containing multiple words
separated by commas or spaces) resulted in several pending tags. These changes
now parse the pasted string to convert all but the last into submitted tags,
leaving the last as a pending tag.

https://user-images.githubusercontent.com/438664/111004071-e99c7f00-834d-11eb-96e8-904727e65b00.mov

### Test
**Scenario A**
1. Open note within Simplenote.
1. Focus "Add tags" input.
1. Type `something`.
1. Press left arrow twice.
1. Press any character key.
1. Click beginning of inputted string placing caret in first position.
1. Press <kbd>Shift</kbd>+<kbd>Delete</kbd>.

_Expected Outcome:_ The cursor never jumps to the end of the inputted value. The cursor
remains in place where you have positioned the cursor either via keyboard arrows or mouse clicks.

**Scenario B**
1. Open note within Simplenote.
1. Focus "Add tags" input.
1. Paste `some-thing, a_gain another thing`.

_Expected Outcome:_ `some-thing`, `a_gain`, and `another` are submitted tags. `thing` is
a pending tag.

### Release
* Improved parsing of multiple pasted tags.
* Fix editing of pending tag input.

### DIFF
--- a/lib/tag-input/index.tsx
+++ b/lib/tag-input/index.tsx
@@ -91,7 +91,7 @@ export class TagInput extends Component<Props> {
       this,
       'inputField.removeEventListener',
       'paste',
-      this.removePastedFormatting,
+      this.parsePastedInput,
       false
     );
     this.inputObserver.disconnect();
@@ -211,16 +211,22 @@ export class TagInput extends Component<Props> {
     this.setState({ isComposing: false }, () => this.onInput(value));
   };
 
-  removePastedFormatting = (event: ClipboardEvent) => {
+  parsePastedInput = (event: ClipboardEvent) => {
     let clipboardText;
 
+    // Remove pasted formatting
     if (get(event, 'clipboardData.getData')) {
       clipboardText = event.clipboardData.getData('text/plain');
     } else if (get(window, 'clipboardData.getData')) {
       clipboardText = window.clipboardData.getData('Text'); // IE11
     }
 
-    this.onInput(clipboardText, { moveCaretToEndOfValue: true });
+    // Convert spaces/commans to submitted tags
+    const tags = clipboardText.split(/\s|,|\n/);
+    const submittedTags = tags.slice(0, tags.length - 1);
+    const [pendingTag] = tags.slice(tags.length - 1);
+    submittedTags.filter(Boolean).forEach(this.props.onSelect);
+    this.onInput(pendingTag, { moveCaretToEndOfValue: true });
 
     event.preventDefault();
     event.stopPropagation();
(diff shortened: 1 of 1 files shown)


lib/tag-input/index.tsx
```diff
@@ -175,11 +195,14 @@ export class TagInput extends Component<Props> {
...
 
-  onInput = (value: string) => {
+  onInput = (value: string, { moveCaretToEndOfValue = false } = {}) => {
     if (this.state.isComposing) {
       return;
     }
 
+    this.caretPosition = moveCaretToEndOfValue
+      ? value.length
+      : getCaretPosition(this.inputField);
```
dcalhoun, line 205: I don't find the addition of `this.caretPosition` or `moveCaretToEndOfValue` to be the most elegant solution. However, given the large number of edge cases for different types of input, keystrokes, clicks, etc, it was the only approach I could find that covered all the cases. I definitely welcome feedback or thoughts on this approach.

dcalhoun (commented):
I definitely recommend pulling this branch down locally to play with the tag input to compare it with the current iteration that is found in `develop`. It's very difficult to showcase all the different nuances of this input via text or video.

sandymcfadden (commented):
I'm going to leave this to have others look as well, specifically to see if they have any thoughts on [your comment here](https://github.com/Automattic/simplenote-electron/pull/2756/files#r593483123).

The testing seems to work very well.

codebykat (approved):
This is working well for me. I don't have a better idea for moving the caret than yours 🤷‍♀️

PR #2775, opened 2021-03-16
Autofocus delete tag button in when dialog opens

### LINKED ISSUE #2757: Moving default focus to 'Delete Tag' button in delete tag confirmation
### What
`Delete Tag` button is focused by default, so the `Enter` keystroke will be a faster alternative of clicking the button:

Screenshot 2021-03-13 at 15 13 43

### Why
Since there is no way to select multiple tags for removal, the process is slow by itself. Currently, the user has to click the `Trash` button for a tag, and then move the cursor to 'Delete Tag' button (or use the `Tab` navigation, which is even slower)

sandymcfadden:
### Fix

This fixes #2757 
Currently when deleting a tag you need to edit the tag list, then click the trash icon next to the tag, then move over and click the delete tag button.
This puts focus on the delete tag button when the dialog is opened and allows you to hit return/enter on the keyboard to confirm the deletion.

### Test

1. Choose to delete a tag.
2. When the confirmation dialog opens press return/enter on the keyboard.
3. Ensure tag deletes and the confirmation dialog closes. 

### Release

- Autofocus the delete tag button when the confirmation dialog is opened.

### DIFF
--- a/lib/dialogs/trash-tag-confirmation/index.tsx
+++ b/lib/dialogs/trash-tag-confirmation/index.tsx
@@ -34,6 +34,7 @@ const TrashTagConfirmation: FunctionComponent<Props> = ({
     </div>
     <div className="dialog-buttons">
       <button
+        autoFocus
         className="button-primary delete-tag"
         onClick={() => trashTag(tagName)}
       >


dmsnell (approved):
Great improvement! I did not test this.

pachlava:
✅ Checked in Simplenote 2.9.0-beta1, tried deleting multiple tags by using the `Enter` and `Space` keys. Works well and is very handy. Tested under:
- Desktop: Ubuntu 20.04, macOS 11.2.3 and Windows 10 Home
- Safari: macOS 11.2.3

PR #2785, opened 2021-03-17
Update the expanded note list preview to be three lines instead of four

sandymcfadden:
### Fix

This will take care of one of the remaining tasks on issue #2560.
Currently, we display 4 lines of a note in the note list preview when the display is set to be expanded. The new designs call for 3 lines instead. This updates that.

### Test

1. Have at least one note with more than four lines of content after the title.
2. Set the note display to be expanded.
3. Ensure the note list preview shows only three lines of the note.

### Release

- Updates the number of lines in the note preview to be three lines in expanded display.

### DIFF
--- a/lib/note-list/style.scss
+++ b/lib/note-list/style.scss
@@ -44,7 +44,7 @@
 
   .expanded .note-list-item-excerpt {
     white-space: normal;
-    max-height: 88px;
+    max-height: 63px;
   }
 
   &.is-empty {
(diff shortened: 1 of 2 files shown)


belcherj (approved):
Tested, works. Code looks good.

pachlava:
✅ Checked under light and dark themes, both with MD rendering and without. 3 lines of text are shown in all cases for Expanded display, checked under:
- Web 2.9.0-beta1: Safari and Chrome on macOS 11.2.3, Chrome on Windows 10 Home
- Desktop 2.9.0-beta1: Ubuntu 20.04 | macOS 11.2.3 | Windows 10 Home

PR #2808, opened 2021-03-24
Testing using WordPress Components

sandymcfadden:
Just for personal testing currently.

This currently updated the Note Info panel on the right to use the WordPress Toggle Control instead of our previous component and removes Material UI.
The changes were pretty straightforward and with some basic styling was able to get the components to look the same as our current designs.

It also adds Webpack Bundle Analyzer to check things out. To run you can now run the following two commands.
- `npm run stats` generates the stats file.
- `npm run analyzer` which will start the bundle analyzer so you can view.

**Build size differences**

Current 2.9.0 build:
- x64.appx: 102.5 MB
- exe: 135.6 MB
- x86_64.rpm: 64.9 MB
- amd64.deb: 102.4 MB

This build with WP Components:
- x64.appx: 112 MB
- exe: 145.3 MB
- x86_64.rpm: 68.9 MB
- amd64.deb: 111 MB

**Webpack bundle analyzer gzipped sizes**
- Current develop branch: 1.66 MB
- With WordPress Components added and Material UI Removed: 1.73 MB


dmsnell:
looks like the `dist` file grew by above 400KB after this patch. that's really pushing it unless we can remove more code. it's possible that a webpack update could help, especially if we're resorting to commonJS modules because of lack of support (I don't know if this is the case)

sandymcfadden:
I took a stab at upgrading webpack to the latest version. That involved upgrading other plugins and dependencies. It also involved adding in some other dependencies and config updates to get everything building properly.
I haven't done any digging yet to see if we can change our build process, basically just followed errors as they came up and fixed them. 
The build sizes have grown even more.

sandymcfadden:
I figured out what was wrong with one of my previous updates and not having source maps set correctly.
I've gone through a few different things and now have removed material UI and replaced with matching WordPress components and one custom one. This has the size down some.

I've updated the main description with size details.

sandymcfadden:
With @dmsnell's help I have webpack bundle anayzler now looking at the gzipped sizes of our bundle. 
With the WordPress components added, and Material UI removed it reports a size of 1.73 MB and our current develop branch at 1.66 MB.

Issue #2777, opened 2021-03-16
Chore: Update Dialogs to handle long content and keep action buttons static at the bottom


sandymcfadden:
There are some dialogs that have the potential to show long content.

Currently, we set the dialog to scroll, but if there are buttons at the bottom they are hidden out of view.
We should go through and update dialogs so the inner content is scrollable but the title, and footer with buttons if relevant, are static and always shown.

fluiddot:
After I merged this [PR](https://github.com/Automattic/simplenote-electron/pull/2815), I checked the rest of dialogs and we should review the following ones:
## About

Not sure if we should display a scrollbar in this dialog because it looks weird, I'm wondering if we should have a larger max height 🤔 .

https://user-images.githubusercontent.com/14905380/114442240-a3487300-9bcc-11eb-8ee4-bd7c9388248b.mp4

## Import notes

For this dialog we could try to apply a similar fix to what we did in [PR](https://github.com/Automattic/simplenote-electron/pull/2816).

https://user-images.githubusercontent.com/14905380/114442258-a9d6ea80-9bcc-11eb-833d-5eb061b91e84.mp4

## Settings

I think the expected behaviour here would be that only the panel content overflows so the tab list is not scrolled.

https://user-images.githubusercontent.com/14905380/114442275-ae030800-9bcc-11eb-9294-e061974bbc15.mp4

pachlava:
@sandymcfadden Please let me know if the case below will be covered here, or it's a separate issue?

https://user-images.githubusercontent.com/73365754/124119146-d1767a00-da7a-11eb-86c7-db65e0128300.mov

sandymcfadden:
Thanks @pachlava this looks like a separate issue. 
Looks like a couple things are going on here. One the error message height is wrong causing that scroll, and the wrong file type error doesn't appear until you add a valid file.

Issue #2716, opened 2021-03-04
Unsynchronized notes warning needs a height and overflow


codebykat:
If you have a long list of unsynchronized notes and attempt to log out, it will take up the full height of the browser window, with no way to see the options.

### Expected
The list of notes should be within a container with an `overflow:scroll` on it and a `max-height`.

### Observed

      
<img width="504" alt="Screen Shot 2021-03-03 at 10 12 45 PM" src="https://user-images.githubusercontent.com/52152/109919971-6d7baa80-7c6e-11eb-8ed9-308da4c540d7.png">
<img width="254" alt="Screen Shot 2021-03-03 at 10 13 46 PM" src="https://user-images.githubusercontent.com/52152/109919979-710f3180-7c6e-11eb-9c95-8f6ac59a0a5e.png">

### Reproduced
Safari/OSX

PR #2815, opened 2021-03-29
Fix: Add max height to dialog's content to prevent cut off

### LINKED ISSUE #2750: Modals become obscured in short windows or with zoomed UI
If a user has a short window height or is leveraging a zoomed UI (i.e. <kbd>Cmd</kbd>+<kbd>+</kbd> to increase size of text), several modals become obscured and do not provide scrolling to see the entirety of their content.

Obviously the impact of this issue is dependent upon _how much_ zooming is applied. So, it may not occur for the majority of users, but could negatively impact users who have smaller screens or require larger type for reading.

### Expected
If a modal overflows the height of the window, one can scroll to reveal the rest of its content.

### Observed
If a modal overflows the height of the window, there is not a way to view the entirety of the modal content.

### Reproduced
1. Shorten the browser window height to less than ~600px.
2. Increase the zoom of the browser window (e.g. <kbd>Cmd</kbd>+<kbd>+</kbd> a couple of times).
3. Click the Menu button in the top-right.
4. Click the About link in the bottom of the sidebar.

This issue occurs on each modal I encountered. There may be more occurrences than the following.

<details>
<summary>About Modal</summary>
modal-about
</details>

<details>
<summary>Review Your Account Modal</summary>
modal-review-account
</details>

<details>
<summary>Keyboard Shortcuts Modal</summary>
modal-keyboard-shortcuts
</details>

### Where did you see the bug
- System Make: Apple
- System Model: MacBook Pro
- OS: macOS
- OS version: 11.2.2
- Browser (if applicable): Safari
- Browser version (if applicable): 14.0.3

fluiddot:
### Fix

Fixes: https://github.com/Automattic/simplenote-electron/issues/2750

If a dialog overflows the height of the window, the content of the dialog is cut off.

https://user-images.githubusercontent.com/14905380/113024622-40e08300-9187-11eb-9ecd-99201374fa6c.mp4

### Test

1. Click the Menu button.
2. Click the "Keyboard Shortcuts" button.
3. Zoom in the window a couple of times (can be done via `View/Zoom In`).
4. Decrease the height of the browser window until it's smaller than the dialog height.
5. Observe that the dialog's content is not cut off and can be scrolled.

### Release

Fixed cut off issue for dialogs when the window is too small.

### DIFF
--- a/lib/dialog/style.scss
+++ b/lib/dialog/style.scss
@@ -1,15 +1,23 @@
+$dialog-max-height: 420px;
+$dialog-title-height: 56px;
+
 .dialog {
   display: flex;
   flex-direction: column;
   width: calc(100vw - 2rem);
   background: white;
   border-radius: 8px;
+  max-height: $dialog-max-height;
+
+  @media only screen and (max-height: $dialog-max-height) {
+    max-height: calc(100vh - 2rem);
+  }
 }
 
 .dialog-title-bar {
   display: flex;
   border-bottom: 1px solid $studio-gray-5;
-  height: 56px;
+  height: $dialog-title-height;
 
   .button {
     border: 0;
@@ -39,7 +47,12 @@
   display: flex;
   flex-direction: column;
   flex: 1 0 auto;
-  max-height: calc(100vh - 56px - 20px);
+  max-height: $dialog-max-height - $dialog-title-height;
+  overflow: auto;
+
+  @media only screen and (max-height: $dialog-max-height) {
+    max-height: calc(100vh - 2rem - #{$dialog-title-height});
+  }
 }
 
 .dialog-actions {
--- a/lib/dialog/style.scss
+++ b/lib/dialog/style.scss
@@ -39,6 +39,7 @@
   display: flex;
   flex-direction: column;
   flex: 1 0 auto;
+  max-height: calc(100vh - 56px - 20px);
 }
 
 .dialog-actions {


lib/dialog/style.scss
```diff
@@ -39,6 +39,7 @@
   display: flex;
   flex-direction: column;
   flex: 1 0 auto;
+  max-height: calc(100vh - 56px - 20px);
```
fluiddot: The max height of the content view is calculated by subtracting the header height (`56px`) and some padding (`20px`) to the window height (`100vh`).

sandymcfadden: Thanks so much for diving in and starting on fixes!
There are a couple of small things. In a [previous PR](https://github.com/Automattic/simplenote-electron/pull/2685#issuecomment-797630876), I was working on one dialog in particular and took a similar approach to what you did here to set the max height. It was decided instead that we could use a height of 420px as the max instead. 
The dialog in your video works well as one of the inner elements of the dialog has an overflow set for it. For other dialogs this can cause the content to overflow out of it. Would be good to add an `overflow: auto;` to this as well.
Not sure how far you want to go with this but #2777 is sort of related to this as well. 
The plan was to have the entire height of the dialog be a max of 420px, then have the inner content be scrollable if needed. The header, and if there were any action buttons with the dialog at the bottom, would remain static and always visible without scrolling.

sandymcfadden (changes requested):

lib/dialog/style.scss
fluiddot: >There are a couple of small things. In a previous PR, I was working on one dialog in particular and took a similar approach to what you did here to set the max height. It was decided instead that we could use a height of 420px as the max instead.

I just checked that the [minimum window is set to `520px`](https://github.com/Automattic/simplenote-electron/blob/f8165a48c35423e6bc83513ad7cf9bd77f9e7aea/desktop/app.js#L67:L67) so `420px` should be enough. However the content is still cut off if you zoom in the UI (see attached video), as far as I checked the only way to prevent it is by using the viewport height with `vh` unit 😞 .

https://user-images.githubusercontent.com/14905380/112873218-a36f4b80-90c1-11eb-87d2-566b2b57302d.mp4

>The dialog in your video works well as one of the inner elements of the dialog has an overflow set for it. For other dialogs this can cause the content to overflow out of it. Would be good to add an overflow: auto; to this as well.

Good point! I think it would be a great idea to add overflow to the content view, this way we assure that it's always scrollable. 

Another thing it would be interesting is to review the dialogs and set overflow to the sections that would require it. As an example, I'm working on this [issue](https://github.com/Automattic/simplenote-electron/issues/2716) related `UnsynchronizedConfirmation` dialog and here I think makes more sense to make the changes list scrollable not the whole content (check follow video).

https://user-images.githubusercontent.com/14905380/112873709-47f18d80-90c2-11eb-85ad-217e794416bf.mp4

>Not sure how far you want to go with this but #2777 is sort of related to this as well.

I'm still getting familiar with the project but I can take a look if you want.

fluiddot: One option I'm considering, although I'm not sure about [potential compatibility issues](https://caniuse.com/css-math-functions), is to use this for calculating the max height:
`max-height: unquote('min(100vh - 20px, 420px)');`.

This will allow us to set the maximum height to `420px` but if for any reason the window gets smaller (looks like this is what's happening when zooming in), we use the window height to prevent the cut off.

Note the use of `unquote` for prevent using the Sass's `min` function because it [doesn't support `vh` units](https://sass-lang.com/documentation/syntax/special-functions#min-and-max).

sandymcfadden: Using the min function seems like a good approach to me here to get the 420px height, but also ensure it works on zoomed in views.

My thoughts on #2777 were very similar to what you're thinking about for the `UnsynchronizedConfirmation` dialog.

fluiddot: After a second thought, I decided to use media queries instead of this approach because it looks way safer ([related commit](https://github.com/Automattic/simplenote-electron/pull/2815/commits/b91a2f115890225a1193997570fc62ac5bad54a5)).

_EDIT: I've also updated the video from the description._

lib/dialog/style.scss
```diff
@@ -39,6 +47,12 @@
   display: flex;
   flex-direction: column;
   flex: 1 0 auto;
+  max-height: $dialog-max-height - $dialog-title-height;
+  overflow: auto;
+
+  @media only screen and (max-height: $dialog-max-height) {
+    max-height: calc(100vh - 2rem - #{$dialog-title-height});
```
sandymcfadden, line 54: I'm not yet had to use Interpolation in Sass, is it needed here? I believe as this is a variable it should work without just as `max-height: calc(100vh - 2rem - $dialog-title-height);`

sandymcfadden (approved):
I left one comment about the use of interpolation, but other than that this looks good, thank you! I like the media query approach.
This tests well on all the dialogs I tried where it doesn't fall out of the screen and I can access all the content in it.

lib/dialog/style.scss
fluiddot: I checked the [documentation for this function](https://sass-lang.com/documentation/syntax/special-functions#calc-clamp-element-progid-and-expression) and looks like that interpolation is required here for using the `calc` function from CSS:
>The calc(), clamp() and element() functions are defined in the CSS spec. Because calc()’s mathematical expressions conflict with Sass’s arithmetic, and element()’s IDs could be parsed as colors, they need special parsing.

In fact if I remove the interpolation, it produces an invalid CSS property (see attached screenshot).

Screenshot 2021-03-31 at 18 15 38

sandymcfadden: Ah thank you! I had tested that and everything looked good in the devtools to me, but I must have been looking at the second rule there and missed the first that has the error. 
All looks good to me.

PR #2816, opened 2021-03-30
Fix: Add overflow to changes list section of unsynchronized confirmation dialog

### LINKED ISSUE #2716: Unsynchronized notes warning needs a height and overflow

fluiddot:
### Fix

Fixes https://github.com/Automattic/simplenote-electron/issues/2716

This PR adds overflow to the changes list section of the unsynchronized confirmation dialog, this will enable scroll if the list is too long while keeping the action buttons visible.

https://user-images.githubusercontent.com/14905380/112972237-571f1c80-9150-11eb-8d72-206f794bcfce.mp4

### Test

1.  Go offline to prevent saving notes.
2. Create 10 or more notes.
3. Make the window height as small as possible.
4. Exit the application.
5. Observe that the unsynchronized confirmation dialog displays all the content and that the changes list is scrollable.

### Release

N/A

### DIFF
--- a/lib/dialogs/unsynchronized/style.scss
+++ b/lib/dialogs/unsynchronized/style.scss
@@ -31,6 +31,7 @@
   .change-list {
     border-bottom: 1px solid $studio-gray-5;
     border-top: 1px solid $studio-gray-5;
+    overflow-y: auto;
 
     ul {
       list-style: none;


lib/dialog/style.scss
```diff
@@ -39,6 +39,7 @@
   display: flex;
   flex-direction: column;
   flex: 1 0 auto;
+  max-height: calc(100vh - 56px - 20px);
```
fluiddot: This change comes from https://github.com/Automattic/simplenote-electron/pull/2815.

lib/dialogs/unsynchronized/style.scss
```diff
@@ -31,6 +31,7 @@
   .change-list {
     border-bottom: 1px solid $studio-gray-5;
     border-top: 1px solid $studio-gray-5;
+    overflow-y: auto;
```
fluiddot, line 34: This is the only change from this PR, the rest are coming from the original [PR](https://github.com/Automattic/simplenote-electron/pull/2815).

sandymcfadden (approved):
This makes that the unsynced notes screen much nicer. Thank you.

I did leave [a question](https://github.com/Automattic/simplenote-electron/pull/2815#discussion_r604844143) on the related PR about the use of interpolation for the sass variable which is used here as well.

PR #2831, opened 2021-04-08
Updates to the sign up form design and account creation copy

### LINKED ISSUE #2823: Update Sign up copy and design
With the recent updates to the sign up flow, there are some copy and design changes to be made. 

The copy changes can be found in the internal thread p2XJRt-2R6-p2

Designs to match these:
https://app.zeplin.io/project/5ff5d9f41ff2fa53ddd0d08c/screen/606c24fd8561ee914adabeeb
https://app.zeplin.io/project/5ff5d9f41ff2fa53ddd0d08c/screen/606c24fe675bec972a37afa6
<details><summary>Light Mode</summary>

</details>
<details><summary>Dark Mode</summary>

</details>

sandymcfadden:
### Fix

With the new sign up flow, there were some copy changes and form design changes that needed to be updated. This updates those.

**Before Screenshots**

**Light mode**

<details>
<summary>Initial form</summary>

</details>
<details>
<summary>Initial form with focus</summary>

</details>
<details>
<summary>Form with email and focus</summary>

</details>
<details>
<summary>Form with email</summary>

</details>
<details>
<summary>Next steps message</summary>

</details>

**Dark mode**

<details>
<summary>Initial form</summary>

</details>
<details>
<summary>Initial form with focus</summary>

</details>
<details>
<summary>Form with email and focus</summary>

</details>
<details>
<summary>Form with email</summary>

</details>
<details>
<summary>Next steps message</summary>

</details>

**After Screenshots**

**Light mode**

<details>
<summary>Initial form</summary>

</details>
<details>
<summary>Initial form with focus</summary>

</details>
<details>
<summary>Form with email and focus</summary>

</details>
<details>
<summary>Form with email</summary>

</details>
<details>
<summary>Next steps message</summary>

</details>

**Dark mode**

<details>
<summary>Initial form</summary>

</details>
<details>
<summary>Initial form with focus</summary>

</details>
<details>
<summary>Form with email and focus</summary>

</details>
<details>
<summary>Form with email</summary>

</details>
<details>
<summary>Next steps message</summary>

</details>

Resolves #2823

### Test

Test in both light and dark mode.

1. Ensure you are not logged in.
2. Choose to sign up for an account.
3. Ensure the form matches the designs.
4. Ensure the copy after requesting matches the updated copy.

### Release

- Updated the design of the login/sign up form as well as clarifying the message shown when requesting an account.

### DIFF
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -222,5 +222,10 @@
       color: $studio-gray-50;
       border-color: $studio-gray-70;
     }
+
+    &:read-only {
+      border-color: $studio-gray-60;
+      background-color: rgba(255, 255, 255, 0.07);
+    }
   }
 }
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -114,6 +114,7 @@
     color: $studio-simplenote-blue-50;
     margin-left: 5px;
     text-decoration: none;
+    white-space: nowrap;
   }
 
   .or {
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -64,12 +64,6 @@
     font-weight: 400;
   }
 
-  input:focus {
-    border-color: $studio-simplenote-blue-50;
-    box-shadow: 0 0 0 1px $studio-simplenote-blue-50;
-    outline: none;
-  }
-
   input:disabled {
     border-color: $studio-gray-5;
     color: $studio-gray-20;
@@ -78,6 +72,12 @@
   input:read-only {
     border-color: $studio-gray-10;
     background-color: rgba(0, 0, 0, 0.05);
+  }
+
+  input:focus {
+    border-color: $studio-simplenote-blue-50;
+    box-shadow: 0 0 0 1px $studio-simplenote-blue-50;
+    outline: none;
   }
 
   .spinner {
@@ -215,11 +215,6 @@
   input {
     border-color: $studio-gray-60;
 
-    &:focus {
-      border-color: $studio-simplenote-blue-30;
-      box-shadow: 0 0 0 1px $studio-simplenote-blue-30;
-    }
-
     &:disabled {
       color: $studio-gray-50;
       border-color: $studio-gray-70;
@@ -228,6 +223,11 @@
     &:read-only {
       border-color: $studio-gray-60;
       background-color: rgba(255, 255, 255, 0.07);
+    }
+
+    &:focus {
+      border-color: $studio-simplenote-blue-30;
+      box-shadow: 0 0 0 1px $studio-simplenote-blue-30;
     }
   }
 }
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -65,8 +65,8 @@
   }
 
   input:focus {
-    border-color: transparent;
-    box-shadow: 0 0 0 2px $studio-simplenote-blue-50;
+    border-color: $studio-simplenote-blue-50;
+    box-shadow: 0 0 0 1px $studio-simplenote-blue-50;
     outline: none;
   }
 
@@ -216,7 +216,8 @@
     border-color: $studio-gray-60;
 
     &:focus {
-      box-shadow: 0 0 0 2px $studio-simplenote-blue-30;
+      border-color: $studio-simplenote-blue-30;
+      box-shadow: 0 0 0 1px $studio-simplenote-blue-30;
     }
 
     &:disabled {
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -65,7 +65,8 @@
   }
 
   input:focus {
-    border: 2px solid $studio-simplenote-blue-50;
+    border-color: transparent;
+    box-shadow: 0 0 0 2px $studio-simplenote-blue-50;
     outline: none;
   }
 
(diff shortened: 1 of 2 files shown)


SylvesterWilmott:
Looks good @sandymcfadden 

Can we add "nowrap" to the "Terms of Service" link or maybe a line break before it so it doesn't wrap mid-link?

When focusing the fields, the text does a small jump, is this because of the increase in border size? Maybe using a box shadow for the focus border would be better in this case (similar to the `validate:invalid` style).

lib/auth/style.scss
```diff
@@ -194,4 +210,22 @@
   svg.icon-mail {
     color: $studio-gray-30;
   }
+  input {
+    border-color: $studio-gray-60;
+
+    &:focus {
+      border: 2px solid $studio-simplenote-blue-30;
+      outline: none;
```
fluiddot: I was wondering if since some `border` properties are already defined in the default theme (light), here we could only have the border color as we have in the other states (disabled and read-only):
`border-color: $studio-simplenote-blue-30;`

And the same for `outline`, I think we could remove it from here too.

fluiddot (commented):
I tested it and looks great 🎊 , I added just a minor comment.

sandymcfadden:
Thanks @SylvesterWilmott!
> Can we add "nowrap" to the "Terms of Service" link or maybe a line break before it so it doesn't wrap mid-link?
I've updated this now to fix the wrapping. 
 
> When focusing the fields, the text does a small jump, is this because of the increase in border size? Maybe using a box shadow for the focus border would be better in this case (similar to the `validate:invalid` style).

You're right it was because of the change in border size. I've updated it to use a box-shadow now instead as you suggested.

lib/auth/style.scss
sandymcfadden: Thanks! You're right that is a good call. Based on some other feedback I've switched to using box-shadow instead of border, and I don't believe there is a way to change only the color.
I did drop the outline though.

lib/auth/style.scss
```diff
@@ -63,6 +64,22 @@
     font-weight: 400;
   }
 
+  input:focus {
+    border-color: transparent;
+    box-shadow: 0 0 0 2px $studio-simplenote-blue-50;
```
fluiddot: I saw that the border is maybe expanded too much when the element is focused (see attached screenshot), does this match the designs? (I can't check them because I don't have access 😄 )
Screenshot 2021-04-13 at 11 16 08

I was wondering if we could set the border color with the same blue and instead of having two pixels in the box shadow, reduce it to one (see attached screenshot):
```
border-color: $studio-simplenote-blue-50;
box-shadow: 0 0 0 1px $studio-simplenote-blue-50;
```

Screenshot 2021-04-13 at 11 19 33

<hr>

Additionally I'm not sure how the element should look like when it's invalid and focused, now the border is set to grey but the focus color should be blue, is this right?

Here you can check how it looks like with the current changes and what I proposed:
Current changes|Proposal
--|--
<video src=https://user-images.githubusercontent.com/14905380/114528023-e516ff00-9c48-11eb-99a8-66f2fa465564.mp4>|<video src=https://user-images.githubusercontent.com/14905380/114528110-fb24bf80-9c48-11eb-85c4-f3e1e908094e.mp4>

lib/auth/style.scss
```diff
@@ -194,4 +212,21 @@
   svg.icon-mail {
     color: $studio-gray-30;
   }
+  input {
+    border-color: $studio-gray-60;
+
+    &:focus {
+      box-shadow: 0 0 0 2px $studio-simplenote-blue-30;
```
fluiddot: I think we have to add here `border-color: transparent;`, otherwise the dark theme sets the border color to grey when it's focused as per line 216.

Here is how it looks like currently, observe the thin inner grey line:
Screenshot 2021-04-13 at 11 28 16

And here are the styles of the email input when it's focused, observe that `.theme-dark .login input` sets the border color to grey:
Screenshot 2021-04-13 at 11 29 34

sandymcfadden: Thanks for catching this. 
I think I'll solve this by going with the recommendation in your other comment to use a combination of border and box-shadow.

lib/auth/style.scss
sandymcfadden: I like the suggestion of going with both border and box-shadow. I'm going to work on that. I have screenshots of the designs in the issue related to this #2823 under the light and dark mode toggles. 

@SylvesterWilmott how would you want the invalid state to look when focused? I had left it as it was previously, but we could update it for sure.

SylvesterWilmott: >I had left it as it was previously, but we could update it for sure.

As it is currently looks good to me!

lib/auth/style.scss
```diff
@@ -63,6 +64,22 @@
...
+
+  input:disabled {
+    border-color: $studio-gray-5;
+    color: $studio-gray-20;
+  }
+
+  input:read-only {
+    border-color: $studio-gray-10;
+    background-color: rgba(0, 0, 0, 0.05);
+  }
```
fluiddot, line 75: I saw that it's possible to have both `focus` and `read-only` states applied at the same time. In this case the `read-only` styles have preference over the `focus` ones because they're located after and therefore the inner border is set to grey. 

If we want to keep the same styles (only have blue border), we should put this part before `input:focus` and the same in the code below for the dark theme.

Here you can check how it looks like when a `readonly` input is focused:
Screenshot 2021-04-14 at 10 05 52

sandymcfadden: Another good catch, thank you! I've updated this appropriately now.

fluiddot (approved):
LGTM 🎊 !

SylvesterWilmott (approved):
Looks good! Thanks @sandymcfadden

PR #2834, opened 2021-04-09
Fix: Remove fixed height from tab panels

fluiddot:
### Fix

When navigating to `Settings/Display` the content was displaying two scroll bars, one set by the dialog content (introduced in this recent [PR](https://github.com/Automattic/simplenote-electron/pull/2815)) and the other by the tab panels component due to having a fixed height.

Since the dialog component is already setting a max height, we don't really need to specify a fixed height in the tab panels so I removed that CSS propery.

Before|After
--|--
Screenshot 2021-04-09 at 16 41 17|Screenshot 2021-04-09 at 16 41 28

### Test

1. Go to Settings
2. Click on Display
3. Observe that the content can be scrolled and that there's only one scroll bar

### Release

N/A

### DIFF
--- a/lib/components/tab-panels/style.scss
+++ b/lib/components/tab-panels/style.scss
@@ -35,7 +35,6 @@
 }
 
 .tab-panels__panel {
-  height: 30.5em;
   overflow: auto;
   -webkit-overflow-scrolling: touch;
 }


sandymcfadden (approved):
Good call and catch, thank you!

PR #2835, opened 2021-04-09
Add: Standard checkbox style

fluiddot:
### Fix

I noticed that a recent [PR](https://github.com/Automattic/simplenote-electron/pull/2622) introduced the standard version of checkbox. Since this could be used in different parts of the app, I decided to make it generic and add a specific class name (`standard`) into the checkbox component.

This PR also introduce a style fix for the size of the check box:

Before|After
--|--
Screenshot 2021-04-09 at 17 48 43|Screenshot 2021-04-09 at 17 49 00

#### Standard checkboxes
1. Open a note
2. Click on actions button on the top-right corner
3. Observe that standard checkboxes are displayed properly

#### Default checkboxes
1. Open settings
2. Click on Display
3. Observe that default checkboxes are displayed properly

### Release

N/A

### DIFF
--- a/lib/note-actions/index.tsx
+++ b/lib/note-actions/index.tsx
@@ -93,8 +93,8 @@ export class NoteActions extends Component<Props> {
               <span className="note-actions-item-control">
                 <CheckboxControl
                   id="note-actions-pin-checkbox"
-                  className="theme-color-border"
                   checked={isPinned}
+                  className="standard"
                   onChange={() => {
                     this.pinNote(!isPinned);
                   }}
@@ -113,6 +113,7 @@ export class NoteActions extends Component<Props> {
                 <CheckboxControl
                   id="note-actions-markdown-checkbox"
                   checked={isMarkdown}
+                  className="standard"
                   onChange={() => {
                     this.markdownNote(!isMarkdown);
                   }}
@@ -158,6 +159,7 @@ export class NoteActions extends Component<Props> {
                 <CheckboxControl
                   id="note-actions-publish-checkbox"
                   checked={isPublished}
+                  className="standard"
                   onChange={() => {
                     this.publishNote(!isPublished);
                   }}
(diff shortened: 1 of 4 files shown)


lib/note-actions/index.tsx
```diff
@@ -93,8 +93,8 @@ export class NoteActions extends Component<Props> {
               <span className="note-actions-item-control">
                 <CheckboxControl
                   id="note-actions-pin-checkbox"
-                  className="theme-color-border"
                   checked={isPinned}
+                  className="standard"
```
fluiddot: I wasn't sure about the best way to set the standard styles, wdyt? would you rather have a prop for picking it?

lib/note-actions/style.scss
```diff
@@ -73,30 +71,13 @@ $checkbox-sprite: 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
   }
 
   .note-actions-item-control .checkbox-control {
-    bottom: 3px;
+    bottom: 1px;
```
fluiddot, line 74: I recalculated this value because the checkbox with standard styles is two pixels bigger.

lib/note-actions/index.tsx
sandymcfadden: Thanks so much for picking this up. This makes good sense to move it into the component. 
I wasn't sure the best way to do this either. It looks like we're going to bring in the WordPress Components to Simplenote Electron/Web at some point soon and it seems they use props to set options like this on components. Maybe we should do that also to be consistent.

fluiddot: Ok, makes sense, I'm going to move it to a prop then 👍 .

fluiddot: I applied this change in the following [commit](https://github.com/Automattic/simplenote-electron/pull/2835/commits/1e09b59c135b11ade2ee96abb081ab3f9ac53d16).

sandymcfadden (approved):
Thanks!
Both the standard and the original circle version test well and the code looks good to me.

PR #2837, opened 2021-04-09
[DO NOT MERGE] Add note's tags to history screen - Use action generator

fluiddot:
This PR shows a potential alternative for generating a Redux action based on data from the state.

The idea I introduced here is similar to [Thunk actions](https://github.com/reduxjs/redux-thunk) but quite simplified just to demonstrate how it could work. Basically I added a middleware that checks if the action is a function, if so it does a call with the current state and uses the return as the action to be passed to the next middleware. This way we could dispatch functions (aka action generators) that produce an action object based on the current state.

This option could help us to free up the middlewares and migrate some logic we're doing in the middlewares to actions. One good example for this would be the [`CREATE_NOTE` action](https://github.com/Automattic/simplenote-electron/blob/2c36f8a97d8a0b8206d49079c6571e35daf55fa1/lib/state/data/middleware.ts#L18-L47), here we’re using a Redux action to generate another one based on data from the state.

### DIFF
--- a/lib/state/data/actions.ts
+++ b/lib/state/data/actions.ts
@@ -1,5 +1,8 @@
 import * as A from '../action-types';
 import * as T from '../../types';
+import * as S from '..';
+import isEmailTag from '../../utils/is-email-tag';
+import { tagHashOf } from '../../utils/tag-hash';
 
 export const addCollaborator: A.ActionCreator<A.AddCollaborator> = (
   noteId: T.EntityId,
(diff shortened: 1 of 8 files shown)


lib/state/data/actions.ts
```diff
@@ -77,13 +80,33 @@ export const toggleAnalytics: A.ActionCreator<A.ToggleAnalytics> = () => ({
...
+  });
+
+  return {
+    type: 'RESTORE_NOTE_REVISION',
+    noteId,
+    note: {
+      ...revision,
+      tags: [...noteEmailTags, ...revisionCanonicalTags],
+    },
+  };
```
fluiddot, line 111: This action contains the logic that we had in the data middleware.

lib/state/index.ts
```diff
@@ -50,6 +50,20 @@ export type State = {
...
 
+const actionGeneratorMiddleware: Middleware = (store) => (
+  next: (action: A.ActionType) => A.ActionType
+) => (action: A.ActionType | ReturnType<ActionGenerator>) => {
+  const state = store.getState();
+  if (typeof action === 'function') {
+    const result = action(state);
+    if (!result || !result.type) {
+      return;
+    }
```
fluiddot, line 61: This gives control to the action generator to prevent the action dispatch, it could be useful if for example we determine that the Redux action shouldn't be dispatched.

dmsnell:
Thanks for exploring this route. I strongly prefer avoiding this kind of change because it fundamentally breaks the Redux model that we still maintain; that is, it prevents us from having a synchronous store update with inspectable and fake-able actions. The middleware gives us a chance to perform operations on the actions that leaves a fairly clear audit trail. Once we hide computation in thunks we lose all that.

fluiddot:
> Thanks for exploring this route. I strongly prefer avoiding this kind of change because it fundamentally breaks the Redux model that we still maintain; that is, it prevents us from having a synchronous store update with inspectable and fake-able actions. The middleware gives us a chance to perform operations on the actions that leaves a fairly clear audit trail. Once we hide computation in thunks we lose all that.

Yeah, unfortunately this option as well as thunk actions comes with their tradeoffs. Ok, let's avoid this option, thank you very much for review 🙇 !

PR #2838, opened 2021-04-09
[DO NOT MERGE] Add note's tags to history screen - Add cross slice reducer

fluiddot:
This PR shows an example of how to share data between reduce slicers.

The idea exposed here is based on the [Redux documentation](https://redux.js.org/recipes/structuring-reducers/beyond-combinereducers#sharing-data-between-slice-reducers) about how to do a cross slice reducer when you use `combineReducers`.

### DIFF
--- a/lib/state/data/reducer.ts
+++ b/lib/state/data/reducer.ts
@@ -6,6 +6,7 @@ import {
   withTag,
   withoutTag,
 } from '../../utils/tag-hash';
+import isEmailTag from '../../utils/is-email-tag';
 
 import type * as A from '../action-types';
 import type * as T from '../../types';
@@ -108,7 +109,6 @@ export const notes: A.Reducer<Map<T.EntityId, T.Note>> = (
 
     case 'NOTE_BUCKET_UPDATE':
     case 'REMOTE_NOTE_UPDATE':
-    case 'APPLY_NOTE_REVISION':
       return new Map(state).set(action.noteId, action.note);
 
     case 'IMPORT_NOTE_WITH_ID': {
@@ -437,18 +437,6 @@ export const tags: A.Reducer<Map<T.TagHash, T.Tag>> = (
       return next;
     }
 
-    case 'APPLY_NOTE_REVISION': {
-      const next = new Map(state);
-      action.note.tags.forEach((tagName) => {
-        const tagHash = t(tagName);
-        if (!next.has(tagHash)) {
-          next.set(tagHash, { name: tagName });
-        }
-      });
-
-      return next;
-    }
-
     case 'TRASH_TAG': {
       const next = new Map(state);
       return next.delete(t(action.tagName)) ? next : state;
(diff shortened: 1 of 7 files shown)


lib/state/data/reducer.ts
```diff
@@ -553,3 +543,56 @@ export default combineReducers({
   preferences,
   tags,
 });
+
+const crossSliceReducer = (
+  state: DataState,
+  action: A.ActionType
+): DataState => {
+  switch (action.type) {
+    case 'RESTORE_NOTE_REVISION': {
```
fluiddot, line 552: We could have this part in a separate function that only receives the required state slices (`notes` and `tags`).

lib/state/data/reducer.ts
```diff
@@ -553,3 +543,56 @@ export default combineReducers({
...
+    default:
+      return state;
+  }
+};
+
+const dataReducer: A.Reducer<DataState> = (state, action) => {
+  const intermediateState = combinedReducer(state, action);
+  const finalState = crossSliceReducer(intermediateState, action);
+  return finalState;
+};
```
fluiddot, line 596: We would have to be careful with this approach because, for the cross slice reducer, we use the resulting state from `combineReducer`.

If for example we watch for the `RESTORE_NOTE_REVISION` action in one of the slices of `combineReducer` and modify the state, we could have an undesired behaviour when executing the cross slice reducer.

lib/state/simperium/middleware.ts
```diff
@@ -351,9 +352,19 @@ export const initSimperium = (
...
-        action.note.tags.map(t).forEach((tagHash) => {
-          if (!prevState.data.tags.has(tagHash)) {
+      case 'RESTORE_NOTE_REVISION': {
+        const revision = nextState.data.noteRevisions
+          .get(action.noteId)
+          ?.get(action.version);
+
+        if (!revision) {
+          return result;
+        }
```
fluiddot, line 362: This part won't be required if we keep it in the data middleware.

dmsnell:
Similar to my comment in #2837 I would prefer we avoid doing this especially by giving the reducer all of the state. There's the danger you point out in that the combined-slice reducer operates on the state tree _after_ some of the updates have already hit. This breaks the fundamental model again which is independent data types operating and producing an atomic state update for the app. I like more what you have going in #2817 but I think one big thing we need to examine is whether we're trying too hard to avoid something that's normal and normative in this app, which is using the middleware for action transformation.

fluiddot:
> Similar to my comment in #2837 I would prefer we avoid doing this especially by giving the reducer all of the state. There's the danger you point out in that the combined-slice reducer operates on the state tree _after_ some of the updates have already hit. This breaks the fundamental model again which is independent data types operating and producing an atomic state update for the app. I like more what you have going in #2817 but I think one big thing we need to examine is whether we're trying too hard to avoid something that's normal and normative in this app, which is using the middleware for action transformation.

Yep, I was also concerned about this. Ok, let's avoid this option too and focus on the ongoing solution in #2817, thanks for taking a look into this 🙇 !

PR #2842, opened 2021-04-13
Disable note preview selection/copy when note actions is open

### LINKED ISSUE #2836: Note action copy buttons won't copy when previewing note
### Expected

When you click to copy the internal link, or the publish URL that those links would copy to the clipboard.

### Observed
If you click to view the markdown preview of a note and then click one of the copy buttons the link is not copied. Depending on the browser sometimes the note content is copied instead.

### Reproduced

1. Create a note with markdown.
2. Click to view the markdown preview.
3. Click the note actions icon.
4. Click on the Copy internal link button.
5. Try to paste the links.

### Where did you see the bug
Happens in multiple browsers, Firefox, Chrome and in the Electron build.

sandymcfadden:
### Fix

Fixes #2836

Currently when you view the markdown preview of a note and then open the note actions and try to use one of the Copy buttons to copy either the publish or internal link the content of the note is copied instead.

This resolves that by first checking to see if note actions is open before selecting the note content in the note preview component.

### Test

1. Open a note with markdown enabled and published.
2. Open the note actions menu for this note.
3. Ensure the Copy Internal Link button copies the link.
4. Ensure the Copy Link button works as well.
5. Toggle to view the markdown preview.
6. Test both buttons again and ensure they both still work.

### Release

None needed.

### DIFF
--- a/lib/components/note-preview/index.tsx
+++ b/lib/components/note-preview/index.tsx
@@ -185,7 +185,7 @@ const mapStateToProps: S.MapState<StateProps, OwnProps> = (state, props) => {
   const note = props.note ?? state.data.notes.get(noteId);
 
   return {
-    isFocused: state.ui.dialogs.length === 0 && !state.ui.showNoteInfo,
+    isFocused: state.ui.dialogs.length === 0 && !state.ui.showNoteActions,
     note,
     noteId,
     notes: state.data.notes,


codebykat (approved):
Nice, this works great 👍 

In my testing with prod (unfixed version) it actually copied nothing rather than the note content, not sure why, but in this PR it does work.

PR #2843, opened 2021-04-13
Fix: Set checkbox component `isStandard` prop optional

fluiddot:
### Fix

This [PR](https://github.com/Automattic/simplenote-electron/pull/2835) introduced a Typescript error in `RadioGroup` component because it's not setting the `isStandard` prop when rendering the checkboxes. I changed this prop to be optional so by default it renders the circle variant.

#### Standard checkboxes
1. Open a note
2. Click on actions button on the top-right corner
3. Observe that standard checkboxes are displayed properly

#### Default checkboxes
1. Open settings
2. Click on Display
3. Observe that default checkboxes are displayed properly
### Release

N/A

### DIFF
--- a/lib/controls/checkbox/index.tsx
+++ b/lib/controls/checkbox/index.tsx
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 type OwnProps = React.HTMLProps<HTMLInputElement> & {
   className?: string;
   onChange: () => any;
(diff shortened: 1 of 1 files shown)


sandymcfadden (approved):
I should have caught this on the last PR, glad you caught it.

PR #2844, opened 2021-04-14
Add note's tags to history screen - Refactor notes revision and revision selector

fluiddot:
Related to comment: https://github.com/Automattic/simplenote-electron/pull/2817#discussion_r611979568

The goal of this refactor is to handle the toggle restoring deleted tags in note revisions component state instead of in Redux.

For this purpose, the revision selector component is now rendered inside the notes revision component and most of its logic has been moved to that component too.

### Test

_Same testing instructions as the [parent PR](https://github.com/Automattic/simplenote-electron/pull/2817)._

### DIFF
--- a/lib/app-layout/index.tsx
+++ b/lib/app-layout/index.tsx
@@ -29,7 +29,6 @@ const NotePreview = React.lazy(() =>
 );
 
 type StateProps = {
-  hasRevisions: boolean;
   isFocusMode: boolean;
   isNavigationOpen: boolean;
   isNoteInfoOpen: boolean;
@@ -37,8 +36,6 @@ type StateProps = {
   isSmallScreen: boolean;
   keyboardShortcuts: boolean;
   keyboardShortcutsAreOpen: boolean;
-  openedNote: T.EntityId | null;
-  openedRevision: T.Note | null;
   showNoteList: boolean;
   showRevisions: boolean;
   showSortBar: boolean;
@@ -87,14 +84,11 @@ export class AppLayout extends Component<Props> {
   render = () => {
     const {
       showNoteList,
-      hasRevisions,
       isFocusMode = false,
       isNavigationOpen,
       isNoteInfoOpen,
       isNoteOpen,
       isSmallScreen,
-      openedNote,
-      openedRevision,
       showRevisions,
       showSortBar,
     } = this.props;
@@ -141,15 +135,10 @@ export class AppLayout extends Component<Props> {
             >
               <NoteToolbar aria-hidden={hiddenByRevisions} />
               {showRevisions ? (
-                <NoteRevisions
-                  aria-hidden={hiddenByRevisions}
-                  noteId={openedNote}
-                  note={openedRevision}
-                />
+                <NoteRevisions aria-hidden={hiddenByRevisions} />
               ) : (
                 <NoteEditor />
               )}
-              {hasRevisions && <RevisionSelector />}
             </main>
           )}
         </Suspense>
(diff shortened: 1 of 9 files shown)


lib/app-layout/index.tsx
```diff
@@ -141,15 +135,10 @@ export class AppLayout extends Component<Props> {
             >
               <NoteToolbar aria-hidden={hiddenByRevisions} />
               {showRevisions ? (
-                <NoteRevisions
-                  aria-hidden={hiddenByRevisions}
-                  noteId={openedNote}
-                  note={openedRevision}
```
fluiddot, line 147: I saw that these values are only used for this component so since the can be fetched from the state in the `connect` of `NoteRevisions` component, I decided to remove them to keep this component cleaner.

lib/note-revisions/index.tsx
```diff
@@ -3,70 +3,179 @@ import { connect } from 'react-redux';
...
+  getTags() {
+    return this.props.noteCanonicalTags
+      .map((tagName) => {
+        const tagHash = tagHashOf(tagName);
+        return { name: tagName, deleted: !this.props.allTags.has(tagHash) };
+      })
+      .filter((tag) => {
+        return this.state.includeDeletedTags || !tag.deleted;
+      });
+  }
```
fluiddot, line 54: These tags were calculated in the `getRevision` selector, now since we have to use the component's state `includeDeletedTags` for the calculation, we have to place it within the component.

lib/note-revisions/index.tsx
```diff
@@ -3,70 +3,179 @@ import { connect } from 'react-redux';
...
+
+    const noteEmailTags = note.tags.filter((tagName) => isEmailTag(tagName));
+    const revisionCanonicalTags = this.getTags().map(({ name }) => name);
+
+    restoreRevision(noteId, {
+      ...revision,
+      tags: [...noteEmailTags, ...revisionCanonicalTags],
+      systemTags: note.systemTags,
+    });
+  };
```
fluiddot, line 71: We calculate the revision object to be restored upon accepting the revision, this way we no longer require the `getRevisions` selector.

lib/note-revisions/index.tsx
```diff
@@ -3,70 +3,179 @@ import { connect } from 'react-redux';
...
+          onCancel={cancelRevision}
+          onSelect={this.onSelectRevision}
+          onToggleRestoreDeletedTags={(toggled) =>
+            this.setState({ includeDeletedTags: toggled })
+          }
+          restoreDeletedTags={this.state.includeDeletedTags}
+          revision={revision}
+          revisionsSize={revisions.size}
+          selectedIndex={revisionIndex}
+        />
```
fluiddot, line 133: I extracted all the required props for disconnecting the `RevisionSelector` component from Redux, this way `NoteRevision` handles the logic related to revision management and `RevisionSelector` only handles the UI logic.

dmsnell:
hi @fluiddot 👋 

wow, this was a lot of work! how do you feel about it? I didn't realize it would take as much as it seems to have and now I wonder if it's wroth it. what would you say?

as a side note, I want you to feel empowered with this stuff. ultimately it's your call on the PR and code and what you think is right is something we're glad to trust you with. I appreciate all the exploring you've done with me in reviewing this feature. as with many elements of the app, things get more complicated as we dig in.

fluiddot:
> hi @fluiddot 👋
> 
> wow, this was a lot of work! how do you feel about it? I didn't realize it would take as much as it seems to have and now I wonder if it's wroth it. what would you say?

Hello @dmsnell 👋 !

To be honest it wasn't too much, there're a lot of changes but it was more a work of moving the code around than actually coding. I knew from the beginning that these changes could be discarded but I wanted to provide a different approach to what I was doing in https://github.com/Automattic/simplenote-electron/pull/2817, explore the option of decoupling a bit the revisions logic from Redux and see how that would look like.

Overall I'd say it was useful to do this exercise and it could help you in the future if something similar comes across.

> as a side note, I want you to feel empowered with this stuff. ultimately it's your call on the PR and code and what you think is right is something we're glad to trust you with. I appreciate all the exploring you've done with me in reviewing this feature. as with many elements of the app, things get more complicated as we dig in.

Thanks! As I posted in this [comment](https://github.com/Automattic/simplenote-electron/pull/2817#discussion_r616011678), I think it's better to keep with the Redux approach as it follows the architecture of the app.

PR #2846, opened 2021-04-15
Fix search selections when content is changed

### LINKED ISSUE #2809: [Search] Match position indicator does not update when matches number changes
### Expected
If the note is edited while search mode is active, the position indicator (`2 of 5` thing) should update the current position digit, if matches number changes.

### Observed
Current position indicator never changes regardless of matches number:

https://user-images.githubusercontent.com/73365754/112330234-86e9a280-8cc0-11eb-895e-d97d49d43ac9.mov

### Reproduced
1. Have a note with multiple instances of `word` in it
2. Search for `word`
3. Move the current match position in a note with `>` button or with `Ctrl/Cmd+G`, so that it becomes greater than one (e.g. `5 of 5`)
4. Start removing the instances of `word` until one remains ~> the indicator will show `5 of 1`

### Where did you see the bug
For example in 
- OS: Windows
- OS version: 10 Home 20H2 build 19042.508
- Simplenote app version: 2.9.0-beta1 desktop

sandymcfadden:
### Fix

Fixes  #2809

Update:
There are still a few issues search selection when the content of the note has changed. I've changed this back to a draft to come back and address sometime in the future. 

Currently, when you search for a term, change the selected search match, then edit the content to remove a search match the selected search indicators can become incorrect. For example, it could say you are at match 4 of 3.
This corrects that.

### Test

1. Have a note with multiple instances of `word` in it
2. Search for `word`
3. Move the current match position in a note with > button or with Ctrl/Cmd+G, so that it becomes greater than one (e.g. 5 of 5)
4. Start removing the instances of `word` until one remains ~> the indicator will show 5 of 1

### Release

- Fixed issue where the search results navigation bar could show incorrect position if the content of a note was changed.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -381,8 +381,9 @@ class NoteContentEditor extends Component<Props> {
       this.state.editor === 'full' &&
       (this.props.selectedSearchMatchIndex ?? 0) > this.matchesInNote.length - 1
     ) {
-      this.setSearchSelection(this.matchesInNote.length - 1);
-      this.props.storeSearchSelection(this.matchesInNote.length - 1);
+      const newIndex = Math.max(this.matchesInNote.length - 1, 0);
+      this.setSearchSelection(newIndex);
+      this.props.storeSearchSelection(newIndex);
     }
 
     if (
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -374,6 +374,17 @@ class NoteContentEditor extends Component<Props> {
       this.setDecorators();
     }
 
+    // If content is changed so that the selected search index is greater then the number of matches
+    // then select the last match instead
+    if (
+      this.editor &&
+      this.state.editor === 'full' &&
+      (this.props.selectedSearchMatchIndex ?? 0) > this.matchesInNote.length - 1
+    ) {
+      this.setSearchSelection(this.matchesInNote.length - 1);
+      this.props.storeSearchSelection(this.matchesInNote.length - 1);
+    }
+
     if (
       this.editor &&
       this.state.editor === 'full' &&


codebykat:
Hmm. So testing this out, the search highlight seems to vanish. Is that expected?

for example, here I have a search and I'm on "8 of 13"
      
Screen Shot 2021-04-15 at 11 57 04 PM

I edit a match prior to the one I'm currently focused on, and it shows "8 of 12", but nothing is highlighted.

Screen Shot 2021-04-15 at 11 57 14 PM

sandymcfadden:
Thank you. I didn't catch that case. I'm not sure how to proceed here when the content changes and there is a search selection. 

Right now we store the index in the array of matched search results. So if the content changes that index may not be valid anymore. In that case, do we reset the selection to the first search match? 

Should we also store the range for the current search match selection? Then if the content changes we could try to select the closest match to the previously selected range?

The case I tried to address here was pretty easy. If the selected index is higher than the number of matches we have, select the last one instead, but the other cases seem more difficult.

codebykat:
Hmm. I would say we reset it to the nearest valid search selection if we can't figure out what was actually selected before the content changed. I am honestly not sure how that would feel. We could try resetting to the first selection too and just see what feels better.

I think it feels buggy to have it still say "8 of 12" or whatever without a selection being highlighted at all, so we should try to handle that case somehow.

> Should we also store the range for the current search match selection? Then if the content changes we could try to select the closest match to the previously selected range?

This is a cool idea.

belcherj:
Removed the reviewers on this PR until we are ready to take another look.

codebykat:
We need to see if we can still reproduce this, I refactored the search decorations in #3183 and I believe they are being recalculated when the model changes, but I'm not sure if the selected index is being properly invalidated.

dmsnell:
it looks like this is fairly stale, but just in case, it would be nice to see if it can be approached through middleware instead of as an action dispatched from the view component. I'm sure there are a number of edges where things jump around unexpectedly because of this.

belcherj:
Hi!

PR #2852, opened 2021-04-18
Test Checklist: Adding headers and more consistency with other platforms

pachlava:
Test checklist for Electron had no headers, which made it's reading harder.

### Change
Added headers, reordered some tests and added a few tests that were missing (e.g. new user signup test) to make the checklist consistent with checklists structure and wording on iOS, macOS and Android.

### Test
Test checklist is easier to navigate than before.

### Release
These changed do not require release notes

### DIFF
--- a/TESTING-CHECKLIST.md
+++ b/TESTING-CHECKLIST.md
@@ -1,36 +1,54 @@
 ## Testing
 
+### Login/Signup
+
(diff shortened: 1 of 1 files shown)


sandymcfadden (approved):
This certainly makes it easier to follow, thanks!

PR #2862, opened 2021-04-20
Fix/hide clear search button

### LINKED ISSUE #2860: Searchbar now contains the `X` button even when empty
### Expected
Search bar should contain the `X` button only if there's an entered text

### Observed
The search bar has the `X` button in all cases:
Screenshot 2021-04-20 at 16 18 37

### Reproduced
1. Remove the text from searchbar, if it had any ~> the `X` button is still shown

### Where did you see the bug
- OS: Windows 10 Home 20H2

sandymcfadden:
### Fix

Fixes #2860

Currently, the `X` icon in the search field, used to clear the search term, is always present. It should only show when there is a search term to clear. This fixes that.

Before:
Screen Shot 2021-04-20 at 10 54 48 AM

After:
Screen Shot 2021-04-20 at 10 54 32 AM

### Test

1. Open Simplenote.
2. Ensure `X` icon does not show in search field.
3. Add a search term.
4. Ensure `X` icon does now appear.
5. Click the `X` icon
6. Ensure the search term is removed.

### Release

NA


belcherj:
Hmm, looks like there are other commits in here. Was that intended?

belcherj:
Tested and change works. Let me know when the commit issue is worked out and I will approve.

sandymcfadden:
> Hmm, looks like there are other commits in here. Was that intended?

Thanks so much @belcherj for catching this.
I've fixed it up now.

belcherj (approved):
Great work.

PR #2867, opened 2021-04-21
Fix extra line breaks appearing in the note preview

### LINKED ISSUE #2866: Note preview contains extra in-line space / line breaks compared to 2.9.0-beta1
### Expected
When it comes to line breaks, preview should look close to what's seen in the editor.

### Observed
2.9.0-beta1 on the left, 2.10.0-beta1 on the right:

https://user-images.githubusercontent.com/73365754/115565883-724df980-a2c2-11eb-818f-9ca66160db34.mov

### Reproduced
1. Have a note with this content:

```
Bulleted List:
- one
- to
- three

Checkoxes:
- [ ] box1
- [x] box2

The text.

Linebreak is above.
New line.
```

2. Open preview ~> the in-line spaces are much bigger than in 2.9.0-beta1

### Where did you see the bug
Video shot in Chrome on Mac, but same in:
- OS: Windows 10 Home 20H2

sandymcfadden:
### Fix

Fixes #2866

While trying to preserve white space in the note preview it had a side effect of showing unintended line breaks as well.
This works to fix it by removing line breaks after closing HTML tags. 

Note: Preview still only shows one empty line even if there are multiple in between, this was consistent with previous behavior.

<details>
<summary>Before</summary>
Screen Shot 2021-04-21 at 1 06 22 PM
</details>

<details>
<summary>After</summary>
Screen Shot 2021-04-21 at 1 05 56 PM
</details>

### Test

1. Create a new note
2. Add sample content such as
```
Bulleted List:
- one
- to
- three

Checkoxes:
- [ ] box1
- [x] box2

The text.

Linebreak is above.
New line.
```
3. View preview
4. Ensure there is not unexpected space between lines

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -13,7 +13,7 @@
 ### Fixes
 
 - Fixed extra line breaks issue when exporting notes [#2819](https://github.com/automattic/simplenote-electron/pull/2819)
-- Fixed displaying extra blank spaces in history screen [#2829](https://github.com/automattic/simplenote-electron/pull/2829)
+- Fixed displaying extra blank spaces in history screen [#2829](https://github.com/automattic/simplenote-electron/pull/2829), [#2867](https://github.com/automattic/simplenote-electron/pull/2867)
 - Fixed to apply selected tag to a new note by default [#2556](https://github.com/automattic/simplenote-electron/pull/2556)
 - Fixed spacing on unsynced notes warning message [#2797](https://github.com/automattic/simplenote-electron/pull/2797)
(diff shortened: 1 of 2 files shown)


sandymcfadden (commented):
I'm not sure this is the best way to do this. But what is happening is that we have applied `white-space: pre-wrap;` to the preview which causes the line breaks after HTML tags to be shown as lines. This attempts to remove line breaks after HTML tags. So instead of output like this
```
<p>Paragraph</p>
<ul>
<li>Item</li>
<li>Item
<ul>
<li>Sub item</li>
</ul>
<li>Item</li>
</ul>
```
It would output like this
```
<p>Paragraph</p><ul><li>Item</li><li>Item<ul><li>Sub item</li></ul><li>Item</li></ul>
```

belcherj (approved):
I agree. I do not totally love this however the behaviour for the user is better. Let's merge this and  we can revisit the CSS to determine if there is a better solution.

PR #2869, opened 2021-04-22
Added a `setup` test, asking to sign up using mixed-capitalization

pachlava:
The testing checklist now contains a new `Setup` section, with one new test:

`Sign up using mixed-capitalization in the email address. Use this account to run tests below.`

### Changes
Besides the change mentioned above, there is a lot of noise, coming from the fact that I've:
1. Added `.` to the end of all sentences
2. Changed tests headers from H3 to H4, because this is what's used in Call for Testing.

### Test
1. The new test is easy to understand, and it really informs the beta-tester to use the new mixed-cap account for all further tests.

### Release
No release notes are required.

### DIFF
--- a/TESTING-CHECKLIST.md
+++ b/TESTING-CHECKLIST.md
@@ -1,99 +1,101 @@
 ## Testing
 
-### Login/Signup
+### Setup 
 
-- [ ] Logout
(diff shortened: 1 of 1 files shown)


sandymcfadden (approved):
This looks good to me.

PR #2873, opened 2021-04-26
Update method of setting theme selection

sandymcfadden:
### Fix

This works to update the way we set the theme being used in the app.
The current method has been to set a class such as `theme-light` or `theme-dark` in a number of places, the main one a couple of elements below the `#root` div of the app.

This had the added effect of not reaching modals opened as they were as part of the `body` element. This meant each modal had to have the class added as well.

This approach adds a data attribute to the `body` element such as `data-theme="dark"` or `data-theme="light"` instead. 

There should be no user-facing changes here.

### Test

1. In light and dark mode ensure that colors appear as they should.
2. Pay special attention to modals and dialogs opened. 

### Release

- Updated the method of setting the theme selection.


codebykat (approved):
Thank you!! This was a MESS.

PR #2874, opened 2021-04-27
Start process to move to using custom CSS properties for colors

sandymcfadden:
### Fix

This is an initial start to moving our CSS to use custom properties for colors throughout the app. 
This focuses on light mode and adds in the first set of CSS variables for colors. Those colors are then updated throughout the app to use the variables instead. 

### Test

- Smoke test in light and dark mode and ensure colors look proper.

### Release

- Updated some of the colors to use CSS variables.


belcherj (approved):
Tested and looks great

PR #2878, opened 2021-04-28
Make search global

### LINKED ISSUE #2626: Search should be global
### What
Search should be global. E.g. when user has a tag selected in sidebar, the search across certain keyword should be executed for 'All Notes'. Of course, search within a tag should be still possible with 'tag:TagName keyword'

### Why
Already done for iOS, Android and macOS. p2XJRt-2FK-p2

sandymcfadden:
### Fix

Fixes #2626
This updates the way search is handled if there is an opened tag to match the other platforms.
Currently, if there is a tag selected the search is restricted to that tag. 
This updates so that whenever a search is placed it is for all notes. Searches can still be restricted to tags using the syntax `tag:recipe` in the search field.

<details>
<summary>Before</summary>
Screen Shot 2021-04-28 at 2 28 03 PM
Screen Shot 2021-04-28 at 2 27 51 PM
Screen Shot 2021-04-28 at 2 27 35 PM
Screen Shot 2021-04-28 at 2 27 26 PM
Screen Shot 2021-04-28 at 2 26 57 PM

</details>
<details>
<summary>After</summary>
Screen Shot 2021-04-28 at 2 31 23 PM
Screen Shot 2021-04-28 at 2 31 04 PM
Screen Shot 2021-04-28 at 2 30 32 PM
Screen Shot 2021-04-28 at 2 30 10 PM
Screen Shot 2021-04-28 at 2 29 55 PM

</details>

### Test

1. Select a tag
2. Ensure only notes with that tag are displayed
3. Perform a search
4. Ensure any note matching shows

### Release

- Updated search so that all notes are searched even if there is a currently selected tag.

### DIFF
--- a/lib/search-field/index.tsx
+++ b/lib/search-field/index.tsx
@@ -74,9 +74,6 @@ export class SearchField extends Component<Props> {
     const { openedTag, searchQuery } = this.props;
     const hasQuery = searchQuery.length > 0;
 
-    const description =
-      'Search ' + (openedTag ? 'notes in ' + openedTag : 'notes and tags');
-
     return (
       <div className="search-field theme-color-fg theme-color-border">
         <button
@@ -89,7 +86,7 @@ export class SearchField extends Component<Props> {
         <input
           ref={this.inputField}
           type="search"
-          placeholder={description}
+          placeholder="Search notes and tags"
           onChange={this.update}
           onKeyUp={this.interceptEsc}
           value={searchQuery}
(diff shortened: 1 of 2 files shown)


codebykat (approved):
Looks good, works fine. I wonder if this is going to be confusing.

sandymcfadden:
> Looks good, works fine. I wonder if this is going to be confusing.

I think it will take some getting used to, but after playing with it for a while and seeing how it works on other platforms, I think this makes sense.

PR #2879, opened 2021-04-29
Add ability to search for tag:untagged to find all untagged notes

### LINKED ISSUE #1644: Search: Allow searching for untagged notes with tag: syntax
[Allegedly](https://github.com/Automattic/simplenote-electron/issues/355) you used to be able to search for untagged notes with `tag:untagged`, that no longer seems to work. We should have some way to search for untagged notes.

See also https://github.com/Automattic/simplenote-electron/issues/1230 (untagged notes in sidebar!)

sandymcfadden:
### Fix

Fixes #1644 

This adds the ability to use the tag search syntax to find all notes without any tags. You can now search with `tag:untagged`.

Screen Capture on 2021-04-29 at 08-58-33

### Test

1. Ensure you have some notes with no tags.
2. Search for `tag:untagged`.
3. Notice all untagged notes show in the note list.

### Release

- Added the option to search with `tag:untagged` to find notes without any tags.

### DIFF
--- a/lib/search/index.ts
+++ b/lib/search/index.ts
@@ -217,18 +217,22 @@ export const middleware: S.Middleware = (store) => {
         continue;
       }
 
+      const showUntagged =
+        collection.type === 'untagged' ||
+        searchTags.has(t('untagged' as T.TagName));
+
+      if (showUntagged && note.tags.size) {
+        continue;
+      }
+
       let hasAllTags = true;
       for (const tagName of searchTags.values()) {
-        if (!note.tags.has(tagName)) {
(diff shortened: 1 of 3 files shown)


belcherj (approved):
This is great. Should we include untagged in the list of tag suggestions?

sandymcfadden:
> This is great. Should we include untagged in the list of tag suggestions?

Good call. I've added that now!

belcherj (approved):
I like it. I think having the untagged being added in the component is better than doing it deeper in the system.

Issue #1799, opened 2019-12-30
In a dark theme, the menu bar, title bar and pop-up menus are displayed with a white background


Rakleed:
Please, be as descriptive as possible.  Issues lacking detail, or for any other reason than to report a bug, may be closed without action.

### Expected
When the dark theme is on, the menu bar, title bar and pop-up menus when clicking on File, Edit, etc. on the menu bar are displayed with a dark background.

### Observed
When the dark theme is on, the menu bar, title bar and pop-up menus when clicking on File, Edit, etc. on the menu bar are displayed with a white background.

### Reproduced
1. Click on File in the menu bar
2. Click on Preferences...
3. Click Display
4. Go to Theme
5. Click on Dark
6. Click on File on the menu bar

![image](https://user-images.githubusercontent.com/19418601/71577108-e67bf100-2b03-11ea-8087-d2c7d2ec9fd8.png)

Make|Model|OS Version|App Version
-|-|-|-
SYSTEM_MAKE|SYSTEM_MODEL|Windows 10 v. 1909|1.12.0

This is due to problem https://github.com/Automattic/simplenote-electron/issues/1036, but I'm also talking about a pop-up window, and it has been a year already, and in applications such as Spotify, Discord, GitHub Desktop, Visual Studio Code, the menu bar, title bar and pop-up windows are dark.

belcherj:
This is not a high priority but PR's are welcome.

rossanafmenezes:
Another user reporting this issue here 3338818-zen

thabotswana:
A user reported this issue with the Windows build: https://forums.simplenote.com/forums/topic/remove-background-of-icon-on-windows-taskbar/?view=all#post-2948

![screenshot-690](https://github.com/user-attachments/assets/01734f13-5e87-491a-ba48-973ef29878c1)

PR #2882, opened 2021-04-30
Add system theme for Windows and Mac Electron builds

sandymcfadden:
### Fix

Picked up from PR #2135

Currently in our Electron builds we don't offer a system theme that responds to the OSs color preference but is available in the web app.

This enables it for both Windows and Mac builds. There appears to be a bug in Electron or Chromium on Linux that the app doesn't respond properly to changing the color preferences so it still won't work there. Possibly this will get resolved in Electron 13 due to come out May 25, 2021.

It also updates so that the system menu for Windows updates to be choice of theme in Simplenote itself. So if you specifically choose Light the menu will be light, same for dark, overriding the system color preference. 

This semi addresses issues #2134 and #1799 as they will both be resolved for Mac and Windows, but still not working on the Linux builds. 

Related issues: 
- https://github.com/electron/electron/issues/28887
- https://github.com/electron/electron/issues/25925

Windows Screen Recording
https://user-images.githubusercontent.com/1326294/116714765-76f26c00-a9ac-11eb-979c-c08ad8aeebf5.mov

### Test
Needs testing on Mac, Windows, and Linux

1. Open Simplenote
2. Ensure on Windows and Mac under the view menu and theme there is a System option
3. Ensure on Linux there is no System option showing for the theme.
4. Ensure option shows under Settings > Display as well for both Windows and Mac, but not Linux
5. Choose System theme on Windows or Mac.
6. Change color preference on the operating system.
7. Ensure Simplenote updates to match.
8. Set either light or dark theme.
9. Change color preferences on the operating system.
10. Ensure Simplenote does not change and stays with your app settings.
11. On Linux ensure you can still switch between Light and Dark themes properly.

### Release

- Added system theme for Mac and Windows Electron apps.

### DIFF
--- a/desktop/menus/view-menu.js
+++ b/desktop/menus/view-menu.js
@@ -5,6 +5,32 @@ const buildViewMenu = (settings, isAuthenticated) => {
   settings = settings || {};
   isAuthenticated = isAuthenticated || false;
 
+  const themeSubMenu = platform.isLinux()
+    ? [
+        {
+          label: '&Light',
+          id: 'light',
+        },
+        {
+          label: '&Dark',
+          id: 'dark',
+        },
+      ]
+    : [
+        {
+          label: '&System',
+          id: 'system',
+        },
+        {
+          label: '&Light',
+          id: 'light',
+        },
+        {
+          label: '&Dark',
+          id: 'dark',
+        },
+      ];
+
   const menu = {
     label: '&View',
     submenu: [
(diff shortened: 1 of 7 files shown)


codebykat (approved):
Awesome! Working great on OSX. Didn't test the Windows build.

desktop/menus/view-menu.js
```diff
@@ -5,6 +5,32 @@ const buildViewMenu = (settings, isAuthenticated) => {
...
+        },
+        {
+          label: '&Light',
+          id: 'light',
+        },
+        {
+          label: '&Dark',
+          id: 'dark',
+        },
+      ];
```
stpoa: I think that below looks cleaner

```
const themeSubMenu = [
  ...(platform.isLinux() ? [] : [{ label: "&System", id: "system" }]),
  { label: "&Light", id: "light" },
  { label: "&Dark", id: "dark" },
];

```

sandymcfadden: Thanks. I can see how that is not the most readable, I've updated it but using a third option that to me is the most readable.

PR #2885, opened 2021-05-03
Continue adding light mode custom color variables

sandymcfadden:
### Fix
This continues on with the work in https://github.com/Automattic/simplenote-electron/pull/2874

It adds more color variables for light mode. Those colors are then updated throughout the app to use the variables instead.

### Test

- Smoke test in light and dark mode and ensure colors look proper.

### Release

NA


belcherj (approved):
Looks great

Issue #2889, opened 2021-05-04
Tags are restored regardless of `Restore deleted tags` state


pachlava:
### Expected
If `Restore deleted tags` is disabled, the tags should not be restored

### Observed
Tags are restored regardless:
https://user-images.githubusercontent.com/73365754/116990681-1fba0780-acdc-11eb-9f4e-bcf436c0cee6.mov

### Reproduced
1. Have a note with a tag and history available
2. Delete the tag
3. Open history for the note
4. Disable `Restore deleted tags`
5. Roll back to the state when deleted tag was still present, and click `Restore` ~> tag is restored

### Where did you see the bug
System Make: MBP 16"
OS: Win10 on Parallels Desktop
Simplenote app version: desktop 2.11.0-beta1

sandymcfadden:
Ah, this does seem to be behaving as expected, but maybe we need to make more clear what the checkbox does.
Tags that have only been removed from a note but still exist within your tag list itself will always be restored. 
The checkbox is meant for tags that have been deleted from your tag list directly. If a note has a tag that has been removed from the tag list it won't be restored unless that checkbox is selected.

I'm trying to think of how to make that more clear, but I'm not thinking of anything concise. @SylvesterWilmott do you have any ideas?

SylvesterWilmott:
I can see how this label can be misunderstood. Some possible alternatives:

- Restore trashed tags
- Recover deleted tags

@pachlava do you think this would be an improvement? If not we can include some expanded description in a tooltip on mouseover.

pachlava:
To be honest, both options still might be understood ambiguously, imo. I think a tooltip would help. Maybe it's just me.

SylvesterWilmott:
Here's an example using the popover component along with a "help" icon target (icon would need to be created for this purpose). Thoughts @pachlava ?

<img width="431" alt="Screenshot 2021-05-04 at 12 56 16" src="https://user-images.githubusercontent.com/36532468/116999843-1f1f7200-acd8-11eb-89c6-ef50686d4eda.png">

pachlava:
@SylvesterWilmott I like the copy, maybe adding a `general` word, like `...and re-added to the general tags list.` will make it even more clear.

mokagio:
Moving to keep track of this 2.12.0. It's time to finalize 2.11.0.

PR #2890, opened 2021-05-04
Fix javascript error on log out

### LINKED ISSUE #2888: JavaScript error on logout
### Expected
Logging out should not cause a JavaScript error.

### Observed
It does:
Screenshot 2021-05-04 at 12 44 18
Screenshot 2021-05-04 at 12 44 55

### Reproduced
1. Be logged in
2. Log out

### Where did you see the bug
- System Make: MBP 16"
- OS: Win10 (Parallels) | Big Sur 11.2.3
- Simplenote app version: desktop 2.11.0-beta1

sandymcfadden:
### Fix

Fixes #2888

When setting the Electron `nativeTheme` with the theme selected in the Simplenote settings it can cause a JavaScript error if the theme setting isn't present, such as when logging out of the app.

This checks to ensure the theme setting is set before updating the Electron `nativeTheme` setting.

### Test

1. Ensure you are logged in
2. Log out.
3. Ensure no error appears.

### Release
NA

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -103,7 +103,9 @@ module.exports = function main() {
       Menu.setApplicationMenu(
         Menu.buildFromTemplate(createMenuTemplate(args), mainWindow)
       );
-      nativeTheme.themeSource = settings.theme;
+      if ('theme' in settings) {
+        nativeTheme.themeSource = settings.theme;
+      }
     });
 
     ipcMain.on('clearCookies', function () {


desktop/app.js
```diff
@@ -103,7 +103,9 @@ module.exports = function main() {
       Menu.setApplicationMenu(
         Menu.buildFromTemplate(createMenuTemplate(args), mainWindow)
       );
-      nativeTheme.themeSource = settings.theme;
+      if ('theme' in settings) {
```
belcherj, line 106: Just want to make sure you know the implications of using `in` over `hasOwnProperty`. In does not care if that property is directly on the object or inherited. There are also some other differences. This seems like a pretty good [article](https://masteringjs.io/tutorials/fundamentals/hasownproperty). (I only skimmed it)

In this case the use of `in` is fine.

belcherj (approved):

desktop/app.js
sandymcfadden: Thank you!

mokagio:
This went into the release branch after beta1 shipped. Should we ship a new beta?

From the milestone list I can see there's nothing else open, so I'd say yes, but thought I'd check first. I'm asking because shipping a new beta involves a bit of manual labor still, so just trying to be efficient 😄 

cc @pachlava

sandymcfadden:
Thanks @mokagio there is one more I'd like to get in. I hadn't updated the milestone yet but I have now https://github.com/Automattic/simplenote-electron/issues/2889
I'd say once that fix goes in I can create the new beta.

mokagio:
👍 thanks @sandymcfadden

PR #2893, opened 2021-05-04
Fix spacing on tag list edit buttons

### LINKED ISSUE #2891: Position of `trash tag` button feels off
### What
When tags are sorted alphabetically, the position of `trash tag` button feels somewhat unpleasant:

Screenshot 2021-05-04 at 13 59 54

It feels like positioning it a bit to the left would be better

### Why
Better UX

sandymcfadden:
### Fix

Fixes #2891

Currently, if you set to sort tags alphabetically and then choose to edit your tag list, the trash icon is hard up to the right side of the list.
If you have it set to manually sort tags then the reorder handle has spacing on the right.

This makes it more consistent and adds the spacing in either situation.

Before:
Screen Shot 2021-05-04 at 8 37 09 AM

After:
Screen Shot 2021-05-04 at 8 36 46 AM

### Test

1. Set tags to be sorted alphabetically
2. Open tag list.
3. Choose to edit tags.
4. Notice trash icon is spaced nicely from the right.
5. Set tags to be no longer sorted alphabetically.
6. Choose to edit tags.
7. Ensure spacing still looks correct.

### Release

- Fixed spacing issue on the trash tag button when set to sort tags alphabetically.

### DIFF
--- a/lib/tag-list/style.scss
+++ b/lib/tag-list/style.scss
@@ -12,6 +12,10 @@
   &.is-selected {
     background-color: $studio-simplenote-blue-5;
   }
+
+  .icon-button:last-child {
+    margin-right: 10px;
+  }
 }
 
 .tag-list-item:not(:last-child) {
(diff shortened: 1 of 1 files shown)


belcherj (approved):
Great!

PR #2896, opened 2021-05-04
Add default title bar space for Mac Electron

sandymcfadden:
### Fix

Previously we had been adding specific styles to different components for only Mac Electron app to prevent things from overlapping the window buttons.

This removes those and instead enables the default title bar.

<details>
<summary>Screenshots</summary>
Screen Shot 2021-05-04 at 1 55 54 PM
Screen Shot 2021-05-04 at 1 55 40 PM

</details>

### Test

1. Run Mac Electron app.
2. Ensure bar appears at the top with the app below it as usual.

### Release

- Added default title bar to Mac Electron app

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -66,7 +66,6 @@ module.exports = function main() {
       height: mainWindowState.height,
       minWidth: 370,
       minHeight: 520,
-      titleBarStyle: 'hidden',
       show: false,
       webPreferences: {
         contextIsolation: true,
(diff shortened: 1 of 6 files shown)


belcherj (changes requested):
Looks good except:

- note-toolbar/style.scss Line 22

It looks a little different. Can you get a review from Sly? We don't ship macOS so It is more of making him aware.

sandymcfadden:
Thanks, @belcherj I was so confused because while searching in my editor I couldn't find any more instances of `is-macos`. There wasn't any, but I hadn't saved that file after changing 🤦

SylvesterWilmott:
Looks good @sandymcfadden

belcherj (approved):
Ship

PR #2899, opened 2021-05-05
Add tooltip for restore deleted tags clarification

### LINKED ISSUE #2889: Tags are restored regardless of `Restore deleted tags` state

sandymcfadden:
### Fix

Fixes #2889

With the new restore deleted tags option in the note history revision selector, there was some confusion around what the action would do. This adds a tooltip with more information explaining what will take place. 

Screen Shot 2021-05-05 at 9 08 21 AM
Screen Shot 2021-05-05 at 9 07 50 AM

### Test

1. Open a note.
2. Choose to view its history.
3. Hover over the new help icon next to Restore deleted tags.
4. Ensure things look proper.
5. Ensure the action is clear.

### DIFF
--- a/lib/icons/help-small.tsx
+++ b/lib/icons/help-small.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+export default function SmallHelpIcon() {
+  return (
+    <svg
+      className="icon-help-small"
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 16 16"
+    >
+      <rect x="0" fill="none" width="16" height="16" />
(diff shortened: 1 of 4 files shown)


pachlava:
My 2 cents: maybe it's worth making the tooltip opaque, because the revision slider 'line' (and possibly the date value, when the slider is positioned behind the tooltip) make it harder to read the description:

Screenshot 2021-05-05 at 15 24 14

sandymcfadden:
Good call @pachlava! I've fixed that up now in my last update.

sandymcfadden:
Going to wait for updated copy from @SylvesterWilmott for this.

SylvesterWilmott:
For the copy, I think we need something a little more explicit to indicate that it wont restore **all** tags but only the ones associated with the restored version of the note:

"Any deleted tags associated with the restored version of this note will be re-added to your list of tags."

sandymcfadden:
That does make it more clear @SylvesterWilmott. I've updated the copy to that now

SylvesterWilmott:
Thanks @sandymcfadden

mokagio:
Pushing this to 2.12.0 because it's time to finalize 2.11.0 for submission.

codebykat (approved):
Tested in light and dark mode in Firefox and Electron, looks great 👍

PR #2900, opened 2021-05-05
Update webpack and related dependancies

sandymcfadden:
### Fix

We have been using an old version of Webpack which should be updated to take advantage of the latest changes.
This involves some configuration changes and the need to install some extra dependencies. 

One of the big things with Webpack 5 is it no longer installs polyfills for core node modules and so they have to be installed directly. 

In order to get things to build these dependencies needed to be added
- postcss
- path
- process
- setimmediate
- stream
- util

### Test

This is a pretty big change and will need some good testing.
We should run through the testing checklist for each of the builds. 

### Release

- Updated Webpack and related dependencies and updated the config as needed.


hamoody0706:
`[chore/update-webpack-and-related-dependancies](https://story.snapchat.com/o/W7_EDlXWTBiXAEEniNoMPwAAYYXdma21qamliAXwbGuSnAXwbGtuxAAAAAA
Duplicate of #)`

codebykat:
Incorporated most of these changes into #3183, the only difference really is I'm using https://www.npmjs.com/package/node-polyfill-webpack-plugin to provide some of the core polyfills from webpack4. Thanks for figuring this stuff out @sandymcfadden, this was a huge help two years later 😂

PR #2901, opened 2021-05-06
Fixes a bug where you had to use ⌘Q twice to quit the mac app.

### LINKED ISSUE #2676: Double ⌘Q needed to quit the application
I saw the same report here: #2445, which is said to be fixed with #2439, but it looks like it isn't. Correct me if I'm wrong.

### Expected
First ⌘Q should close the Simplenote process.

### Observed
On Mac, the first ⌘Q closes the window (similar to ⌘W), and only the second ⌘Q closes the process.

### Reproduced
1. Have the app window shown and be logged in
2. Hit ⌘Q ~> the window is closed, but the process is not closed.

### Where did you see the bug
- System Make: Mac
- System Model: MBP 16"
- OS: macOS
- OS version: 11.2 

sandymcfadden:
### Fix

Fixes #2676

In Mac OS there the expected behavior is that when you close the window of an app that it stays running and you can open it again from the dock. Then if you quit the app using ⌘Q it completely closes.

Currently in order to fully quit you need to press ⌘Q twice which is unexpected. This resolves that by allowing you to close the window and reopen it, but also quit the app as expected.

### Test

1. Open Mac app.
2. Close the window.
3. App should still be in the dock.
4. App should open again when you click it.
5. Quit the app with ⌘Q or using the context menu when clicking on the doc icon.
6. App should fully quit.

### Release

- Fixed a bug where the Mac app wouldn't quit on the first request.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -28,6 +28,14 @@ module.exports = function main() {
   // be closed automatically when the JavaScript object is GCed.
   let mainWindow = null;
   let isAuthenticated;
+  let shouldQuit = false;
+
+  // Checks to see if the application was asked to quit instead of just close the window
+  // we then use this variable to check if we should quit the app.
+  // Important for MacOS so it will quit when pressing CMD+Q
+  app.on('before-quit', () => {
+    shouldQuit = true;
+  });
 
   app.on('will-finish-launching', function () {
     setTimeout(updater.ping.bind(updater), config.updater.delay);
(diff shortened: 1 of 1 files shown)


codebykat (approved):
Works great! Tested the build on OSX, I did not test on other platforms but I assume the platform check prevents it from affecting anything else. One note - it might be nice to add a comment for our future selves. The `on('before-quit')` => `shouldQuit = true;` bit is something I am 100% not going to understand why it's there in 6 months 😂

sandymcfadden:
Good call @codebykat thank you. I've added a comment now.

PR #2907, opened 2021-05-15
Add 2.12.0 release notes

sandymcfadden:
~Draft: Parent branch should be release/2.12.0~ Done

https://github.com/Automattic/simplenote-electron/milestone/54

https://github.com/Automattic/simplenote-electron/compare/v2.11.0...release/2.12.0

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [v2.12.0]
+
+### Enhancements
+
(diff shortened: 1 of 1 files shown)


mokagio (approved):
No one from the @Automattic/simplenote-web-reviewers approved this yet. 

I read them twice and all seems good but I might miss some context (I didn't drill into each PR).

If something's not right, we can easily update the notes on the GitHub release that will be generated. I'm not sure if it's possible for the Microsoft Partner Center, though. Anyways... these should stay up for only two weeks so no drams 😄 

Thanks @sandymcfadden !

Issue #2839, opened 2021-04-10
Make the New Version notification dialog non-modal or less obtrusive.


Kizaemon:
### What
(Windows version 2.8.0)
On each launch the app is checking for the new version and shows modal dialog box, announcing a new version (e.g. New Version 2.9.0 is available, your version is 2.8.0).

Please make this dialog box non-modal.
_- - or - -_
Please make the notification of the new version during application exit, not during application launch.

### Why
The modal dialog box disrupts user's workflow, forcing to take an immediate action (e.g cancel or upgrade).
There are some situations where the user (i.e. me) would like to defer the update to a later point of time.

Please consider my usual workflow scenario:
1) I'm opening the Simplenote app to get some reference information.
2) The popup box prevents me from using the app.
3) I cancel the new version box, because my immediate priority (the working context) is to get some reference data from a note.
4) Now that the box is cancelled, I can get the data from the note and continue with the workflow.
5) Five minutes later I close simplenote app, because now I'm cleaning up my working context.
6) I FORGET about the simplenote notification about the new version.
I should have checked and applied the new version, but there is no reminder anymore.
Because of the time gap and context switches, I'm not aware of the update anymore.
7) Goto 1

anandnalya:
A somewhat related issue was raised in `4297834-zd`.

PR #2913, opened 2021-05-19
10 seconds delay for the non-user-initiated checks

Kizaemon:
### Fix
Upon the program launch the user is instantly presented with a modal box, informing of a new app version.
For a smoother UX, delay the new version check by 10 seconds.

A graceful delay has a positive effect of reducing the resource contention for the application and for the user.

### Test
1. Make sure a new version of the program is available for distribution. (e.g. Version 2.12.0)
2. Close and then re-launch the program.
3. The "new version available" modal box should appear with a 10 seconds delay.

### Release
Checking for a new program version is delayed by 10 seconds on program launch.

### DIFF
--- a/desktop/updater/auto-updater/index.js
+++ b/desktop/updater/auto-updater/index.js
@@ -27,7 +27,8 @@ class AutoUpdater extends Updater {
   // Check and download in the background, and only notify the user if
   // an update exists and has completed downloading.
   ping() {
-    autoUpdater.checkForUpdates();
+    // introduce a slight delay for the non-user-initiated checks
+    setTimeout(() => autoUpdater.checkForUpdates(), 10 * 1000);
   }
 
   // For user-initiated checks.


Kizaemon:
Never mind, please.
Thank you for your attention.

sandymcfadden:
Hey @Kizaemon sorry for not responding here previously. I did take a quick look at this when you first submitted and your open issue as well https://github.com/Automattic/simplenote-electron/issues/2839

I agree that the upgrade notification could be better, but I'm not sure a delay is the best way. To me, this could interrupt the flow of writing more as you could get the app open, start writing, and then have the dialog come up and take focus causing a break in the typing.

One of the things we have on the roadmap is to add better notifications in the app itself. I think once we get that in we can look at improving how we notify about available upgrades as well. 

Thanks for the submission and opening the issue. We'll keep the issue open so it can be improved in the future.

PR #2936, opened 2021-06-22
Update Dark mode colors to use to css variables

sandymcfadden:
### Fix

This is a continuation of previous work to update the CSS to use variables for colors.
This moves all colors for dark mode to use CSS variables instead. 

There should be no user impact or any noticeable changes. It's an intermediate step to simplifying the CSS used throughout the app. 

### Test

1. Smoke test in dark mode
2. Ensure that everything looks good

### Release

- Update colors in dark mode to use CSS variables

### DIFF
--- a/lib/alternate-login-prompt/style.scss
+++ b/lib/alternate-login-prompt/style.scss
@@ -73,6 +73,6 @@
 
 body[data-theme='dark'] {
   .alternate-login__button-row .button-borderless {
-    color: $studio-simplenote-blue-20;
+    color: var(--accent-color);
   }
 }
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -67,7 +67,7 @@
 
   input:read-only,
   input:disabled {
-    border-color: var(--readonly-input-bg-color);
+    border-color: var(--readonly-input-highlight-color);
     background-color: var(--readonly-input-color);
   }
 
@@ -214,7 +214,7 @@ body[data-theme='dark'] .login {
 
     &:read-only,
     &:disabled {
-      border-color: var(--readonly-input-bg-color);
+      border-color: var(--readonly-input-highlight-color);
       background-color: var(--readonly-input-color);
     }
 
--- a/lib/auth/style.scss
+++ b/lib/auth/style.scss
@@ -202,7 +202,7 @@ body[data-theme='dark'] .login {
   }
 
   .or {
-    background: $studio-gray-90;
+    background: var(--background-color);
     color: $studio-gray-20;
   }
   .or-line {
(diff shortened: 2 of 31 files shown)


codebykat (changes requested):
Phew that's a lot of files changed 😂

It looks pretty good to me. I did note one difference in comparing this build to what's currently on app.simplenote.com, a slight change in the text colors in some of the Settings panels. It's visible on Collaborate (the text color got brighter while the example text in the form field got dimmer) and also on the analytics informational text in Settings.

app.simplenote.com:

Screen Shot 2021-06-24 at 5 21 54 PM

Screen Shot 2021-06-24 at 5 24 10 PM

this branch in browser:

Screen Shot 2021-06-24 at 5 21 58 PM

Screen Shot 2021-06-24 at 5 24 12 PM


The "Move to trash" link in the dropdown has also changed.

app.simplenote.com:

Screen Shot 2021-06-24 at 5 26 01 PM

this branch in browser:


Screen Shot 2021-06-24 at 5 26 05 PM

sandymcfadden:
Thanks so much @codebykat This is a lot of changes the next PRs that follow this will be much more manageable. 

- For the placeholder text on the collaborate screen I do see it the difference, I think we should leave it as in this PR though, it matches the login form inputs. What do you think @SylvesterWilmott?
- I've fixed up the text on the settings panels
- I see what you mean about the Move to trash button as well. I used the same color here that was used on the delete forever button when in the Trash. It kind of makes sense to me that those destructive buttons should match. What do you think of this one too @SylvesterWilmott?

SylvesterWilmott:
Thanks for spotting those @codebykat 

> For the placeholder text on the collaborate screen I do see it the difference, I think we should leave it as in this PR though, it matches the login form inputs.

In this case, the placeholder text in dark mode should be gray20. As time goes on I would prefer to use transparency over colours since dark mode will contain a lot of different surface colours and we will need to make sure that all input states have sufficient contrast. For now, can we update the instances where the placeholder text is not gray20.

> It kind of makes sense to me that those destructive buttons should match.

For sufficient contrast, coloured text should be the [color]20 version. So in this case, any red text on a dark background should be red20 so we should update the "Delete forever" label.

It may be helpful if I reiterate the color rules on text here:

**Light mode:**

Primary text: Gray80
Secondary text: gray50
Coloured text: [color]50
Placeholder text: gray50

**Dark mode:**

Primary text: White
Secondary text: gray20
Coloured text: [color]20
Placeholder text: gray20

sandymcfadden:
Thank you @SylvesterWilmott that's very helpful. I believe I have everything updated appropriately now.

SylvesterWilmott (approved):
:shipit:  if we spot any more inconsistencies we can fix as we go.

PR #2943, opened 2021-06-30
Update top components in app to use css variables for colors

sandymcfadden:
### Fix

This updates the components which physically make up the top of the app which are:
- Note toolbar
- Menu bar
- Note editor
- Search field

### Test

1. Smoke test in light and dark mode
2. Ensure those components (basically the very top of the app) still appear the same

### Release

- Updated the Note toolbar, Menu bar, Note editor, and Search field components to use CSS variables for colors

### DIFF
--- a/lib/menu-bar/index.tsx
+++ b/lib/menu-bar/index.tsx
@@ -64,7 +64,7 @@ export const MenuBar: FunctionComponent<Props> = ({
   const CmdOrCtrl = isMac ? 'Cmd' : 'Ctrl';
 
   return (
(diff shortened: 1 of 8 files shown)


codebykat (approved):
This looks great. The only difference I noted is the placeholder text in the search field in dark mode is brightened up a bit (I think we talked about that on a different PR as well and decided it is correct behavior).

sandymcfadden:
Thanks @codebykat you're right. That is the correct color now.

PR #2946, opened 2021-07-02
Fix the font stack outside of the editor

sandymcfadden:
### Fix

Fix the font stack to add in BlinkMacSystemFont. This was included in the font stack for the editor, but not the rest of the app.

This adds it in.

### Test
1. Ensure the font used in the editor matches font used in the rest of the app.

### Release

- Fixed font stack used so the font is consistent everywhere


SylvesterWilmott:
Thanks for the fix! LGTM

PR #2955, opened 2021-07-09
Update tag field / input to use css variables for colors

sandymcfadden:
### Fix

This updates the tag field and related components to use the variables instead of classes. The components updated are:

- Note Detail
- Tag Field
- Tag Input
- Tag Suggestion

### Test

1. Smoke test in light and dark mode
2. Ensure these components still appear the same.

### Release

- Updated the Note Detail, Tag Field, Tag Input, and Tag Suggestion components to use CSS variables for colors

### DIFF
--- a/lib/note-detail/style.scss
+++ b/lib/note-detail/style.scss
@@ -5,6 +5,7 @@
   flex-direction: column;
   flex: 1 1 auto;
   padding-top: 0;
+  background-color: var(--background-color);
 }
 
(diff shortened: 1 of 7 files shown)


codebykat (approved):
Looks great! Same note as before with the input placeholder text in dark mode (it's a bit brighter now and I think we decided that's correct). Didn't see any other differences in light or dark mode.

PR #2937, opened 2021-06-24
Update display of logo for Windows start menu tile

sandymcfadden:
### Fix

Fixes #2935

Currently, if you install the appx or Windows store version of Simplenote and then pin it to the start menu the tile generated has a large logo and doesn't show the app name at all. This is different from most other apps installed.

This shrinks the size of the logo and displays the name on the tile. 

Before:
Screen Shot 2021-06-24 at 10 32 48 AM

After:
Screen Shot 2021-06-24 at 10 31 54 AM

### Test

1. Install the appx version of SImplenote
2. Pin the app to the start menu
3. Ensure things look proper with a smaller icon and the app title

### Release

- Updated the logo used when pinning to the Windows start menu.

PR #2965, opened 2021-07-23
Update Windows Tile Icons

SylvesterWilmott:
### Fix
Update Windows tile icons by adjusting the size and shadow so that our icon is consistent with apps by Microsoft. Mentioned in #2937 See below for illustration.

Screenshot 2021-07-23 at 13 51 18

### To Test
1. Verify that the icons more closely match the icons of apps by Microsoft.


sandymcfadden (approved):
Tested on Windows and the tile icons look nicer now with the shadow.

PR #2969, opened 2021-07-30
Update note list components to use CSS variables for colors

sandymcfadden:
### Fix

This updates the note list and related components

- Note List
- Note Cell
- No Notes

It also fixes a small issue with the tag suggestions, but more will be required for that component in future PRs.
It also fixes an issue with the search highlighting in the note list where a background color was being set on matches when it shouldn't be.

### Test

- Switch to All notes and ensure components look the same
- Switch to a tag and ensure components look the same
- Switch to the trash and ensure components look the same
- Perform a search and ensure components look the same

### Release

- Updated the Note List, Note Cell, and No Notes, components to use CSS variables for colors

### DIFF
--- a/lib/note-list/index.tsx
+++ b/lib/note-list/index.tsx
@@ -74,7 +74,7 @@ const renderNote = (
         rowIndex={index}
       >
         <div className="note-list is-empty" style={{ ...style, height: 200 }}>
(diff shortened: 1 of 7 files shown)


codebykat (approved):
Looks great! I tested in the browser, light and dark modes, and didn't see any style differences between this branch and production.

PR #2972, opened 2021-08-07
Add support to handle compromised password at login

sandymcfadden:
### Fix

This will add support to show an error message and prompt to reset the password if the given password has been found in known data breaches and is therefore not secure. It required updating the `node-simperium` dependency to do that as well.

### Testing

This is rather hard to test currently to see the new error message as changes to the API endpoint haven't been made to show the error yet.

We can ensure that login still works as expected that you can log in with a valid email and password, and you can't log in with an invalid one.

In order to test this, I had to set up Charles Proxy and use it to intercept the API call and change the response text to be `compromised password` 

### Release

- Add support for handling passwords found in known data breaches.

### DIFF
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -13,6 +13,7 @@ type OwnProps = {
   accountCreationRequested: boolean;
   authPending: boolean;
   emailSentTo: string;
+  hasCompromisedPassword: boolean;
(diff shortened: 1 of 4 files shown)


nsakaimbo (approved):
Changes look good, and this makes sense to me. 👍

PR #2976, opened 2021-08-15
Add support to handle unverified accounts at login

sandymcfadden:
### Fix

This will add support to show an error message with a link to start verification of account email when logging in with an unverified account.

### Todo

- [x] Get clarification on error message copy

### Test

This is rather hard to test currently to see the new error message as changes to the API endpoint haven't been made to show the error yet.

We can ensure that login still works as expected that you can log in with a valid email and password, and you can't log in with an invalid one.

In order to test this, I had to set up Charles Proxy and use it to intercept the API call and change the response text to be `verification required`

### Release

- Added support for handling failed logins due to having an unverified account email

### DIFF
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -17,6 +17,7 @@ type OwnProps = {
   hasInsecurePassword: boolean;
   hasInvalidCredentials: boolean;
   hasLoginError: boolean;
+  hasUnverifiedAccount: boolean;
   login: (username: string, password: string) => any;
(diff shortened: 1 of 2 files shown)


nsakaimbo (approved):
Looks good! Small nit: please add a comment around the order-of-priority for the password check vs. the unverified accounts, i.e. the user updating their password will lead to automatic account verification (since the password update is routed through their email).

Otherwise changes are pretty similar to #2972 and seem pretty straightforward.

PR #2977, opened 2021-08-16
Save note scroll position and restore when returning to note

### LINKED ISSUE #2861: `Account` dialog scroll position is reset when `Share Analytics` is toggled
### Expected
Changes in switches states should not affect the scroll position of the dialog.

### Observed
When `Share Analytics` switch is clicked, the scroll position is set to 0. Also, I've noticed that switch animation is instant for `Share Analytics`, but slow for the other switches:

https://user-images.githubusercontent.com/73365754/115406414-d0fa7100-a1f7-11eb-9ba3-188873f4f303.mov

### Reproduced
1. Open `Settings` ~> `Account`
2. Scroll down
3. Click `Share Analytics` switch

### Where did you see the bug
- OS: Windows 10 Home 20H2

### LINKED ISSUE #2951: Maintain Scroll Position When Switching from Markdown Preview and Edit Mode
### Expected
Similar to the Android and macOS apps, when switching between the markdown preview and edit modes, your scroll position should be maintained.

### Observed
Instead, when switching between markdown preview and edit modes, you will always be scrolled to the top of your note.

### Reproduced
1. Open a long note with markdown enabled.
2. Scroll down.
3. Click on the Preview mode — note that you are now scrolled to the top of the note.
4. Scroll down again within the preview.
5. Click on the preview button again to switch back to edit mode — note that you are now scrolled to the top of the note.

### Where did you see the bug
- System Make: Apple
- System Model: MacBook Pro
- OS: macOS Catalina
- OS version: 10.15.7
- Browser (if applicable): Firefox
- Browser version (if applicable): 89.0.2
- Simplenote app version: Web app

sandymcfadden:
### Fix

This will help fix #2951 and help fix #2861

Currently, if you have a longer note that needs to be scrolled when you leave the note and come back the scroll position is set back to the top. This includes if you are returning from a Markdown preview, or even if you change a setting that causes the editor to re-render.

This saves the line and scroll position of a note when you leave and then restores it when you come back. This is only saved for the current session. 

### Test

1. Open a note which needs to be scrolled
2. Scroll to a point in the note
3. Leave the note
4. Open the note again and ensure you are taken to the same spot in the note

### Release

- Updated to save the scroll position of a note so you can be restored when the note is viewed again

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -220,12 +220,10 @@ class NoteContentEditor extends Component<Props> {
   }
 
   componentWillUnmount() {
-    const { noteId } = this.props;
-    const position = {
+    setNotePosition(this.props.noteId, {
       line: this.editor?.getPosition()?.lineNumber ?? 0,
       scroll: this.editor?.getScrollTop() ?? 0,
-    };
-    setNotePosition(noteId, position);
+    });
 
     if (this.bootTimer) {
       clearTimeout(this.bootTimer);
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -205,9 +205,63 @@ class NoteContentEditor extends Component<Props> {
       } while ((node = node.next));
     };
     removeById(contextMenuLinks, removableIds);
+
+    const notePositions = localStorage.getItem('note_positions');
+    let currentSavedPositions: any;
+    if (notePositions) {
+      try {
+        currentSavedPositions = JSON.parse(notePositions);
+      } catch (e) {
+        // pass - we only care if we can successfully do this,
+        //        not if we fail to do it
+      }
+    } else {
+      currentSavedPositions = {};
+    }
+
+    this.editor?.revealLineInCenter(currentSavedPositions[noteId].line);
+    window.setTimeout(() => {
+      this.editor?.setScrollPosition({
+        scrollTop: currentSavedPositions[noteId].scroll,
+      });
+    }, 200);
+
+    // const currentState = currentSavedPositions[noteId].state ?? {};
+    // if (currentState) {
+    //   this.editor?.restoreViewState(currentState);
+    //   window.setTimeout(() => {
+    //     this.editor?.setScrollPosition({
+    //       scrollTop: currentSavedPositions[noteId].scroll,
+    //     });
+    //   }, 200);
+    // }
   }
 
   componentWillUnmount() {
+    const notePositions = localStorage.getItem('note_positions');
+    let currentSavedPositions;
+    if (notePositions) {
+      try {
+        currentSavedPositions = JSON.parse(notePositions);
+      } catch (e) {
+        // pass - we only care if we can successfully do this,
+        //        not if we fail to do it
+      }
+    } else {
+      currentSavedPositions = {};
+    }
+
+    const { noteId } = this.props;
+    currentSavedPositions[noteId] = {
+      //state: this.editor?.saveViewState(),
+      line: this.editor?.getPosition()?.lineNumber,
+      scroll: this.editor?.getScrollTop(),
+    };
+    localStorage.setItem(
+      'note_positions',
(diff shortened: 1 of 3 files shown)


lib/note-content-editor.tsx
```diff
@@ -205,9 +206,25 @@ class NoteContentEditor extends Component<Props> {
       } while ((node = node.next));
     };
     removeById(contextMenuLinks, removableIds);
+
+    const position = getNotePosition(noteId);
+
+    if (position) {
+      this.editor?.revealLineInCenter(position.line);
+      window.setTimeout(() => {
+        this.editor?.setScrollPosition({
```
codebykat: * I'm curious why we're calling both `revealLineInCenter` and `setScrollPosition`? `revealLineInCenter` [will scroll the note to the position requested](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandalonecodeeditor.html#reveallineincenter). Similarly, do we really need to save both the position and the `scrollTop` value?
* Did you test this without the timeout? It seems odd that we need it for `setScrollPosition` but not for `revealLineInCenter`.
* See also line 1106 where we use a promise that immediately resolves to update the cursor position; perhaps that would be faster than the 200ms timeout?

codebykat (commented):
Code looks good and mostly tests well. I noticed a few oddities that may or may not be fixable...

* When you switch back to a note, it loads at the top and then jumps to the scroll position. I'm wondering if there's any way to get that to happen when we pass new content to the editor? Does it need the timeout? Can we move it higher up in the `componentDidMount` function (maybe even inside the `bootTimer` where we set the editor content)?

* Switching into Markdown preview obviously doesn't know to scroll to the same part of the note, which is unfortunate. I am not sure there's any easy way to do that but we might want to consider it in future.

* Left an in-line comment about `revealLineInCenter` which I think has the potential to cause additional jumpiness.

lib/note-content-editor.tsx
sandymcfadden: You're right, there really isn't a need to keep both and call both. In this case I think keeping and setting the scroll position is the best way. The reason for this is the selected line may not match where you are scrolled to. Also if the line is long and wraps it will really only bring you to the top of that line and not where you were scrolled to.

The downside with scroll position and something I haven't accounted for yet is if the window size changes from the time when the position is saved and when you open the note again. There is a good chance it doesn't take you to the same spot in the note in this case.

sandymcfadden: I've updated to use only scroll position now and clear the saved positions of things are resized as I think it is likely better to go to the top of a note rather than a random place if the size changes.

sandymcfadden:
Thanks @codebykat!

> When you switch back to a note, it loads at the top and then jumps to the scroll position. I'm wondering if there's any way to get that to happen when we pass new content to the editor? Does it need the timeout? Can we move it higher up in the `componentDidMount` function (maybe even inside the `bootTimer` where we set the editor content)?

I've moved it inside the `bootTimer` as you suggested and it seems there we don't need the timeout at all. You still get a bit of a jump, but I think the content has to load into the editor before you scroll so I'm not sure there is a way to eliminate it completely.

> Switching into Markdown preview obviously doesn't know to scroll to the same part of the note, which is unfortunate. I am not sure there's any easy way to do that but we might want to consider it in future. 

Yes, this would require more work and I think a different PR. The same scroll position wouldn't work in preview especially if you have things like markdown images or something that are going to take up a lot of space. 

Saving the line I don't think it will really work because of what I mentioned in another comment about long lines that wrap. So I think to do that well would require a fair bit more work.

codebykat (approved):
Looks good, I think this is working better without the timeout (or it's a placebo effect).

Nice thinking on the resize. I suppose if we wanted to try to make it more bulletproof, we could store both the scroll position and the line, and use revealLineInCenter instead of the scroll position after a resize... but it's probably an edge case and I'm not sure it would work that well anyhow.

Issue #3000, opened 2021-09-10
Chore: Update standard checkbox component


sandymcfadden:
The checkbox component we use currently allows us to display checkboxes in two different formats. One a circle and one a more standard-looking checkbox.

The standard-looking checkbox uses background images to give us more control over the look of the checkbox. This makes it difficult to update the colors of checkboxes when needed.

The component should be updated to allow us to be able to change the color of the checkboxes with CSS color variables instead.

PR #3001, opened 2021-09-10
Updated note action component to use css variables for colors

sandymcfadden:
### Fix

This updates the Note Action component.

~Note: The checkboxes used in this component and elsewhere will need to be updated separately still to allow changing colors more easily. Tracked in https://github.com/Automattic/simplenote-electron/issues/3000~

### Test

- Click the note action icon in the upper right
- Ensure the open panel looks the same
- Toggle publish ensure spinner and copy still appear the same

Release

- Updated the Note Action components to use CSS variables for colors

### DIFF
--- a/lib/note-actions/index.tsx
+++ b/lib/note-actions/index.tsx
@@ -78,11 +78,8 @@ export class NoteActions extends Component<Props> {
           onDeactivate: this.handleFocusTrapDeactivate,
         }}
       >
-        <div
-          className="note-actions theme-color-bg theme-color-fg theme-color-border"
-          ref={this.containerRef}
-        >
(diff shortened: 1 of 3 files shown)


codebykat (changes requested):
This seems to have left the component with a transparent background.

Screen Shot 2021-09-16 at 8 18 45 PM

Screen Shot 2021-09-16 at 8 19 00 PM

sandymcfadden:
That's kind of embarrassing, thank you so much for catching that @codebykat.
I guess I only tested in test notes that didn't have content over that far.
It is updated now.

codebykat (approved):
I don't know why the build is failing but it looks right now! Thanks ❤️

PR #3002, opened 2021-09-13
Update the standard checkbox component to use individual SVG images

### LINKED ISSUE #3000: Chore: Update standard checkbox component

sandymcfadden:
### Fix

Fixes #3000 

The current implementation of the standard checkboxes uses an SVG sprite as background to display the correct color and state of a checkbox.

This makes it difficult to modify the color of the checkbox and could get hard to manage if adding more potential colors.

This updates the component to add the appropriate SVG image to the HTML. The color can then be set with CSS. 

<table>
<tr>
<td>
Screen Shot 2021-09-10 at 8 45 30 PM
</td>
<td>
Screen Shot 2021-09-10 at 8 45 44 PM
</td>
</tr>
</table>

### Test

- Ensure checkboxes in the Note Actions panel work and look as expected
- Ensure the checkbox in the Note History panel works and looks as expected

### Release

- Updated the standard checkbox to be easier to manage

### DIFF
--- a/lib/controls/checkbox/index.tsx
+++ b/lib/controls/checkbox/index.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
 import classNames from 'classnames';
 
+import CheckedCheckbox from '../../icons/checkbox-checked';
+import UncheckedCheckbox from '../../icons/checkbox-unchecked';
+
 type OwnProps = React.HTMLProps<HTMLInputElement> & {
   className?: string;
   onChange: () => any;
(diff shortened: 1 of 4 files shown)


lib/controls/checkbox/index.tsx
```diff
@@ -7,18 +7,43 @@ type OwnProps = React.HTMLProps<HTMLInputElement> & {
...
-      <span className="checkbox-control-base">
-        <span className="checkbox-control-checked" />
-      </span>
+      <input type="checkbox" {...props} checked={checked} />
+      {!isStandard ? (
+        <span className="checkbox-control-base">
+          <span className="checkbox-control-checked" />
+        </span>
+      ) : checked ? (
+        <svg
```
belcherj: Should we create components for them as we do the icons: https://github.com/Automattic/simplenote-electron/blob/develop/lib/icons/connection.tsx

The folder name doesn't make a ton of sense. But the contents are all svg's.

lib/controls/checkbox/index.tsx
```diff
@@ -7,18 +7,43 @@ type OwnProps = React.HTMLProps<HTMLInputElement> & {
...
         className,
         { 'checkbox-standard': isStandard },
       ])}
     >
-      <input type="checkbox" {...props} />
-      <span className="checkbox-control-base">
-        <span className="checkbox-control-checked" />
-      </span>
+      <input type="checkbox" {...props} checked={checked} />
+      {!isStandard ? (
```
belcherj, line 27: What is `isStandard`? What cases does it appear this way?

belcherj (changes requested):

lib/controls/checkbox/index.tsx
sandymcfadden: The checkbox component is used in two different forms right now. The default so with `isStandard` set to false, or not supplied as a prop, would be the toggles you see in places like Settings such as Share Analytics toggle.
Screen Shot 2021-09-13 at 9 36 42 AM

The `isStandard` has it show up instead as a more standard-looking checkbox such as in the Note Action panel like the screenshots included in the main description, and in the Revision Slider.

lib/controls/checkbox/index.tsx
sandymcfadden: That's a good call, it cleans up the checkbox component quite a bit by creating separate components for the svgs. I've done that now.

codebykat (approved):
Looks good 👍

Daviontaylor355 (commented):
Daviontaylor

PR #3006, opened 2021-09-17
Update CircleCI xcode version

sandymcfadden:
### Fix

The Xcode image version we were using in CircleCI has been deprecated and removed. This caused builds to fail as the Mac build could not be completed.
This updates the Xcode image to the [latest stable version](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions).

### Test

- Download the built Mac artifact and ensure it installs and works as expected

### Release

- [Internal] Updated the Xcode version for CircleCI to a newer supported version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,7 @@ jobs:
 
   mac:
     macos:
-      xcode: '10.0.0'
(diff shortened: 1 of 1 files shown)


belcherj (approved):
Ship It.

PR #3017, opened 2021-09-26
Update tag hash algorithm to make it consistent with other platforms

### LINKED ISSUE #3279: Fix escaping of special characters in tag names
Please see https://github.com/Automattic/simplenote-android/issues/1394 over at the Android repo. I _think_ the web app is likely the cause, but it could be Android as well.

**To Reproduce**
* Create a note with a tag on it with a dash, like `test-ing`.
* Head over to the web and create another note, and use the `test-ing` tag. I did the tab autocomplete to add the tag.
* You should now have two `test-ing` tags, but one of them is actually using a url escaped dash `%2D`.

sandymcfadden:
Commandeering this PR! --Kat

### Fix

There were a few characters that were not being encoded in the tag hash algorithm which were being encoded on other platforms. This causes inconsistency and duplicate tags when working across different platforms.

This will fix #3279 and might fix some of our longer-standing issues with duplicate tags. However it will not clean up any preexisting tags with special characters that were improperly encoded.

### Test

Follow testing directions on #3279, i.e. it's best to create a tag with special characters on a different platform and then add it to a note here. You can also create a tag here and then log into Simperium to confirm it is being properly escaped:

Screenshot 2025-01-23 at 07 06 29

### Release

Release notes were updated with:

`- Fixed inconsistent tag encoding of special characters that led to duplicate tags across platforms [#3017](https://github.com/Automattic/simplenote-electron/pull/3017)
`

### DIFF
--- a/lib/tag-field/index.tsx
+++ b/lib/tag-field/index.tsx
@@ -8,7 +8,11 @@ import EmailToolTip from '../tag-email-tooltip';
 import TagChip from '../components/tag-chip';
 import TagInput from '../tag-input';
 import classNames from 'classnames';
-import { tagHashOf } from '../utils/tag-hash';
+import {
+  isTagInputKey,
+  tagHashOf,
+  MAX_TAG_HASH_LENGTH,
+} from '../utils/tag-hash';
 import { noteCanonicalTags } from '../state/selectors';
 
 import type * as S from '../state';
@@ -201,6 +205,14 @@ export class TagField extends Component<Props, OwnState> {
   onKeyDown = (e: React.KeyboardEvent) => {
     if (this.state.showEmailTooltip) {
       this.hideEmailTooltip();
+    }
+
+    if (
+      tagHashOf(this.state.tagInput as T.TagName).length >
+        MAX_TAG_HASH_LENGTH &&
+      isTagInputKey(e.which)
+    ) {
+      e.preventDefault();
     }
 
     return this.interceptKeys(e);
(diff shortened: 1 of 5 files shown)


lib/tag-field/index.tsx
```diff
@@ -203,6 +203,14 @@ export class TagField extends Component<Props, OwnState> {
       this.hideEmailTooltip();
     }
 
+    if (
+      tagHashOf(this.state.tagInput as T.TagName).length >
+        MAX_TAG_HASH_LENGTH &&
+      String.fromCharCode(e.which).match(/([^,\s])/g)
+    ) {
```
codebykat, line 210: Sandy's original code looked at the keycode of the event in order to determine if it was a keypress that would add a character to a tag (AKA any character except whitespace and a comma). But keycodes are notoriously fragile and vary with different locales and keyboard layouts. Converting to a string and using a regex is mildly better, but I'm still a little concerned about the fact that we're not handling meta/modifier keys, clipboard actions, or (I believe) Unicode codepoints (there is a String.fromCodePoint function, but I don't think `which` is giving us a codepoint).

codebykat:
In testing I noticed that you can't paste more than one string into the tag field, which appears to be an existing bug, not something introduced by this PR. FYI @roundhill

roundhill (approved):
Tested well for me! The characters in the tag id were properly escaped. I also tested clicking outside of a tag and such as well and that seemed to work fine, too.

PR #3020, opened 2021-09-26
Update the revision selector and slider components to use css variables for colors

sandymcfadden:
### Fix

This updates the Revision Selector and Slider components.

### Test

- Open the history for a note

### Release

- Updated the Revision Selector and Slider components to use CSS variables for colors


belcherj (approved):
🚢

PR #3021, opened 2021-09-26
Update the note info compontent to use css variables for colors

sandymcfadden:
### Fix

This updates the Note Info components.

### Test

- Have a note with references pointing to it
- Open the Note Info panel using the (i) icon in the upper right
- Ensure the note info panel looks the same

### Release

- Updated the Note Info components to use CSS variables for colors


belcherj (approved):
🚢

PR #3022, opened 2021-09-27
Update the alternative login, verify email, and error boundary components to use css variables for colors

sandymcfadden:
### Fix

This updates the Alternative Login, Email Verification, and Error Boundary components.

### Test

The easiest way to test these is to force the modals to appear by temporarily editing the `app.tsx` file to always show them.
To display the error boundary you need to introduce an error in code. 

- Enable the Alternative Login Prompt
- Enable the Email Verification Modal
- Trigger the Error Boundary

### Release

- Updated the Alternative Login, Email Verification, and Error Boundary components to use CSS variables for colors

### DIFF
--- a/lib/dialog/index.tsx
+++ b/lib/dialog/index.tsx
@@ -23,6 +23,7 @@ export class Dialog extends Component {
       children,
       onDone,
     } = this.props;
+
     return (
       <div
         className={classNames(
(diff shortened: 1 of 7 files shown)


lib/dialog/index.tsx
```diff
@@ -23,7 +23,6 @@ export class Dialog extends Component {
       children,
       onDone,
     } = this.props;
-
```
belcherj: was this change intentional?

belcherj (approved):

lib/dialog/index.tsx
sandymcfadden: It was not at all, thank you! I've fixed it.

PR #3023, opened 2021-09-27
Update importer dialog to use css variables for colors

sandymcfadden:
### Fix

This updates the Import Dialog components. 

### Test

- Open the note importer
- Go through the process of importing notes
- Ensure everything still appears the same

### Release

- Updated the Importer dialog components to use CSS variables for colors

### DIFF
--- a/lib/dialogs/import/source-importer/executor/index.tsx
+++ b/lib/dialogs/import/source-importer/executor/index.tsx
@@ -115,7 +115,7 @@ class ImportExecutor extends Component<Props> {
               disabled={locked}
             />
           </label>
-          {hint && <p className="theme-color-fg-dim">{hint}</p>}
+          {hint && <p className="hint">{hint}</p>}
         </section>
         <TransitionFadeInOut shouldMount={Boolean(errorMessage)}>
           <div role="alert" className="source-importer-executor__error">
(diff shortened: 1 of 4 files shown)


lib/dialogs/import/source-importer/executor/index.tsx
```diff
@@ -115,7 +115,7 @@ class ImportExecutor extends Component<Props> {
               disabled={locked}
             />
           </label>
-          {hint && <p className="theme-color-fg-dim">{hint}</p>}
+          {hint && <p className="hint">{hint}</p>}
```
belcherj, line 118: I assume the CSS for hint exists elsewhere?

belcherj (approved):
Love these deletions! Ship it.

lib/dialogs/import/source-importer/executor/index.tsx
sandymcfadden: I actually don't think this hint is used anymore at all so never gets displayed. I added the class `hint` just in case we needed to style it in the future. But I believe it could be removed altogether.

sandymcfadden: PR to remove it here https://github.com/Automattic/simplenote-electron/pull/3027

PR #3024, opened 2021-09-27
Update settings and other dialogs to use CSS variables for colors

sandymcfadden:
### Fix

This is a large one that updates most of the Settings and other dialogs.

These are the components that have been updated:
- Panel Title
- Tab Panels
- Dialog
- Button Group
- Keybindings
- Settings Group
- Settings Panels
- Collaborating
- Trash Tag Confirmation
- Unsynchronized Notes

### Test

- Go through the Settings dialogs
- Ensure they still look the same
- Open the keybindings dialog
- Open the collaborate dialog
- Start to delete a tag to open the Trash Tag dialog
- Have unsynced notes and try to log out
- Ensure that the dialog warning still looks the same

### Release

- Updated the Dialogs used throughout the app to use CSS variables for colors

### DIFF
--- a/lib/components/tab-panels/style.scss
+++ b/lib/components/tab-panels/style.scss
@@ -4,39 +4,30 @@
   justify-content: center;
   margin: 0;
   padding: 0;
-  border-bottom: 1px solid;
+  border-bottom: 1px solid var(--secondary-color);
   list-style: none;
 
   li {
     margin: 0 0.5em;
     border-bottom: 2px solid;
     padding: 0.75em;
-    color: var(--primary-color);
+    color: var(--accent-color);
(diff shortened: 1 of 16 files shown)


lib/components/tab-panels/style.scss
```diff
@@ -4,41 +4,32 @@
...
+      }
     }
+
     &:active {
       color: var(--background-color);
     }
   }
 }
 
-body[data-theme='light'] .tab-panels__tab-list li.is-active {
```
belcherj, line 32: Feels sooo good having this gone.

belcherj (approved):
Code looks good, tested well.

PR #3025, opened 2021-09-27
Update App Layout and Note Revisions components to use CSS variables for colors

sandymcfadden:
### Fix

This updates the App Layout and Note Revisions components.

### Test

- Ensure the main layout of the app still appears the same
- Open note history
- Ensure the main content still appears the same 

### Release

- Updated the App Layout and Note Revision components to use CSS variables for colors


belcherj (approved):
🚢

PR #3027, opened 2021-09-28
Remove unused hint from importer

sandymcfadden:
### Fix

This removes an unused hint option that was included in the importer dialogs. It came to light in [this comment](https://github.com/Automattic/simplenote-electron/pull/3023#discussion_r716749274).

### Test

- Attempt to import some notes
- Ensure the import works as expected

### Release

- Removed unused hint option in the importer dialog

### DIFF
--- a/lib/dialogs/import/source-importer/executor/index.tsx
+++ b/lib/dialogs/import/source-importer/executor/index.tsx
@@ -19,9 +19,6 @@ type OwnProps = {
   locked: boolean;
   onClose: Function;
   onStart: Function;
-  source: {
-    optionsHint: string;
-  };
 };
 
 type DispatchProps = {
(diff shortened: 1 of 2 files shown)


belcherj (changes requested):
Are you sure this is unused?

lib/dialogs/import/index.tsx:L25 defines source

sandymcfadden:
> Are you sure this is unused?
> lib/dialogs/import/index.tsx:L25 defines source

`source` itself is still used here https://github.com/Automattic/simplenote-electron/blob/5d8f2ec5733d2fbe1bf73877bcad49f5be30350c/lib/dialogs/import/source-importer/index.tsx#L27

But the hint part of it was removed when refactored in https://github.com/Automattic/simplenote-electron/pull/2685/ it used to be defined in  `lib/dialogs/import/sources.ts`

codebykat (approved):
Tests fine for me 👍

PR #3029, opened 2021-09-30
Add handling of 429 HTTP errors on authentication

sandymcfadden:
### Fix

There are times when rate limiting will be applied when trying to login multiple times unsuccessfully causing future login attempts to fail as well. Currently, there is no way to show this in the app leaving things confusing if you get rate limited. This updates to handle that error case when logging in.

To do this it also includes an update to the Simperium dependency. It is right now a temporary package for testing.

Screen Shot 2021-09-30 at 1 01 12 PM

### Test

In addition to the steps below, please ensure that syncing, publishing, and other features work as expected.

1. Attempt to log in with valid credentials and ensure login is successful
2. Attempt to log in with a compromised password and ensure it is unsuccessful
3. Attempt to log in with an unverified account and ensure it is unsuccessful
4. Attempt to log in multiple more times with invalid credentials.
5. Ensure you are not able to log in and get the appropriate error message "Too many log in attempts. Try again later."

### Release

- Updated to handle too many login attempts during authentication

### DIFF
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -17,6 +17,7 @@ type OwnProps = {
   hasInsecurePassword: boolean;
   hasInvalidCredentials: boolean;
   hasLoginError: boolean;
+  hasTooManyRequests: boolean;
   hasUnverifiedAccount: boolean;
   login: (username: string, password: string) => any;
   requestSignup: (username: string) => any;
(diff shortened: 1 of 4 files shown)


codebykat (approved):
Looks great! My only quibble is, "log in" or "login"? I think it's a noun here so it's "too many login attempts" but it's late and I'm not 100% sure.

sandymcfadden:
@codebykat I think you are right with "log in" vs "login" but I'm going to leave it as it is for now as that is how the messaging is in GAE and it is the messaging Android is going with as well.

SylvesterWilmott:
@codebykat you are right. In this instance "login" should be used.

Issue #3071, opened 2022-02-03
End to end encryption


jrcacd:

charliescheer:
Thanks for the suggestion.  I have gone ahead and added this to our feature request list, we will consider it for potential future updates to Simplenote.

I am gonna go ahead and close this issue.

PR #3086, opened 2022-03-27
Why is there no cryptography like p2p? 

seriousfreezing:
Fix: Having encryption or even extra security for the notes

Test: idk


codebykat:
This should be an issue, not a spurious PR. Tracked in a few other places including #3071. TLDR we hope to be able to implement this someday, but it is not something on the immediate horizon.

PR #3097, opened 2022-06-18
CI: use supported image

FelicianoTech:
- `circleci/*` images are deprecated. This PR replaces them with the newer supported images.
- line 267 is removed as it's not needed when the config version is set to `2.1`.


codebykat:
Cool, thanks @felicianotech!

PR #3183, opened 2024-03-20
Upgrade everything

codebykat:
### Changes and problems so far

- node version bumped to `lts/hydrogen` (`v18.19.1`)
- replaced `husky` with normal git hook and a line in `npm prepare`
- replaced `misty` with `concurrently` and `wait-on` in the `Makefile`
- e2e tests disabled; [Spectron is deprecated](https://www.electronjs.org/blog/spectron-deprecation-notice)
- Webpack 4->5 migration has [significant configuration differences](https://webpack.js.org/blog/2020-10-10-webpack-5-release/) (polyfills especially)
- Webpack `file-loader` [replaced with asset modules](https://webpack.js.org/guides/asset-modules/)
- Fixed up linting (eslint / stylelint / prettier), for now ignoring changes that prettier and the linter want to make, to avoid making this PR hard to review
- `npm audit` now yields `found 0 vulnerabilities` 😎 
- Fixed: Our hack to disable default keybindings stopped working because there's now an official way to disable keybindings
- ~Monaco editor can't be upgraded past 0.45.0 due to platform detection issue (https://github.com/microsoft/monaco-editor/issues/4378) - this will be fixed in 0.48.0~ Update: Bumped to 0.48.0 to circumvent this
- Fixed: new note not focusing editor (actually editor not focusing on load) - this is an issue with react-monaco-editor above version 0.50.1
- Replaced Enzyme with react-testing-library (ref: [Enzyme is dead](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl))
- React to version 17
- MaterialUI v4 -> v5
- CircleCI builds... code signing
- Fix Monaco completion provider (for internal note links)
- Removed unwanted items from the Monaco context menu without CSS hacks! 🤩 

### Follow-up work

- possibly need to do a PR to upgrade vulnerable packages in `simperium` (e.g. `uuid`)

### References

- Electron breaking changes: https://github.com/electron/electron/blob/main/docs/breaking-changes.md
- Webpack [upgrade guide](https://webpack.js.org/migrate/5/) and [cheat sheet](https://gist.github.com/ef4/d2cf5672a93cf241fd47c020b9b3066a)
- Babel polyfill deprecation: https://babeljs.io/docs/babel-polyfill
- Monaco changelog: https://github.com/microsoft/monaco-editor/blob/main/CHANGELOG.md
- MaterialUI migration v4->v5: https://mui.com/material-ui/migration/migration-v4/
- Stylelint migration guides: [v14](https://github.com/stylelint/stylelint/blob/HEAD/docs/migration-guide/to-14.md), [v15](https://github.com/stylelint/stylelint/blob/HEAD/docs/migration-guide/to-15.md), [v16](https://github.com/stylelint/stylelint/blob/HEAD/docs/migration-guide/to-16.md)

### Test

Setup: Decrypt the config using `openssl aes-256-cbc -md md5 -d -in ./resources/secrets/config.json.enc -out ./config.json -k [key from the secret store]`

To test: basically everything 😭 

* `npm run test`
* `npm run dev`
* `make build` etc


roundhill:
Here's some bugs I've found:

- This error message upon first load:

Screenshot 2024-03-22 at 3 32 52 PM

- If you select `History` from the menu at the top-right, it crashes.

roundhill:
One lil' bug, if you click on the new note button, it should focus on the editor so you can start typing right away.

codebykat:
> One lil' bug, if you click on the new note button, it should focus on the editor so you can start typing right away.

Okay yeah fine, I wasn't sure prod did this either, but I just checked and it's a regression 😅 

Might be some of the Monaco code I commented out due to package loading issues, I'll check. Good find!

roundhill (approved):
I think we're at the point to land this! I've tested it a bunch in local dev electron and on web browsers.

Windows/Linux/Mac builds are also tested as well 🥳

PR #3184, opened 2024-03-21
Fix monaco recursion error

roundhill:
### Fix
The latest Monaco was not happy with a recursive call: `Invoking deltaDecorations recursively could lead to leaking decorations.` I've replaced it with `changeDecorations` which I think is the same thing under the hood 🤷. But it removes the error :)

### Test
1. On web or electron, edit a task list. You should not see any errors about the recursion.
2. In electron only, press cmd+g to search notes, press it again a few times. It should work and no recursion errors should be in the console.


codebykat (approved):
Nice, thanks!

PR #3185, opened 2024-03-22
Fix monaco decorations (again)

roundhill:
### Fix
Follow up from https://github.com/Automattic/simplenote-electron/pull/3184 that actually fixes the search highlighting in the editor ;)

### Test
Same as #3184


codebykat (approved):
👍

PR #3188, opened 2024-04-18
Experiment with overriding the menu with css.

roundhill:
Enables the `useShadowDOM` flag, and tweaks the menu with CSS to add styling and remove the last menu item.

Screenshot 2024-04-18 at 4 54 47 PM

**To Test**
* Right-click on web and electron in light and dark mode.
* The "Command Palette" menu should be gone.
* The menu should look nice and work :)

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -1243,6 +1243,7 @@ class NoteContentEditor extends Component<Props> {
               selectionHighlight: false,
               suggestOnTriggerCharacters: true,
               unusualLineTerminators: 'auto',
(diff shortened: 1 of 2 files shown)


codebykat:
Okay this was a good backup plan 🙇  buuuuut I feel like hacking the CSS is a little too fragile (and I'm not sure what other effects the removal of the shadow DOM might have later on). I managed to find a different hack that feels a tiny bit less terrible: https://github.com/Automattic/simplenote-electron/pull/3183/commits/b6e8303638eafa8b7e80c98e24a4ea8b29abc41c

PR #3192, opened 2024-05-09
Fix up editor padding and Safari right-click behavior

codebykat:
### Fix

When testing the context menu fixes, I noticed that I wasn't able to display the right-click menu in Safari, though it worked in Electron and other browsers. It would flicker for a moment and then vanish again.

I don't have a real mouse and use ctrl+tap on my trackpad to right click. It turns out that this does not create a "true" right click event, and was therefore being erroneously picked up by the click and mousedown handlers.

In the process of fixing this I found a few opportunities to clean up the padding, though there is sadly no official way to add left/right padding via Monaco. I also tested and fixed the word-wrap setting in Safari in a separate PR, #3191 .

### Test

In multiple browsers and on Electron:
* Click in various places on the editor and verify that it focuses the text editor.
* Right-click (with various mice if possible?) and verify that the context menu is shown appropriately.

n.b. We do not show the context menu after a right-click on the left gutter, because we are tricking Monaco into putting padding there by abusing the lineDecorationsWidth option.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -977,7 +977,10 @@ class NoteContentEditor extends Component<Props> {
     });
 
     editor.onMouseDown((event: Editor.IEditorMouseEvent) => {
-      if((event.event.ctrlKey && event.event.leftButton) || event.event.rightButton) {
+      if (
+        (event.event.ctrlKey && event.event.leftButton) ||
+        event.event.rightButton
+      ) {
         // ignore right clicks and OSX trackpad right clicks (ctrl+click)
         return;
       }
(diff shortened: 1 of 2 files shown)


lib/note-content-editor.tsx
```diff
@@ -938,6 +938,10 @@ class NoteContentEditor extends Component<Props> {
     });
 
     editor.onMouseDown((event: Editor.IEditorMouseEvent) => {
+      if((event.event.ctrlKey && event.event.leftButton) || event.event.rightButton) {
```
roundhill: Will `event.event` always exist? Maybe we should also add check for that.

roundhill: Also maybe a space after the `if`? Not sure what our coding standards are here but I see spaces after other ifs in the file.

roundhill (approved):
Tested fine, I had a few thoughts that aren't blockers

lib/note-content-editor.tsx
codebykat: Hah - `prettier` did fix this once I merged the other PR.

As for `event.event` always existing, I believe it must, and TypeScript didn't think it could be `null` either, so perhaps I'll leave that be 🤔

PR #3194, opened 2024-05-09
Update: Let Prettier do its thing

codebykat:
### Fix

In #3183 I added a list of files to the `.prettierignore` so that we wouldn't mix functional code updates with linting. This PR removes those files and runs `make format` to allow Prettier to format these files according to its wishes.

### Test

Code review and smoke test. This should include no functional changes.


roundhill (approved):
Still works! And... code is prettier.

PR #3195, opened 2024-05-09
Build: Update the deploy script and config file encryption

codebykat:
### Fix

1. We need `npm build --legacy-peer-deps` everywhere we were previously using `--force` or no arguments. This is required for the deploy script to succeed.
2. vip/package.json needs to be upgraded due to a security flaw in `express.js` (see #3193)
3. Also upgraded the VIP Go helper ([release notes](https://github.com/Automattic/vip-go-node/releases) show no breaking changes)
4. Add make commands to decrypt and encrypt the config file, and add info to the README, including the lack of support for community development at this time. These encrypt/decrypt commands are the ones we use on App Engine.
5. Re-encrypt the config file with newer encryption (again, same command from App Engine)

### Test

* CircleCI builds should succeed with this config
* `make decrypt_conf` should succeed and should create `config-local.json` with production secrets
* `npm run deploy development` should succeed


roundhill (approved):
* CircleCI builds should succeed with this config ✅
* make decrypt_conf should succeed and should create config-local.json with production secrets ✅
* npm run deploy development should succeed ✅ (I think this worked, no Slack confirmation but does it not run for development deploys?)

PR #3201, opened 2024-05-22
Update: react-monaco-editor and the hooks issue

codebykat:
### Fix

In the course of the multiple upgrades in #3183, we noticed that the editor sometimes failed to focus the text area on load or on creating a new note. I eventually traced this to the version bump of `monaco-react-editor`, specifically the [refactor to use hooks](https://github.com/react-monaco-editor/react-monaco-editor/issues/660), which was included in versions higher than 0.50.1.

After some experimentation I reordered the code in such a way that it seems to work (moving the initialization from `componentDidMount` into `editorReady`), but I'm a little worried about this change, because the existing codebase is written with a Redux (action dispatch) architecture that largely predates React hooks, and I'm not sure whether it might introduce unintended side effects. I'm also not sure if we still need to use a boot timer, or if this refactor incidentally solves that.

The discussion at the above-referenced issue does not contain any real workarounds nor is there an update to the component that reverts or fixes the behavior, which is somewhat concerning. However a quick search for alternatives to `react-monaco-editor` didn't unearth any viable options (the other major package is written to load the editor from a CDN on load to avoid configuration, which is definitely a choice).

### Test

Since this change affects the loading order of the editor component, we need to make sure it doesn't have any unintended side effects, so smoke-test all the things: switching notes, creating a new note, syncing changes, etc.

### Release

Fixed a bug causing the text area not to be focused on load or on new note creation.

### DIFF
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -186,20 +186,6 @@ class NoteContentEditor extends Component<Props> {
 
   componentDidMount() {
     const { noteId } = this.props;
-    // this.bootTimer = setTimeout(() => {
-    //   if (noteId === this.props.noteId) {
-    //     this.setState({
-    //       editor: 'full',
-    //       content: withCheckboxCharacters(this.props.note.content),
-    //     });
-    //     const position = getNotePosition(noteId);
-    //     if (position) {
-    //       this.editor?.setScrollPosition({
-    //         scrollTop: position,
-    //       });
-    //     }
-    //   }
-    // }, SPEED_DELAY);
     window.addEventListener('resize', clearNotePositions);
     window.addEventListener('toggleChecklist', this.handleChecklist, true);
     this.toggleShortcuts(true);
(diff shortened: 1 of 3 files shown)


dmsnell (approved):
If this tests well I don't see a problem with the proposed patch. These are still imperative APIs from the editor, no orthogonal to React and its hooks. Redux actions and updates still work fine in a hooks world too; the only thing to watch out for are the normal inherent gotcha's with hooks and stale data, when supplied to Redux code, as improper use of hooks can lead to corruption deeper in the code.

If the delay isn't causing noticeable harm maybe it's best to leave it in. Take it out in another PR where it can be tested by itself. I don't remember the reason it's there but I would hope it's in a commit message.

codebykat:
Thanks @dmsnell!

> I don't remember the reason it's there but I would hope it's in a commit message.

Unfortunately, as with so many things around here, almost all the blame points to a giant changeset that was itself a refactor of something else: https://github.com/Automattic/simplenote-electron/commit/fc8921291b6b702db35b076d65e4632203ff4574 🫠

dmsnell:
@codebykat the answer is in the following

Screenshot 2024-05-23 at 11 56 42 AM

https://github.com/Automattic/simplenote-electron/blame/d036a087370589a884f03ef32955e1842caf26ac/lib/note-content-editor.tsx#L126-L137

We were dealing with UI jank when navigating through notes because of the shift that occurs when initializing Monaco. It would render the checkboxes and that bogged down scrolling. The initial state then was the "fast" mode, which rendered plaintext, and when stopping at a single note, after the delay, would shift into "full" mode with the replaced checkboxes and full Monaco functionality.

So it's not necessary if the experience of switching between notes remains fast. It's a workaround to not having a faster UI.

mokagio:
@codebykat jus a ping because I landed on this approved but still open PR while looking at the PRs list for the upcoming 2.23.2 release.

I see `react-monaco-editor` is still at the version prior to this PR on `trunk`:

https://github.com/Automattic/simplenote-electron/blob/a1fe102878fd23b71719167e7a4d8d66dace55df/package-lock.json#L48

codebykat:
Thanks for the ping @mokagio! This is old enough that I'm not confident in it. There was something I wanted to revisit I think... anyway I'll set it back to draft for now.

PR #3204, opened 2024-05-29
Move from CircleCI to Buildkite

mokagio:
What is says in the title. Please forgive the long commits list—This was quite the trial and error. I underestimated the challenge of understanding (part of) `electron-builder` and getting everything it needs to work in a new CI environment.

The latest build shows everything green https://buildkite.com/automattic/simplenote-electron/builds/89

And I used this branch to tag `v2.22.0-beta6` and verify [the GitHub release gets created](https://github.com/Automattic/simplenote-electron/releases/tag/untagged-657b39dc85c254543422).

Unfortunately, the diff is substantial. I left a few comments in there.

PR #3206, opened 2024-05-30
Run lint and tests in Buildkite CI

mokagio:
Moves the lint and unit tests CI steps from CircleCI to Buildkite.

It uses an experimental approach to `npm` that:

1. Runs `npm ci` instead of `npm install`, as recommended for CI jobs
2. Uses caching, but since we're using `npm ci`, which recreates `node_modules`, only caches the NPM cache.
3. Has a gate at the start of the pipeline to create the cache if necessary, so that we'll avoid race conditions where multiple jobs might attempt to upload the NPM cache simultaneously.

I'm a bit unsure about the approach to be honest. Is the cache gate necessary? What are the chances of race condition? Besides, the performance of using vs not using the cache with `npm ci` do not seem to be that different, which makes me question whether it's worth having in the first place.

Keen to hear what you think. I'd still propose to merge as is and see how it behaves. This investigation has a broader scope that just this project, as we have new Electron apps that might use this new approach. 

### Test
See green CI.

### Release
N.A.


.buildkite/commands/install_node_dependencies.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -eu
+
+# --legacy-peer-deps is necessary because of react-monaco-editor.
+# See README for more details
+install_npm_packages_clean --legacy-peer-deps
```
mokagio, line 5: See https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/96

mokagio:
Requested an @Automattic/apps-infrastructure review for feedback on the `npm` approach.

@codebykat the head is `release/2.22.0` because I'm wishfully hoping to port the rest of the setup before it's time to cut the final build. At worst, we'll ship 2.22.0 on the old infra and continue the upgrade later.

.xcode-version
```diff
@@ -0,0 +1 @@
+15.4
```
codebykat, line 1: This snuck in here I think? Unless xcode is somehow required for Buildkite. I'd prefer this be removed and added to `.gitignore` if you like.

codebykat (commented):
Okay by me, if it works - I don't have an opinion about the npm caching. I know CircleCI is doing a bunch of stuff to sign the builds already, but it's finicky, maybe Buildkite is less finicky (however I have not worked with Buildkite at all). I'm happy for us to do whatever integrates best with the apps infra processes.

I did leave a note about the `.xcode-version` which probably shouldn't be there.

.xcode-version
mokagio: This is here intentionally. It's sort of like `.nvmrc`

It is "required" in that Buildkite reads it to know which macOS-Xcode VM to load. Other tooling might read this, although I don't think any would at the moment.

https://github.com/Automattic/simplenote-electron/pull/3206/files#diff-0259de1994ebca16ff7f60a3198680b690e517788c1b01be7c039206801f6ce5R7

mokagio:
Superseded by #3204

PR #3208, opened 2024-05-31
Fix: Notarize build

codebykat:
Import the correct package to notarize build, use `--legacy-peer-deps` everywhere, notarize even when not on a tag

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ references:
       command: |
         source $HOME/.nvm/nvm.sh
         nvm use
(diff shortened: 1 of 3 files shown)


mokagio:
Closing in favor of having picked this up on `release/2.22.0` directly, see f476598b3bf97d7db1e1431a35cff5cc00f713ec and apologies for stealing attribution @codebykat, I used `git cherry-pick` but with the `-n` option. I thought the command would have maintained the author but maybe because of `-n` the commit resulted as mine.

PR #3209, opened 2024-05-31
Exploration on the "can't resolve buffer" error

mokagio:
2bf1dd662b5134bc0c8d3d4d508406ab32a27773 shows that using `--force` instead of `--legacy-peer-deps` doesn't result in the failure in CI.

But why is the failure not occurring locally when using `--legacy-peer-deps`?

---

40b4250c74926e0c34da2edff9779928f1cbe321 and a6282429 are to verify whether the `react-monaco-editor` update, which if I remember well was the reason for using the legacy deps flag, solves the issue.

Still to answer: why is the failure not occurring locally when using `--legacy-peer-deps`??

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ references:
       command: |
         source $HOME/.nvm/nvm.sh
         nvm use
-        npm install --force
+        npm install --legacy-peer-deps
   npm_restore_cache: &npm_restore_cache
     restore_cache:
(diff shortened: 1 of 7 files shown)


mokagio:
Closing after establishing that @codebykat fix in a6613931fea70001bd2a7db60f590fb2635c90c7 addresses the issue.

**Still to answer why CircleCI and Buildkite see this, but local builds do not**

PR #3223, opened 2024-06-23
2.22.0 hotfix: Try to fix the build config + dependencies update

codebykat:
### Fix
* Use webpack config to fallback from config-local.json to config.json, this uses the correct config file without throwing an error from attempting to `require` a missing file
* `npm audit fix` for two high severity vulnerabilities (`braces` and `ws`)

### Test

Download build artifacts and verify you can successfully log in

### Release

* ~Use webpack config to fallback from config-local.json to config.json~ (including this in release notes with other build updates)


codebykat:
@mokagio Updated release notes, let's merge this if the builds succeed, and we can proceed with the release 🙇

mokagio (approved):

PR #3224, opened 2024-06-24
Attempt to fix build on top of #3223

mokagio:


.buildkite/pipeline.yml
```diff
@@ -92,10 +92,10 @@ steps:
       - $NVM_PLUGIN
     command: |
       .buildkite/commands/install_node_dependencies.sh
+      echo "--- Decrypt secrets"
+      make decrypt_conf
       echo "--- Build"
       make build
-      echo "--- Decrypt secrets"
-      make decrypt_conf_production
```
mokagio, line 98: @codebykat I think this is the reason the Linux build started [failing](https://buildkite.com/automattic/simplenote-electron/builds/107#019046f6-5eda-416c-a606-72c233bbb45a) in #3223.

I somehow mixed up the order here and attempted to decrypt the secrets only after running the build. 🤦‍♂️ 

It worked before #3223 because `get-config.js` would fallback to `config.json`. I guess we never noticed the config file was wrong because of our mutual issues with the Linux VM?

One thing I'm not sure of is why the `resolve.fallback` didn't kick in:

```js
resolve: {
      fallback: {
        './config-local': require.resolve('./config'), // fallback to config.json if config-local.json is missing
      },
```

I'm not familiar enough with Webpack to be able to say why 😞

codebykat: Aha!

And, yeah... it's possible that's not working how I think it is working, though it seemed to do what I expected locally. I'm okay with it at the moment, because as long as it's building with the secrets in `config-local`, the fallback is just there as an example. We need to rethink the example config, honestly, because it's not really possible to develop the app without those secrets anymore.

codebykat:
This is beautiful thank you!!

.buildkite/pipeline.yml
mokagio: > https://github.com/Automattic/buildkite-ci/tree/d95608759a43595b2493f3e88e8b4adab664f6a1/src/buildkite

Would adopting a `.env` file approach be suitable? We have (Ruby, and [WIP standalone binary](https://github.com/automattic/configure) tooling to manage that in the mobile apps and it's been working well.

Of course, it all depends on how much work is expected for this in the future 🙃

PR #3231, opened 2024-07-11
Create an app-config.json file in webpack

### LINKED ISSUE #3228: Upgraded to 2.22.0 and just shows white blank page and nothing else
### Expected

Expected SimpleNote to launch normally.

### Observed

Just a white blank screen:
image

I saved application logs as well but they are empty (0 bytes).

### Reproduced

I've tried quitting and relaunching, same thing happens.

### Where did you see the bug

- System Make: AMD
- System Model: Ryzen 7
- OS: Windows
- OS version: 11
- Browser (if applicable): 
- Simplenote app version: 2.22.0

roundhill:
### Fix
This fixes #3228. It does not appear possible to do dynamic requires in webpack, so instead I'm proposing the app use an `app-config.json` file that gets created in the webpack build, using either the contents of `config-local.json` if it is present, otherwise it will use `config.json` using `fs` to detect if the file is present or not.

### Test
* `npm run dev`
* Verify that an `app-config.json` file was created properly. (Try deleting your `config-local.json` file if present, it should update to whatever is in `config.json`)
* The electron and web app should both load up without errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules/**/*
+app-config.json
 config-local.json
 desktop-build/
 dist/
--- a/get-config.js
+++ b/get-config.js
@@ -1,7 +1,6 @@
 function readConfig() {
-  const configPath = './config-local';
   try {
-    const config = require(configPath);
+    const config = require('./app-config');
     if (typeof config === 'function') {
       throw new Error('Invalid config file. Config must be JSON.');
     }
@@ -9,7 +8,7 @@ function readConfig() {
   } catch (e) {
     // eslint-disable-next-line no-console
     console.error(
-      `Could not load the required configuration file at ${configPath}.\n` +
+      `Could not load the required configuration file.\n` +
         'Please consult the project README.md for further information.'
     );
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -86,9 +86,6 @@ module.exports = () => {
       ],
     },
     resolve: {
-      fallback: {
-        './config-local': require.resolve('./config'), // fallback to config.json if config-local.json is missing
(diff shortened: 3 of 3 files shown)


codebykat (approved):

twstokes:
Buildkite artifact testing:
- [macOS binary artifact](https://buildkite.com/automattic/simplenote-electron/builds/124#0190a2dd-0e48-490d-82dd-36172079cdd9) ✅ 
- [Windows binary artifact - x64 with installer](https://buildkite.com/automattic/simplenote-electron/builds/124#0190a2dd-0e4e-4b1c-ba1b-a3e63e6319a9): ✅ 
- [Linux binary artifact - x64 Debian package](https://buildkite.com/automattic/simplenote-electron/builds/124#0190a2dd-0e56-4a72-87be-7d67d5176ae8): ✅ 

On both macOS and Windows I tested that I could log in (to further ensure that the config contained the proper keys).

codebykat (approved):
Oh this is clever. I am not sure the code to create the file belongs in the webpack config?

Also, just want to note that technically using fs to check if the file exists could create a race condition. I don't think we're worried about it, but technically.

Did you find any info on why that fallback doesn't work? Do I have a different version of webpack installed or something?

I accidentally hit "approve" already but let me test this a bit more thoroughly.

roundhill:
> Did you find any info on why that fallback doesn't work? Do I have a different version of webpack installed or something?

No I'm not sure either, but it also seems that you can't have a `require` with a bad path anywhere in the code, as webpack needs to look for all of the required files to do its compression with.

roundhill:
Closing in favor of #3232

PR #3232, opened 2024-07-11
Fix: Webpack fallback release config

codebykat:
### Fix

Alternative to #3231, trying to see if this fixes the build artifacts.

Ref: https://medium.com/@6465660/webpack-compilation-error-critical-dependency-the-request-of-a-dependency-is-an-expression-13350139f2a5

And I'm mad about it 😠


twstokes:
✅ I was able to successfully run and log in (via WordPress.com) using the Mac artifact from [this job](https://buildkite.com/automattic/simplenote-electron/builds/130#0190a319-b584-42d4-a966-42a8208af1d9).

codebykat:
I also downloaded the Mac artifact and was able to log in. Can we find someone to test these on the Linux and Windows VMs?

Assuming they work, @roundhill how do you feel about proceeding with this fix?

twstokes:
> Can we find someone to test these on the Linux and Windows VMs?

I can be that someone, with real machines. 👍

twstokes:
@codebykat is starting the app and logging in enough to test for this change?

codebykat:
Yes! If you're able to log in that definitely shows it has the correct config.

twstokes:
[Windows 64-bit installer](https://buildkite.com/automattic/simplenote-electron/builds/130#0190a319-b58f-47e7-a952-10be666581eb): ✅ 
[Debian 64-bit](https://buildkite.com/automattic/simplenote-electron/builds/130#0190a319-b59a-4a3b-997f-2eccd5f25dbd): ⚠️ Application loaded and the WordPress.com auth flow worked up until Simplenote was brought back to the foreground and it showed an auth error. Force-quitting and reloading seemed to resolve it.

roundhill:
Here's an approach for loading the config for webback from the file system:

```
function getConfigFromFileSystem() {
  const configPath = fs.existsSync('./config-local.json')
    ? './config-local.json'
    : './config.json';

  return fs.readFileSync(configPath, 'utf8');
}

module.exports = () => {
  const isDevMode = process.env.NODE_ENV === 'development';
  const config = JSON.parse(getConfigFromFileSystem());
  ...
  ```
  
  This fixed the issue where `getConfig()` was being called before we set the `fallback`.

**Comment by codebykat** [comment]
Ugh. I've updated this PR to have webpack require `fs` and look to see which of the config files exist so that it can fall back without throwing a warning. I spent a VERY long time trying to allow `get-config.js` to require `fs`, but absolutely could not get webpack to provide it in the bundle no matter what I tried. As a result, the function `getConfig` is no more, and all uses of it now need to rely on the global `config` which is being magically provided by webpack. I made some config adjustments so TS is aware of the situation and won't complain.

I included a package upgrade as I got a notification about a high risk vulnerability in `electron-updater`, allowing a code signing bypass for Windows, which seemed important to address.

Once the artifacts build and we have tested them on all platforms, I think we've got our next (and hopefully final??) release candidate.

File: webpack.config.js [hunk 1]
```diff
@@ -1,13 +1,40 @@
...
+function readConfig() {
+  const configPath = fs.existsSync('./config-local.json')
+    ? './config-local.json'
+    : './config.json';
+
+  try {
+    const config = fs.readFileSync(configPath, 'utf8');
+    // if (typeof config === 'function') {
+    //   throw new Error('Invalid config file. Config must be JSON.');
+    // }
```
roundhill: This is probably ok to remove since it is in a try/catch block now. If there's something "not json-y" in the config file it will throw.

webpack.config.js
```diff
@@ -26,7 +53,7 @@ module.exports = () => {
         publicPath: config.web_app_url + '/',
       }),
     },
-    // target: 'node',
+    target: 'web', // this seems like it should be "node" or "electron-renderer" but those both crash
```
roundhill: Do you think we can just remove this line? I was getting some crashing because of it for a while, I can't reproduce now but just in case maybe we should keep this as it was. From the docs it looks like it should use the `browserslist` by default.

roundhill (changes requested):
Overall working well, I saw a weird thing with the webpack `target` for a while...

webpack.config.js
roundhill: https://webpack.js.org/configuration/target/

codebykat: Ah yeah I will comment it again.

roundhill (approved):
Our config woes seem resolved 😅 

* Build works with and without `config-local.json` present
* Sign-in works on both web and electron builds

lib/auth/index.tsx
```diff
@@ -48,7 +47,6 @@ export class Auth extends Component<Props> {
 
   render() {
     // Don't render this component when running on the web
-    const config = getConfig();
     if (config.is_app_engine) {
```
mokagio, line 50: What happens here? 

With my limited Electron + Webpack + TS understanding, I think at this point `config` would be `undefined`, making the if below unreachable. Unless... `config` is somehow injected?

webpack.config.js
```diff
@@ -1,13 +1,37 @@
...
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const MonacoWebpackPlugin = require('monaco-editor-webpack-plugin');
 const NodePolyfillPlugin = require('node-polyfill-webpack-plugin');
+const fs = require('fs');
 
-// Object.assign(global, { WebSocket: require('ws') });
+function readConfig() {
+  const configPath = fs.existsSync('./config-local.json')
+    ? './config-local.json'
+    : './config.json';
```
mokagio, line 13: Is it worth checking for `config.json` existence here and fail if it's not available? Or... can we trust the code later on to fail appropriately if this happens?

mokagio: > Or... can we trust the code later on to fail appropriately if this happens?

I should have kept scrolling... There is a clear `console.error` below that handles the case.

mokagio:
@codebykat @roundhill I update the target branch for this PR to `release/2.22.1`, #3233, as discussed internally. TL;DR given 2.22.0 has already been tagged, any fix on top of it needs to come in a new version, in this case, a patch version or hotfix.

lib/auth/index.tsx
roundhill: It's because `config` is actually set as a Webpack plugin, defined here:

https://github.com/Automattic/simplenote-electron/blob/13cb5d1b70ac095b2d41339f6287e7ecd350a371/webpack.config.js#L157

With my limited knowledge that is why it exists everywhere :)

codebykat: Yep exactly, as long as you *don't* define it inside the file, webpack will inject it. (If it IS defined in the file then webpack will assume you want a locally-defined version instead.)

mokagio: Thanks for the info @roundhill @codebykat

PR #3235, opened 2024-07-19
Windows signing fix experiments

mokagio:


mokagio:
d3ff05935fbaf78387383baeefeee0047d866460 was based on what's written in the docs

> If the AppX package is meant for Windows Store distribution, no need to sign the package with any certificate. The Windows Store will take care of signing it with a Microsoft certificate during the submission process.

I guess "no need to sign" might be different from "don't pass any parameter"?

image

mokagio:
With `c824f57bfcfc907235e539c627aa05a63bf8147e` we finally got back to an all green CI and a package that uploads fine to the Microsoft Partner Center:

image

However, my Windows VM showed all sorts of red flag when attempting to run the `.exe`, such as:

image

I'll keep iterating, but wanted to leave a record of this commit as the new green state.

Issue #3057, opened 2021-12-21
Windows 10/11 taskbar Icon


Saikur0n:
### Expected

Windows store version of app icon on the white taskbar should be without blue rectangle background and also the same size of other application icons.
### Observed

Windows store version of app icon on the white taskbar theme shows icon with a blue rectangle tile background and also on Windows 11 icon in start menu little bit smaller then other app icons.
### Reproduced

1. Set white taskbar theme on Windows 10/11
2. Download app from microsoft store
3. Launch app and check taskbar icon

### Where did you see the bug

- System Make:
- System Model:
- OS: Windows
- OS version: 10/11
- Browser (if applicable): 
- Browser version (if applicable): 
- Simplenote app version: 2.21

xmha97:
# Dark theme ✔️
![image](https://user-images.githubusercontent.com/23485114/148116629-0c283633-7a4a-4fb4-938a-ac6290397912.png)
# Light theme ❌
![image](https://user-images.githubusercontent.com/23485114/148116524-d141f122-9f6d-4283-bd89-518611d45364.png)

zdenys:
Another case in 4686314-zd-woothemes

![simplenote_win11_icon_missing](https://user-images.githubusercontent.com/39308239/150103118-e8266aef-0072-4b44-8a2d-4bfb7f025b82.png)

erricgunawan:
> … when Simplenote is pinned to the taskbar, it will have a square background of the Windows accent color. Is it possible to remove this background?
>
> Strangely, the background only appears in light mode but not dark mode.

https://forums.simplenote.com/forums/topic/remove-background-of-icon-on-windows-taskbar/?view=all#post-2661

thabotswana:
![simplenote-taskbar-accent-colour-showing-as-icon-background](https://github.com/user-attachments/assets/825670a0-f0c5-4a63-a669-8e011dce8346)


>Yes, I’d like to see that annoyance of my accent colour appearing as the background to the Simplenote taskbar icon please. As you can see above, it looks shite!

https://forums.simplenote.com/forums/topic/remove-background-of-icon-on-windows-taskbar/?view=all#post-2874

PR #3239, opened 2024-07-25
Update Windows configuration to remove AppX ambiguity

mokagio:
## What this does

In hindsight, it sounds pretty simple... This fix removes the `appx` node from the default configuration, `electron-builder.json`, so that CI generates only one set of `appx` files. The `appx` are then generated using a dedicated configuration. More on that later.

If you check [the logs for the `v2.22.1` build](https://buildkite.com/automattic/simplenote-electron/builds/141#0190c362-dd17-4b36-a186-9fd826668e27), you'll notice that two pairs of `appx` are being built. First as part of the `npx electron-builder --win` call which uses `electron-builder.json`...:

image

...then from the `npx electron-builder --win --config=electron-builder-appx.json` call:

image

You'll also notice that the first `appx` are code signed while the last are not. This sent me down a bit of a rabbit hole, more on that later, too.

So, it turns out that the code signing was not strictly broken, it just so happened that I downloaded the wrong `appx` artifacts from the release... 🤦‍♂️ That's an error only on my part, as our process checklist clearly [states](https://github.com/Automattic/mobile-toolbox/blob/d5416cbcef6c860c20bdb99d378fe4000ecedcff/desktop/simplenoteelectron-release-scenario.sh#L69-L70) to download the ones with `win-store` in the name.

## About code signing AppX

The short story is: We don't need to code sign AppX if we want to upload them to the Microsoft Partner Center. Upon uploading, Microsoft signs them for us. See also https://www.electron.build/configuration/appx#appx-package-code-signing.

Things got confusing when in the migration between CircleCI and Buildkite I made use of the certificate with publisher `CN="Automattic, Inc.", O="Automattic, Inc.", S=California, C=US` where before we were using `CN=E2E5A157-746D-4B04-9116-ABE5CB928306`. 

So, even though I was able to code sign the `appx` with the new certificates (despite code siging being unnecessary, which I only learned afterwards) Microsoft rejected them because of the mismatch between the expected publisher.

## About using two configurations

The reason for this setup is that, as far as I able to find, the configuration required to generate a code signed `exe` via the `nsis` target will conflict the `appx` configuration.

You can see the various experiments I went through to try and make both configurations play well together in #3235.

In practice, `"certificateSubjectName": "Automattic, Inc."` is required to sign the `exe` — which we need for it to run without looking dodgy, given it is distributed outside the store – but if that setting is present and so are the `appx` settings, there will be a failure, e.g. https://buildkite.com/automattic/simplenote-electron/builds/176#0190e45b-0298-4168-b47c-6e2aef7caec3, similar to what is reported in https://github.com/electron-userland/electron-builder/issues/6698

As a personal note, I wish I'd been smarter and dig into why we had two configurations sooner. Also worth noting that WordPress Desktop also uses this approach, and I knew it, too, use `electron-builder`. The fact that two projects use the same odd configuration setup would have pointed more clearly to the config incompatibility.

## "Proof" that it works

Here's a screenshot of the Partner Center where the `appx` packages from the latest build in this branch have passed validation:

image

And here is a screenshot of my Windows VM where the `exe` from the same build passes the Edge safety check etc.

image

image

## Where to go from here

Things I'd like to look into:

- Convert the config to YML so we can have comments and document why the setup exists close to where the setup is used. (In the meantime, I added a little note about code signing in the `make` task output)
- Look into the crash on the `electron-builder` side, as well as the configuration hierarchy that results in the clash. Maybe there is a reason, but I don't see why we cannot set code sining settings at the `nsis` level so that only the `exe` will code sign, instead of having `certificateSubjectName` in the `win` root and resulting in it code signing the `appx`, too
- If we really have to stick with the two config approach, then we should generated the ad hoc config at runtime with a script that
  1. Reads `electron-builder.json`
  2. Strips the unrelated settings
  3. Injects the appropriate `appx` settings


mokagio:
@codebykat @roundhill I wanted to add additional details on the implementation in the PR description but I run out of time today. 

Here's a ping to have a look and maybe test the `exe` on your end. I'll put the finishing touches tomorrow.

codebykat:
Oh, nice. I haven't tested this, but I'm looking at that transparent icon setting and wondering about these reported issues: https://github.com/Automattic/simplenote-electron/issues/3057 and https://github.com/Automattic/simplenote-electron/pull/3157

Possibly we want the icon to be *not* transparent? 🤔

roundhill:
The .x64 exe worked great for me on Windows 10. I tried installing the .appx but it said there was a code sign error, which I assume is because that has to be done on the Partner Center Portal?

mokagio:
>  I tried installing the .appx but it said there was a code sign error, which I assume is because that has to be done on the Partner Center Portal?

Correct @roundhill . The AppX are explicitly not signed by CI and we the Partner Center do that for us.

Makefile
```diff
@@ -141,14 +141,10 @@ package: build-if-changed
 
 .PHONY: package-win32
 package-win32:
-	@echo Packaging .exe...
+	@echo "Packaging exe..."
 	@npx electron-builder --win -p $(PUBLISH)
-ifeq ($(IS_WINDOWS),true)
```
mokagio, line 146: I decided to remove the platform check under the assumption these will mostly run in CI.

I'm not aware of any requirement for building Windows on a non Windows machine either.

This way, the script is simpler to understand.

Makefile
```diff
@@ -141,14 +141,10 @@ package: build-if-changed
...
-	@echo Packaging .appx as well...
+	@echo "Packaging appx with dedicated configuration to work around code signing conflicts..."
```
mokagio: A tiny attempt of help future readers understand the need for the dedicated config

electron-builder-appx.json
```diff
@@ -29,7 +29,5 @@
       "name": "simplenote",
       "schemes": ["simplenote"]
     }
-  ],
-  "afterSign": "./after_sign_hook.js",
-  "afterAllArtifactBuild": "./after_sign_hook.js"
```
mokagio, line 34: Notice that these were the same script. If you look into it, the only thing it does is notarizing for macOS. I decided to remove it to keep the config as small as possible.

spencertransier: To check my understanding of this file: it's fine to remove the calls to the `after_sign_hook.js` script because they only do macOS notarization and this file only deals with the Windows AppX build, correct?

mokagio: @spencertransier correct

Makefile
AliSoftware:
```suggestion
	# Note: the configuration required to generate a code signed exe via the `nsis` target will conflict with the `appx` configuration.
	# In practice, "certificateSubjectName": "Automattic, Inc." is required to sign the exe, but if that setting is present and so are the `appx` settings, there will be a failure.
	# Hence the need for a separate configuration here.
	# See also in https://github.com/electron-userland/electron-builder/issues/6698
	@echo "Packaging appx — with dedicated configuration to work around code signing conflicts..."
```
 - Borrowed some parts of your explanations from the PR description (👌 ) to add them as comments here for posterity.
 - Added the `–` to make it clearer that the thing that works around code signing conflict is the use of a dedicated configuration (as opposed to my uncaffeinated brain initially reading this as if packaging the appx was the thing that was working around code signing conflicts, e.g. by packaging as appx instead of another format, or using a tool like `fastlane` for the code signing, or other than `electron-builder`…)

electron-builder.json
```diff
@@ -54,17 +54,9 @@
       {
         "target": "nsis",
         "arch": ["ia32", "x64"]
-      },
-      {
-        "target": "appx",
-        "arch": ["ia32", "x64"]
       }
```
AliSoftware, line 57: @mokagio I was about to suggest adding some comments to mention that the `.appx` is instead generated with a custom config via `make package-win32` so it's properly code-signed, + link to this PR's description for the breadcrumbs…

…but then I just realized this is a `.json` file, which doesn't accept comments 😢 

That being said, [it seems that `electron-builder` supports configuration files in YAML (or js) format](https://www.electron.build/configuration/configuration.html). So WDYT of migrating the project to a `electron-builder.yml` file instead of `.json` in a future PR, so we can start adding such comments to it?

AliSoftware (approved):

electron-builder.json
mokagio: Sure, 100% agree. I thought I had written it somewhere as a TODO myself, but I couldn't find a reference to share here.

Every time I see a JSON used for config these days my first instinct is to ask if we can make it into a YAML. Being able to comment and use anchors and aliases makes the config work simpler.

Makefile
mokagio: Thanks!

Issue #3241, opened 2024-07-30
Version 2.22.1 not installing on Linux Mint 22


jannotti-glaucio:
### Expected
Install simplenote version 2.22.1 on a linux minnt 22 distribution, using the .deb installation

### Observed
When I try to install it I receive the following error:
![image](https://github.com/user-attachments/assets/e4b349eb-d89c-434a-a6bb-364344bd52e2)

### Where did you see the bug
- System Make: Dell
- System Model: Latitude 5510
- OS: Linux Mint
- OS version:  22
- Simplenote app version:  2.22.1

DanielNoga:
Same for Kubuntu 24.04 and latest beta 2.22.2 beta2

fishinnabarrel:
Same problem on Debian Trixie ...
![Screenshot_20240804_142211](https://github.com/user-attachments/assets/ba77fe6f-854a-4304-90e7-b6e18e9ee321)

codebykat:
Hm. @mokagio do you happen to know if this dependency is still necessary? Some prior discussions [here](https://github.com/Automattic/wp-calypso/issues/75452) and [here](https://github.com/Automattic/simplenote-electron/issues/3048).

codebykat:
Thanks, this dependency has been removed in the upcoming 2.22.2 version. If you'd like to test this ahead of the release, you can download the beta from here: https://github.com/Automattic/simplenote-electron/releases/tag/v2.22.2-beta.4

PR #3244, opened 2024-08-08
Remove `gconf2` dependency

mokagio:
### Fix

This is an attempt to fix the issue reported by users who cannot install Simplenote on various Linux systems because the `gconf2` dependency cannot be satisfied, https://github.com/Automattic/simplenote-electron/issues/3241

The [`electron-builder` docs](https://www.electron.build/configuration/linux.html) mention that `gconf2` can give issues if one wants to support [KDE](https://kde.org/).

> If need to support KDE, `gconf2` and `gconf-service` should be removed as it’s no longer used [by GNOME](https://packages.debian.org/bullseye/gconf2).

Moreover, [the `gconf2` entry in the Debian packages repository](https://packages.debian.org/bullseye/gconf2) states:

> This package is for legacy applications and no longer used by GNOME.

Given the above, this PR is an attempt to see what the app would do without the dependency. My guess is that everything should run fine in most modern systems, because all information points to `gconf2` not being in use there.

### Test

To test this we'll need to download the artifacts from CI and try to install them on Linux distros. If it works on our end, it might be good to reach out to the users in #3241 and see how they go.

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -110,8 +110,7 @@
     }
   ],
   "deb": {
-    "compression": "gz",
-    "depends": ["gconf2"]
+    "compression": "gz"
   },
   "afterSign": "./after_sign_hook.js",
   "afterAllArtifactBuild": "./after_sign_hook.js"


mokagio:
All the Linux binaries have been built, which is already a good sign. I didn't have the time to set up VMs, but I plan to do so next week.

In the meantime pinging @codebykat and @roundhill in case you have access and bandwidth.

codebykat:
This got lost in the Slack backlog, noting @roundhill was able to successfully test this on Fedora in an arm64 VM.

codebykat:
Hmm. We can't share the build artifacts from Buildkite, they require a login. I'm going to merge this in hopes of getting a testable beta.

mokagio:
>  I'm going to merge this in hopes of getting a testable beta.

👍

PR #3246, opened 2024-08-16
Add Magic link login

codebykat:
### Fix

Adds the option to login via a link or code to the Electron apps.

Screenshot 2024-08-27 at 12 17 37 PM

Screenshot 2024-08-27 at 12 20 00 PM

### Test
1. You should download a build from this PR to test with. (In the build artifacts)
2. Enter a valid account email.
3. You should get the code or link via email
4. Entering the code should work (Clicking the link was tested on the GAE PR already)
5. Try ALL the other things on the sign in form: password entry, creating an account, etc. Should still work!

### Release

Added the option to log in via "magic" link or code.

### DIFF
--- a/desktop/preload.js
+++ b/desktop/preload.js
@@ -15,7 +15,7 @@ const validChannels = [
   'wpLogin',
 ];
 
-contextBridge.exposeInMainWorld('electron', {
+export const electronAPI = {
   confirmLogout: (changes) => {
     const response = remote.dialog.showMessageBoxSync({
       type: 'warning',
@@ -61,4 +61,6 @@ contextBridge.exposeInMainWorld('electron', {
   },
   isMac: process.platform === 'darwin',
   isLinux: process.platform === 'linux',
-});
+};
+
+contextBridge.exposeInMainWorld('electron', electronAPI);
--- a/desktop/preload.js
+++ b/desktop/preload.js
@@ -15,7 +15,7 @@ const validChannels = [
   'wpLogin',
 ];
 
-export const electronAPI = {
+const electronAPI = {
   confirmLogout: (changes) => {
     const response = remote.dialog.showMessageBoxSync({
       type: 'warning',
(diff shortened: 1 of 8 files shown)


roundhill:
We might need to make sure we're sending `Access-Control-Allow-Origin` from the new endpoints?

lib/boot-without-auth.tsx
```diff
@@ -118,6 +121,74 @@ class AppWithoutAuth extends Component<Props, State> {
...
 
+  requestLogin = (email: string) => {
+    const username = email.trim().toLowerCase();
+    if (!username) {
+      return;
+    }
+    this.setState({ authStatus: 'submitting' }, async () => {
+      try {
+        const response = await fetch(
+          'https://app.simplenote.com/account/request-login',
```
jleandroperez, line 132: Is it a regular practise to leave URLs like this one?

desktop/preload.js
```diff
@@ -61,4 +61,10 @@ contextBridge.exposeInMainWorld('electron', {
   },
   isMac: process.platform === 'darwin',
   isLinux: process.platform === 'linux',
-});
+};
+
+contextBridge.exposeInMainWorld('electron', electronAPI);
+
+module.exports = {
```
roundhill, line 68: Needed to change this to `module.exports` to avoid a fatal

jleandroperez (approved):
✅ Logging in with Username + PW works
✅ Logging in with Username + Code works
⚠️ Code text field isn't the first responder
✅ Logging in with Username + Magic Link
✅ Quit after Sign In works well

Works exactly as advertised Dan!!

:shipit:

lib/boot-without-auth.tsx
roundhill: Ehh probably not, I'll see if there's some globals or a config or something!

roundhill: Other urls were just hard code like that so I left these in!

roundhill:
> ⚠️ Code text field isn't the first responder

All sorted! Thanks @jleandroperez

Issue #3048, opened 2021-11-26
Remove dependency on gconf2


dimitry-ishenko:
### What

Replace gconf2 with dconf.

### Why

gconf2 is deprecated.

From wikipedia:

> GConf was a system used by the GNOME desktop environment for storing configuration settings for the desktop and applications. It is similar to the Windows Registry.
> 
> It was **deprecated** as part of the GNOME 3 transition. Migration to its replacement, GSettings and dconf, is ongoing.

From [https://gitlab.gnome.org/Archive/gconf](https://gitlab.gnome.org/Archive/gconf):

![Screenshot from 2021-11-25 21-54-45](https://user-images.githubusercontent.com/2858291/143519944-0ee7944d-4456-42e7-b49f-288beae83f68.png)

### How

By modifying the code to use dconf instead of gconf2?... Yep.

dimitry-ishenko:
FWIW it seems whatever the thing that `Simplenote` is built on, which used to require `gconf2`, no longer needs it.

In other words, I was able to install both `Simplenote-1.21.1` and `Simplenote-2.21.0` (one at a time) with:
```shell
dpkg -i Simplenote-linux-...-amd64.deb
```
and they both worked fine without `gconf2` being installed on my system.

Not sure if whatever that thing was that needed it (_electron_, or _chromium_, or whichever), has since switched to `dconf` or just altogether doesn't need it anymore.

codebykat:
Thanks, yeah, looks like [Electron has removed the dependency](https://github.com/electron/electron/pull/19498).

dimitry-ishenko:
When was this completed? Latest Linux version (2.21.0) still depends on gconf2.

PR #3248, opened 2024-08-27
Merge 2.22.2-beta.3 into `trunk`

mokagio:
Routine PR to ensure `trunk`'s version reflects the latest available (beta) version.

All changes have already gone through code review, see https://github.com/Automattic/simplenote-electron/pull/3239. Also notice that the change for beta 3, #3048, was accidentally merged into `trunk` instead of `release/2.22.2`, which is why it's not in the diff.

Finally, ignore the commits list, it appears like that because we use "Squash and merge".


roundhill (approved):
Thanks @mokagio!

PR #3257, opened 2024-09-12
Meta: Update RELEASE NOTES instructions in the PR template

codebykat:
### Fix

This updates the RELEASE NOTES section of the PR template to copy [the one used in the simplenote-android repo](https://raw.githubusercontent.com/Automattic/simplenote-android/trunk/.github/PULL_REQUEST_TEMPLATE.md), because I just opened a PR over there and I liked it. I feel our current one creates some confusion by leaving it ambiguous when the release notes themselves will be updated, and that is currently an extra manual step in the release cycle.

In other words **this represents a slight change to our process for generating release notes**; rather than collating them all and doing a separate commit for release notes when we cut a new beta, this will keep release notes updates with the relevant PRs. Hopefully it won't become annoying if we don't know what version a PR will land in and have to move things around, but we can iterate on it.

FYI for @mokagio particularly!

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,11 +1,13 @@
 ### Fix
+
 <!--
-**_(Required)_** Add a concise description of what you fixed. If this is related 
+**_(Required)_** Add a concise description of what you fixed. If this is related
 to an issue, add a link to it. If applicable, add screenshots, animations, or
 videos to help illustrate the fix.
 -->
 
 ### Test
+
 <!--
 **_(Required)_** List the steps to test the behavior. For example:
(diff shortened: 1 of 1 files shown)


mokagio (approved):
This sounds good to me.

> Hopefully it won't become annoying if we don't know what version a PR will land in and have to move things around, but we can iterate on it.

Most, if not all, the mobile apps and libraries have some version of this workflow and:

- I haven't had reports of it being annoying or toil for devs
- I haven't felt it to be toil myself
- Our PRs tend to land relatively fast which means the changes for them being out of sync with `RELEASE-NOTES.md` on the main branch is low. Usually when that happens there is a merge conflict anyway, it's rare that the notes become "corrupted".

Keen to see how this goes for y'all.

PR #3256, opened 2024-09-09
Add label/link to log in with email view to manually log in.

roundhill:
### Fix #3255 

Adds a label with a link on the email log in form so that you can get to the password entry if desired.

I've also removed the `isDev` check for showing password entry for dev mode - since you can click the link now to sign in with password that should work ok for devs to sign in.

Screenshot 2024-09-09 at 3 54 28 PM

### Test
1. Sign out.
2. `npm run dev`
3. You should see email log in form, click the new `manually log in` link. 
4. You should now be at the password log in form.
5. Try click back to email log in, sign up, etc. You should only see the new label when on the email log in form.
6. Signing in should work (best to download a build of this PR so magic links work)

PR #3254, opened 2024-09-05
Update: Disable sticky scroll option

codebykat:
### Fix

Potential fix for #3250. Also removes this, and some unused features, from the package.

This is a hotfix for 2.22.2 release.

(I updated the list of possible features by importing metadata and logging it to console, as recommended in the linked README.)

### Release

Disabled "sticky scrolling" option ([#3254](https://github.com/Automattic/simplenote-electron/pull/3254))

PR #3266, opened 2024-10-10
Merge 2.22.2 release branch back into trunk

codebykat:
### Fix

Hmm realized we never did this, I feel like we should have by now?

cc @mokagio in case I did it wrong. I'm not sure if I should squash and rebase or create a merge commit for this one.

There are not many changes that were merged as hotfixes, basically just #3254, #3256, and some release notes.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,10 +1,27 @@
 # Changelog
 
+## [v2.22.2]
+
+### Enhancements
+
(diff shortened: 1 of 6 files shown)


mokagio (approved):
Thanks @codebykat 

My bad for not opening this when I shipped 2.22.2.

Regarding merging vs squash and merge... 🤷  I configured auto-merge here with squash because that's what we usually do for other PR, right?

PR #3269, opened 2024-10-30
Update: Clear tracks data when logging out

codebykat:
### Fix

Fixes this issue about the `tk` cookie sticking around: https://hackerone.com/reports/2793499

Will close https://github.com/Automattic/simplenote-gae/issues/644

### Test

1. Log in and have analytics enabled (if it was disabled, you might want to log out and log back in to make sure the cookie gets set)
2. Log out
3. Verify that the `tk_id` cookie no longer has the signed-in email address

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [v2.23.0]
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


roundhill (approved):
Tested well, the `tk_` cookies were removed! Thank you @codebykat

PR #3281, opened 2025-01-03
Fix WPCC redirect

### LINKED ISSUE #3278: "Log in with WordPress" fails in macos 2.22.2
### Expected

Redirected to browser, log in, application is opened and authenticated

### Observed

Redirected to browser, log in, a tab opens with logged in browser SimpleNote. Going back to the app and clicking the button again results in a page that does try to open the app, but does not log it in. 

### Reproduced
1. Click "Log in with Wordpress.com" in the app
2. Log in in browser
3. You are not logged in in the app.

### Where did you see the bug

- System Make: Apple 
- System Model: 2019 MacBook pro
- OS: macOS Catalina 
- OS version: 10.15.7
- Browser (if applicable): chrome
- Simplenote app version: 2.22.2

codebykat:
### Fix

Fixes #3278 

A bit of a typo / copypasta was introduced in #3246 that caused the auth state (checksum) to be overwritten before the request was handled and verified. If we keep it as a locally-scoped `const` rather than a state variable, authentication works again.

### Test

1. n.b. Make sure you are logged out of app.simplenote.com, or it won't redirect back to the app
1. Download the build artifact from this PR
2. Sign in via WordPress.com
3. Approve the request and login should succeed

### Release

> Fix WordPress.com login on the Electron builds

### DIFF
--- a/lib/auth/index.tsx
+++ b/lib/auth/index.tsx
@@ -543,7 +543,7 @@ export class Auth extends Component<Props> {
 
   onSubmitCode = (event: React.FormEvent) => {
     event.preventDefault();
-    this.setState({ authState: 'login-requested' });
+    // this.setState({ authStatus: 'login-requested' });
     this.setState({
       passwordErrorMessage: '',
     });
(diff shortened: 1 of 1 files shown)


roundhill (approved):
Worked great and code LGTM!

PR #3288, opened 2025-01-16
Bump version to 2.23.0 stable

mokagio:
Release flow PR, no additional description needed.


mokagio:
Admin merging so it doesn't get stale and I can also close the loop on the release.

PR #3291, opened 2025-02-12
Update: use `simplenote:user_id` instead of `wpcom:user_id` for Tracks

### LINKED ISSUE #3290: Tracks event are using wrong user id type
Tracks events from the app are sending `wpcom:user_id` value for the  `useridtype` property. We should update it to be `simplenote:user_id`

Internal ref: p4qSXL-8qM

codebykat:
### Fix

Will close #3290.

After testing, @roundhill and I concluded that we only care about the Simplenote ID, and don't need to associate these accounts with WPCOM accounts. At any rate, the code path that attempts to load user information from the WPCOM login cookie never seems to be hit. So for consistency with the mobile apps (and the expectations of the Tracks API) this removes that code entirely and sends the Simplenote ID (email address) as the ID type `simplenote:user_id`.

### Release

> Updated the ID type used for analytics to avoid sending invalid WPCOM IDs

### DIFF
--- a/lib/analytics/tracks.ts
+++ b/lib/analytics/tracks.ts
@@ -139,7 +139,7 @@ function buildTracks() {
     }
     userId = loadWpcomIdentity();
     if (userId) {
-      userIdType = 'simplenote:user_id';
+      userIdType = 'wpcom:user_email';
     } else {
       userIdType = 'anon';
       userId = get(userAnonCookie);
--- a/lib/analytics/tracks.ts
+++ b/lib/analytics/tracks.ts
@@ -139,7 +139,7 @@ function buildTracks() {
     }
     userId = loadWpcomIdentity();
     if (userId) {
-      userIdType = 'wpcom:user_id';
+      userIdType = 'simplenote:user_id';
     } else {
       userIdType = 'anon';
       userId = get(userAnonCookie);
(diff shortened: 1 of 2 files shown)


roundhill:
I don't think we want to try and send wpcom info at all - makes no sense. This feels like we went about it all wrong originally - it should be sending up the Simplenote user's email if they are signed in, and not read any wpcom cookies or anything... weird.

lib/analytics/tracks.ts
```diff
@@ -139,7 +139,7 @@ function buildTracks() {
     }
     userId = loadWpcomIdentity();
     if (userId) {
-      userIdType = 'wpcom:user_id';
+      userIdType = 'wpcom:user_email';
```
roundhill: Even with wpcom login this path was never reached during my testing.

roundhill (approved):
I think this is the change we need to fix the error on the tracks side 👍 I tested with regular and wpcom login and they were both sent with `simplenote:user_id` which matches it up to the other apps.

codebykat:
Cool - removed the WPCOM code to avoid future confusion. I am pretty sure this Tracks code was originally a library written for WPCOM and that code was likely vestigial, but if it ever worked it has been broken for some time.

PR #3240, opened 2024-07-29
fix: no button outline

wenfangdu:
### Fix

Button outline looks bad, not sure why it's added:

https://github.com/user-attachments/assets/88a754f3-93cd-46e9-b0f9-9a314b787c69

### Test

1. Click the note entry
2. No button outline

https://github.com/user-attachments/assets/593540ce-f71a-483b-9ca7-b2ee8bb627f8

### Release

Removed button outline.

PR #3294, opened 2025-02-12
Chore: Update sass, stylelint, focus rings, general SCSS cleanup

codebykat:
### Fix

Upgrading dependencies for sass and stylelint I realized there was some tidying up to do.

* sass now uses dart which requires some workarounds, such as string interpolation for rgba() with variables ([ref](https://stackoverflow.com/a/47625605))
* `@import` is deprecated in sass so we need to migrate to `@use` ([ref](https://sass-lang.com/documentation/breaking-changes/import/))
* Fix weird focus rings. We don't need a polyfill for focus-visible anymore since it's part of modern css ([ref](https://github.com/WICG/focus-visible?tab=readme-ov-file#support)). This was raised in #3240 but no one ever got around to fixing it. ([ref](https://github.com/Automattic/simplenote-electron/pull/3240#issuecomment-2315836096))
* Allow WP style class names in stylelint ([ref](https://github.com/WordPress/gutenberg/issues/28616))

### Test

1. `make lint-scss`
2. `npm run dev` and smoke-test the app
3. test in both dark and light themes
4. test keyboard navigation and make sure focus rings appear on elements as appropriate

### Release

> Updated styles, sass, style linting


roundhill (approved):
Tested well, the focus ring thing was indeed annoying!

Issue #3053, opened 2021-12-03
Option to Maintain Indention after line overflow/word wrap, and Indention Guidelines


jatoran:
### What
Maintain Indention after line overflow, and Indention Guidelines

### Why

These 2 changes will greatly improve the readability of notes, especially for people who maintain lots of different topic'd notes within one note document (such as an inbox of notes that has not been sorted yet)

Maintaining Indention:  This will improve readability of notes, especially when multiple different notes are near eachother.  When indenting, if you are writing a few sentences, the new line will lose indention after word wrapping, which makes the notes look very jumbled.

Indention Guidelines:  Due in part to the empty space before the writing area, it is difficult to see where the indentions end and as such multiple nested indentions within a single note topic can become hard to read or discriminate from other indention levels.

### How
Sublime text Indention Guidelines: 
![image](https://user-images.githubusercontent.com/16913820/144634783-6cf1809c-4f29-4594-a3a9-e88c71c22ab8.png)

Maintaining Indentions after new lines/word wrap:
![image](https://user-images.githubusercontent.com/16913820/144635136-b2919965-1e02-4705-9032-1c854e36878a.png)

d-walsh:
This is really important and it was requested here also: https://forums.simplenote.com/forums/topic/feature-request-option-to-preserve-indent-when-text-wrapping/#post-78

michaeltrollan:
This would be a huge improvement. Would love to see it implemented.

erricgunawan:
Might be related:

> It appears that in all of my notes, indenting a paragraph does not work any more. It has been this way for a while and I have made sure everything is up to date but the issue persists.
>
> Any notes that existed prior to this issue are appear indented properly upon first opening the note for just a split second and then flash to being all aligned. If I try to indent a paragraph now, it indents the entire block rather than just the first line as it is supposed to.
>
> Note, I use Simplenote through the online browser version

> It seems like this is the view provided by the Simplenote web app. In the iOS app, a paragraph created with a leading tab only indents the first line. When I view this note in the web app, all lines of the paragraph appear indented. Changing the text in the web, and switching back to iOS, and the text is back to normal. It works the same with leading spaces on the first line.

https://forums.simplenote.com/forums/topic/paragraph-indenting-not-working/?view=all#post-2678

codebykat:
Hmm. Well, I do agree we should fix the hanging indent issue - I have a fix for that in #3216.

[...]

PR #3302, opened 2025-03-04
Update: Indentation woes

codebykat:
### Fix

In #3216, I changed Monaco's `wrappingIndent` to `none` in order to address a request to support hanging indents on paragraphs. Unfortunately this also removes the indentation from lists, making lists less readable.

Some further research in https://github.com/Automattic/simplenote-electron/issues/3053#issuecomment-2696467125 but basically I think we were better off with the default of `same`. There's no way to indent paragraphs, sadly, as even a single space gets matched by the word wrapping algorithm (which is not affected by the settings for detecting indentation settings, tab size, etc.).

Monaco/VS Code does not support these things I've thought of as workarounds:
- Different word wrap settings for lists and paragraphs
- Using CSS to apply styling to only the first line
- Different word wrap settings for different languages
- Using the tokenizer to adjust the indent size
- Language settings for indent/outdent applying to the word wrapping indent

Here are some relevant upstream issues: https://github.com/microsoft/monaco-editor/issues/3914, https://github.com/microsoft/vscode/issues/219680, https://github.com/microsoft/vscode/issues/169804, https://github.com/microsoft/vscode/issues/164267

Anyhow, this PR represents a partial revert of #3216, as maybe the least terrible of the available options. I think that `indent` is a deal-breaker due to representing a serious departure from expectations for long-form writing, whereas a hanging indent on paragraphs is more of a "nice to have" for that use-case. Meanwhile for lists, `none` is pretty bad, probably(?) worse than the weird behavior of indenting entire paragraphs that we get with `same`.

Which is a complicated way of saying that `same` is optimal for lists with word-wrapping, `none` is optimal for paragraphs, and we can't switch our strategy depending on which it is we have, so we have to pick a side.

Other considerations:
- An initial indent gets perceived by the Markdown preview renderer as an indication of a code block. List items that aren't indented, and their nested children, are safe -- but indented list items that don't have an un-indented parent are not.
- The native apps apparently support an initial indent on paragraphs, which then indents the entire paragraph when viewed on web/desktop.

## Screenshots

Here are the three options and their trade-offs.

Screenshot 2025-03-04 at 01 03 09

Screenshot 2025-03-04 at 01 34 39

Screenshot 2025-03-04 at 02 26 31

### Release

Release notes updated with:

> Remove hanging indent support due to its detrimental effect on word-wrapped list items.

n.b. I've added this to a new minor version, 2.23.3, but it could arguably be a hotfix for 2.23.2.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [v2.23.3]
+
+### Fixes
+
+- Remove hanging indent support due to its detrimental effect on word-wrapped list items [#3302](https://github.com/Automattic/simplenote-electron/pull/3302)
+
 ## [v2.23.2]
 
 ### Other Changes
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -1286,7 +1286,7 @@ class NoteContentEditor extends Component<Props> {
               unusualLineTerminators: 'auto',
               wordWrap: 'bounded',
               wordWrapColumn: 400,
-              wrappingIndent: 'none',
+              wrappingIndent: 'same',
               wrappingStrategy: 'advanced',
             }}
             value={content}


mokagio:
@dmsnell what do you think of this old PR? it's stale but the diff is minimal, really just a change in one of the configs. Shall we merge?

dmsnell (commented):
@mokagio if you’d like, go for it. I don’t really have the capacity to give it the review and testing it deserves. but it can always be reverted. if you have tested and feel confident, then it should be fine.

PR #3306, opened 2025-03-07
Bump version to 2.23.2 stable

mokagio:
Process breadcrumb PR for release 2.23.2.

If CI is green for both the commit and the tags that will be associated to it, I will admin-merge. The team ping is for FYI only.


roundhill (approved):
Thank you Gio!

PR #3316, opened 2025-03-30
Fix: Paste issues in the context menu

### LINKED ISSUE #2994: Enable "Paste" option in Monaco contextual menu
## Issue

The contextual menu accessible by right-click on the editor does not contain a `paste` action.

## Fix

Add `Paste` contextual menu item below `Copy`

### LINKED ISSUE #3315: Windows: Can't use right-click context menu to paste
Reported by four different users so far in this forum thread: https://forums.simplenote.com/forums/topic/windows-unable-to-paste-using-right-click-in-version-2-23-2/

### Expected
Users should be able to use the right-click context menu to paste content into the Simplenote app.

### Observed
Users cannot use the right-click context menu to paste content into the Simplenote app.

### Reproduced
1. Copy text from another app or a website.
2. Open the Simplenote Windows app.
3. Open or create a note.
4. Right-click on the note editor.
5. Click "Paste" to try and paste the previously copied text.
6. Notice that the text is not pasted into the note.

### Where did you see the bug

Replicated on virtual machine:
- Windows 11
- Simplenote 2.23.2

codebykat:
### Fix

Will fix #3315 
Will fix #2994

Monaco/VSCode [now support](https://github.com/microsoft/vscode/issues/82604) paste in the browser.

We do still need to re-add the keybindings if we want labels for the keybindings. See: https://github.com/microsoft/monaco-editor/issues/2882#issuecomment-2101997854

### Test

This needs testing in multiple browsers and platforms (browser and Electron).

1. Right click should show Cut/Copy/Paste with labels appropriate to the platform (cmd key / ctrl key)
2. Cut/Copy/Paste should work from context menu
3. Should also work from keyboard shortcuts
4. Should also wrok from top-level Edit menu

### Release

> Fixed paste not working in context menu and enabled it in browsers

### DIFF
--- a/lib/boot-without-auth.tsx
+++ b/lib/boot-without-auth.tsx
@@ -54,10 +54,10 @@ class AppWithoutAuth extends Component<Props, State> {
     window.electron?.receive('appCommand', this.onAppCommand);
     document.body.dataset.theme = this.systemTheme;
 
-    window.electron?.receive('tokenLogin', (url) => {
+    window.electron?.receive('tokenLogin', (url: string) => {
       const { searchParams } = new URL(url);
       const simperiumToken = searchParams.get('token');
-      const email = atob(searchParams.get('email'));
+      const email = atob(searchParams.get('email') || '');
       this.tokenLogin(email, simperiumToken);
     });
   }
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [v2.23.3]
+
+### Fixes
+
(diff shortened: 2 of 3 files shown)


lib/boot-without-auth.tsx
```diff
@@ -54,10 +54,10 @@ class AppWithoutAuth extends Component<Props, State> {
     window.electron?.receive('appCommand', this.onAppCommand);
     document.body.dataset.theme = this.systemTheme;
 
-    window.electron?.receive('tokenLogin', (url) => {
+    window.electron?.receive('tokenLogin', (url: string) => {
       const { searchParams } = new URL(url);
```
codebykat, line 58: drive-by typescript fixes

roundhill (approved):
Tested well for me on Windows, Web and macOS 👍

PR #3328, opened 2025-07-22
Importer: Supporting ZIP Imports

jleandroperez:
### Fix
In this PR we're enhancing the Importer, so that users can import the `zip` files, exactly the way they've exported them.

### Test
1. Export your notes (zip) at Simplenote.com, any test account would work!
2. Setup Simplenote Electron secrets in your system (DM me if anything!)
3. Run `npm run dev`
4. Open `http://localhost:4000/`
5. Log into your account
6. Open `Settings > Tools > Import Notes`

- [ ] Verify you see a reference to Simplenote Zip files
- [ ] Verify you can import your notes!
- [ ] Verify the Unit Tests are green (`npm test`)

### Release

> Implemented support for importing ZIP files (https://github.com/Automattic/simplenote-electron/pull/3328)

### DIFF
--- a/lib/utils/import/simplenote/index.ts
+++ b/lib/utils/import/simplenote/index.ts
@@ -38,13 +38,13 @@ class SimplenoteImporter extends EventEmitter {
 
     if (endsWith(fileName, '.zip')) {
       this.processZipFile(file);
+      return;
     }
 
     this.emit('status', 'error', 'File must be a .json or .zip file.');
   };
 
   processJsonFile = (file) => {
-    const coreImporter = new CoreImporter(this.addNote);
     const fileReader = new FileReader();
 
     fileReader.onload = (event) => {
@@ -55,14 +55,16 @@ class SimplenoteImporter extends EventEmitter {
         return;
       }
 
-      this.parseAndImportJson(fileContent, coreImporter);
+      this.parseAndImportJson(fileContent);
     };
 
     fileReader.readAsText(file);
   };
 
-  parseAndImportJson = (jsonContent, coreImporter) => {
+  parseAndImportJson = (jsonContent) => {
+    const coreImporter = new CoreImporter(this.addNote);
     let dataObj;
+
     try {
       dataObj = JSON.parse(jsonContent);
     } catch (error) {
@@ -83,6 +85,53 @@ class SimplenoteImporter extends EventEmitter {
         note_count: noteCount,
       });
     });
+  };
+
+  processZipFile = (file) => {
+    const fileReader = new FileReader();
+
+    fileReader.onload = (event) => {
+      const fileContent = event.target.result;
+
+      if (!fileContent) {
+        this.emit('status', 'error', 'File was empty.');
+        return;
+      }
+
+      import(/* webpackChunkName: 'jszip' */ 'jszip')
+        .then(({ default: JSZip }) => JSZip.loadAsync(fileContent))
+        .then((zip) => {
+          // Look for JSON files in the ZIP
+          const jsonFiles = Object.keys(zip.files).filter((filename) =>
+            filename.toLowerCase().endsWith('.json')
+          );
+
+          if (jsonFiles.length === 0) {
+            this.emit('status', 'error', 'No JSON files found in ZIP archive.');
+            return;
+          }
+
+          // Process the first JSON file found
+          const jsonFile = zip.files[jsonFiles[0]];
+          return jsonFile.async('text');
+        })
+        .then((jsonContent) => {
+          if (!jsonContent) {
+            return;
+          }
+
+          this.parseAndImportJson(jsonContent);
+        })
+        .catch((error) => {
+          this.emit(
+            'status',
+            'error',
+            'Failed to process ZIP file: ' + error.message
+          );
+        });
+    };
+
+    fileReader.readAsArrayBuffer(file);
   };
 }
 
--- a/lib/utils/import/simplenote/index.ts
+++ b/lib/utils/import/simplenote/index.ts
@@ -31,10 +31,10 @@ class SimplenoteImporter extends EventEmitter {
       return;
     }
 
-    // if (endsWith(fileName, '.json')) {
-    //   this.processJsonFile(file);
-    //   return;
-    // }
+    if (endsWith(fileName, '.json')) {
+      this.processJsonFile(file);
+      return;
+    }
 
     if (endsWith(fileName, '.zip')) {
       this.processZipFile(file);
@@ -43,7 +43,7 @@ class SimplenoteImporter extends EventEmitter {
     this.emit('status', 'error', 'File must be a .json or .zip file.');
   };
 
-  processJsonFile = (file, coreImporter) => {
+  processJsonFile = (file) => {
     const coreImporter = new CoreImporter(this.addNote);
     const fileReader = new FileReader();
 
--- a/lib/utils/import/simplenote/index.ts
+++ b/lib/utils/import/simplenote/index.ts
@@ -72,6 +72,16 @@ class SimplenoteImporter extends EventEmitter {
       return;
     }
 
+    if (!dataObj.activeNotes || !Array.isArray(dataObj.activeNotes)) {
+      this.emit('status', 'error', 'Invalid Simplenote JSON format.');
+      return;
+    }
+
+    if (!dataObj.trashedNotes || !Array.isArray(dataObj.trashedNotes)) {
+      this.emit('status', 'error', 'Invalid Simplenote JSON format.');
+      return;
+    }
+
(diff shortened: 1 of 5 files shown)


lib/utils/import/simplenote/index.ts
```diff
@@ -17,26 +17,34 @@ class SimplenoteImporter extends EventEmitter {
...
+    const fileName = file.name.toLowerCase();
 
     // Limit file size we will read to 5mb
     if (file.size > 5000000) {
       this.emit('status', 'error', 'File should be less than 5 MB.');
       return;
     }
 
+    if (endsWith(fileName, '.json')) {
+      this.processJsonFile(file);
```
jleandroperez, line 35: JSON Import logic has been extracted into `processJsonFile`

lib/utils/import/simplenote/index.ts
```diff
@@ -47,31 +55,83 @@ class SimplenoteImporter extends EventEmitter {
...
+    coreImporter.importNotes(processedNotes, this.options).then(() => {
+      this.emit('status', 'complete', noteCount);
+      this.recordEvent('importer_import_completed', {
+        source: 'simplenote',
+        note_count: noteCount,
+      });
+    });
+  };
+
+  processZipFile = (file) => {
```
jleandroperez, line 95: And this would be the new processor!

lib/utils/import/simplenote/index.ts
```diff
@@ -17,26 +17,34 @@ class SimplenoteImporter extends EventEmitter {
...
     const file = filesArray[0];
-
-    if (!endsWith(file.name.toLowerCase(), '.json')) {
-      this.emit('status', 'error', 'File name does not end in ".json".');
-      return;
-    }
+    const fileName = file.name.toLowerCase();
 
     // Limit file size we will read to 5mb
     if (file.size > 5000000) {
```
dmsnell, line 29: deary me. why do we limit imports to 5 MB?

lib/utils/import/simplenote/index.ts
```diff
@@ -47,31 +55,93 @@ class SimplenoteImporter extends EventEmitter {
...
+      this.emit('status', 'error', 'Invalid JSON file.');
+      return;
+    }
+
+    if (!dataObj.activeNotes || !Array.isArray(dataObj.activeNotes)) {
+      this.emit('status', 'error', 'Invalid Simplenote JSON format.');
+      return;
+    }
+
+    if (!dataObj.trashedNotes || !Array.isArray(dataObj.trashedNotes)) {
```
dmsnell: not sure if this was here before or not, but do we know if we always get trashed notes, even when no notes have ever been trashed?

lib/utils/import/simplenote/index.ts
```diff
@@ -47,31 +55,93 @@ class SimplenoteImporter extends EventEmitter {
...
+      this.emit('status', 'error', 'Invalid Simplenote JSON format.');
+      return;
+    }
```
dmsnell: if we wanted, we could create a JSONSchema validator for the note format and then report errors in the JSON format. probably not worth it here, just ear-marking it

lib/utils/import/simplenote/index.ts
```diff
@@ -47,31 +55,93 @@ class SimplenoteImporter extends EventEmitter {
...
-        activeNotes: convertModificationDates(dataObj.activeNotes),
-        trashedNotes: convertModificationDates(dataObj.trashedNotes),
-      };
-
-      coreImporter.importNotes(processedNotes, this.options).then(() => {
-        this.emit('status', 'complete', noteCount);
-        this.recordEvent('importer_import_completed', {
-          source: 'simplenote',
-          note_count: noteCount,
+      import(/* webpackChunkName: 'jszip' */ 'jszip')
```
dmsnell, line 106: nice job lazy-loading this module. was this library already included in the project?

lib/utils/import/simplenote/index.ts
jleandroperez: Not sure, but... perhaps just as a failsafe? (nobody ever complained about this AFAIK)

lib/utils/import/simplenote/index.ts
jleandroperez: YES! there's a snippet just like this in `to-zip.js`

lib/utils/import/simplenote/index.ts
jleandroperez: Wasn't here! dropped, thank you!

lib/utils/import/simplenote/index.ts
dmsnell: coooool. can we verify that this lazy-loads and isn’t pulled in as soon as the app loads? I think we should be able to do that via the dev server and the local browser dev tools

lib/utils/import/simplenote/index.ts
dmsnell: fine to leave in the `activeNotes`. I just wasn’t sure if we would exclude valid import files simply because an account owner never trashed any notes

jleandroperez: There we go, restored just that part! 😆

lib/utils/import/simplenote/index.ts
jleandroperez: I'm **only** seeing this line after clicking `Import` in `Settings > Tools`. I believe we're good!

image

lib/utils/import/simplenote/test.ts
```diff
@@ -42,6 +50,90 @@ describe('SimplenoteImporter', () => {
...
+
+      expect(importer.processZipFile).toHaveBeenCalledWith(zipFile);
+    });
+
+    it('should emit error when ZIP file is empty in processZipFile', () => {
+      const zipFile = new File([''], 'test.zip', { type: 'application/zip' });
+      const mockFileReader = new FileReader();
+      (FileReader as any).mockImplementation(() => mockFileReader);
+
+      importer.processZipFile(zipFile);
```
dmsnell: this could be calling `importer.importNotes()` as in the previous test, but as it is, it’s reaching into an internal method making the test brittle and at risk of diverging from the actual app code

lib/utils/import/simplenote/index.ts
```diff
@@ -47,31 +55,88 @@ class SimplenoteImporter extends EventEmitter {
...
+        .then(({ default: JSZip }) => JSZip.loadAsync(fileContent))
+        .then((zip) => {
+          // Look for JSON files in the ZIP
+          const jsonFiles = Object.keys(zip.files).filter((filename) =>
+            filename.toLowerCase().endsWith('.json')
```
dmsnell: we can look specifically for `source/notes.json` since that’s what the official exporter produces.

we don’t want ZIPs with random JSON destroying accounts with junk data

lib/utils/import/simplenote/test.ts
```diff
@@ -42,6 +50,90 @@ describe('SimplenoteImporter', () => {
...
+        type: 'application/zip',
+      });
+      const mockFileReader = new FileReader();
+      mockFileReader.readAsArrayBuffer = jest.fn();
+      (FileReader as any).mockImplementation(() => mockFileReader);
+
+      importer.processZipFile(zipFile);
+
+      expect(mockFileReader.readAsArrayBuffer).toHaveBeenCalledWith(zipFile);
+    });
```
dmsnell: we really don’t care about this; it’s a junk test. we care more that the notes are imported, or even that `processJsonFile` is called when we supply a ZIP to `importNotes()`

lib/utils/import/simplenote/test.ts
```diff
@@ -42,6 +50,90 @@ describe('SimplenoteImporter', () => {
...
+
+    it('should call readAsArrayBuffer for ZIP files', () => {
+      const zipFile = new File(['content'], 'test.zip', {
+        type: 'application/zip',
+      });
+      const mockFileReader = new FileReader();
+      mockFileReader.readAsArrayBuffer = jest.fn();
+      (FileReader as any).mockImplementation(() => mockFileReader);
+
+      importer.processZipFile(zipFile);
```
dmsnell: again with the tests of fragile internal details…

dmsnell (commented):
From a pairing review session

jleandroperez:
@dmsnell Ready for another look, thank you!

dmsnell (approved):
Ready for merge and deployment. We should be ready and reactive when we do so, however, to ensure that we don’t break things accidentally.

What testing have you performed on these changes? I have performed none.

jleandroperez:
I've tested the steps outlined in the PR (importing a ZIP!).

Let's go forwards, reverting takes a sec, if anything. Thank you Dennis!!

PR #3341, opened 2025-11-14
Bump CI toolkit to version 5.7.0 to leverage new Windows AMI

mokagio:
These changes make sure we can make the most of the new Windows CI.

See https://linear.app/a8c/issue/AINFRA-1472

### DIFF
--- a/.buildkite/commands/package_windows.ps1
+++ b/.buildkite/commands/package_windows.ps1
@@ -1,7 +1,7 @@
 # Stop script execution when a non-terminating error occurs
 $ErrorActionPreference = "Stop"
 
-& "prepare_windows_host_for_app_distribution.ps1" # via CI toolkit Buildkite plugin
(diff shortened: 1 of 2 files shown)


.buildkite/commands/package_windows.ps1
```diff
@@ -1,7 +1,7 @@
 # Stop script execution when a non-terminating error occurs
 $ErrorActionPreference = "Stop"
 
-& "prepare_windows_host_for_app_distribution.ps1" # via CI toolkit Buildkite plugin
+& "setup_windows_code_signing.ps1" # via CI toolkit Buildkite plugin
```
mokagio, line 4: Given the new CI image for Windows comes with the main dev settings and tools already installed, we can simply call the script to set up code signing and move on.

AliSoftware (approved):
:shipit:

Issue #3329, opened 2025-07-27
Paste (ctrl-v) not working on webapp with Windows/Firefox


MikeRaffa:
Keyboard shortcut ctrl-v for paste not working on webapp with Windows/Firefox.

### Expected
Expected ctrl-v to paste the text from the clipboard.

### Observed
Ctrl-v does not appear to do anything. Ctrl-shift-v and shift-insert do work for pasting.

### Reproduced
1. Go to https://app.simplenote.com/ and log in if necessary.
2. Copy some text with ctrl-c, either from within simplenote, within another webpage in Firefox, or from another desktop app.
3. In the body of a note in simplenote do the ctrl-v keyboard shortcut (no result seen). 

### Where did you see the bug
- System Make: Windows 11 desktop
- System Model: Windows 11 desktop
- OS: Windows 11 Pro
- OS version: 24H2, build 26100.4652
- Browser (if applicable): Firefox
- Browser version (if applicable): 141.0 
- Simplenote app version: 2.23.2

surfer0627:
I’m experiencing the same issue.

emilyofficial:
Also experiencing this issue, for any simplenote employees seeing this, please know this isn't a minor issue, this is a major usability issue affecting a lot of people, there's an entire thread about it on the forums (https://forums.simplenote.com/forums/topic/ctrl-v-not-working/) and has been going on for nearly a week, so it may be worth bumping up the priority of a fix on this. Using ctrl-shift-v does work as a temporary fix but not a lot of people are aware of that. This is the error I see in the console.

‘Error: command ‘editor.action.clipboardPasteAction’ not found’

ghost:
As a workaround until they find the time to fix it, go to [Simplenote](https://app.simplenote.com/), press `Ctrl`+`i`  → `Permissions`  → uncheck `Override keyboard shortcuts` and block it:

<img width="816" height="407" alt="Screenshot" src="https://github.com/user-attachments/assets/5f303ca6-36f6-4d71-8c95-a2068b7881fe" />

michaelwexler:
The "blocking" of keyboard shortcuts will work, but you'll lose Ctrl (Cmd) Z to undo, and potentially other commands as well.  Shift-Ctrl (Cmd) V appears to work without having to change that setting, as a workaround.

michaelwexler:
In my Firefox, Ctrl (cmd) V throws the below error (FF 141.0 on MacOS Seq 15.5):

```
[...]

PR #3349, opened 2026-02-22
Fix Firefox web Ctrl+V paste regression

Epikaigle:
### Fix

Fixes Firefox web paste regressions in the note editor (`#3329`).

After `#3316`, Simplenote re-added Monaco clipboard keybindings in the browser and continued using Monaco’s context menu there. In Firefox web mode, those Monaco clipboard paths are not reliable.

This PR keeps the existing Monaco clipboard behavior for Electron and non-Firefox browsers, but for Firefox web it now:
- skips the Monaco paste shortcut so Firefox handles `Ctrl+V` natively
- disables the Monaco editor context menu so Firefox uses its native clipboard context menu instead

That moves Firefox web back to the browser-native clipboard flow while keeping the current behavior elsewhere.

### Test

1. In Firefox web, focus the note editor and verify `Ctrl+V` pastes using the browser-native clipboard flow.
2. In Firefox web, right-click in the editor and verify the browser-native context menu is used instead of Monaco’s context menu.
3. In Chromium/Electron flows, verify cut/copy/paste shortcuts still work.
4. Run:
   - `npx eslint lib/note-content-editor.tsx lib/utils/platform.ts lib/utils/platform.test.ts`
   - `npx jest --config=./jest.config.js lib/utils/platform.test.ts`
   - `make build`

### Release

> Fixed Firefox web paste to use the browser's native clipboard shortcuts and context menu

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [v2.24.1]
+
+### Fixes
+
+- Fixed Firefox web paste to use the browser's native clipboard shortcuts and context menu [#3349](https://github.com/Automattic/simplenote-electron/pull/3349)
+
 ## [v2.24.0]
 
 ### New Features
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -26,7 +26,7 @@ import {
   getNotePosition,
   setNotePosition,
 } from './utils/note-scroll-position';
-import { isMac, isSafari } from './utils/platform';
+import { useMonacoPasteShortcut } from './utils/platform';
 import {
   withCheckboxCharacters,
   withCheckboxSyntax,
@@ -729,8 +729,9 @@ class NoteContentEditor extends Component<Props> {
       },
     });
 
-    // re-add keybindings for cut/copy/paste so they show labels
-    monaco.editor.addKeybindingRules([
+    // Firefox fails to resolve `editor.action.clipboardPasteAction` in web mode.
+    // Let the browser handle Ctrl+V natively there.
+    const clipboardKeybindingRules = [
       {
         keybinding: monaco.KeyMod.CtrlCmd | monaco.KeyCode.KeyX,
         command: 'editor.action.clipboardCutAction',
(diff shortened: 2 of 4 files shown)


mdawaffe (changes requested):
@Epikaigle, @thabotswana - neither #3349 (this PR) nor #3350 work for me in Firefox 148.0.2.

The only thing that works for me is to just never do anything for Firefox :) For this PR (#3349), that means always returning `false` from `canUseMonacoPasteShortcut()` for Firefox regardless of whether the clipboard API or the paste command is supported.

That browser detection isn't very satisfying, but it's the simplest thing I found that worked. Note that it doesn't help the multiple-paste bug when pasting from the context menu (https://github.com/Automattic/simplenote-electron/issues/3329#issuecomment-4059671969)

Do either of you have any other thoughts?

Epikaigle:
@mdawaffe I updated this PR to take the simpler Firefox-specific approach you suggested.

For Firefox web, `canUseMonacoPasteShortcut()` now always returns `false`, so `Ctrl+V` is handled natively by the browser there regardless of clipboard API / paste command support.

I also disabled the Monaco editor context menu for Firefox web, so right-click paste uses Firefox's native context menu as well. Electron and non-Firefox browsers keep the existing Monaco clipboard behavior.

The current changes are in:
- `lib/utils/platform.ts`
- `lib/note-content-editor.tsx`
- `lib/utils/platform.test.ts`

I reran:
- `npx eslint lib/note-content-editor.tsx lib/utils/platform.ts lib/utils/platform.test.ts`
- `npx jest --config=./jest.config.js lib/utils/platform.test.ts`
- `make build`

I also updated the PR description to match the current behavior.

This is intended to fix the Firefox web clipboard path in the editor. The separate multiple-paste/context-menu behavior mentioned in `#3329` may still need confirmation from your Firefox 148.0.2 setup, although this branch now also avoids Monaco's context menu on Firefox web.

Epikaigle:
The original issue addressed here was caused by routing Ctrl+V through Monaco's paste action in Firefox. Since #3358, paste handling has been reworked so browser paste events are used directly and the Monaco Ctrl+V paste keybinding is no longer re-added. That makes the Firefox-specific workaround in this PR obsolete against current trunk.

PR #3358, opened 2026-05-13
Allow copying from history view; copy rich-text and plaintext MIME types.

### LINKED ISSUE #2898: Improve the accessibility of note History dialog
### What
When the history dialog is opened, and the note is big enough, the dialog will cover a part of the note in this way:

https://user-images.githubusercontent.com/73365754/117119256-771ead00-ad9a-11eb-8721-16e77167551a.mov

To see the changes in the covered part, the only available method is scrolling. In a case when scrolling is impossible, there is no possibility to see the changes in this part of the note, because the dialog is closed on click-out, and because the focus is trapped inside the dialog.

### Why
Improve the accessibility of the app.

### [How](p1620139812028900-slack-C01KY08KQSY)

> 1. A small win would be to disable the click outside from closing the history slider, but that introduces other oddities we’d need to address as well like you wouldn’t be able to click on another note in the note list or any menu items until you cancel the note history, but it looks like you should be able to.

> 2. Currently, when the revision slider value changes, the revision date is announced. We could announce additional information if it is deemed helpful. It should likely be explored/debated whether announcing the entire selected note revision or a "diff" of the revision is a helpful announcement. Either one could potentially be a lot of text and overwhelming. However, that may be the "norm" or expected as a screen reader user — I'm not sure. We'd likely need to build it and play with it to get a sense for how helpful it is.

> 3. As you mentioned, we could stop hiding the Note Editor UI and allow the user to navigate into that UI while the Revision Selector is open. This would allow the user to "read" the entire note, and scroll if needed. We would need to figure out a way to focus trap the Note Editor + Revision Selector, which may require restructuring HTML or additional JavaScript.

> 4. If we ultimately allow navigating into the Note Editor, we likely should no longer declare the Revision Selector a dialog, but something else (i.e. change role="dialog"). Dialogs are generally defined as UI that overlays all other UI preventing access to anything else.

### LINKED ISSUE #3042: [Accessibility] Don't close `History` feature on outside click.
### What
It came to my mind while working in the Simplenote queue that clicking outside the history area in order to close it might be a major inconvenience for people who don't use/have a mouse wheel or a possibility to use touchpad/trackpad gestures in order to scroll.

### Why
If the note is long, the user might need to scroll the note while History is open, in order to see the whole content. If they do this with click-and-drag the scroll slider, or clicking the up and down buttons on the sides of the scrollbar, it's practically impossible.

### How
Not closing History if click outside it occurs. Have a dedicated close button for it. If this will be done, please let me know, I'll create a sibling for native macOS.

dmsnell:
Resolves #3042 
Resolves #2898

## Summary

This 100% vibe-coded change addresses two frustrations with copying in the editor:

 1. It’s not possible to copy content from a revision because clicking outside of the revision slider closes the history view.
 2. Copying and pasting is rich-text unaware and only copies Markdown source as plaintext.

These changes fix the history viewer and require an explicit click to close (or a keypress of the `Escape` key, or by navigating to another note). This is a different behavior than the click-outside-to-close model, but it may be more ergonomic, because the old behavior prevents copying parts of an old version of a post.

Secondarily, several issues arose during the work on this feature relating to copy and paste behaviors. The app should be more aware of HTML/richly-formatted content when pasting, and will send _two_ copies of copied content into the paste buffer: a `text/plain` version containing the markup source; and a `text/html` version containing the rendered note. This makes it possible for other applications to use the rendered content after copying.

A side-effect of this change is that it’s now possible to preserve basic formatting when pasting HTML content into the app, as it parses the inbound HTML, sanitizes it, and creates Markdown syntax from the HTML.

## Before / After

These videos demonstrate both aspects of this change:

 1. They attempt to select and copy a portion of an old revision of a note.
 2. They copy a note’s content and then paste it into another app, showing how the copied content, even if copied from the Markdown source itself, retains its formatting through the copy.

As a note on this, most applications provide an alternative paste mode where it’s still possible to paste the plaintext content. In the _After_ video, the plaintext content is automatically pasted into the search box since it doesn’t support rich content, but in something like an email, it would likely be possible with a combination like Cmd + Opt + V or “Paste without formatting.”

https://github.com/user-attachments/assets/0ddd166b-ef47-4d6f-968b-ad0747f7940d

https://github.com/user-attachments/assets/4db160af-e0dd-4084-9415-60d2e3f727ce

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [v2.25.0]
+
+### New Features
+
+- Added support for selecting and copying text from older versions of notes in History [#3358](https://github.com/Automattic/simplenote-electron/pull/3358)
+
+### Enhancements
+
+- Improved Markdown copy and paste support so copied notes include rich HTML for external apps and pasted HTML is converted into readable Markdown [#3358](https://github.com/Automattic/simplenote-electron/pull/3358)
+
 ## [v2.24.0]
 
 ### New Features
--- a/lib/app-layout/index.tsx
+++ b/lib/app-layout/index.tsx
@@ -4,7 +4,6 @@ import { connect } from 'react-redux';
 
 import MenuBar from '../menu-bar';
 import NoteToolbar from '../note-toolbar';
-import RevisionSelector from '../revision-selector';
 import SearchField from '../search-field';
 import SimplenoteCompactLogo from '../icons/simplenote-compact';
 import NoteRevisions from '../note-revisions';
@@ -29,7 +28,6 @@ const NotePreview = React.lazy(
 );
 
 type StateProps = {
-  hasRevisions: boolean;
   isFocusMode: boolean;
   isNavigationOpen: boolean;
   isNoteInfoOpen: boolean;
@@ -86,7 +84,6 @@ export class AppLayout extends Component<Props> {
   render = () => {
     const {
       showNoteList,
-      hasRevisions,
       isFocusMode = false,
       isNavigationOpen,
       isNoteInfoOpen,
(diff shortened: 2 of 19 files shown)


dmsnell:
I have performed checks on this and ran a (primitive, overly-naive) fuzzer on it to ensure that the changes aren’t introducing obvious flaws. My plan is to merge, try to deploy to staging, then deploy to production.

PR #3357, opened 2026-05-07
[codex] Add source-visible Markdown decorations

mzorz:
## Summary

Adds light source-visible Markdown formatting in the Monaco note editor.

Changes:
- Adds a Markdown decoration helper for headings, `**bold**`, `__bold__`, `*italic*`, `_italic_`, `***bold italic***`, `___bold italic___`, `~~strikethrough~~`, and `-` / `*` / `+` unordered-list markers, including indented list markers.
- Decorates inline bold, italic, bold-italic, and strikethrough spans inside unordered list item text while keeping the list marker decoration intact.
- Aligns asterisk and underscore inline delimiter matching more closely with the existing Showdown renderer settings, including literal mid-word underscore behavior.
- Skips Markdown inline decorations inside URL spans, inline code spans, and fenced code blocks so the editor does not preview formatting that the rendered Markdown would keep literal/code-formatted.
- Wires the helper into the existing Monaco decoration lifecycle with an IME-composition guard.
- Adds editor styles for heading levels, italic spans, bold spans, strikethrough spans, and list markers.
- Adds unit coverage for heading, list marker, inline formatting inside list items, italic, underscore italic, underscore bold, combined bold/italic, bold, strikethrough, escaped marker, whitespace-padded marker, underscore-in-word, URL-skip, inline-code-skip, and fenced-code-skip behavior.
- Keeps raw Markdown text unchanged; this is decoration-only source-visible formatting.

## Context

This is for Multica issue LAV-24: Desktop Command Palette and Markdown. It targets `trunk` directly and is intentionally separate from the command palette PR (#3360). The older Cmd+K-only search shortcut PR (#3356) was closed per direction.

Video:

https://github.com/user-attachments/assets/03529b95-7271-48c3-812e-1a5f1c57626e

## Testing

- `npx jest lib/markdown-decorations.test.ts --runInBand`
- `make test`
- `make lint-js` (passes with existing warnings)
- `make lint-scss`
- `git diff --check`
- `make build NODE_ENV=development`

PR #3356, opened 2026-05-07
[codex] Add Cmd+K global search shortcut

mzorz:
## Summary

Adds `Cmd+K` / `Ctrl+K` as another way to focus the existing global search field.

Changes:
- Handles `Cmd/Ctrl+K` in the existing browser capture-phase search shortcut handler.
- Adds an Electron Edit menu accelerator that dispatches the same `focusSearchField` action.
- Documents the shortcut in the keyboard shortcuts dialog.

Linear: https://linear.app/a8c/issue/SIMPL-48/add-cmdk-ctrlk-as-an-alias-for-global-search

## Testing

- `npx eslint lib/app.tsx desktop/menus/edit-menu.js lib/dialogs/keybindings/index.tsx`

Note: targeted lint passes with one existing `no-shadow` warning in `lib/dialogs/keybindings/index.tsx`.

PR #3360, opened 2026-05-20
[codex] Add Cmd+K command palette modal

mzorz:
## Summary

- Adds a command palette modal that opens from the note editor with `Cmd+K` / `Ctrl+K`.
- Removes the previous `/` detection while editing: typing `/` in note content now stays plain note text and does not open or control the palette.
- Keeps palette search state outside Monaco, with a focused modal input, arrow-key selection, Enter execution, Escape dismissal, mouse selection, and scroll-into-view for the selected command.
- Reuses existing app actions for command execution instead of adding new parallel flows.

## Commands Included

- Search: `search`, `s`
- Keyboard shortcuts: `shortcuts`
- Focus mode: `focus`
- Settings / preferences: `settings`, `pref`, `preferences`
- Export: `export`
- New note: `new`, `create`
- All notes: `all`, `allnotes`
- Untagged notes: `untagged`, `untaggednotes`
- Trash: `trash`
- Help & Support: `help`, `support`
- About: `about`

## Context

This is for Multica issue LAV-24: Desktop Command Palette and Markdown. It is intentionally separate from the Markdown decorations PR (#3357), which targets `trunk` directly. The older Cmd+K-only search shortcut PR (#3356) was closed per direction, and Cmd/Ctrl+K now opens this command palette modal.

## Testing

- `npx jest lib/slash-commands.test.ts --runInBand`
- `make test`
- `make lint-js` (passes with existing warnings)
- `make lint-scss`
- `git diff --check`
- `make build NODE_ENV=development`
- `make dev-server PORT=4010` + `curl -I http://localhost:4010` returned HTTP 200

### DIFF
--- a/lib/dialogs/keybindings/index.tsx
+++ b/lib/dialogs/keybindings/index.tsx
@@ -52,6 +52,9 @@ export class AboutDialog extends Component<DispatchProps> {
                   <Keys keys={[CmdOrCtrl, '/']}>Show keyboard shortcuts</Keys>
                 </li>
                 <li>
+                  <Keys keys={[CmdOrCtrl, 'K']}>Show command palette</Keys>
+                </li>
+                <li>
                   <Keys keys={[CmdOrCtrl, 'Shift', 'F']}>
                     Toggle focus mode
                   </Keys>
--- a/lib/note-content-editor.tsx
+++ b/lib/note-content-editor.tsx
@@ -46,6 +46,7 @@ import {
   getNotePosition,
   setNotePosition,
 } from './utils/note-scroll-position';
+import { viewExternalUrl } from './utils/url-utils';
 import { isMac, isSafari } from './utils/platform';
 import {
   withCheckboxCharacters,
@@ -137,13 +138,18 @@ type StateProps = {
 
 type DispatchProps = {
   clearSearch: () => any;
+  createNote: () => any;
   editNote: (noteId: T.EntityId, changes: Partial<T.Note>) => any;
   exportNotes: () => any;
   focusSearchField: () => any;
   insertTask: () => any;
   openNote: (noteId: T.EntityId) => any;
+  selectTrash: () => any;
+  showAbout: () => any;
+  showAllNotes: () => any;
   showKeyboardShortcuts: () => any;
   showPreferences: () => any;
+  showUntaggedNotes: () => any;
   storeEditorSelection: (
     noteId: T.EntityId,
     start: number,
@@ -860,17 +866,35 @@ class NoteContentEditor extends Component<Props> {
     this.clearSlashCommand();
 
     switch (command) {
+      case 'new':
+        this.props.createNote();
+        return;
       case 'search':
         this.props.focusSearchField();
+        return;
+      case 'all-notes':
+        this.props.showAllNotes();
+        return;
+      case 'untagged-notes':
+        this.props.showUntaggedNotes();
+        return;
+      case 'trash':
+        this.props.selectTrash();
+        return;
+      case 'settings':
+        this.props.showPreferences();
         return;
       case 'shortcuts':
         this.props.showKeyboardShortcuts();
         return;
+      case 'help':
+        viewExternalUrl('http://simplenote.com/help');
+        return;
+      case 'about':
+        this.props.showAbout();
+        return;
       case 'focus':
         this.props.toggleFocusMode();
-        return;
-      case 'pref':
-        this.props.showPreferences();
         return;
       case 'export':
         this.props.exportNotes();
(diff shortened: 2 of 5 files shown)


mzorz:
Addressed reviewer feedback in `ea0a463` and `9c4bd06`:

- Removed the dead Monaco assignment/property.
- Kept Tab completion in the keydown path and removed the duplicate addCommand branch.
- Removed the redundant post-Tab microtask update.
- Measures the rendered slash-command palette for vertical placement instead of carrying the mismatched 190px constant.
- Added controller-level tests for key handling and selection wrapping.
- Fixed the caret-inside-token edge case by only activating slash commands at the end of the token.

Revalidated with `npx jest lib/slash-commands.test.ts --runInBand`, `make test`, `make lint-js`, `make lint-scss`, `git diff --check`, and `make build NODE_ENV=development`.

mzorz:
Multica reviewer-agent feedback copied here for the PR record.

Reviewer 1 (Claude) reviewed the original slash-command palette implementation and confirmed the requested `/command` behavior worked: Tab completes, Enter executes and removes the token, Space leaves literal text, Escape cancels, arrows navigate, ghost text appears on the right, and commands resolve to existing app actions.

Should-fix findings from that review:

1. Remove the unused `this.monaco = monaco` assignment.
2. Avoid handling Tab in two separate paths; keep only one completion path.
3. Remove the redundant `Promise.resolve().then(this.updateSlashCommand)` after Tab completion.
4. Avoid carrying a hardcoded palette height that can drift from CSS; measure the rendered popup.
5. Add coverage for the higher-risk key handling and state transitions around completion, execution, cancellation, selection, and token cleanup.

Reviewer 2 (Codex) found one actionable issue: if the caret sat inside an existing slash-command token, the trigger range only covered text up to the cursor, so completion/execution could leave the remainder of the token behind. The trigger should activate only at the end of a token or replace the full token range.

Resolution:

- The should-fix items and caret-inside-token bug were addressed in follow-up commits `ea0a463` and `9c4bd06`.
- I also addressed the remaining practical accessibility note in `78e1276` by wiring the Monaco input's `aria-activedescendant` / `aria-controls` to the selected slash-command option.
- Remaining lower-priority notes are intentionally not blocking this PR: title-substring matching is kept for discoverability in the expanded command set, telemetry can be added in a follow-up if there is a preferred event path for editor-only command execution.

Revalidated after the latest change with `npx jest lib/slash-commands.test.ts --runInBand`, `make test`, `make lint-js` (existing warnings only), `make lint-scss`, `git diff --check`, and `make build NODE_ENV=development`.

dmsnell:
What’s the rationale for taking over the `/` and adding slash commands? it seems like none of them relate to editing activity, so unlike Gutenberg where slash commands are part of the editing sequence, all of these slash commands take the author _out_ of the editing flow, and there’s no way to go back (for example, hitting the tab key does not return the cursor to the editing canvas).

For some of these like `/search` I would expect to be able to continue typing to create the search query, such as `/search pizza`, but instead of searching for pizza that instead leaves the slashcode as literal text.

With <kbd>⌘</kbd>+<kbd>k</kbd> this idiom makes more sense, but then it’s very awkward that it inserts its pop-up dialog wherever the cursor is, and only if the cursor is in a spot where the flanking characters allow it. this is confusing since the shortcut feels like it should be a kind of global action. but in fact, if the cursor is anywhere else, like the search bar or the triple-dot menu, then pressing <kbd>⌘</kbd>+<kbd>k</kbd> surprisingly starts making an edit to the note and synchronizes it with the server.

https://github.com/user-attachments/assets/c86594a9-5626-4ca7-b0ad-03863cc524a8

---

What are your thoughts on this? I think that the mechanism in this PR leaves a high risk of confusing customers, “corrupting” notes because of accidental shortcut keypresses when people expected the global action to not edit their notes, and of muddying a simpler UX with the dialog appearing at the cursor.

The heart of the command palette work is good, but I wonder if these changes could make this less confusing:

 - skip the slashcodes since none of them relate to editing (the _one_ slashcode that might relate to editing is inserting a link to another post, and that’s not in this PR).
 - make the command palette an app-wide modal more akin to Alfred or Raycast or Spotlight. remove all of the complexity from the editor instance (Monaco) and decouple editing a note from navigating the app.
 - give space for appropriate commands to type additional context, e.g. "search pizza" or "tag books". this amounts to an optional input field when running search that would provide an auto-complete list, again like other command palettes do, which when selected would convey in the app the appropriate action (select this tag, search for this query, create a new note called this…)

mzorz:
Oh I see what you mean, the entry point being the editing area is what is confusing and flawed - and among other things what you type there becomes part of the editing history. Thank you for pointing that out! 

I agree having a separate context / field (reusing the search field for example) to start from there could be a great option so we don't mess up with the edit history and set clear expectations and separation of concerns ("this area is for commands, app operation", "this other area is where your writing flows").

I'd like to explore moving all of that to only the search (so, in the search, entering / as the first character would trigger the palette), as a first iteration only.

dmsnell:
> in the search, entering / as the first character would trigger the palette

this feels like it’s moving the confusion to the search bar. how then does someone disambiguate searching for `/all` in a note and triggering the `/all` command

maybe a good start is _not_ having the additional input and just having the command palette working in a separate area from an existing editing surface

mzorz:
Thanks for the clarifying conversation on Slack @dmsnell 

For visibility, we'll remove any editing parts and keep cmd+k to trigger the command palette modal as a first iteration and take it from there.

mzorz:
this is how it works now after the update in [dcb82b7](https://github.com/Automattic/simplenote-electron/pull/3360/commits/dcb82b72183ef34bcf0acedca724f23bed3934ad) 

https://github.com/user-attachments/assets/4f448ed6-5fdc-48a6-9084-d511bcb7cb87

dmsnell:
This seems much better. The code is full of messes, but this is the age of AI. We can toss this to the future.

PR #3374, opened 2026-05-30
Clear critical dependency alerts

lancewillett:
## Summary
- Refreshes package-lock.json to clear the npm critical alerts for pbkdf2, sha.js, cipher-base, and form-data.
- Updates fastlane-plugin-wpmreleasetoolkit from the 11.x line to the 12.x line so Bundler can resolve Nokogiri above the patched floor.
- Refreshes Gemfile.lock to use Nokogiri 1.19.3 for the existing macOS platforms.

## Validation
- npm ci --legacy-peer-deps
- npm audit --audit-level=critical reports critical: 0; non-critical findings remain.
- npm run lint passed with existing warnings.
- npm test passed: 21 suites, 151 tests.
- npm run build passed with existing bundle-size warnings.
- Ruby 3.2.2 container: bundle install passed.
- Ruby 3.2.2 container: fastlane action configure_apply loaded successfully.
- git diff --check

## Reviewer signal
Requesting @mokagio because they handled the recent Simplenote release merges and the latest Fastlane maintenance PR, and they have admin permission on this repo.


mokagio (commented):
Thanks @lancewillett !

For the record, I run this past Opus 4.7 to confirm that the Node modules updates where compatible with a current limitation in the app with installing certain legacy dependencies. The answer was positive, which is consistent with the validation steps in the PR description.

<details>
<summary>Full anwser</summary>

```
  What the PR actually is: #3374 "Clear critical dependency
  alerts" (fix/drain-critical-dependabot-alerts → trunk). It
  touches only Gemfile, Gemfile.lock, and package-lock.json.
  package.json is not changed — no declared dependency is added
  or removed. The lockfile diff is purely transitive version
  bumps (@babel/*, plus pbkdf2/sha.js/cipher-base/form-data per
  the body) to drain the critical alerts.

  The --legacy-peer-deps need is documented, in three places:

  - README.md step 3 — npm install --legacy-peer-deps
  - .buildkite/commands/install_node_dependencies.sh — npm ci
  --legacy-peer-deps
  - bin/deploy.sh:40 — npm install --legacy-peer-deps

  Why it's needed is not "deps are stale in general" — it's
  specific: react-monaco-editor pins an exact monaco-editor
  version and npm treats sub-1.0.0 minor bumps as breaking, so
  strict peer resolution fails. (That's the README's stated
  reason.)

  Why there's no conflict:

  - --legacy-peer-deps is an install-time resolver relaxation. It
  doesn't live in the lockfile, so refreshing the lock can't
  contradict it.
  - The PR doesn't touch package.json or react-monaco-editor, so
  the peer-dep situation that requires the flag is unchanged.
  - The PR's own validation used npm ci --legacy-peer-deps, i.e.
  the regenerated lock was produced and verified under the same
  flag CI/deploy use. It's consistent end to end.

  One thing to keep in mind (not introduced by this PR, but
  relevant): there's no .npmrc, so the flag isn't persisted —
  every entry point has to pass it explicitly. CI, deploy, and
  the README all do. A bare npm ci/npm install would still fail
  regardless of this PR. If you want to harden that, adding
  legacy-peer-deps=true to a committed .npmrc would make every
  install path correct by default — but that's a separate
  cleanup, not a blocker here.

  So: the npm change is a lockfile-only security refresh that
  rides on the existing --legacy-peer-deps workflow rather than
  fighting it.
```

</details>

PR #3380, opened 2026-06-08
Add Azure Trusted Signing as an opt-in Windows signing path

mokagio:
Part of [AINFRA-2472](https://linear.app/a8c/issue/AINFRA-2472).

Prepares the app for Azure code signing, which we'll enable in July, while also preloading the CN for the Azure identity so that PFX -> Azure upgrades (for builds installed outside of the Microsoft Store) will still work.

Used `CSC_FOR_PULL_REQUEST` to verify the builds worked, see https://buildkite.com/automattic/simplenote-electron/builds/426, then removed in commit [d601290](https://github.com/Automattic/simplenote-electron/pull/3380/commits/d6012909ca67847fb491ca01709c1f80de87f588). Then, I realized that all the other steps still have `CSC_FOR_PULL_REQUEST: true` so I don't know how useful that was...

<details>
<summary>Additional AI-generated details</summary>

### Fix

Add Azure Trusted Signing as an **opt-in** Windows signing path, gated behind
`USE_AZURE_TRUSTED_SIGNING`, while the **PFX cert stays the default**. Part of the
[Azure Trusted Signing migration](https://linear.app/a8c/project/replace-our-windows-code-signing-certificate-6a4cc099fda6) (the Sectigo PFX cert expires 2026-07-05).

- The default build signs the NSIS exe via the existing PFX cert
  (`certificateSubjectName`) — unchanged behaviour.
- With `USE_AZURE_TRUSTED_SIGNING` set, the exe signs via Azure Trusted Signing
  through a `win.sign` callback (`scripts/azure-sign.cjs`); the Store AppX stays
  unsigned (its config is untouched).
- The pipeline runs both: the default **Package on Windows** (PFX, ships) and an
  opt-in **Package on Windows (Azure Trusted Signing)** that signs and gates on
  `signtool verify /pa`.
- CI Toolkit plugin `5.7.0 → 6.1.0` (adds `setup_azure_trusted_signing.ps1`,
  keeps the PFX provisioner — see commit bodies for the per-tag audit).

⚠️ **Why Azure is opt-in, not the default (AINFRA-2472):** Azure changes the
Authenticode publisher CN (`Automattic, Inc.` → `Automattic Inc.`), which breaks
`electron-updater` for existing users. A PFX-signed *bridge* release carrying the
dual `publisherName` (pinned here) must ship before any Azure-only build.

### Test

1. Both Windows steps are green on the build. The Azure step signs and verifies
   with `signtool verify /pa`; I downloaded its `Simplenote-win-2.26.0-x64.exe`
   and confirmed on macOS with `osslsigncode`: SHA256, signer `Automattic Inc.`,
   issuer `Microsoft ID Verified CS EOC CA 03`, Microsoft RFC3161 timestamp.
2. `make test` covers the new `scripts/azure-sign.test.js`.

### Release

---

*Opened by Claude (Opus 4.8) on behalf of @mokagio with approval.*

</details>

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -87,6 +87,26 @@ steps:
       - release\*.blockmap
       - release\*.yml
 
+  # Validates the Azure Trusted Signing path alongside the default PFX build above. The PFX build
+  # remains what ships until the auto-update publisher-name handover lands (AINFRA-2472); this step
+  # signs the exe with Azure and verifies it with `signtool verify /pa`.
+  - label: Package on Windows (Azure Trusted Signing)
+    key: package-windows-azure
+    agents:
+      queue: windows
+    plugins:
+      - $CI_TOOLKIT_PLUGIN
+      - $NVM_PLUGIN
+    command: .buildkite/commands/package_windows.ps1
+    env:
+      CSC_FOR_PULL_REQUEST: true
+      USE_AZURE_TRUSTED_SIGNING: 1
+    artifact_paths:
+      - release\*.exe
+      - release\*.appx
+      - release\*.blockmap
+      - release\*.yml
+
   - label: Package on Linux
     key: package-linux
     plugins:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -99,7 +99,6 @@ steps:
       - $NVM_PLUGIN
     command: .buildkite/commands/package_windows.ps1
     env:
-      CSC_FOR_PULL_REQUEST: true
       USE_AZURE_TRUSTED_SIGNING: 1
     artifact_paths:
       - release\*.exe
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -48,8 +48,7 @@
   },
   "win": {
     "icon": "resources/images/simplenote.ico",
-    "sign": "./scripts/azure-sign.cjs",
-    "signingHashAlgorithms": ["sha256"],
+    "certificateSubjectName": "Automattic, Inc.",
     "publisherName": ["Automattic, Inc.", "Automattic Inc."],
     "artifactName": "Simplenote-win-${version}-${arch}.${ext}",
     "target": [
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -50,6 +50,7 @@
     "icon": "resources/images/simplenote.ico",
     "sign": "./scripts/azure-sign.cjs",
     "signingHashAlgorithms": ["sha256"],
+    "publisherName": ["Automattic, Inc.", "Automattic Inc."],
     "artifactName": "Simplenote-win-${version}-${arch}.${ext}",
     "target": [
       {
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -48,7 +48,8 @@
   },
   "win": {
     "icon": "resources/images/simplenote.ico",
-    "certificateSubjectName": "Automattic, Inc.",
+    "sign": "./scripts/azure-sign.cjs",
+    "signingHashAlgorithms": ["sha256"],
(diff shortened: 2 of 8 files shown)


electron-builder.json
```diff
@@ -49,6 +49,7 @@
   "win": {
     "icon": "resources/images/simplenote.ico",
     "certificateSubjectName": "Automattic, Inc.",
+    "publisherName": ["Automattic, Inc.", "Automattic Inc."],
```
mokagio, line 52: Crucial to ensure auto-upgrade continues to work from a PFX-signed to an Azure-signed build.

Caveat: Users need builds with this setting for the auto-upgrade to work. See also https://github.com/Automattic/wp-calypso/pull/111456

.buildkite/pipeline.yml
```diff
@@ -87,6 +87,26 @@ steps:
...
+  - label: Package on Windows (Azure Trusted Signing)
+    key: package-windows-azure
+    agents:
+      queue: windows
+    plugins:
+      - $CI_TOOLKIT_PLUGIN
+      - $NVM_PLUGIN
+    command: .buildkite/commands/package_windows.ps1
+    env:
+      CSC_FOR_PULL_REQUEST: true
```
mokagio:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -87,6 +87,25 @@ steps:
...
+  - label: Package on Windows (Azure Trusted Signing)
+    key: package-windows-azure
+    agents:
+      queue: windows
+    plugins:
+      - $CI_TOOLKIT_PLUGIN
+      - $NVM_PLUGIN
+    command: .buildkite/commands/package_windows.ps1
+    env:
+      USE_AZURE_TRUSTED_SIGNING: 1
```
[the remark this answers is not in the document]
mokagio: Fair but unnecessary. See also https://linear.app/a8c/issue/AINFRA-2473

iangmaia, line 102: To make sure I follow:

1. This script always runs `make package-win32 SKIP_BUILD=true`
2. Then the Electron Builder wil run with: `npx electron-builder --win -p $(PUBLISH)`
3. `PUBLISH` is set by the `Makefile`:
```
     ifeq ($(strip $(BUILDKITE_TAG)),)
         PUBLISH ?= never
     else
         PUBLISH ?= always
     endif
```

So on a tag build the `BUILDKITE_TAG` is set and both Windows steps publish with `-p` BOTH the intended PFX-signed release artifacts AND the validation-only artifacts?

So maybe a fix could be something like:

```suggestion
    env:
      USE_AZURE_TRUSTED_SIGNING: 1
      PUBLISH: never
```

scripts/azure-sign.cjs
```diff
@@ -0,0 +1,151 @@
...
+      nonBlank(env.SIGNTOOL_PATH),
+      buildAzureArgs(file, env),
+      file,
+      'Azure Trusted Signing'
+    );
+    return;
+  }
+
+  const pfx = pfxInputs(env);
+  if (env.CI && process.platform === 'win32') {
```
iangmaia, line 122: If I read it correctly, if `setup_azure_trusted_signing.ps1` exits successfully but doesn't export one of the expected env vars, the signing won't fail Azure validation and will sign with the PFX certificate instead?

iangmaia (approved):
A couple of remarks / questions, but otherwise it looks solid 👍

iangmaia:
Ah, didn't realize auto-merge was on. I'll leave it to you @mokagio on following up on the points I mentioned (I can review it quickly if you create a new PR based in what is done here).

mokagio:
Thanks @iangmaia !

> 'll leave it to you @mokagio on following up on the points I mentioned (I can review it quickly if you create a new PR based in what is done here).

See #3393

PR #3382, opened 2026-06-10
Upgrade to React 18

Usiel:
### Fix

This is required for future work that potentially integrates Lexical for live markdown editing. I thought it may be nicer to split out the React upgrade to a separate PR, since it will be useful regardless of other downstream changes.

React 18 [retires the ReactDOM.render API](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis). Apps must create a root via createRoot, and React warns when createRoot is called repeatedly on the same container. Our boot phases (platform warning, auth screen, main app, logging out) each rendered into #root, so they now share a single root (lib/app-root.ts) and replace each other via root.render().

React 18 also drops support for findDOMNode, which two of our dependencies rely on:

- react-virtualized is replaced with the react-virtualized-compat fork which is compatible with React 18. At some point, a switch to react-window makes more sense
- react-transition-group's CSSTransition avoids findDOMNode when given an explicit nodeRef. We pass one everywhere, which meant adding a wrapper div (with flex styles to keep the Suspense placeholder centered) in TransitionDelayEnter.

### Test

Automated tests still work + manual testing.

### Release

`RELEASE-NOTES.md` wasn't updated yet. It should contain

> ### Other Changes
> - Upgrade React to v18.3.1


Usiel:

PR #3384, opened 2026-06-11
Warn user when note sync fails

Usiel:
While experimenting with Simplenote and a new editor, I was testing performance with very large notes. I noticed that Simperium actually stops syncing notes of a certain size, but the user is never notified. This PR introduces an error indicator for 413 (doc too large) and other errors. 

### Fix

When note failed to sync we see the warning triangle:
CleanShot 2026-06-11 at 15 10 43@2x

When user changes the note (e.g. shortens it), the pending spinner shows until we either succeed or again get an error:
CleanShot 2026-06-11 at 15 11 01@2x

### Test

Get a sufficiently large random note:

```
head -c 1M /dev/urandom | pbcopy
```

- Add a new note and paste in the random note text.
- Wait for sync to fail
- **Expected**: Warning triangle shows
- Make a change to the note
- **Expected**: Pending spinner shows (triangle stays) and disappears when sync fails again
- Remove most of the text to make it sufficiently small
- **Expected**: Pending spinner comes up and once the sync completes the warning indicator disappears

### Release

Enhancements:
- Warn user when note sync fails

### DIFF
--- a/lib/note-list/note-cell.test.tsx
+++ b/lib/note-list/note-cell.test.tsx
@@ -1,0 +1,141 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import { NoteCell } from './note-cell';
+import { getSyncErrorMessage } from '../utils/sync-error-message';
+import type * as T from '../types';
+
+const noteId = 'note-1' as T.EntityId;
+
+const minimalNote: T.Note = {
+  content: 'Hello world',
+  creationDate: 0,
+  deleted: false,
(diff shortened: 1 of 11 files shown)


Usiel:
This PR is better with https://github.com/Automattic/simplenote-electron/pull/3385, since this PR alone only fixes the error signalling in an active session. When the user restarts Simplenote with an usnyncable note, then the error handling isn't great without #3385.

Usiel:

PR #3385, opened 2026-06-11
Re-enable NoteDoctor to sync notes on startup

Usiel:
NoteDoctor scans for local/ghost discrepancies on idle and queues sync attempts. It was disabled during the 2020 rewrite and never turned back on. Looks to me like it was just forgotten in the already very complex rewrite, but please correct me if I'm wrong @dmsnell. The doctor would be quite useful in combination with https://github.com/Automattic/simplenote-electron/pull/3384.

### Fix

Uncommented code and added some automated tests.

### Release

Other changes:
- Re-enable NoteDoctor to sync notes on startup


Usiel:

PR #3386, opened 2026-06-12
perf: fix bucket queue timer loop

Usiel:
Noticed this while profiling for other work.

### Fix

When there is no work for the BucketQueue it ends up rescheduling itself constantly without any delays, which causes 2.9% of the main thread to spend time on just setting and clearing timeouts. On my machine I get roughly 650 tasks per second due to the BucketQueue's constant rescheduling.

We can optimize this by only running the scheduler when there is work in the bucket and avoid causing any work on the main thread when there is no work to do.

### Test

I added tests that are green for both the new and old implementations, demonstrating that the change doesn't actually change any relevant behavior.

### Release

Other changes:
- Improve performance and energy usage by reducing work executed when idle

### DIFF
--- a/lib/state/simperium/functions/bucket-queue.test.ts
+++ b/lib/state/simperium/functions/bucket-queue.test.ts
@@ -113,6 +113,9 @@ describe('BucketQueue', () => {
 
     queue.add('note-1', Date.now());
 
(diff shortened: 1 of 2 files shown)


Usiel:

PR #3389, opened 2026-06-12
Upgrade to React 18

Usiel:
### Fix

This is required for future work that potentially integrates Lexical for live markdown editing. I thought it may be nicer to split out the React upgrade to a separate PR, since it will be useful regardless of other downstream changes.

React 18 [retires the ReactDOM.render API](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis). Apps must create a root via createRoot, and React warns when createRoot is called repeatedly on the same container. Our boot phases (platform warning, auth screen, main app, logging out) each rendered into #root, so they now share a single root (lib/app-root.ts) and replace each other via root.render().

React 18 also drops support for findDOMNode, which two of our dependencies rely on:

- react-virtualized is replaced with the react-virtualized-compat fork which is compatible with React 18. At some point, a switch to react-window makes more sense
- react-transition-group's CSSTransition avoids findDOMNode when given an explicit nodeRef. We pass one everywhere, which meant adding a wrapper div (with flex styles to keep the Suspense placeholder centered) in TransitionDelayEnter.

### Test

Automated tests still work + manual testing.

### Release

`RELEASE-NOTES.md` wasn't updated yet. It should contain

> ### Other Changes
> - Upgrade React to v18.3.1


lib/note-list/index.tsx
```diff
@@ -91,7 +91,9 @@ const renderNote =
           parent={parent}
           rowIndex={index}
         >
-          <TagSuggestions style={{ ...style }} />
+          <div style={style}>
+            <TagSuggestions />
+          </div>
```
dmsnell, line 96: why are these wrapping `DIV` elements necessary? how do they impact the visual apperance?

dmsnell (commented):
This looks mostly fine. I didn’t test it, as I’m relying on your own testing and confidence. Would be nice to explain the new wrapping elements and see if they are fully necessary. Otherwise if you are ready you can push this out. Might be ideal if we do this first and in isolation so we can release it and quickly react if something goes wildly wrong.

lib/note-list/index.tsx
Usiel: This relates to the following:

> react-transition-group's CSSTransition avoids findDOMNode when given an explicit nodeRef. We pass one everywhere, which meant adding a wrapper div (with flex styles to keep the Suspense placeholder centered) in TransitionDelayEnter.

Here's some additional context: https://github.com/reactjs/react-transition-group/issues/918 - I didn't want to switch to some patched version or a completely different dependency, as the `div` workaround seems to work well enough.

Usiel:
> Might be ideal if we do this first and in isolation so we can release it and quickly react if something goes wildly wrong.

Yep, that sounds good to me.

dmsnell (approved):
Not sure if you were waiting for my approval, but here it is. This is all on you! I have not tested anything.

Usiel:
> Not sure if you were waiting for my approval, but here it is. 

I was indeed! 

> This is all on you! I have not tested anything.

Understood, feels safe to me. I've been running my local Simplenote with this change for a week now.

PR #3408, opened 2026-07-04
Only apply markdown decorations when the note has markdown enabled

### LINKED ISSUE #3379: Markdown formatting applied even when markdown disabled
I don't usually use markdown for my notes, so I was surprised after upgrading to v2.26.0 that some markdown formatting is being applied automatically.

Given there is a specific option to turn markdown on or off, this feels like an oversight/regression?  I guess I could live with it (or find another note app use).. the issue is many of my notes contain various unix commands/config files, which specifically use "#" as a code comment, and it's super distracting to see those lines now in a huge font.

### Expected

With the Markdown option disabled, markdown formatting should not be applied

In v2.25.0 and below:
Image

### Observed

Limited markdown formatting is applied no matter what.

Same note in v2.26.0:
Image

### Reproduced

1. In the latest SimpleNote, add a note with any sorta markdown, like "# header", and see special formatting applied to that line.  Click the "..." Actions button and make sure 'Markdown' is disabled.
2. Downgrade your SimpleNote to 2.25.0
3. That same note is now just plain formatted text (like it had been for every version before 2.26.0)

### Where did you see the bug

- OS: Windows
- OS version: 11
- Simplenote app version: 2.26.0

gunjanjaswal:
Fixes #3379.

Markdown decorations (headings, bold, italic, list markers, and so on) were applied in the editor even when the note had Markdown disabled. Every other Markdown behavior in `note-content-editor.tsx` gates on the note's `markdown` system tag, but the decoration path in `setDecorators` did not, so a plain note with a line like `# comment` or `*.conf` still got heading/emphasis styling. This is the v2.26.0 regression reported in #3379.

## Changes proposed

- `markdown-decorations.ts`: `getMarkdownDecorations` now takes a `markdownEnabled` flag (defaults to `true`) and returns `[]` when it is false.
- `note-content-editor.tsx`: pass `note.systemTags.includes('markdown')`. The existing `clear()` stays unconditional, so decorations are removed when Markdown is toggled off on an open note.
- `markdown-decorations.test.ts`: add a case asserting no decorations are returned when markdown is disabled.

## Testing instructions

`npx jest lib/markdown-decorations.test.ts`

Or manually: open a note with Markdown disabled and a body containing `# heading` or `**bold**`, confirm the text stays plain, then toggle Markdown on and confirm the styling appears.

### DIFF
--- a/lib/markdown-decorations.test.ts
+++ b/lib/markdown-decorations.test.ts
@@ -36,6 +36,15 @@ describe('getMarkdownDecorations', () => {
     ]);
   });
 
+  it('returns no decorations when markdown is disabled', () => {
+    const decorations = getMarkdownDecorations(
+      modelFromLines(['# Title', '**bold**']),
+      false
+    );
+
+    expect(decorations).toEqual([]);
+  });
+
   it('decorates supported unordered list markers', () => {
     const decorations = getMarkdownDecorations(
       modelFromLines(['* list item', '- second item', '  + nested item'])
--- a/lib/markdown-decorations.ts
+++ b/lib/markdown-decorations.ts
@@ -118,9 +118,10 @@ const addInlineSpansForPattern = (
 };
 
 export const getMarkdownDecorations = (
-  model?: Editor.ITextModel | null
+  model?: Editor.ITextModel | null,
+  markdownEnabled = true
 ): Editor.IModelDeltaDecoration[] => {
-  if (!model) {
+  if (!model || !markdownEnabled) {
     return [];
   }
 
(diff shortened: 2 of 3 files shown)


gunjanjaswal:
Gentle bump on this one. It only applies the markdown decorations when the note actually contains markdown, so plain notes stop getting decorated. It's been sitting a while and is still green and mergeable. Happy to rebase if it's drifted from trunk, just flagging it's ready for a look whenever someone has a moment.

gunjanjaswal:
Friendly bump on this. It's been open a while without a first look. The change is small and self-contained: only apply the markdown decorations when the note actually has markdown enabled, so plain notes stop getting them. Glad to rebase or adjust anything if that makes review easier.

